PR #20, opened 2013-01-16
Fix for corrupted targets when using placeables

oschwald:
Without this, editing a target in an XLIFF with placeholders will result in repeated text that is not fixable within Pootle.


unho:
Closing this since the codebase greatly changed and more work is required.

PR #23, opened 2013-02-18
Sync terminology store when creating or updating

rojkov:
This way translators get a chance to edit terminology offline.
Also syncing is required for updating terminology translations against
templates.

This patch depends on the pull request #22. So probably you can't merge it before the release of 2.5.0. But it means that the terminology feature doesn't really work in the pre-release. Especially the part "It is also possible to add extra entries".

### DIFF
--- a/pootle/apps/pootle_terminology/views.py
+++ b/pootle/apps/pootle_terminology/views.py
@@ -18,6 +18,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, see <http://www.gnu.org/licenses/>.
 
+import os
+
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_terminology/views.py
```diff
@@ -18,6 +18,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, see <http://www.gnu.org/licenses/>.
 
+import os
```
unho: This is not necessary.

julen:
Terminology extraction was removed from Pootle, so no point in keeping this around.

PR #248, opened 2014-04-21
Split extension from database file format

diorcety:
This depends on https://github.com/translate/translate/pull/3270

Currently Pootle uses the file extension in order to identify which format use.
- The issue is that an extension(.properties) can be used by multiple format (Java properties / Java UTF-8 properties / GWT properties)
- We need to set the source and the target language of the store before handling it. Indeed it can be useful for the converter to know for which target language the call is made.
  For example, in order to export a .properties file from Pootle we need to know which language to use in order to know which form is each elements of the multistring provided for the translation

(Need proper testing, i only made it works in my env)

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -184,7 +184,7 @@ def fix_monolingual(oldunit, newunit, monolingual=None):
     if monolingual is None:
         monolingual = is_monolingual(type(newunit._store))
 
-    if monolingual and newunit.source != oldunit.source:
+    if monolingual and newunit.source != oldunit.source and not newunit.target:
         newunit.target = newunit.source
         newunit.source = oldunit.source
 


diorcety:
Hi,
Any news of that PR?
Regards,

unho:
@diorcety Since last review we merged back the Evernote fork with our work, which will be released as 2.7.0. Since the new version is based on Evernote codebase this PR might not apply cleanly, so you might need to review it.

However I am not sure if it is possible to reapply this PR since the new codebase removed the support for monolingual formats. But please review the PR status yourself and check the new codebase. Contact us here on in the IRC channel if you need any help.

pootle/apps/pootle_store/fields.py
```diff
@@ -187,6 +187,7 @@ def _update_store_cache(self):
                 from pootle_store.filetypes import factory_classes
 
                 store_obj = factory.getobject(self.path,
+                                              localfiletype=self.instance.translation_project.project.localfiletype,
```
unho: `factory.getobject` doesn't have a `localfiletype` argument (see https://github.com/translate/translate/blob/master/translate/storage/factory.py#L165 ) so this commit doesn't apply. What do you tried to achieve here?

diorcety: Hi,
Please see https://github.com/translate/translate/pull/3270
The issue is that the filetype is resolved from the extension. The issue is that some filetype, for example gwt properties files (java properties files + plurals handling) use the same extension that java properties files. So pootle use the wrong format

unho: Thanks. This was really helpful. Next time please warn about dependencies in the PR. I just added this one myself.

unho:
Hi again.

We cherry-picked **"localfiletype in the model is not anymore the file extension but..."** commit, but as it turned out (see https://github.com/translate/pootle/pull/248/files#r47065913 ) picking it depends on another PR to be merged into TTK and then a new TTK release to be made. So unfortunately this won't make to next Pootle release.

The other four commits can't be picked in the current `master` as they rely on monolingual format support which is currently dropped (yet marked to be recovered) and other massive code changes on Pootle codebase. So those will be picked once monolingual support is brought back to Pootle.

diorcety:
Okay, no problem

unho:
This also seems to depend on https://github.com/translate/translate/pull/126 and https://github.com/translate/translate/pull/3269

PR #298, opened 2014-07-07
RFC: Cleanup: Drop @ajax_required decorator

jleclanche:
Reasoning: All it does is check the method is ajax (through's django's `is_ajax()` method, which checks the `HTTP_X_REQUESTED_WITH` header) and error out if it's not (and we're not in DEBUG mode)

We don't *actually* need that. All of these bits are used as internal apis but requesting them normally should not be an issue (makes it harder to debug, or simply use curl).


unho:
@jleclanche Not sure if this is still applicable.

jleclanche:
The PR isn't but I still want to remove @ajax_required.

unho:
@jleclanche Closing this PR as it doesn't cleanly applies anymore. It would be easier to just start from scratch and file a new PR.

PR #3338, opened 2014-11-07
Templates: implemented generic about page

jleclanche:
Brings back stock Pootle content.

Fixes evernote/pootle#262.


jleclanche:
obsolete

PR #3439, opened 2015-01-03
Travis: enabled caching

julen:
Hope this provides some speed-ups, especially when we enable more environments.

References:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
http://docs.travis-ci.com/user/caching/#Caching-directories-%28Bundler%2C-dependencies%29

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 # https://travis-ci.org/#!/translate/pootle
+sudo: false
 language: python
 python:
   - 2.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,9 @@ matrix:
     - env: DJANGO_VERSION=1.6.5 DATABASE_BACKEND=postgres
     - env: DJANGO_VERSION=1.6.5 DATABASE_BACKEND=mysql_innodb
 cache:
-  - apt
+  apt: true
+  directories:
+    - node_modules
 before_install:
   - travis_retry sudo apt-get install -qq bc
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,11 +44,9 @@ cache:
   directories:
     - node_modules
     - .pip_download_cache
-before_install:
-  - travis_retry sudo apt-get install -qq bc
 install:
   - CFLAGS="-O0" travis_retry pip install -r requirements/travis.txt --timeout=240
-  - travis_retry pip install "Django>=$DJANGO_VERSION,<"$(echo $(echo $DJANGO_VERSION | cut -d"." -f1,2) + 0.1 | bc) --timeout=240
+  - travis_retry pip install "Django==$DJANGO_VERSION" --timeout=240
   - pip freeze  # Print all installed versions for reference.
 before_script:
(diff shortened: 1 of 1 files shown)


julen:
This is currently failing because our *.travis.yml* does `sudo apt-get install -qq bc`.

`bc` is only used to calculate the next major Django version (float arithmetic).
```bash
pip install --use-mirrors "Django>=$DJANGO_VERSION,<"$(echo $(echo $DJANGO_VERSION | cut -d"." -f1,2) + 0.1 | bc) --timeout=240
```

This ensures the latest bugfix release will be installed, but in this scenario couldn't we get rid of that in favor of a simple `pip install --use-mirrors "Django=$DJANGO_VERSION"`? (cc @dwaynebailey)

Another alternative is replacing `bc` with some other scriptable tool that is already available in the Travis CI container.

A last resource would be uploading the `bc` Debian package to S3 and installing it in user-land (seems overkill).

jleclanche:
> This ensures the latest bugfix release will be installed, but in this scenario couldn't we get rid of that in favor of a simple pip install --use-mirrors "Django=$DJANGO_VERSION"?

I proposed this. Dwayne was against it. Let me propose it again more sternly: Let's do it, please. We've been trying to speed up travis in various ways and this is a pretty big bottleneck for pretty much zero gain.

julen:
Ok, let's hear from @dwaynebailey, but the improvement is already noticeable. We could also potentially cache the *virtualenv* directory (it'd make a bigger difference when the matrix is bigger).

dwaynebailey:
@jleclanche lets deal with [real data points](https://docs.google.com/a/translate.org.za/spreadsheets/d/11FQMa5QoZWQsDk8w7U5WPQHkORcIIn2shE4KCOG4PXc/edit#gid=1347765808) not guesswork please.  ``bc`` is not the bottleneck at all, its expensive (10s because of apt-get) but not as time wasting as ``pip install`` or ``make assets``.

@julen the improvements in those real bottlenecks are significant though, so simply from the point that:

1. We start sooner - a bigger issue I think then running long at the moment.
2. Can run shorter - with some potential improvements in docs and virtualenv not yet explored.

We should land this.

However, the ``bc`` cludge is there for a reason. With ``==`` we end up not installing the version that our users will install, namely the latest bugfix version, rather we're running and testing against an older obsolete versions of Django.  From experience we're rubbish at updating ``requirements/base.txt`` and even worse at updating ``.travis.yml`` to match.

Can I rather propose that we just do something with ``cut`` and such to provide the ``<N.N+1`` option.  I'll try craft it and paste it here.

jleclanche:
Let's be realistic though, Django minor versions should *never* be backwards-incompatible in a way that will break travis. The odds are so remote that it would be that it's not worth the trouble. And if a Django version ends up being backwards incompatible in such a way, we would most likely know about it.

That's why this isn't worth it imo.

dwaynebailey:
@julen This patch https://github.com/dwaynebailey/pootle/commit/582c59834d360f5818dfc6305b91cc75ea53f9f1 does the right thing without using ``bc``.

To be honest just fixing this was easier than debating this. It works and I'd prefer our testing to not debend on chance.

jleclanche:
I'm ok with this.

Shouldn't you be resting? :)

PR #3442, opened 2015-01-05
Pick error reporting changes from stable

unho:
This is to merge into core branch.


jleclanche:
lgtm

unho:
@jleclanche The name of the app is unimportant. Functionality is. Anyway you wanted to remove all those `pootle_` prefixes from app_names, so I didn't saw the need to rename the app.

unho:
I would like to hear something from Evernote people. @julen @ta2-1 @iafan

pootle/apps/evernote_contact/forms.py
```diff
@@ -64,3 +65,26 @@ def from_email(self):
...
+
+    report_email = forms.EmailField(
+        max_length=254,
+        required=False,
+        widget=forms.HiddenInput(),
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(PootleReportForm, self).__init__(*args, **kwargs)
+        self.fields.keyOrder += ['report_email']
```
julen: Can we add a note to mention this is necessary because `keyOrder` is altered in the parent form, please? Someone might think this line is unnecessary.

pootle/apps/pootle_project/models.py
```diff
@@ -154,6 +154,13 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
 
     directory = models.OneToOneField('pootle_app.Directory', db_index=True,
             editable=False)
+    report_email = models.EmailField(
```
julen: Would it be possible to stick to the style of the current file?

julen:
Beside the two nits commented inline, looks good. Also feel free to drop the `evernote_` prefix if needed, it's harmless and straightforward as no models are involved.

pootle/apps/pootle_project/models.py
unho: Sure. Despite I hate that style.

Issue #3343, opened 2014-11-12
Core: Reimplement allauth


jleclanche:
Core uses the old Pootle authentication while 2.6 moved to allauth to provide a more modular approach to authentication.  The divergent authentication methods is a big blocker to merging the code.

The approach will be to land allauth on core as follow:

1. [x] Translate integrates allauth in core, using a feature branch for doing so (#3447 & #3468)
2. [x] Translate implements the workflow with Evernote's help (#3517)
3. [x] Evernote implements the allauth Evernote backend (#3515)
4. [ ] Evernote addresses styling UI issues (#3518)
5. [x] Verify previously-existing email when using a social account (#3728)
6. [x] Sync-up email addresses when editing users (#3737)

Also depends on:
* [x] Must be able to disable user sign ups (#3449)
* [x] CSRF cookie might be missing in private browsing sessions (#3774)

dwaynebailey:
Some observations of things that maybe need clarification on the mockups vs current screens/workflow

- 'Remember me' is missing from the mockup signups (do we still want that? My guess is yes)
- Settings->'Link with Evernote' (this needs to go but I assume kept with EN backend)

iafan:
We can add the 'remember me' option (and have it checked by default).

As for sign up not asking for the user's full name, I think it's ok. I'd try to make sign up flow as simple/fast as possible, and then provide a way for the user to discover the convenient form to fill in their profile (and probably add some score when one fills in their profile).

ta2-1:
I've created #3515 where we adjust Evernote backend to allauth branch.
the UI work can be left for the future.

ta2-1:
The question we haven't discussed: how do we connect different social accounts?

julen:
> The question we haven't discussed: how do we connect different social accounts?

This is a question to be answered in #3517.

julen:
FTR: I've split the separate concerns described here into more manageable subtasks.

PR #3447, opened 2015-01-07
Move core to Allauth

jleclanche:
Bit of a WIP but open for reviews.


jleclanche:
cc @julen @unho @dwaynebailey

jleclanche:
@iafan This is actually ready for UI review. There shouldn't be any actually changes in the templates from core currently, just a bit of shuffling around with the template names.

jleclanche:
merged to `allauth` branch

dwaynebailey:
This is a testing review.  I setup Pootle on mysql with MyISAM, using syncdb, migrate, initdb to get the server setup.

I noted these issues in testing:

- "Table 'pootle.account_emailaddress' doesn't exist".  This occurs when trying to signin and also when trying submitting the lost password form.
- b1e045f86bf0ac31c865e07a2a9958acdf8d68b4 LOGIN_REDIRECT_URL = "/" - currently we redirect to the place from which the user requested a login.  Is reverting to / something planned temporary or in error?
- CAN_REGISTER = True, but I can't find any access to a registration form from the main login Window
- The pages don't expose the required functionality outlines in issue #3343. Referring to the mockups:
  - The login page (3) doesn't expose all the needed functionality
  - Not sure how to enable social authentication to validate operation of (1)
  - Didn't test (4), login when singups are disabled, as without signins having needed (3) functionality I assumed it isn't implemented
  - Can't test singups so can't test what is exposed for (5) (form for signing up as a new user)
  - Reset password (6) doesn't expose all the needed items.  It also hits the Table issue mentioned above so can't test any of the results of a reset request.
  - Couldn't test activation of an account (8), or password reset (9) without being able to login or create new accounts.

jleclanche:
> "Table 'pootle.account_emailaddress' doesn't exist". This occurs when trying to signin and also when trying submitting the lost password form.

I had that and migrate took care of the issue. Hmm...

jleclanche:
> b1e045f LOGIN_REDIRECT_URL = "/" - currently we redirect to the place from which the user requested a login. Is reverting to / something planned temporary or in error?

This is the default redirect url. It redirects to ?next if it exists.

jleclanche:
I kept social auth disabled because it's disabled in master and we don't have a good UI for it. It's easy enough to enable when we get to that.

Closed signups are implemented in allauth with the help of 6981a540ee580b44110f19ca71595abc3a9930d0 and only require a setting.

unho:
Just did a very quick check and I found that it would be easier if we can keep a 2 spaces indent for HTML as told in the styleguide since that would imply less diff to review.

dwaynebailey:
> I kept social auth disabled because it's disabled in master and we don't have a good UI for it. It's easy enough to enable when we get to that.

OK, but that's needed for any styling executed by EN so I'd assume since its in the basic mockups that we'd have it on by default.

> Closed signups are implemented in allauth with the help of 6981a54 and only require a setting.

Which setting? There are no signups at the moment with ``CAN_REGISTER = True``.  Either that needs to go and use the allauth setting, or the signups needs proper exposure as per mockups.

dwaynebailey:
Another batch of review feedback.  *The missing Table issue was bustage in ``django-allauth==0.19.0``*

Workflow issues (@jleclanche mostly)
- Settings->Reset Password - goes to reset password dialogue which asks for an email address to initiate password reset.  Current core goes to the actual reset password dialogue.  I think the former is more desirable here.
- Register screen - @jleclanche says is at /accounts/allauth/register/ - gives a 404.
- Signin and Signout produce notices - not present in https://core-translate.stage.evernote.com/

Styling related issues (@iafan and @jleclanche):
- Signin and Signout all post redundant message boxes, that need to be dismissed, to declare that the action has been successful.  Lets rather report errors and assume success doesn't need any reporting.
- Others reported on #3343

What does work:
- Login works
- Logout works
- Reset password via emails send correct link
- Password reset works after getting the reset link

jleclanche:
> Signin and Signout produce notices - not present in https://core-translate.stage.evernote.com/

Noticed that. I'm not sure how to get rid of them - it's possibly something we'll have to ignore in js. Im gonna leave it in for now because it could simply be integrated as part of the signin workflow instead of removed.

jleclanche:
@julen @iafan This can either be merged imminently, or it can wait for further styling while it's still in upstream/allauth. What's your take on it?

jleclanche:
@iafan Re signin/signout messages: We want you guys to take a decision about what we do with these notices, since they could present themselves useful for the UI. If they are to be removed, I am guessing they just have to be ignored in javascript and not displayed.

dwaynebailey:
@iafan Just so that things are doubly clear:

1. This is now ready for you to take over and style and its ready for Evernote to look at the Evernote backend
2. The implementation has been done such that it could land now on core, only issue is the login logout message that @jleclanche mentioned which are I think a regression.  Thus this can land now on core.  Doing that would assist us in that it would make migrations easier, not sure though how it could/would live alongside current EN login system

iafan:
@jleclanche, let's keep the development in the branch and merge this into core once this is fully implemented and ready for production (keep in mind that we do frequent releases off the core, so it should be in a stable state at any time).

iafan:
@ta2-1, can you please have a look at this rather sooner than later and see how we can re-integrate Evernote flow according to our UI specs and whether we will need any changes/tweaks/help from @jleclanche to make this happen?

julen:
Some notes from a quick overview:
* In some template *Sign In* has been renamed to *Log In*. We have been using the former, so this is a reminder for ourselves to keep that consistent.
* Re. messages: they come from the messages framework and [according to the docs](http://django-allauth.readthedocs.org/en/latest/advanced.html#messages) they can be ignored by overriding its templates. The truth is the [messages framework](https://docs.djangoproject.com/en/1.6/ref/contrib/messages/) is only used in a [single place](https://github.com/translate/pootle/blob/b1329ce44c8bfb7f4b74aa23e37dd7a41ad55411/pootle/apps/evernote_auth/views.py#L61) currently, and since we'll be integrating EN auth as an allauth provider, this might be a good time to revisit if we actually need it.
* In `/accounts/allauth/*` URLs *allauth* is superfluous, can we get rid of that in favor of `/accounts/`?
* Moving `pootle_profile` URLs to `/profiles/` breaks our URLs. We want to keep having public user-related pages available at `/user/<username>/`.
* As @unho said, please keep 2 space indents in HTML.

jleclanche:
Thanks, will fix all that today. Use Sign In everywhere then?

dwaynebailey:
> Thanks, will fix all that today. Use Sign In everywhere then?

The mockups use Sign in everywhere, so lets follow the mockups if any changes where made.

iafan:
> Settings->Reset Password - goes to reset password dialogue which asks for an email address to initiate password reset. Current core goes to the actual reset password dialogue. I think the former is more desirable here.

I'm kind of inclined to keep it the way Jerome did, since this simplifies the code and doesn't introduce any new UI. People seldom change their passwords unless forced to, so it's not the feature I'd invest much time into.

PR #3467, opened 2015-01-13
Move core to Allauth

jleclanche:
Moved from #3447


jleclanche:
Oops, misclicks. And github doesn't let me edit it.

PR #3476, opened 2015-01-14
README: removed redundant badges

julen:
Travis status is well enough.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,22 +5,6 @@ Pootle
     :alt: Build Status
     :target: https://travis-ci.org/translate/pootle
 
-.. image:: https://pypip.in/download/Pootle/badge.png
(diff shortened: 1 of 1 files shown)


unho:
@julen Despite them being redundant, I think they provide useful information in one sight.

julen:
So perhaps you can add them somewhere else, but bloating the README with them is redundant as you say.

jleclanche:
I'm +1 on this, badge bloat looks as professional as blinking "New" gifs on a website.

unho:
@julen The idea behind the badges was to see them as soon as you visit the project main page, which prominently displays a rendered version of the README, so if those are going to be removed from there, they are definitely not being put in another place.

Since you both want this to be gone, then merge the PR.

PR #3481, opened 2015-01-15
Allow admins to access disabled projects

### LINKED ISSUE #3451: Allow admins to access disabled projects
Currently admins cannot access disabled projects, it returns a 404 error. They should be able to access them, because they might be preparing a project (adjusting permissions etc.) prior to making it public.

For a enhanced user experience and to avoid admins shooting themselves in the foot, a message needs to be displayed when browsing disabled projects, indicating that while disabled, the project is only visible to admins. This is basically the same thing we do for inactive static pages.

julen:
This allows admins to access disabled projects, which currently 404s for everyone.

Fixes #3451.

### DIFF
--- a/pootle/apps/pootle_misc/context_processors.py
+++ b/pootle/apps/pootle_misc/context_processors.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright 2009-2014 Zuza Software Foundation
+# Copyright 2014-2015 Evernote Corporation
 #
 # This file is part of Pootle.
 #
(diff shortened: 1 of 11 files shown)


julen:
Would appreciate a review/test @iafan.

iafan:
Testing on stage: the only minor thing I currently noticed is styling. The disclaimer box needs some margin between its top and the header. Can you adjust it to match the left/right margins between that box and the edges of the page?

julen:
Thanks for reviewing, just fixed that and rebased on top of current master. Should be good to merge, then?

iafan:
GTM

PR #3503, opened 2015-01-19
Pick and update docs

unho:
This is for picking INSTALL, installation, upgrade and releasing docs changes from stable branch, merge those into current master branch, and update those documents to reflect the changes introduced in core.


unho:
@dwaynebailey @jleclanche No longer a work in progress. Please review so it can be landed on master as soon as possible.

dwaynebailey:
gtm

PR #3516, opened 2015-01-20
Docs: drop changelog

dwaynebailey:
This drops the old changelog and merges any information as needed into the older changelogs.


unho:
@dwaynebailey gtm. This was super easy to review.

PR #3526, opened 2015-01-21
Switch to Django 1.7

### LINKED ISSUE #3354: Be Django 1.7 compatible
We must ensure we're Django 1.7 compatible, let's track that here.

The blockers are:
* [x] Move to built-in Django migrations (#3355)
* [x] Set sane defaults for `SESSION_CACHE_ALIAS` (#3356)
* [x] Rename `Submission` model's `check` field (#3462)
* [x] Pages pending user agreement cannot be retrieved (#3506)

There are other cleanups that can be done and niceties introduced in Django 1.7 but those are out of the scope of this issue.

Resources:
* [Django 1.7 release notes](https://docs.djangoproject.com/en/dev/releases/1.7/)
* [Django deprecation timeline](https://docs.djangoproject.com/en/dev/internals/deprecation/)

julen:
This PR puts in place all the necessary bits to have current master up & running in Django 1.7.

The most notable part is the switch to built-in Django migrations. This, along with a conflictive field (it was necessary to rename `Submission.check`), requires some steps to migrate existing DBs in master. Instructions follow:

1. Make sure your virtualenv is still on Django 1.6 (`pip freeze --local | grep Django` is your friend)
2. `manage.py migrate` to apply the migrations required pre-1.7
3. Switch to 1.7, `pip install -U "Django<1.8"`
4. Ensure no old migration-related bytecode exist: `rm $(find . -type f -name "*.pyc")`
5. `manage.py migrate --fake` to fake initial 1.7 built-in migrations

I'd appreciate testing this PR with real world DBs. @ta2-1 do you think it's worth testing `update_stores`/`sync_stores` just in case?

Fixes #3354.


dwaynebailey:
Did a high level review and looks gtm, didn't test on any database as I don't have any master ones at the moment.

I suspect that the two stage migration needs to be replicated for the 2.6.0 migration path as follows:

1. Get ``stable/2.6.0`` up to speed with these migrations
2. Reset migration again in master

Would appreciate some input from @jleclanche on this.

unho:
@julen I assume you didn't squash the South migrations because those are planned to be dropped in the future, isn't it?

unho:
Didn't test this, but looks gtm.

julen:
@unho *south_migrations* is a simple `git mv` for the *migrations* directory

unho:
@dwaynebailey Yes, we will need to backport the migration for the check field rename.

ta2-1:
Ran `refresh_stats` command on migrated DB (full EN production copy) and got a hanging query:
```sql
SELECT
  `pootle_store_unit`.`store_id`,
  MAX(`pootle_store_unit`.`creation_time`) AS `max_creation_time`
FROM 
  `pootle_store_unit` 
INNER JOIN 
  `pootle_store_store` ON (`pootle_store_unit`.`store_id` = `pootle_store_store`.`id` )
GROUP BY 
  `pootle_store_unit`.`store_id`
ORDER BY 
  `pootle_store_store`.`pootle_path` ASC, `pootle_store_unit`.`index` ASC
```

On 1.6 it was 
```sql
SELECT
  `pootle_store_unit`.`store_id`,
  MAX(`pootle_store_unit`.`creation_time`) AS `max_creation_time`
FROM 
  `pootle_store_unit` 
LEFT OUTER JOIN 
  `pootle_store_store` ON (`pootle_store_unit`.`store_id` = `pootle_store_store`.`id` )
GROUP BY 
  `pootle_store_unit`.`store_id`
ORDER BY 
  `pootle_store_store`.`pootle_path` ASC, `pootle_store_unit`.`index` ASC
```

1.6 query takes less than 1 minute,  when 1.7 query can take more than 10 minutes. 
Link to the code: https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/management/commands/refresh_stats.py#L437-L439

julen:
I've ran `refresh_stats` several times locally today and it finished without any issues. Also tried both queries against MySQL and it didn't complain. Stage works fine too, and we suspect the issue is in @ta2-1's machine. Waiting for his feedback before merging.

jleclanche:
Missing index maybe? That's the only thing that could cause a difference like that on two machines.

julen:
Got the green light from @ta2-1 via IM to merge this.

dwaynebailey:
@ta2-1 can we have such comments in the PR in future please

ta2-1:
@dwaynebailey, yes of course

PR #3546, opened 2015-01-26
Cleanup: simplified code by using `setdefault()`

julen:
While reading over `refresh_stats`' code found out this to be an easy cleanup.

Waiting for your approval @ta2-1.


ta2-1:
GTM

PR #3558, opened 2015-01-29
Store: use custom file system storage class

julen:
This fixes an issue with migrations, where it would always report changed models because the `location` of `FileSystemStorage` changed — yes, even if it's not DB-related it's used to allow using code in `RunPython` calls.

As a side note, it also drops the `upload_to` parameter from `TranslationStoreField` (subclass of `FileField`), which had a dummy value because it was required, but [starting in Django 1.7 it's not required anymore](https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.FileField.upload_to) and it's safe to be dropped.

Please, test initializing from scratch works on your side. Also please check that it doesn't report changed fields anymore.

### DIFF
--- a/pootle/apps/pootle_store/migrations/0001_initial.py
+++ b/pootle/apps/pootle_store/migrations/0001_initial.py
@@ -37,7 +37,7 @@ class Migration(migrations.Migration):
             name='Store',
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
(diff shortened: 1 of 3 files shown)


julen:
Sure, done.

ta2-1:
It works well.

dwaynebailey:
Performed an install on sqlite from scratch.

* migrate + initdb - all fine
* update_stores - fine
* runserver - fine, tested up until issues related to mysql specific queries.

lgtm

julen:
Thanks for all the confirmations guys.

dwaynebailey:
@julen did the same test but changing PODIRECTORY to po2, all works no errors in migrate of initdb

Issue #3462, opened 2015-01-12
Rename `Submission` model's `check` field


julen:
In order to switch to Django 1.7 the `check` field of the `pootle_statistics.Submission` model needs to be renamed, because now the name conflicts with a new method introduced as part of the [system check framework](https://docs.djangoproject.com/en/1.7/ref/checks/#models).

More specifically, we'll be hitting `models.E020`:

```python
ERRORS:
pootle_statistics.Submission: (models.E020) The 'Submission.check()' class method is currently overridden by <django.db.models.fields.related.ReverseSingleRelatedObjectDescriptor object at 0x10f845d10>.
```

PR #3559, opened 2015-01-29
BACKPORT: Django 1.7: renamed `check` field to `quality_check`

### LINKED ISSUE #3550: Backport rename `Submission` model's `check` field to 2.6.0
See #3462.

jleclanche:
Django 1.7+ includes a new system check framework which adds a `check`
method on models, so the `check` field has been renamed to resolve the
conflict.

References #3462.
Fixes #3550.

Cherry-Picked-From: 1453757472bb6252a12b18e57c303b0088e53ad7


dwaynebailey:
Looks GTM

I'm with @jleclanche on the deps issue.  South is designed to work out the deps, we only need them when it gets it wrong.  Sounds like we're trying to fix a problem that doesn't exist.  I'd rather add deps when we see it getting it wrong and we have a number of DBs that we can test to see exactly that.

unho:
@jleclanche lgtm.

PR #3565, opened 2015-01-29
Bump to 2.7.0-alpha1

dwaynebailey:
In other words lets lie less about Pootle's current version.

@julen wanted to check with you before landing in case there is EN fallout.

### DIFF
--- a/pootle/__version__.py
+++ b/pootle/__version__.py
@@ -20,6 +20,6 @@
 
 """This file contains the version of Pootle."""
 
-build = 25101
-sver = "2.5.1-alpha1"
-ver = (2, 5, 1)
+build = 27001
+sver = "2.7.0-alpha1"
+ver = (2, 7, 0)


julen:
Can we use the occasion to be more (*de facto*) "standard" and put the version information in `pootle.__init__`? `__version__` should be a module/package attribute as specified by PEP8, not a module.

Also ensuring the version scheme matches [PEP 440](https://www.python.org/dev/peps/pep-0440/) wouldn't hurt. Defining the version information in the tuple should be enough, then the string can be generated from there ([see how Django does it](https://github.com/django/django/blob/stable/1.7.x/django/utils/version.py)).

And answering the concern, version numbers don't affect us and any change is safe from that perspective. But at the same time, we'd very likely start using this module if we can directly get the git revision/branch information from there ([see example](https://github.com/getsentry/sentry/blob/master/src/sentry/__init__.py)). We currently do some deployment hackery to get this information for our staging server (see the ribbon on the upper-right corner).

jleclanche:
@julen Maybe leave those changes for Pootle 3.0 (which we're going to discuss a bit at fosdem I think). They are backwards incompatible and there is a potential change in version scheme.

dwaynebailey:
@jleclanche I don't think backward compatibility is an issue as no outside app is querying or should be querying the version info.  Maybe in TTK that is true.

@julen yes PEP 440 and other cleanups would be great, but I don't have time to do that just yet.  I'll file and issue to track this.

unho:
I know this is closed, but leaving this as a reminder: Do we still need `build` now that we dropped the old upgrade machinery?

Issue #3547, opened 2015-01-26
Projects initialisation does not make stores and stats available


dwaynebailey:
In the following conditions:
1. You have a project folder with various PO files in non-GNU layout
2. You create the project within Pootle
3. Pootle initialises the project

Result
1. All TPs exist and all of the directory structure and files exist
2. There are no stats for the stores
3. Running `refresh_stats` across the stores does not create the stats.  You can only create them by running `update_stores`

Expected
- The stores are initialised and all word counts are correct

jleclanche:
@ta2-1 status ping

ta2-1:
@jleclanche it will be fixed in #3583.

jleclanche:
@ta2-1 Are you planning to land #3583 any time soon?

phlax:
i think this is fixed - @ta2-1 any thoughts?

unho:
Reopened so @dwaynebailey can confirm which is the expected behavior. From my reading it is that translations should be automatically pulled in after project is created.

dwaynebailey:
Think I see now why @unho reopened, did an initdb on master and we're in an odd state

phlax:
this was related to old refresh_stats mechanism

PR #3567, opened 2015-01-30
rq: fix queue running indicator for > 1 workers

### LINKED ISSUE #3552: RQ status indicator on Admin->Info page only works for one worker
If you have more than one worker running then the status line will show "Stopped".  It only shows the correct status of "Running" if there is only one worker running.

dwaynebailey:
This indicator is on the admin dashboard and was showing as not running if there was more than one worker.  Took the opportunity to simplify the logic a bit.

Fixes #3552

### DIFF
--- a/pootle/apps/pootle_app/views/admin/dashboard.py
+++ b/pootle/apps/pootle_app/views/admin/dashboard.py
@@ -293,9 +293,7 @@ def rq_stats():
     queue = get_queue()
     failed_queue = get_failed_queue()
     workers = Worker.all(queue.connection)
-    is_running = False
-    if len(workers) == 1:
-        is_running = not workers[0].stopped
+    is_running = len(workers) >= 1 and not workers[0].stopped
 
     result = {
         'job_count': queue.count,


ta2-1:
It's a wrong situation if more than one worker exist.
What do you think if we show something like "Multiple workers started!" error message as a state?
Note that we'll have a separate worker for `update_stores` tasks (according #3547).
And admin UI should be adjusted too.

dwaynebailey:
@ta2-1 its perfectly fine to have more than one worker.  If the queue gets long you can spin up workers to assist on the queue.  I did that when testing ``refresh_stats --calculate-wordcount``, in fact I spun up 5 workers.

I don't think you want a separate worker for ``update_stores``, you want a separate queue.  The same worker can service all queues or only one queue.  I would recommend we not create a queue specific for a given command, but rather have ``low`` and ``high`` priority queues or similar concept.

The only admin part I see changing would be to list jobs in each of the named queues.

ta2-1:
@dwaynebailey technically it's possible to have more than one worker.
But I have to say that RQ worker has been implemented to make sure we update stats from one point (process).
It allows us to avoid conflicts when same branches of our tree (Projects->TPs->Directories->Stores) are updated simultaneously.
As I know one worker is a one process. Thus if we have a lot of stats jobs we can get a stuck with update_stores jobs (either on the contrary).
There is constantly running sync cycle in EN where sync/update stores jobs are executed.
I'd prefer to separate these processes. (cc @iafan)

dwaynebailey:
@ta2-1 if it is the case that you outline that we can't use more than one worker for stats, then I'm afraid the implementation of stats queues is seriously flawed and perpetuating that flaw into ``update_stores`` isn't the right approach.

It sounds like you're missing the concept of creating dependencies when creating stats jobs i.e. higher level rollup jobs need to depend on lower level ones completing first.  RQ provides such ability.

We will see more uses for RQ, if we have a queue for each use and a worker for each queue it just doesn't scale properly.  It makes more sense to have queues that relate to priority rather than being job specific.  But regardless if you do create a job specific queue then we can't afford to have ourselves manage the workers, that is what RQ should be doing.  We should be able to run all quese on one worker or many workers serving all queues.

unho:
This PR looks gtm.

@dwaynebailey @ta2-1 @iafan The discussion here looks like it is unrelated to the original issue. So maybe landing the PR and creating a separate issue for this discussion is the best approach.

iafan:
@dwaynebailey, on your last comment: in theory, I agree with you that the beauty of RQ is that if you spawn multiple workers, they would do the work faster, and the code behind our jobs should better be done in the way that allows parallel execution.

However, in our specific case, we decided to move our statistics calculation code into a *single* worker to simplify the code behind stats calculation and make it more reliable. prior to RQ, we would update stats in real time and in multiple places, and this proved to cause side effects that were extremely hard to debug. Now that we have RQ and a single worker that processes stats-related jobs sequentially, the code seems to be more stable (at least this is as good as we could get up to this point). Now, if we want to have multiple workers doing stats calculation in parallel, we will need to rethink the whole approach behind the calculation logic and make sure we don't introduce the same problems we tried to escape originally. In our case, we didn't reach the point we would necessarily need multiple workers.

We can think of a way when if multiple workers are running, only one is working on the stats queue, while other jobs, if they can be easily paralleled, can be processed by multiple workers.

dwaynebailey:
@iafan there is a design decision embedded in how stats queues have been implementing which I'm concerned is going to be perpetuated into update_stores and into all other uses we will want to make of queues which is fundamentally wrong. I'm concerned that such a fundamental decision wasn't discussed or documented.

There are cases where the stats queue does get filled up.  I hit that with ``--calculate-wordcount``, this was on small projects that created 200,000 jobs.  It took 5 hours to clear that queue.  The only way to clear such a queue in a reasonable time is to launch new workers.

What do you suggest?

ta2-1:
@dwaynebailey `--calculate-wordcount` case is a bug as we discussed before and I believe it shouldn't be a criteria for conclusions.
RQ job count is estimated with number of updated stores. If store.state >= parsed every `store.save()` triggers one RQ job that updates stats for all TreeItems from a leaf (Store) to the root (Project).
We understand that there are a lot of possible optimizations.
I like your idea to calculate stats in `update_stores` job. Let's think how to avoid the problems we tried to escape originally.

julen:
5eca1dd fixed this, closing.

Issue #3351, opened 2014-11-14
Redis is a hard-dependency for RQ and for cache


julen:
Redis now being a hard-dependency (job queue, revision counter also in the near future), we need to update docs, depchecks and code assumptions accordingly.

Developers also agreed that it makes more sense to use it more effectively, therefore:

1. It'll be enforced as The caching backend. People won't need to have both Redis + memcached, but the former will be enough.
2. Developers will be able to use lower-level Redis capabilities when it's required/justified, be it for improved footprint or performance reasons. For any standard use of the cache, general Django APIs are enough.

* [x] #3585 Setup default stats cache
* [x] #3590 Move from django-redis-cache to django-redis
* [x] #3591 Drop memcached as preferred cache and change all docs accordingly
* [x] Reflect all of this in the docs.

dwaynebailey:
This also needs updates in:

1. depcheck
2. dashboard

Essentially ``grep memcache``

dwaynebailey:
@julen busy looking at this.  Was there any reason to choose [django-redis-cache](https://github.com/sebleier/django-redis-cache) over [django-redis](https://github.com/niwibe/django-redis)?

julen:
Nothing specific that I'm aware of, probably the first that we found for the basic caching needs at the time. Anything else you could think of @ta2-1?

julen:
But also check my comment on https://github.com/translate/pootle/issues/3585#issuecomment-73847522 if you haven't do so yet.

PR #3568, opened 2015-01-30
Redis is a default option (dashboard reporting)

dwaynebailey:
This is the dashboard component of ensuring that Redis is part of the default Pootle setup.

This adds Redis to the critical checks and adds the needed depcheck functions.  We check that we can:

1. Access Redis
2. Have at least 1 worker running

The documentation that is required for #3351 will come via its own PR. 

Partial fix of #3351


unho:
lgtm

PR #3571, opened 2015-02-02
l10n: Bring back l10n

dwaynebailey:
Ready for review.

This does a number of changes:

- Fixes scripts that have slightly rotted
- Does extraction of .jsx and related template files more correctly
- Fixes a few issues about strings not being extracted


unho:
@dwaynebailey Other than the nit (which can be fixed later) this lgtm.

PR #3572, opened 2015-02-02
Move path substring logic to Python

### LINKED ISSUE #3539: Make resource list work on sqlite and Postgres (SUBSTRING_INDEX)
In https://github.com/translate/pootle/blob/master/pootle/apps/pootle_project/models.py#L329 we make use of SUBSTRING_INDEX.  [SUBSTRING_INDEX](http://dev.mysql.com/doc/refman/5.6/en/string-functions.html#function_substring-index) is not available on sqlite or on Postgress so we'll need to have custom SQL there for those databases.

dwaynebailey:
The SUBSTRING_INDEX function is specific to MySQL.
This moves the logic to Python where we can easily get the relevant
parts of the path and remove all duplicates without adding
database-specific code for postgres and sqlite.

Fixes #3539

Thanks, @jleclanche ;) who didn't do Postgres but decided to do my work instead

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -324,10 +324,7 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         resources_path = ''.join(['/%/', self.code, '/%'])
 
         sql_query = '''
-        SELECT DISTINCT
-            REPLACE(pootle_path,
-                    CONCAT(SUBSTRING_INDEX(pootle_path, '/', 3), '/'),
-                    '')
+        SELECT pootle_path
         FROM (
             SELECT pootle_path
             FROM pootle_store_store
@@ -339,13 +336,17 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         '''
         cursor = connection.cursor()
         cursor.execute(sql_query, [resources_path, resources_path])
-
         results = cursor.fetchall()
 
         # Flatten tuple and sort in a list
-        resources = list(reduce(lambda x,y: x+y, results))
-        resources.sort(key=get_path_sortkey)
+        resources = set()
+        for result in results:
+            s = result[0]
+            s = s[s.find("/", s.find("/", 1) + 1) + 1:]
+            resources.add(s)
 
+        resources = list(resources)
+        resources.sort(key=get_path_sortkey)
         cache.set(cache_key, resources, settings.OBJECT_CACHE_TIMEOUT)
 
         return resources


dwaynebailey:
@julen you probably want to look at our approach.  I know you mentioned that the current SUBSTRING_INDEX approach was there because of performance issues so it would be good to get your feedback as to if this approach would be problematic.

unho:
I must say I am a little worried about performance impact of moving part of the query to Python code.

Just for reference the move to only SQL was picked in the past in the stable/2.6.0 as https://github.com/translate/pootle/commit/270f7280 then was altered for supporting PostgreSQL in https://github.com/translate/pootle/commit/a592fa83 which was then altered in https://github.com/translate/pootle/commit/a0f9c424 to get back an alternative for SQLite.

jleclanche:
There should be no performance impact. In fact, I believe this way is
actually faster (substr function has performance concerns... some of them
in fact quite critical: http://bugs.mysql.com/bug.php?id=65243). There's no
way python set and list manipulation is slower here.

J. Leclanche

2015-02-03 9:03 GMT+01:00 Leandro Regueiro <notifications@github.com>:

pootle/apps/pootle_project/models.py
```diff
@@ -339,13 +337,16 @@ def resources(self):
...
         # Flatten tuple and sort in a list
-        resources = list(reduce(lambda x,y: x+y, results))
-        resources.sort(key=get_path_sortkey)
+        resources = set()
+        for result in results:
+            s = result[0]
+            s = s[s.find("/", s.find("/", 1) + 1) + 1:]
+            resources.add(s)
 
+        resources = list(resources).sort(key=get_path_sortkey)
```
julen: With this code you never get any results: `.sort()` *has to* go in a separate line because it works in place.

```python
resources = list(resources)
resources.sort(key=get_path_sortkey)
```

julen:
Did anyone care to test the PR before commenting?

As mentioned on the line comment, in its current state it doesn't work.

Also got some numbers after several runs. The best case is listed here.

* Timing with the current custom SQL query
  * cursor.execute 0.179949998856
  * reduce + sort 0.00205397605896
  * TOTAL 0.182003974915

* Timing with PR 3572
  * cursor.execute 0.215694189072
  * set + sort 0.0124309062958
  * TOTAL 0.228125095367

ta2-1:
I have a question if you allow.
Why do we keep `SELECT DISTINCT` in `SELECT` ? 
`SELECT DISTINCT` make a sense with `REPLACE` as I understand.

julen:
Agreed with @ta2-1 — that's why you're using a `set()` in the first place, right?

dwaynebailey:
@julen yeah sorry we did some cleanup after the initial landing and introduced the sort messup. @ta2-1 dropped the DISTINCT from the initial query.

@julen not sure from your timings if the slight perf drop is acceptable?

pootle/apps/pootle_project/models.py
```diff
@@ -339,13 +337,16 @@ def resources(self):
...
         cursor.execute(sql_query, [resources_path, resources_path])
-
         results = cursor.fetchall()
 
         # Flatten tuple and sort in a list
-        resources = list(reduce(lambda x,y: x+y, results))
-        resources.sort(key=get_path_sortkey)
+        resources = set()
+        for result in results:
+            s = result[0]
```
julen: What does `s` mean? Can we be more explicit please?

julen:
If it does solve your issue it's an acceptable drop. Posted the numbers FYI — in a better life everyone stops making assumptions...

pootle/apps/pootle_project/models.py
dwaynebailey: You mean compare to list(reduce(rocket(scientist(coder())))) ;)

Its a temporary var that is a string.  Giving it a sensible name like 'path' just makes the second line less readable... trust me I tried.

jleclanche:
Fixed the various issues. The DISTINCT is likely to slow the query down -
did your numbers include it?

(and yes, this is an assumption - I think it's pretty clear when what I say
is based on assumptions vs hard numbers...)

J. Leclanche

2015-02-03 14:50 GMT+01:00 Julen Ruiz Aizpuru <notifications@github.com>:

ta2-1:
@dwaynebailey actually I think we can get rid of the initial query.

Even we can try to compare timings if we don't use SQL directly.
```python
Store.objects.filter(pootle_pah_contains=resources_path).values_list('pootle_path')
Directory.objects.filter(pootle_pah_contains=resources_path).values_list('pootle_path')
```
and add to the set these lists both.

julen:
Speed was acceptable and not a concern before, but now that `DISTINCT` has been removed, I'm getting even better results than the previous implementation.

* cursor.execute 0.153011083603
* set + sort 0.0144200325012
* TOTAL 0.167431116104

pootle/apps/pootle_project/models.py
julen: Nope, actually compare the whole loop to:

```python
resources = list({to_tp_relative_path(result[0]) for result in results})
```

At the same time realize that the implementation for `to_tp_relative_path()` already exists as part of the `Directory.path` property, and it is way more readable and understandable than the `s.find()...` approach proposed in this PR. Will share a follow-up commit soon.

julen: Check this out https://github.com/julen/pootle/commit/32990d94e52d1101b4c6b656740a971ec9f4ffc0

jleclanche: I like it a lot.

J. Leclanche

2015-02-03 23:23 GMT+01:00 Julen Ruiz Aizpuru <notifications@github.com>:

dwaynebailey:
In the quest for some real data I made a [test scripts](https://gist.github.com/dwaynebailey/251737b064b5e0ebb8f4) to test the three options.  Forgive the hackery.

The three tests are:

1. Current SUBSTRING one
2. The proposed one in this PR - though I've reduced the SQL query to eliminate the ``UNION``
3. Tara's idea in the above comment - though I've had to use ``.extra`` as we want a more complex ``LIKE``

The script needs to be run in the Django shell.  I didn't check anything specific about ``list()`` ``set()`` etc but simply the function as it stands.

Some results based on 5 runs on the firefox project (I've tried more runs but they're masked by hot caches) which has a large number of stores:

subtring | pr | taras | comment
-----------|-----|--------|--------------
5.1078710556 | 1.67964100838 | 3.09581613541 | 5 x run from cold
0.0277471542358 | 1.67618608475 | 3.01708602905 | 5 x hot runs
0.0346581935883 | 1.68674588203 | 2.95028781891 | 5 x hot runs
0.00684380531311 | 0.325207948685 | 0.555719137192 | 1 x hot run

Conclusion
--------------
SUBSTRING is great when hot. But when cold its really slow, and since our results are cached in the real app its likely slowest of the lot, ironically.  The PR is fastest.  Taras's examples is slightly more costly about .2s. 

@julen @ta2-1 @jleclanche @unho would appreciate if you could look at my test script (there is always the chance I got like a monkey) but also run this against your databases to confirm that the results are similar.

Thanks gents.

julen:
I wont dive deep into details because it's not worth it, but my observation with the test script aligns with yours in terms of timing: substring > Taras > PR.

As said in the previous comment, this wasn't a real concern with the PR, but actually it's cool that besides moving the code away from being backend-specific, we now have a faster and easier-to-understand implementation.

julen:
Pushed the follow-up in 9dfdd9ff.

pootle/apps/pootle_project/models.py
dwaynebailey: @julen indeed! Nicely done.

PR #3580, opened 2015-02-05
TM results: hide context if unavailable

### LINKED ISSUE #3541: Local TM UI doesn't work for amaGama results
We should integrate local TM and amaGama results. Local TM has access to other information such as the project name and person who translated the suggestion since it has access to the data from Pootle.

When using amaGama results the string in the UI is confusing: 

*Translated by someone in project ""*

screen shot 2015-01-23 at 17 55 13

We should either:

1. Include some of that information in amaGama results e.g. the project name
2. Use a different comment when we have results from amaGama e.g. just remove it

julen:
Please test and report @jleclanche @dwaynebailey.

Fixes #3541.

### DIFF
--- a/pootle/templates/editor/units/xhr_tm.html
+++ b/pootle/templates/editor/units/xhr_tm.html
@@ -9,7 +9,9 @@
           dir="<%= store.source_dir %>"><%= PTL.editor.doDiff(unit.source[0], sugg.source) %></div>
         <div class="suggestion-translation" lang="<%= store.target_lang %>"
           dir="<%= store.target_dir %>"><%= PTL.utils.fancyHl(sugg.target) %></div>
+        <% if (sugg.project) { %>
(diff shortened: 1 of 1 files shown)


jleclanche:
Thanks, will test.

jleclanche:
Seems to work fine, gtm

dwaynebailey:
lgtm.  The message is certainly gone in my results which are from amaGama.  I'll assume the local TM side works.

jleclanche:
@dwaynebailey codacy is sure being annoying. I dont think we should enable it until we're actually clean...

dwaynebailey:
@jleclanche yeah, I've had to message them as I can't seem to disable it.  Can't actually recall enabling it  in fact either.

julen:
I disabled the webhook for the time being; it's just being useless noise.

PR #3582, opened 2015-02-06
Cleanup: Simpler and slightly faster SQL

unho:
According to my numbers the new SQL takes around 0.026 seconds while
the previous one takes 0.028 seconds.

@julen @jleclanche @dwaynebailey @ta2-1 Please review.


julen:
Feel free to merge it.

jleclanche:
lgtm. speedup is insignificant but yes, let's have more of those sql queries simplified.

PR #3584, opened 2015-02-06
Revision: switched to Redis counter

### LINKED ISSUE #3364: Switch revision counter storage to Redis
The current version of revision counter is supported by MyISAM engine only. Since we have Redis as a hard dependency let's move it there.

julen:
The DB model has been replaced by a thin wrapper around the caching API of Django, which in turn we are forcing to be a Redis backend, thus enabling atomic counters.

This also requires switching tests to use the `LocMemCache` backend, because the operations that require the revision to be incremented will fail otherwise — the `DummyCache` is a no-op.

If for some external reason the revision counter has been lost, it's possible to restore it via the `revision --restore` command.

This includes tests but any other manual testing is welcome.
* @ta2-1 please double-check our data migration works fine; it should be trivial.
* @jleclanche @dwaynebailey please make sure the revision is initialized on `initdb` and subsequent translation changes update the revision. You can query it from a shell with `from pootle.core.models import Revision; Revision.get()`.

**NOTE**: having a Redis caching backend is a hard requirement.

Fixes #3364.


dwaynebailey:
From my testing:

- Redis is now a hard dep (I'm working on documenting that) but we probably want to catch if it isn't running and we're migrating.  Running ``./manage.py migrate`` without Redis running causes a [traceback](https://gist.github.com/dwaynebailey/686de94fe6cc234e1d64)
- Before ``migration`` count is zero
- After ``initdb`` count is zero
- Add a translation, edit a translation, fix a check that requires an edit in all cases the revision increase by one.

I didn't check anything in the case of migration from e.g. a 2.6.0 install.  But for a new install it works well and correctly.

julen:
> * Redis is now a hard dep (I'm working on documenting that) but we probably want to catch if it isn't running and we're migrating. Running ./manage.py migrate without Redis running causes a traceback

That is #3351.
 
> I didn't check anything in the case of migration from e.g. a 2.6.0 install. But for a new install it works well and correctly.

The revision has only been available since we spooned into core, therefore the migration (both data and schema) only applies to EN.

julen:
@ta2-1 could you share your input on this please?

unho:
@julen Just reviewed the code, but looks gtm.

ta2-1:
@julen everything works for me. gtm.

PR #3588, opened 2015-02-11
db: REXEXP is not present on PostgreSQL

dwaynebailey:
Please view.

PG uses !~ while the other DBs use NOT REGEXP. I'm not sure if there is a more elegant way of doing this.


jleclanche:
This is an ok solution but I think there is a way to drop the regex entirely from the query. It's so nasty.

unho:
Looks gtm.

dwaynebailey:
@jleclanche can I get a GTM on this.  I also think there is something more elegant, but I don't think that is critical.  I can add FIXMEs to the code if needed, but I'd like to land this.

jleclanche:
gtm but fix the typo please ;)

Issue #3500, opened 2015-01-17
Document clear_stats command


dwaynebailey:
``flush_cache`` was added in 98241580063576d3ed06b01248e955af54b210c5 but not documented.  See also #2927 for description and rationale that could help documentation.

ta2-1:
Current implementation of `flush_cache` command has no sense now. Filled Redis stats cache is a requirement and if we need to refresh stats we use `refresh_stats` command.
I see two options:
- Drop this command;
- Change it for flushing something like Django templates cache.

@julen?

julen:
Now that more named caches are in use, I wouldn't mind adjusting the command to run `cache.clear()` for a given named cache. I'd also be in favor of renaming the command to `clear_cache` (`flush` might result ambiguous).

These would be valid cases:

```python
$ python manage.py clear_cache stats
$ python manage.py clear_cache template_fragments
```

Although the command is pretty simple and anyone can implement it, the point of having it would be being an utility when upgrading deployments.

julen:
Actually as I'm working on #3364, what I was proposing in the previous point cannot be done blindly, because we want to have some data always available on Redis (the revision counter this time), and clearing a cache would wipe out all the existing data.

Having said that, blindly clearing **most** of the caches is fine, but not all of them.

For convenience, the keys that need to be **preserved at any time** will need to go to the same cache (let's say `default`), and we'd need to keep these keys in a whitelist so we'd be able to save-restore them at the time of doing the clear.

Does this make sense @ta2-1? Any other thoughts?

julen:
As a side note and for the specific case of the revision counter, I'm thinking about providing a way to restore it from the DB too.

ta2-1:
Hmm... I thought revision counter shouldn't be saved via django redis cache API.
And can I ask one more time?
Why do we need to keep `clear_cache` command for stats cache?

julen:
> Hmm... I thought revision counter shouldn't be saved via django redis cache API.

Please move any revision-related discussion into its relevant issue.

> And can I ask one more time?

I'm missing any questions in your previous comment.

> Why do we need to keep clear_cache command for stats cache?

I see your point on `clear_cache` being useless for stats as it's a requirement to have it filled in with values (even if they are zero) — what I wrote was just meant to be an example on the use of `clear_cache`, so disregard any actual cache names.

dwaynebailey:
Busy reviewing documentation.  Was about to add doc for this but based on the discussion I won't do that.

[...]

PR #3589, opened 2015-02-12
Redis required for caching and drop memcached

dwaynebailey:
This needs review and I think is good to land as is.  This removes all references and suggestions of using memcached from the code.  It adds all the needed caches and forces the use of Redis caching.  All the depchecks have been adapted or de-memcached as needed.

The only reference to memcached will be in the existing PO files, old release notes and upgrade instructions.

@julen I'd like some clarity on why we're using django-redis-cache instead of django-redis.  I also added hiredis package as that seems to be suggested by the django-redis-cache default setup and comments.

### DIFF
--- a/docs/server/cache.rst
+++ b/docs/server/cache.rst
@@ -51,86 +51,11 @@ You can specify which backend to use by overriding the value of
 :setting:`CACHES` in your configuration file.
 
 
-.. _cache#memcached:
-
-Memcached
-^^^^^^^^^
-
-::
-
-    CACHES = {
-        'default': {
-            'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-            'LOCATION': '127.0.0.1:11211',
-        }
-    }
-
-:ref:`Memcached <django:memcached>` is the **recommended cache backend**, it
-provides the best performance.  And works fine with multiprocessing servers
-like Apache. It requires the `python-memcached` package and a running
-memcached server. Due to extra dependencies it is not enabled by default.
-
-
-.. _cache#memcached_on_unix_sockets:
-
-Memcached on Unix sockets
-"""""""""""""""""""""""""
-
-::
-
-    CACHES = {
-        'default': {
-            'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-            'LOCATION': 'unix:/path/to/memcached.sock',
-        }
-    }
-
-If you don't want Pootle using TCP/IP to access memcached then you can use Unix
-sockets.  This is often a situation in hardened installations using SELinux.
-
-You will need to ensure that memcached is running with the ``-s`` option. ::
-
-    memcached -u nobody -s /path/to/memcached.sock -a 0777
-
-
-.. _cache#database:
-
-Database
-^^^^^^^^
-
-::
-
-    CACHES = {
-        'default': {
-            'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
-            'LOCATION': 'my_cache_table',
-        }
-    }
-
-|Database caching|_ relies on a table in the main Pootle database for storing
-the cached data, which makes it suitable for multiprocessing servers, with the
-added benefit that the cached data remains intact after a server reboot
-(unlike memcached) but it is considerably slower than memcached.
-
-.. versionchanged:: 2.1.1
-
-This is the default cache backend. On new installs and upgrades the required
-database will be created.
-
-Users of older versions need to create the cache tables manually if they would
-like to switch to the database cache backend using this :doc:`management command
-<commands>`::
-
-    $ pootle createcachetable pootlecache
-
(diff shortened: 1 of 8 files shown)


julen:
@dwaynebailey https://github.com/translate/pootle/issues/3351#issuecomment-73963970

docs/server/cache.rst
```diff
@@ -26,19 +26,13 @@ Without a well functioning cache system, Pootle could be slow.
...
-The cache backends are at least configured with a 'default' cache.  If this is
-the only cache that exists then all caching is placed into this cache.  The
-Pootle cache can also be configured to take advantage of certain named caches.
+Pootle is configured with a these named caches:
 
-Current named caches:
-
-- ``'default'`` -- all non specified cache data and all cache data if only one
-  cache is defined.
+- ``'default'`` -- all non specified cache data and all cache data.
```
julen: The revision counter goes here too. This adds the risk of people clearing the default cache and also wiping out the revision counter — even if there's a way to restore it, we might want to prevent removal in the first place. I previously raised this concern in https://github.com/translate/pootle/issues/3500#issuecomment-73204160

docs/server/cache.rst
```diff
@@ -47,81 +41,11 @@ Cache Backends
...
-            'LOCATION': 'unix:/path/to/memcached.sock',
-        }
-    }
-
-If you don't want Pootle using TCP/IP to access memcached then you can use Unix
-sockets.  This is often a situation in hardened installations using SELinux.
+However, Redis is the only cache backend supported by Pootle.  We use some
+custom features of Redis so cannot support other backends. You can customise
+the Redis cache settings by overriding the value of :setting:`CACHES` in your
+configuration file.
```
julen: Might be worth mentioning there's an example configuration in *90-local.conf.sample*.

pootle/apps/pootle_app/views/admin/dashboard.py
```diff
@@ -197,33 +197,25 @@ def optimal_depcheck():
...
-                             "engine.")
-                    optimal.append({'dependency': 'session', 'text': text})
-        else:
+        if not depcheck.test_cache_server_connection():
+            # Server configured but connection failing
             optimal.append({
                 'dependency': 'cache',
-                'text': _("Pootle is configured to use memcached as caching "
-                          "backend, but Python support for memcached is not "
-                          "installed. Caching is currently disabled.")
```
julen: Actually it's not only that caching is impossible, but the overall behavior of the system is undefined. Having no Redis is similar to not having a DB backend.

pootle/settings/20-backends.conf
```diff
@@ -20,28 +20,47 @@ DATABASES = {
...
+        'BACKEND': 'redis_cache.RedisCache',
+        'LOCATION': 'localhost:6379',
         'TIMEOUT': 60,
         'OPTIONS': {
-            'MAX_ENTRIES': 65536,
-            'CULL_FREQUENCY': 16,
-        }
-    }
+            'DB': 1,
+            'PARSER_CLASS': 'redis.connection.HiredisParser',
```
julen: The redis-py parser should work fine, and unless there's a measured bottleneck here I wouldn't worry about doing premature optimizations at the cost of an extra C dependency that needs to be compiled.

requirements/base.txt
```diff
@@ -10,6 +10,7 @@ django-rq
 # Libraries
 cssmin
 diff-match-patch>=20121119
+hiredis
```
julen: See comment on parser.

docs/server/cache.rst
unho: @julen Just a question, is it not possible to have a separate cache for just the revision counter?

pootle/apps/pootle_app/views/admin/dashboard.py
unho: Yes, these checks must be moved to `required_depcheck()` function above.

docs/server/cache.rst
julen: Yes it's possible but it needs a separate `DB` entry, not only a named cache entry. But this is a separate discussion (see previous link), I'm just raising awareness here.

dwaynebailey: @julen I've opened #3592 to track this concern.  I'm just trying to limit scope here but not lose major concerns.

pootle/apps/pootle_app/views/admin/dashboard.py
dwaynebailey: @julen @unho I'll fix these apart as their own commit otherwise there is just too much noise.

dwaynebailey:
@julen @unho I've made the changes required.  Some I've pushed out to other bugs if they're a little large, I'd prefer to handle them there then in this PR as its already growing a bit too large.

julen:
Great, feel free to merge this, we'll address the rest separately.

PR #3593, opened 2015-02-13
Redis: DRY settings for persistent storage

### LINKED ISSUE #3592: Mitigate the risk of revision counter also being in 'default' Redis cache
The revision counter goes into the 'default' cache.  This adds the risk of people clearing the default cache and also wiping out the revision counter — even if there's a way to restore it, we might want to prevent removal in the first place. This concern was raised in https://github.com/translate/pootle/issues/3500#issuecomment-73204160

julen:
Also ensures the revision counter will use the `redis` named cache, which
in turn will also be used for RQ and any other future uses of Redis as a
persistent data store.

Fixes #3592.

### DIFF
--- a/pootle/core/cache.py
+++ b/pootle/core/cache.py
@@ -19,8 +19,13 @@
 # along with this program; if not, see <http://www.gnu.org/licenses/>.
 
 
+from django.conf import settings
 from django.core.cache import caches, cache as default_cache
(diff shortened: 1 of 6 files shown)


ta2-1:
Default value for TIMEOUT is None. But I agree that it isn't worth to rely on default values.
lgtm (except travis alert).

dwaynebailey:
@julen gtm

Issue #3594, opened 2015-02-14
Use Django's System check framework to get rid of depcheck


dwaynebailey:
Django 1.7 introduces the [System check framework](https://docs.djangoproject.com/en/1.7/topics/checks/), this would allow us to elegantly get rid of depcheck.

The Django checks are run with runserver and manage.py commands and would allow us to prevent Pootle running in cases where there is no Redis server, no caching or no rqworker running.  Thus our 'required' dependencies could actually raise real Errors.

jleclanche:
Moved redis version check to this as well.

jleclanche:
Ok well I got bored and ported most remaining checks too.

julen:
This was done in the linked PR.

PR #3595, opened 2015-02-14
Checks: Moved several check to Django 1.7 checks framework

jleclanche:
References #3594


jleclanche:
This is an example of the new checks framework.

If you ran `manage.py check` or `runserver` without redis running, you would get the following:

```
(translate)[3:32:59] adys@azura ~/src/translate/pootle % ./manage.py check        ✗ master:5703ce3 - pootle 
2015-02-14 01:33:00,923 INFO Using Python PO
CommandError: System check identified some issues:
                                                                                                                                                                       
CRITICALS:                                                                                                                                                             
?: (pootle.E001) Could not connect to Redis (Error 111 connecting to localhost:6379. Connection refused.)                                                              
        HINT: Make sure Redis is running on localhost:6379                                                                                                             

System check identified 1 issue (0 silenced).
```

With redis running but no rq workers:

```
System check identified some issues:
                                                                                                                                                                       
WARNINGS:                                                                                                                                                              
?: (pootle.E002) No RQ Worker running.                                                                                                                                 
        HINT: Run new workers with manage.py rqworker                                                                                                                  

System check identified 1 issue (0 silenced).
```

With this on runserver I don't think it's necessary in the admin anymore. I would actually like us to phase out admin checks entirely, but I was thinking doing it for 3.0 instead.

Note that we need to import pootle.checks somewhere. I do it in manage.py but it's not the right place.

dwaynebailey:
@jleclanche a bit distracted?

Yes I was thinking 3.0 also, which was the same reason julen didn't do the same on Redis cache checking.  Its definitely much more elegant.  And I just didn't try one cause it was late.  It does need a better home.  You implementation forgets to do the version check.  I'd go for ``check_xxxx`` instead of ``test_xxx`` to make it clear its a check not a unit test.

jleclanche:
The version check can be tagged as a compatibility check and make it less critical, too.

dwaynebailey:
@jleclanche yeah I read the docs.  Their switched off when using WSGI so that need still needs to be addressed in some way (which might not be in Pootle come to think of it)

unho:
I like the idea of moving critical checks out of Pootle dashboard. I agree on it be named `check_xxxx`.

And I agree the critical checks must be run always, even when using WSGI, to avoid launching the server without the necessary stuff being ready. I also think the other non-critical checks still are good to be reported in some way.

pootle/checks.py
```diff
@@ -0,0 +1,138 @@
...
+        queue = get_queue()
+        workers = Worker.all(queue.connection)
+    except Exception as e:
+        conn_settings = queue.connection.connection_pool.connection_kwargs
+        errors.append(checks.Critical("Could not connect to Redis (%s)" % (e),
+            hint="Make sure Redis is running on %(host)s:%(port)s" % (conn_settings),
+            id="pootle.E001",
+        ))
+    else:
+        if not workers or workers[0].stopped:
```
julen: I believe there needs to be a test to check if the current command being issued is `rqworker`. Otherwise, when running `rqworker` for the first time it'll display a redundant warning, which can be annoying, and more importantly alarming for some admins:
```python
$ ./manage.py rqworker
System check identified some issues:

WARNINGS:
?: (pootle.E002) No RQ Worker running.
	HINT: Run new workers with manage.py rqworker
```

pootle/checks.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if settings.DEBUG:
+        errors.append(checks.Warning("DEBUG mode is on. Do not do this in production!",
+            hint="Set DEBUG = False in settings.py",
+            id="pootle.E009"
+        ))
+    elif "sqlite" in settings.DATABASES.get("default", {}).get("ENGINE"):
+        # We don't bother warning about sqlite in DEBUG mode.
+        errors.append(checks.Warning("The sqlite database backend is unsupported",
+            hint="Set your default database engine to postgresql_psycopg2 in settings.py",
```
julen: `settings.py` is the wrong hint.

jleclanche: Dur, good catch

pootle/checks.py
jleclanche: I noticed that too. Im not sure there's a way to do that... I'll investigate if I can hack something up.

jleclanche:
ready to land

manage.py
```diff
@@ -24,7 +24,7 @@
 
 from django.core.management import execute_from_command_line
 
-from pootle import syspath_override
+from pootle import checks, syspath_override
```
julen: So you were complaining this is nasty... From what I see the cleanest (and suggested) way to register these checks is using an [`AppConfig`](https://docs.djangoproject.com/en/1.7/ref/applications/#django.apps.AppConfig), more specifically its [`ready()` method](https://docs.djangoproject.com/en/1.7/ref/applications/#django.apps.AppConfig.ready). Since 1.7 this is the place to run initialization code like this.

To learn more about this, [see the example from django.contrib.auth](https://github.com/django/django/blob/0ed7d155635da9f79d4dd67e4889087d3673c6da/django/contrib/auth/apps.py).

dwaynebailey:
I notice that these are labeled as E001 E002 etc when a lot are simply wanting so should be W001 etc.

jleclanche:
They're wanting what, @dwaynebailey? :)

julen:
In the context of management commands, the check for RQ worker should only be enabled for some commands, since there are others that make no use of the RQ machinery and the warnings can be misleading.

This is the full list: `['start', 'initdb', 'revision', 'sync_stores', 'run_cherrypy', 'refresh_stats', 'update_stores', 'calculate_checks', 'retry_failed_jobs']`

dwaynebailey:
@jleclanche A warning W is not an error E so some need to change from e.g E001 to W001

jleclanche:
We have decided to drop checks from the dashboard altogether.

- The dashboard is usually as accessible as ssh.
- ssh is probably required to fix anything wrong by the checks warning the users
- We eliminate an additional maintenance burden

jleclanche:
Final review, please.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+
+# Minimum lxml version required for Pootle to run.
+LXML_MINIMUM_REQUIRED_VERSION = (2, 2, 2, 0)
+
+# Minimum Redis server version required.
+# Initially set to some minimums based on:
+# 1. Ubuntu 12.04LTS's version 2.8.4 (10.04LTS was too old for RQ)
+# 2. RQ requires >= 2.6.0, and
+# 3. Wanting to insist on at least the latest stable that devs are using i.e.
+#    2.8.* versions of Redis
```
unho: I don't think we need this detailed explanation here.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+REDIS_MINIMUM_REQUIRED_VERSION = (2, 8, 4)
+
+
+@checks.register()
+def test_library_versions(app_configs, **kwargs):
```
unho: @jleclanche Can we please name this `check_*` to avoid confusing this with actual tests? I see that there several other functions named like this.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+
+@checks.register()
+def test_optional_dependencies(app_configs, **kwargs):
+    errors = []
+
+    try:
+        import Levenshtein
+    except ImportError:
+        errors.append(checks.Warning(_("Can't find python-levenshtein package. Updating against"
+            " templates is faster with python-levenshtein."),
```
unho: Nit: Avoid starting strings with whitespace.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
+from django.core import checks
```
unho: No license notice?

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+    if lxml_version < LXML_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Warning("Your version of lxml is too old.",
+            hint="Try pip install --upgrade lxml",
+            id="pootle.E005",
+        ))
+
+    if ttk_version < TTK_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Critical("Your version of Translate Toolkit is too old.",
+            hint="Try pip install --upgrade translate-toolkit",
+            id="pootle.E006",
```
unho: IMHO these three should be critical, not regular errors. Or at least the Django and TTK ones.

dwaynebailey:
lgtm after minor comments resolved

pootle/checks.py
unho: Let me correct myself. We are using `checks.Critical` for TTK, but the id provided is `pootle.E006` where it should be `pootle.C006`. The other `id` should also be corrected.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+    errors = []
+
+    try:
+        queue = get_queue()
+        workers = Worker.all(queue.connection)
+    except Exception as e:
+        conn_settings = queue.connection.connection_pool.connection_kwargs
+        errors.append(checks.Critical(_("Could not connect to Redis (%s)") % (e),
+            hint=_("Make sure Redis is running on %(host)s:%(port)s") % (conn_settings),
+            id="pootle.E001",
```
unho: Also correct the `id` letter here.

pootle/checks.py
dwaynebailey: I do, so that nobody has to figure out why its an arbitrary number, because it is arbitrary without that justification.  At some future date it might not be needed.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+@checks.register()
+def test_optional_dependencies(app_configs, **kwargs):
+    errors = []
+
+    try:
+        import Levenshtein
+    except ImportError:
+        errors.append(checks.Warning(_("Can't find python-levenshtein package. Updating against"
+            " templates is faster with python-levenshtein."),
+            hint="Try pip install python-levenshtein",
```
unho: Hint not localized.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if lxml_version < LXML_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Warning("Your version of lxml is too old.",
+            hint="Try pip install --upgrade lxml",
+            id="pootle.E005",
+        ))
+
+    if ttk_version < TTK_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Critical("Your version of Translate Toolkit is too old.",
+            hint="Try pip install --upgrade translate-toolkit",
```
unho: Hint not localized. The same with the other two above.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+        ))
+
+    if lxml_version < LXML_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Warning("Your version of lxml is too old.",
+            hint="Try pip install --upgrade lxml",
+            id="pootle.E005",
+        ))
+
+    if ttk_version < TTK_MINIMUM_REQUIRED_VERSION:
+        errors.append(checks.Critical("Your version of Translate Toolkit is too old.",
```
unho: Check message not localized. The same for the two above.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+        errors.append(checks.Critical(_("Could not connect to Redis (%s)") % (e),
+            hint=_("Make sure Redis is running on %(host)s:%(port)s") % (conn_settings),
+            id="pootle.E001",
+        ))
+    else:
+        if not workers or workers[0].stopped:
+            # We need to check we're not running manage.py rqworker right now..
+            import sys
+            if len(sys.argv) > 1 and sys.argv[1] != "rqworker":
+                errors.append(checks.Warning(_("No RQ Worker running."),
```
unho: This was on the required depchecks, why is now only a Warning?

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+                    hint=_("Run new workers with manage.py rqworker"),
+                    id="pootle.W001",
+                ))
+
+        redis_version = queue.connection.info()["redis_version"].split(".")
+        if tuple(int(x) for x in redis_version) < REDIS_MINIMUM_REQUIRED_VERSION:
+            errors.append(checks.Warning(_("Your version of Redis is too old."),
```
unho: This also was on the required depchecks. Why is now only a Warning?

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+                 "Current settings: %r") % (settings.CACHES.get("default")),
+                id="pootle.W004",
+        ))
+    else:
+        from django_redis import get_redis_connection
+
+        if not get_redis_connection():
+            errors.append(checks.Critical(_("Could not initiate a Redis cache connection"),
+                hint=_("Double-check your CACHES settings"),
+                id="pootle.E007",
```
unho: id doesn't match. Should be C.

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@checks.register()
+def test_settings(app_configs, **kwargs):
+    from django.conf import settings
+
+    errors = []
+
+    if "RedisCache" not in settings.CACHES.get("default", {}).get("BACKEND"):
+        errors.append(checks.Warning(_("Cache backend is not set to Redis."),
```
unho: This was on the required depchecks as well. Why is now only a Warning?

pootle/checks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if settings.DEBUG:
+        errors.append(checks.Warning(_("DEBUG mode is on. Do not do this in production!"),
+            hint=_("Set DEBUG = False in Pootle settings"),
+            id="pootle.W005"
+        ))
+    elif "sqlite" in settings.DATABASES.get("default", {}).get("ENGINE"):
+        # We don't bother warning about sqlite in DEBUG mode.
+        errors.append(checks.Warning(_("The sqlite database backend is unsupported"),
+            hint=_("Set your default database engine to postgresql_psycopg2 in Pootle settings"),
```
unho: Why PostgreSQL and not MySQL?

pootle/depcheck.py
```diff
@@ -1,183 +0,0 @@
...
-
-def test_debug():
-    return not settings.DEBUG
-
-
-def test_webserver():
-    """Test that webserver is apache."""
-    return ('apache' in sys.modules or
-            '_apache' in sys.modules or
-            'mod_wsgi' in sys.modules)
```
unho: Maybe this can be replaced with a check to ensure it is not being run using `runserver` which is something we want to avoid.

pootle/apps/pootle_app/views/admin/dashboard.py
```diff
@@ -42,231 +41,6 @@
...
-                     settings.MARKUP_FILTER[0])
-        elif filter_args == 'invalid':
-            text = _("Invalid value '%s' in MARKUP_FILTER. Falling back to "
-                     "HTML.", settings.MARKUP_FILTER[0])
-
-        if text is not None:
-            optional.append({
-                'dependency': filter_args + '-markup',
-                'text': text
-            })
```
unho: We are losing all the markup checking. Please bring it back.

unho:
@jleclanche Review complete. Please address the reported issues.

pootle/checks.py
jleclanche: Because it's not critical.

pootle/checks.py
jleclanche: Because we now support postgres so we might as well recommend it.

pootle/depcheck.py
jleclanche: No, let's not.

pootle/checks.py
jleclanche: This is literally the same string as the one I removed.

pootle/checks.py
jleclanche: `test_*` is the recommended naming.

pootle/checks.py
jleclanche: I don't care about license notices until the proper ones land.

pootle/apps/pootle_app/views/admin/dashboard.py
jleclanche: Check the commit messages please.

pootle/checks.py
julen: The [official docs](https://docs.djangoproject.com/en/1.7/topics/checks/) do not make any recommendations re. naming, and even more, all the examples (and Django's source code) use `check[_]` all the time. Seeing that, @unho's suggestion is reasonable.

julen:
We find the dashboard checks useful for several reasons:
* It confirms in effect that the deployment meets all the required dependencies (I don't see this PR makes any checks while on a WSGI environment). While this should be checked at deployment time, there are chances to miss out on dependencies.
* Not all of our users with admin access know what ssh is, but they can still see and report any potential issues.
* Even if admins know what ssh is, access to the machine is not always possible (VPN etc.), and this allows them also to notice any mismatches.

julen:
I kindly provided a whitelist of commands that require RQ explicitly, which I still find missing here. Without that in place, RQ-related messages will be displayed for commands that don't need RQ, that being both confusing and misleading.

pootle/checks.py
julen: This provides the wrong message IMO, making people think PG is required.

jleclanche: Alright, I'll add mysql to it too.

ta2-1:
Agree with @julen. 
Don't understand why all checks should be dropped.

dwaynebailey:
@jleclanche why license headers? Please don't waste our cycles on something this silly.

pootle/checks.py
dwaynebailey: It is, but its still wrong

pootle/apps/pootle_app/views/admin/dashboard.py
unho: If the current markup depchecks are not suitable to be ported as is, then they should be adjusted. I do agree that maybe not all these checks are worth porting, but we should at least ensure that if `MARKUP_FILTER` tells to use reStructuredText (or Textile or Markdown) the required package is available.

jleclanche:
license headers, mysql, check args and localization issues all fixed.

julen:
The `DEBUG` check is utterly annoying during development, and the check that uses `RQWORKER_WHITELIST` is doing the opposite of what it should be doing.

On the other hand, there are lines above the 84 char soft limit.

Also, I wanna stress again the convenience to have (at least) the critical checks in the dashboard.

dwaynebailey:
@jleclanche the checks in [pootle/core/cache.py](https://github.com/translate/pootle/blob/master/pootle/core/cache.py#L64-L73) should really be migrated into these system checks.  The current Redis cache check is not actually correct and won't catch this issue whose consequences are quite serious.  I've found and pointed out the code in question.

jleclanche:
@julen I agree re the DEBUG check and I want it gone, to be honest. Maybe we should make running in debug mode more visible overall, eg by adding a `DEBUG MODE` warning in the footer (visible on all pages).

Regarding checks in the dashboard, this is a discussion I had with Dwayne and I'm happy to include you in it once he comes back.

@dwaynebailey This isn't so much a check as a runtime error. I guess I can remove the error and add a check for it instead?

iafan:
> We have decided to drop checks from the dashboard altogether.
> 
> The dashboard is usually as accessible as ssh.
> ssh is probably required to fix anything wrong by the checks warning the users
> We eliminate an additional maintenance burden

@jleclanche, @dwaynebailey, I second what Julen have said — we think having a quick 'health' status accessible from Admin UI is very helpful for Pootle admins (which are not the server admins and have no access to SSH).

What I would suggest to do is to at least simply run the 'check' command internally and render its output in the Admin UI (if that command doesn't take much time to execute). I don't care much about the final presentation of the output — we can probably live without the fancy green check marks, but the idea is simple: Pootle admins must have a clear indication if the current running copy is healthy or not, and I bet there is a pretty easy way to implement this on top of the work that you've already done.

dwaynebailey:
@julen @iafan @jleclanche So I've looked through the whole discussion and want to weigh in with what I've found looking at code and taking in other people's views.

1. We took a position at FOSDEM to say no to everything.  This has made us maybe a bit hard to deal with but its made us have to honestly motivate things.  Its lead to some good improvements in how we implemented some features.  So our default on the dashboard is no.
2. I'm still of the position that there are better ways to monitor the setup of a Pootle instance, munin, new relic are all the things that should be checking status, not Pootle.
3. I think however that we have underestimated the SSH aspect.  I have two machines that I've only recently gained SSH access and I see in Mozilla that's not always an easy thing to get.  I would like honestly for someone to be able to give me some heads up when checking something broken in the install when they might not have SSH access or the knowledge on how to setup ``virtualenv`` and run ``./manage.py check``
4. Looking at the code and how we could implement this I actually don't think its a maintenance burden at all.  We'd need to lift code from the actual ``manage.py check`` and intead of outputting to stdout it would create the HTML needed for the dashboard.  The ``check`` code already does that for text output. We only need to parse the list of errors and display them in HTML in the same way as ``check`. The colouring and icons are already there and I suspect we already have the iterator needed.  Unlike the old system, once we have that in place we essentially have any failing check exposed without more effort.  To be honest I think outputting in raw text is more effort than repurposing the dashboard.
 
So my conclusion for a way forward is this.

1. There is a strong enough use case to keep the dashboard
2. We implement it as a new tab or on the old tab
3. We use the existing styling and existing icons to highlight critical errors vs warnings in HTML and not hacking a static page.
4. We base the code on what's in ``./manage.py check``
5. After that if you write a Django check you get the dashboard for free.
6. More serious installs need to actually have monitoring of critical services on which Pootle depends

julen:
@dwaynebailey sounds reasonable to me. This is not meant to replace any monitoring infrastructure by any means (as it wasn't before), and it's just a basic health report, nothing more than that.

jleclanche:
I reimplemented the checks on the dashboard. Looking for a new review please, I'd like this to land asap.

pootle/apps/pootle_app/views/admin/dashboard.py
```diff
@@ -339,13 +113,20 @@ def rq_stats():
     return result
 
 
+def checks():
+    from django.core.checks.registry import registry
+
+    ret = []
```
julen: What about `return sum([check() for check in registry.registered_checks], [])`?

pootle/checks.py
julen: It's critical, this should error.

pootle/checks.py
```diff
@@ -0,0 +1,187 @@
...
+    elif "sqlite" in settings.DATABASES.get("default", {}).get("ENGINE"):
+        # We don't bother warning about sqlite in DEBUG mode.
+        errors.append(checks.Warning(_("The sqlite database backend is unsupported"),
+            hint=_("Set your default database engine to postgresql_psycopg2 or mysql"),
+            id="pootle.W006",
+        ))
+
+    if settings.SESSION_ENGINE.split(".")[-1] not in ("cache", "cached_db"):
+        errors.append(checks.Warning(_("Not using cached_db as session engine"),
+            hint=_("Set SESSION_ENGINE to django.contrib.sessions.backend.cached_db\n" +
```
julen: `+` is unnecessary to continue strings within parenthesized expressions.

pootle/templates/admin/dashboard.html
```diff
@@ -57,30 +57,19 @@
...
   </div>
   <div class="bd">
-    <ul id="required" class="icons">
-      {% for dep in required %}
-      <li class="{{ dep.dependency }}"><i class="icon-{{ dep.state }}"></i> {{ dep.text }}</li>
-      {% endfor %}
+    <ul>
+    {% for check in checks %}
+      <li>
+        <i class="icon-{% if check.is_serious %}reject{% else %}warning{% endif %}"></i>
```
julen: Just wondering: when *serious* checks fail, is there a chance we ever see this UI? If so, you can use `icon-error` as it was before.

pootle/templates/admin/dashboard.html
```diff
@@ -57,30 +57,19 @@
...
+        {{ check.msg }} <em>{{ check.hint }}</em>
+      </li>
+    {% empty %}
+      <li><i class="icon-accept"></i> {% trans "No issue detected." %}</li>
```
julen: Let's use `icon-tick` as before.

Also, *issues*.
Anyhow, being more verbose can reassure the system checks passed: *No issues have been detected. All is good!*

julen:
Few more notes on the current state:
* Please review string continuations — they don't contain any whitespace at the end of the line before the continuation.
* Please mind the line length. Most of them can be avoided by [appending check failures like in Django](https://github.com/django/django/blob/stable/1.7.x/django/contrib/auth/checks.py)
* New license headers are missing

pootle/apps/pootle_app/views/admin/dashboard.py
unho: @jleclanche Any input on this?

pootle/checks.py
unho: @jleclanche This nit still needs to be fixed.

pootle/checks.py
unho: @jleclanche This hint should be marked as translatable.

pootle/checks.py
unho: @jleclanche Any reply to this question?

pootle/apps/pootle_app/views/admin/dashboard.py
unho: @julen Is not enough with `return sum([check() for check in registry.registered_checks])`?

julen:
Oh, forgot to say: please also add `SILENCED_SYSTEM_CHECKS = ['pootle.W005']` in the default settings to suppress warnings for `settings.DEBUG` when using runserver during development.

pootle/depcheck.py
unho: ok

pootle/apps/pootle_app/views/admin/dashboard.py
unho: @jleclanche Any input on the suggested approach?

pootle/apps/pootle_app/views/admin/dashboard.py
julen: No, you need to flatten lists (sum lists), not sum the actual values of the lists themselves.

jleclanche: I wanted to be a bit more verbose but ok.

pootle/templates/admin/dashboard.html
jleclanche: Serious checks failing usually means that page will 500 anyway.

jleclanche:
Nits and comments fixed, looking for a greenlight.

pootle/checks.py
```diff
@@ -0,0 +1,176 @@
...
+        # We don't bother warning about sqlite in DEBUG mode.
+        errors.append(checks.Warning(_("The sqlite database backend is unsupported"),
+            hint=_("Set your default database engine to postgresql_psycopg2 or mysql"),
+            id="pootle.W006",
+        ))
+
+    if settings.SESSION_ENGINE.split(".")[-1] not in ("cache", "cached_db"):
+        errors.append(checks.Warning(_("Not using cached_db as session engine"),
+            hint=_("Set SESSION_ENGINE to django.contrib.sessions.backend.cached_db\n"
+                 "Current settings: %r") % (setting.SESSION_ENGINE),
```
unho: @jleclanche Please align the content as told in http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#continuing-long-statements

unho:
@jleclanche I still see several issues that need to be fixed. Check all the comments in the PR.

pootle/apps/pootle_app/__init__.py
```diff
@@ -0,0 +1 @@
+default_app_config = "pootle_app.apps.PootleConfig"
```
unho: @jleclanche New license header missing.

jleclanche: No.

pootle/checks.py
```diff
@@ -0,0 +1,176 @@
...
+# Maybe tagging can improve this?
+RQWORKER_WHITELIST = [
+    "start", "initdb", "revision", "sync_stores", "run_cherrypy",
+    "refresh_stats", "update_stores", "calculate_checks", "retry_failed_jobs"
+]
+
+
+
+@checks.register()
+def test_library_versions(app_configs=None, **kwargs):
```
unho: @jleclanche I still think all these functions must be named `check_xxxx`. They are used for checks, not for tests.

pootle/settings/90-dev-local.conf.sample
```diff
@@ -248,3 +248,6 @@ MT_BACKENDS = [
 
 # URL for the amaGama TM server.
 AMAGAMA_URL = 'https://amagama-live.translatehouse.org/api/v1/'
+
+# Silence the DEBUG check on dev servers
+SILENCED_SYSTEM_CHECKS = ["pootle.W005"]
```
julen: As I mentioned when I proposed this, I think the real value is silencing this by default, i.e. in any other conf file and not this sample file.

julen:
GTM after fixing line lengths.

pootle/checks.py
jleclanche: As I said three times now, critical checks are for stuff where we don't even want to allow pootle to run with that setup. "Outdated redis" is not part of that.

pootle/apps/pootle_app/views/admin/dashboard.py
jleclanche: Repeating myself here for clarity: The markup checks are extremely unclear, at least to me. I brought them up before and got no reply. We can bring them back later once we know what we want to do with them, but this is completely unproductive and they aren't blockers, so let's land without them and figure it out later.

pootle/settings/90-dev-local.conf.sample
jleclanche: After discussion on IRC, we decided to leave it like this and consider changing it or removing it if people complain about it.

dwaynebailey:
Nice seeing this landed

PR #3600, opened 2015-02-23
Initial virtual folders

unho:
Ready to review.

This is meant to somehow recover our old goals functionality, but using the new vfolders.

This provides a management command to create and update virtual folders, adds the ability to sort the units by (vfolder) priority on the editor, and injects the priority sorting into the translate URLs in the translation project overview.

Things that differ from specs:

- We don't use any filtering rules. Instead we provide lists of filenames relative to vfolder location.

### DIFF
--- a/docs/features/virtual_folders.rst
+++ b/docs/features/virtual_folders.rst
@@ -36,7 +36,8 @@ The priority defaults to ``1`` and accepts any value greater than ``0``,
 including numbers with decimals, like ``0.75``. Higher numbers means higher
 priority.
 
-By default virtual folders can be browsed.
+By default virtual folders can be browsed. If they are not browsable then they
+won't be displayed, but they are still used for sorting.
 
 Also the virtual folders can have a description which might be useful to
 explain the contents of the folder or provide additional instructions. This
@@ -60,6 +61,18 @@ This command imports a JSON file holding a list of virtual folders, and the
 files included on each virtual folder along with all their attributes. Check
 the specs for the :ref:`JSON format <virtual_folders#json-format>` in order to
 know how to craft a JSON file that fits your needs.
+
+
+.. _virtual_folders#translate:
+
+Translating virtual folders
+---------------------------
+
+If a virtual folder applies in the current location, then clicking on the links
+on the overview page will provide the units in priority order when translating
+in the editor. The priority sorting on the translation editor is calculated
+taking into account all the applicable virtual folders in the current location,
+including the not browsable ones.
 
 
 .. _virtual_folders#json-format:
--- a/docs/features/virtual_folders.rst
+++ b/docs/features/virtual_folders.rst
@@ -46,3 +46,34 @@ The filtering rules specify which translation units are included within a
 virtual folder. Currently the only supported filtering rule consists of a list
 of file paths relative to the virtual folder location. It is possible to not
 specify any filtering rule.
+
+
+.. _virtual_folders#apply:
+
+Adding and updating virtual folders
+-----------------------------------
+
+To add or modify the properties of virtual folders use the
+:ref:`add_vfolders <commands#add-virtual-folders>` management command.
+
+This command imports a JSON file holding a list of virtual folders, and the
+files included on each virtual folder along with all their attributes. Check
+the specs for the :ref:`JSON format <virtual_folders#json-format>` in order to
+know how to craft a JSON file that fits your needs.
+
+
+.. _virtual_folders#json-format:
+
+Format for the JSON file
+------------------------
+
+The JSON file used to import virtual folders consists of a list of virtual
+folder definitions with the :ref:`same fields <virtual_folders#attributes>` as
+the virtual folders, except for two differences:
+
+- If the **description** includes newlines those must be escaped.
+
+The following example depicts a basic JSON file:
+
+.. literalinclude:: virtual_folders.json
+   :language: json
--- a/docs/features/virtual_folders.rst
+++ b/docs/features/virtual_folders.rst
@@ -1,0 +1,48 @@
+.. _virtual_folders:
+
+Virtual Folders
+===============
+
+.. versionadded:: 2.7
+
+Virtual folders provide a way to group translations based on any criteria,
+including a file across all the languages in a project, or files on specific
+locations. Virtual folders have priority, so they can be used to allow
+translators to focus on the most important work.
+
+
+.. _virtual_folders#attributes:
+
+Virtual folders' attributes
+---------------------------
+
+Virtual folders have several attributes:
+
+- A mandatory lowercase name,
+- A mandatory location,
+- An optional priority,
+- An optional browsability flag,
+- An optional description,
+- A field accepting several optional filtering rules.
+
+The location indicates the root place where the virtual folder applies. It can
+use placeholders for language (``{LANG}``) and project (``{PROJ}``).
+
+Each virtual folder must have a unique combination of name and location. This
+means that there can exist two different virtual folders with the same name if
+they have different locations.
+
+The priority defaults to ``1`` and accepts any value greater than ``0``,
+including numbers with decimals, like ``0.75``. Higher numbers means higher
+priority.
+
+By default virtual folders can be browsed.
+
+Also the virtual folders can have a description which might be useful to
+explain the contents of the folder or provide additional instructions. This
+might be handy when using the virtual folders as goals.
+
+The filtering rules specify which translation units are included within a
+virtual folder. Currently the only supported filtering rule consists of a list
+of file paths relative to the virtual folder location. It is possible to not
+specify any filtering rule.
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -346,6 +346,36 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
 against.
 
 
+.. _commands#vfolders:
+
+Virtual Folders
+---------------
+
+These commands allow you to perform tasks with virtual folders from the command
+line.
+
+
+.. _commands#add-virtual-folders:
+
+add_vfolders
+^^^^^^^^^^^^
+
+.. versionadded:: 2.7.0
+
+This command allows you to create :ref:`virtual folders <virtual_folders>` from
+a JSON file. If the specified virtual folders already exist then they are
+updated with the provided data, but only if the data differs.
+
+Check the specs for the :ref:`JSON format <virtual_folders#json-format>` in
+order to know how to craft a JSON file that fits your needs.
+
+This command requires a mandatory filename argument.
+
+.. code-block:: bash
+
+    $ pootle add_vfolders virtual_folders.json
+
+
 .. _commands#manually_installing_pootle:
 
 Manually Installing Pootle
--- a/pootle/apps/virtualfolder/management/commands/add_vfolders.py
+++ b/pootle/apps/virtualfolder/management/commands/add_vfolders.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2014, 2015 Zuza Software Foundation
+#
+# This file is part of Pootle.
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 2 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, see <http://www.gnu.org/licenses/>.
+
+import json
+import logging
+import os
+from optparse import make_option
+
+# This must be run before importing Django.
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from django.core.exceptions import ValidationError
+from django.core.management.base import BaseCommand, CommandError
+
+from virtualfolder.models import VirtualFolder
+
+
+class Command(BaseCommand):
+    help = "Add virtual folders from file."
+
+    def handle(self, *args, **options):
+        """Add virtual folders from file."""
+
+        if not args:
+            raise CommandError("You forgot to provide the mandatory filename.")
+
+        try:
+            inputfile = open(args[0], "r")
+            vfolders = json.load(inputfile)
+            inputfile.close()
+        except IOError as e:
+            raise CommandError(e)
+        except ValueError as e:
+            raise CommandError("Please check if the JSON file is malformed. "
+                               "Original error:\n%s" %e)
+
+        added_count = 0
+        updated_count = 0
+        errored_count = 0
+
+        for vfolder_item in vfolders:
+            vfolder_item['name'] = vfolder_item['name'].lower()
+
+            # Put all the files for each virtual folder as a list and save it
+            # as its filter rules.
+            try:
+                vfolder_item['filter_rules'] = ','.join(vfolder_item['filters']['files'])
+            except KeyError:
+                vfolder_item['filter_rules'] = ''
+
+            if 'filters' in vfolder_item:
+                del vfolder_item['filters']
+
+            # Now create or update the virtual folder.
+            try:
+                # Retrieve the virtual folder if it exists.
+                vfolder = VirtualFolder.objects.get(
+                    name=vfolder_item['name'],
+                    location=vfolder_item['location'],
+                )
+            except VirtualFolder.DoesNotExist:
+                # If the virtual folder doesn't exist yet then create it.
+                try:
+                    vfolder = VirtualFolder(**vfolder_item)
+                    vfolder.save()
+                except ValidationError as e:
+                    errored_count += 1
+                    logging.error(e.message)
+                else:
+                    added_count += 1
+            else:
+                # Update the already existing virtual folder.
+                changed = False
+
+                if vfolder.filter_rules != vfolder_item['filter_rules']:
+                    vfolder.filter_rules = vfolder_item['filter_rules']
+                    changed = True
+                    logging.info("Filter rules for virtual folder '%s' will "
+                                 "be changed.", vfolder.name)
+
+                if ('priority' in vfolder_item and
+                    vfolder.priority != vfolder_item['priority']):
+
+                    vfolder.priority = vfolder_item['priority']
+                    changed = True
+                    logging.info("Priority for virtual folder '%s' will be "
+                                 "changed to %f.", vfolder.name,
+                                 vfolder.priority)
+
+                if ('is_browsable' in vfolder_item and
+                    vfolder.is_browsable != vfolder_item['is_browsable']):
+
+                    vfolder.is_browsable = vfolder_item['is_browsable']
+                    changed = True
+                    logging.info("is_browsable status for virtual folder '%s' "
+                                 "will be changed.", vfolder.name)
+
+                if ('description' in vfolder_item and
+                    vfolder.description.raw != vfolder_item['description']):
+
+                    vfolder.description = vfolder_item['description']
+                    changed = True
+                    logging.info("Description for virtual folder '%s' will be "
+                                 "changed.", vfolder.name)
+
+                if changed:
+                    try:
+                        vfolder.save()
+                    except ValidationError as e:
+                        errored_count += 1
+                        logging.error(e.message)
+                    else:
+                        updated_count += 1
+
+        logging.info("\nErrored: %d\nAdded: %d\nUpdated: %d\nUnchanged: %d",
+                     errored_count, added_count, updated_count,
+                     len(vfolders)-errored_count-added_count-updated_count)
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -71,6 +71,18 @@ class VirtualFolder(models.Model):
         unique_together = ('name', 'location')
         ordering = ['-priority', 'name']
 
+    @classmethod
+    def get_matching_for(cls, pootle_path):
+        """Return the matching virtual folders in the given pootle path.
+
+        Not all the applicable virtual folders have matching filtering rules.
+        This method further restricts the list of applicable virtual folders to
+        retrieve only those with filtering rules that actually match.
+        """
+        return VirtualFolder.objects.filter(
+            units__store__pootle_path__startswith=pootle_path
+        ).distinct()
+
     def __unicode__(self):
         return ": ".join([self.name, self.location])
 
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright 2015 Zuza Software Foundation
+#
+# This file is part of Pootle.
+#
+# Pootle is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <http://www.gnu.org/licenses/>.
+
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from pootle.core.markup import get_markup_filter_name, MarkupField
+from pootle_language.models import Language
+from pootle_project.models import Project
+from pootle_store.models import Store, Unit
+
+
+class VirtualFolder(models.Model):
+
+    name = models.CharField(_('Name'), blank=False, max_length=70)
+    location = models.CharField(
+        _('Location'),
+        blank=False,
+        max_length=255,
+        help_text=_('Root path where this virtual folder is applied.'),
+    )
+    filter_rules = models.TextField(
+        # Translators: This is a noun.
+        _('Filter'),
+        blank=False,
+        help_text=_('Filtering rules that tell which stores this virtual '
+                    'folder comprises.'),
+    )
+    priority = models.FloatField(
+        _('Priority'),
+        default=1,
+        help_text=_('Number specifying importance. Greater priority means it '
+                    'is more important.'),
+    )
+    is_browsable = models.BooleanField(
+        _('Is browsable?'),
+        default=True,
+        help_text=_('Whether this virtual folder is active or not.'),
+    )
+    description = MarkupField(
+        _('Description'),
+        blank=True,
+        help_text=_('Use this to provide more information or instructions. '
+                    'Allowed markup: %s', get_markup_filter_name()),
+    )
+    units = models.ManyToManyField(
+        Unit,
+        db_index=True,
+        related_name='vfolders',
+    )
+
+    class Meta:
+        unique_together = ('name', 'location')
+        ordering = ['-priority', 'name']
+
+    def __unicode__(self):
+        return ": ".join([self.name, self.location])
+
+    def save(self, *args, **kwargs):
+        # Force validation of fields.
+        self.clean_fields()
+
+        self.name = self.name.lower()
+
+        super(VirtualFolder, self).save(*args, **kwargs)
+
+        # Clean any existing relationship between units and this vfolder.
+        self.units.clear()
+
+        # Recreate relationships between this vfolder and units.
+        for location in self.get_all_pootle_paths():
+            for filename in self.filter_rules.split(","):
+                vf_file = "".join([location, filename])
+
+                qs = Store.objects.filter(pootle_path=vf_file)
+
+                if qs.exists():
+                    self.units.add(*qs[0].units.all())
+
+    def clean_fields(self):
+        """Validate virtual folder fields."""
+        if not self.priority > 0:
+            raise ValidationError(u'Priority must be greater than zero.')
+
+        elif self.location == "/":
+            raise ValidationError(u'The "/" location is not allowed. Use '
+                                  u'"/{LANG}/{PROJ}/" instead.')
+
+    def get_all_pootle_paths(self):
+        """Return a list with all the locations this virtual folder applies.
+
+        If the virtual folder location has no {LANG} nor {PROJ} placeholders
+        then the list only contains its location. If any of the placeholders is
+        present, then they get expanded to match all the existing languages and
+        projects.
+        """
+        # Locations like /project/<my_proj>/ are not handled correctly. So
+        # rewrite them.
+        if self.location.startswith("/projects/"):
+            self.location = self.location.replace("/projects/", "/{LANG}/")
+
+        if "{LANG}" in self.location and "{PROJ}" in self.location:
+            locations = []
+            for lang in Language.objects.all():
+                temp = self.location.replace("{LANG}", lang.code)
+                for proj in Project.objects.all():
+                    locations.append(temp.replace("{PROJ}", proj.code))
+            return locations
+        elif "{LANG}" in self.location:
+            try:
+                project = Project.objects.get(code=self.location.split("/")[2])
+                languages = project.languages.iterator()
+            except:
+                languages = Language.objects.iterator()
+
+            return [self.location.replace("{LANG}", lang.code)
+                    for lang in languages]
+        elif "{PROJ}" in self.location:
+            try:
+                projects = Project.objects.filter(
+                    translationproject__language__code=self.location.split("/")[1]
+                ).iterator()
+            except:
+                projects = Project.objects.iterator()
+
+            return [self.location.replace("{PROJ}", proj.code)
+                    for proj in projects]
+
+        return [self.location]
--- a/pootle/core/browser.py
+++ b/pootle/core/browser.py
@@ -21,6 +21,8 @@
 
 from django.utils.translation import ugettext_lazy as _
 
+from virtualfolder.models import VirtualFolder
+
 
 HEADING_CHOICES = [
     {
@@ -85,14 +87,14 @@ def get_table_headings(choices):
     return filter(lambda x: x['id'] in choices, HEADING_CHOICES)
 
 
-def make_generic_item(path_obj):
+def make_generic_item(path_obj, **kwargs):
     """Template variables for each row in the table."""
     return {
         'href': path_obj.get_absolute_url(),
         'href_all': path_obj.get_translate_url(),
-        'href_todo': path_obj.get_translate_url(state='incomplete'),
-        'href_sugg': path_obj.get_translate_url(state='suggestions'),
-        'href_critical': path_obj.get_critical_url(),
+        'href_todo': path_obj.get_translate_url(state='incomplete', **kwargs),
+        'href_sugg': path_obj.get_translate_url(state='suggestions', **kwargs),
+        'href_critical': path_obj.get_critical_url(**kwargs),
         'title': path_obj.name,
         'code': path_obj.code,
         'is_disabled': getattr(path_obj, 'disabled', False),
@@ -100,7 +102,13 @@ def make_generic_item(path_obj):
 
 
 def make_directory_item(directory):
-    item = make_generic_item(directory)
+    filters = {}
+
+    if VirtualFolder.get_matching_for(directory.pootle_path).count():
+        # The directory has virtual folders, so append priority sorting to URL.
+        filters['sort'] = 'priority'
+
+    item = make_generic_item(directory, **filters)
     item.update({
         'icon': 'folder',
     })
--- a/pootle/core/helpers.py
+++ b/pootle/core/helpers.py
@@ -27,8 +27,9 @@ from pootle_app.models.permissions import check_permission
 from pootle_misc.checks import check_names, get_qualitycheck_schema
 from pootle_misc.forms import make_search_form
 from pootle_misc.stats import get_translation_states
-from pootle_store.models import Unit
+from pootle_store.models import Store, Unit
 from pootle_store.views import get_step_query
+from virtualfolder.models import VirtualFolder
 
 from .url_helpers import get_path_parts, get_previous_url
 
@@ -154,9 +155,17 @@ def get_overview_context(request):
     resource_obj = request.resource_obj
     resource_path = getattr(request, 'resource_path', '')
 
-    url_action_continue = resource_obj.get_translate_url(state='incomplete')
-    url_action_fixcritical = resource_obj.get_critical_url()
-    url_action_review = resource_obj.get_translate_url(state='suggestions')
+    filters = {}
+
+    if (not isinstance(resource_obj, Store) and
+        VirtualFolder.get_matching_for(request.pootle_path).count()):
+        filters['sort'] = 'priority'
+
+    url_action_continue = resource_obj.get_translate_url(state='incomplete',
+                                                         **filters)
+    url_action_fixcritical = resource_obj.get_critical_url(**filters)
+    url_action_review = resource_obj.get_translate_url(state='suggestions',
+                                                       **filters)
     url_action_view_all = resource_obj.get_translate_url(state='all')
 
     return {
--- a/pootle/templates/editor/_toolbar.html
+++ b/pootle/templates/editor/_toolbar.html
@@ -4,6 +4,7 @@
   <div class="unit-sort">
     <select id="js-filter-sort">
       <option selected value="default">{% trans "Default order" %}</option>
+      <option value="priority">{% trans "Decreasing priority" %}</option>
       <option value="newest">{% trans "Newest first" %}</option>
       <option value="oldest">{% trans "Oldest first" %}</option>
     </select>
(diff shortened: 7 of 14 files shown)


dwaynebailey:
@unho I thought we were going to change to use floats for the priority to allow 0.N<1(default)<N

To reviewers note:
There is no vfolder UI.  The translators are directed by priority if they choose to translate, address checks or reviews.  UI will come later but for now we can direct translators to work on the top priority vfolders.

@julen @iafan this needs your input I think especially since we now pass ordering info in the URLs.

unho:
@dwaynebailey If I remember correctly we discussed which changes would be necessary in case we actually decided to do such change. We didn't agree on doing such change, and I strongly disagree.

julen:
Any pointers for testing?

Can we have some tests as well, please?

pootle/templates/editor/_toolbar.html
```diff
@@ -4,6 +4,7 @@
   <div class="unit-sort">
     <select id="js-filter-sort">
       <option selected value="default">{% trans "Default order" %}</option>
+      <option value="priority">{% trans "Decreasing priority" %}</option>
```
julen: Does this mean *higher priority*?

pootle/core/browser.py
```diff
@@ -100,7 +100,12 @@ def make_generic_item(path_obj):
 
 
 def make_directory_item(directory):
-    item = make_generic_item(directory)
+    filters = {}
+
+    if len(VirtualFolder.get_matching_for(directory.pootle_path)):
```
julen: As I see this will incur in N*M+1 queries, where:
* N is the number of directories being displayed in the overview,
* M is the number of filtering rules in each browsable `VirtualFolder`, and
* 1 is the query needed for retrieving browsable `VirtualFolder`s

Looks excessive on a first sight, even more taking into account there's no caching of any sort.

pootle/templates/editor/_toolbar.html
unho: It means top priority on the top which decreases as we approach to the bottom.

dwaynebailey:
@unho I assume you disagree with float for priority?  Right, so this came up in the F2F.  We discussed why it was needed because of 0.N level priorities.  You said we can't do 0.N because we're using ints.  @jleclanche said we could simply use floats to achieve that.  So what are we doing to achieve 0.N type priorities? Apart from that you said nothing about disagreeing, especially since I'd articulated why it was needed. You really do need to make yourself heard, especially if we're doing a F2F.

unho:
@dwaynebailey I strongly disagree with using floats. According to https://docs.djangoproject.com/en/1.7/ref/models/fields/#positiveintegerfield we have a range of positive integers between 0 and 2147483647 to specify the priority. I see no need to use floats, that according to your comments on the specs doc can confuse people that are not good at maths. Reading the specs, the use of floats comes from some use of priority for scoring, but IMHO that can be achieved by multiplying the priority by some multiplier like 0.1 (yes, a float).

dwaynebailey:
@unho how are we addressing goals that have a priority of below normal?

In most projects you would have:

1. Some high priority goals
2. Units/Stores that have no special priority set at all.
3. Some items you want to be done last

pootle/core/browser.py
unho: This has been replaced by a new call that triggers only 1 query.

Regarding the cache, the plan is to get something that works, and then optimize it using cache.

unho:
@dwaynebailey I have been working on altering this to allow priorities between 0 and 1 as told on the specs https://docs.google.com/document/d/16g_ustDX2Pxm8yasqIw_kXXi2g1oU25iOnjOBCYcYWc

@julen I have looking at adding tests, but running py.test with InnoDB is proving not to be so easy. So since this looks like it is going to be time consuming, I am going to leave tests for this feature out for now.

julen:
@unho if you are running into trouble when running tests I think it's wiser to come and ask for help instead of disregarding tests altogether.

unho:
Cleaned the PR. Ready to review again.

It still has no tests. I might look into that tomorrow.

I am working on a list of steps to locally verify that this works. Maybe tomorrow.

unho:
In order to check this locally you will have to follow these steps:

- Ensure you have a project with several files on it (per language). It is recommended to have enough files to include some of them in a vfolder while leaving some of them out
- Run migrations
- Craft a JSON file to add some sample vfolders. Example in https://github.com/translate/pootle/pull/3600/files#diff-32d9c83230475a8ff37b1695b1606581R386 It is recommended to have a directory with 4 files at least, and include 3 of them on vfolders (one vfolder with 3 files, and the other two vfolders with 2 files, sharing only one of them).
- Run `add_vfolders` management command to import such vfolders. Example in https://github.com/translate/pootle/pull/3600/files#diff-32d9c83230475a8ff37b1695b1606581R456
- Run the server and go to a path where we know there is some of the files matched by a vfolder
- Check that the translate links above the stats table have the `sort=priority` bit on the URLs
- Check that the translate links on the stats table for the files or directories with files included in a vfolder also include that `sort=priority` bit on the URLs
- Click on any of those translate links and ensure the units are sorted by priority (since we are currently including the units on vfolders using the stores, we just need to ensure that the vfolders are in the appropriate order).

unho:
@dwaynebailey @iafan Please review the PR (or rereview it).

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
+.. _virtual_folders:
+
+Virtual Folders
+===============
+
+.. versionadded:: 2.7
+
+Virtual folders provide a way to group translations based on any criteria,
+including a same file across all the languages in a project, or specific files.
```
julen: What does *a same file* mean here? Can you maybe reword it?

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+Virtual folders have several attributes:
+
+- A mandatory lowercase name,
+- A mandatory location,
+- An optional priority,
+- An optional browsability flag,
+- An optional description,
+- An optional filtering rule.
+
+Each virtual folder must have a different name and location pair. This means
```
julen: Do you mean that name + location must be unique? or that name and location must differ?

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+that there can exist two different virtual folders with the same if they have
```
julen: *name* is missing after *with the same*.

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+different locations.
+
+The priority defaults to ``1`` and accepts any value greater than ``0``,
+including numbers with decimals, like ``0.75``.
+
+By default the virtual folders are not browsable. This means that are not
```
julen: What about: *By default virtual folders cannot be browsed: they won't be displayed and it won't be possible to use them for sorting.*

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+displayed and can't be used for sorting.
+
+Also the virtual folders can have a description which might be useful to tell
+the translators which is the grouping criteria or provide additional
+instructions. This might be handy when using the virtual folders as goals.
```
julen: I don't believe you need to use the article as in *the virtual folders*. @dwaynebailey?

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+
+The filtering rule tells which translation units are included within a virtual
```
julen: What about: *The filtering rule specifies*

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+folder. It is meant to be used internally in order to calculate the units that
+are included within the virtual folder. Currently it consists on a list of file
```
julen: *consists of*

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+
+By default the virtual folders are not browsable. This means that are not
+displayed and can't be used for sorting.
+
+Also the virtual folders can have a description which might be useful to tell
+the translators which is the grouping criteria or provide additional
+instructions. This might be handy when using the virtual folders as goals.
+
+The filtering rule tells which translation units are included within a virtual
+folder. It is meant to be used internally in order to calculate the units that
```
julen: This sentence doesn't seem to add any extra information?

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. _virtual_folders#translate:
+
+Translating virtual folders
+---------------------------
+
+If any virtual folder applies in the current location, the links on the
+overview will automatically point to the translation editor providing the
+priority sorting flag. The priority sorting on the translation editor is
+calculated having in account all the applicable virtual folders in the current
```
julen: *taking into account*

docs/server/commands.rst
```diff
@@ -346,6 +346,129 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
...
+add_vfolders
+^^^^^^^^^^^^
+
+.. versionadded:: 2.7.0
+
+This command allows you to create **virtual folders** from a JSON file. If the
+specified virtual folders already exist and the data differs, then they are
+updated with the provided data.
+
+The JSON file consists in a list of virtual folder definitions with the
```
julen: *consists of*

docs/server/commands.rst
```diff
@@ -346,6 +346,129 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
...
+following fields:
```
julen: Can we split the expected format docs away from the command reference and put it in its own section? Since there's already a *docs/features/virtual_folders.rst* file and nothing links to the format spec, it's rather surprising to find this information along the commands.

docs/server/commands.rst
```diff
@@ -346,6 +346,129 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
...
+
+- A mandatory lowercase name for the virtual folder,
```
julen: For reader's parsing ease, you'll want to backtick-quote the attribute names or make a definition list out of them.

docs/features/virtual_folders.rst
unho: The former one. Rewritten.

docs/features/virtual_folders.rst
unho: Fixed. Thanks!

docs/features/virtual_folders.rst
unho: It is better. Picked it. Thanks.

docs/features/virtual_folders.rst
unho: Maybe I am wrong, but I think it is ok.

docs/features/virtual_folders.rst
unho: Better. Picked. Thanks.

docs/features/virtual_folders.rst
unho: Think in galician, write in english, read @julen comments. :)

Fixed. Thanks!

docs/features/virtual_folders.rst
unho: Fixed. Thanks.

docs/server/commands.rst
unho: Fixed. Thanks!

docs/server/commands.rst
unho: Tried backticking, but I ended up using bold. I think it is more readable. Thanks!

docs/server/commands.rst
unho: @julen Since this JSON file is tied to the command, I think it makes more sense for it to be here.

docs/features/virtual_folders.rst
unho: It means that a file named `file.po` is matched in all languages in a project. For example `/af/proj/file.po` or `/eu/proj/file.po`.

Rephrased to avoid confusion.

docs/features/virtual_folders.rst
unho: Removed. Thanks!

docs/server/commands.rst
unho: @julen What do you mean with "format spec"? Format for this JSON file?

unho:
Fixed all the nits except the one about where the JSON example is included in the docs.

unho:
@julen @dwaynebailey @iafan @jleclanche More reviews and gtms are welcome.

dwaynebailey:
The following are results from my testing.  I setup a simple test with a files containing one each: translated, untranslated, fuzzy and obsolete entries.  Placed in directories a, b and c.

- You get a traceback if you don't supply a gaols file, I'd hazard that if it isn't JSON you'd also get a problem.
- Sorting is busted in that we're sorting on priority and state i.e high priority untranslated, then high priority fuzzy, then normal untranslated, normal fuzzy, low, translated low fuzzy.
- When at http://localhost:8000/projects/firefox2/browser/ i.e. project overview there are no priority links to the languages when I've filtered to a goal location.
- Setup three dirs a, b and c. Setup goals priority of c=2.0, b=0.5.  Correct ordering would then be c, a, b.  Yet when stepping through it goes c, b, a

As suggestion to make testing easier: show priority and goal that applies to the unit on the editor.  Then its easy for a tester to evaluate why a unit is being displayed.

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Virtual folders provide a way to group translations based on any criteria,
+including a file across all the languages in a project, or files on specific
+locations. Virtual folders have priority, so they can be used to allow
+translators to focus on the most important work.
+
+
+.. _virtual_folders#attributes:
+
+Virtual folders attributes
```
julen: *folders'*

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+-----------------------------------
```
julen: Underline unnecessarily long.

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+means that there can exist two different virtual folders with the same name if
+they have different locations.
+
+The priority defaults to ``1`` and accepts any value greater than ``0``,
+including numbers with decimals, like ``0.75``.
+
+By default virtual folders cannot be browsed: they won't be displayed, but
+they are still used for sorting.
+
+Also the virtual folders can have a description which might be useful to tell
```
julen: *useful to explain the contents of the folder or provide additional instructions*.

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+If a virtual folder applies in the current location, then clicking on the links
+on the overview page will provide the units in priority order when translating
+in the editor. The priority sorting on the translation editor is calculated
+taking into account all the applicable virtual folders in the current location,
+including the not browsable ones.
+
+
+.. _virtual_folders#json-format:
+
+Format for the JSON file
```
julen: If you read through the whole doc, this section sounds like a repetition of *Virtual folders' attributes*. Also, in the first section you miss out part of the information contained in this section. I'd suggest merging both sections.

pootle/apps/virtualfolder/management/commands/add_vfolders.py
```diff
@@ -0,0 +1,128 @@
...
+                    vfolder.description = vfolder_item['description']
+                    changed = True
+                    logging.info("Description for virtual folder '%s' will be "
+                                 "changed.", vfolder.name)
+
+                if changed:
+                    vfolder.save()
+                    updated_count += 1
+
+        logging.info("\nSucessfully added %d virtual folders.\nSuccessfully "
```
julen: There's a typo (*sucessfully*), but besides, it looks easier to parse an output such as:
```
Added: M
Updated: N
Unchanged: P
```

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+
+- A mandatory lowercase **name** for the virtual folder,
+- A mandatory **location** in Pootle for the virtual folder that can use
+  placeholders for language (``{LANG}``) and project (``{PROJ}``),
+- An optional float **priority** for the virtual folder, being 1 the default
+  priority and accepting only values greater than 0, where higher numbers means
+  higher priority,
+- An optional ``true`` or ``false`` indicating if the virtual folder is
+  **browsable** (defaults to ``false``),
+- An optional **description** that can include newlines (that must be escaped),
```
julen: *(must be escaped)*.

docs/features/virtual_folders.rst
```diff
@@ -0,0 +1,96 @@
...
+
+.. _virtual_folders#json-format:
+
+Format for the JSON file
+------------------------
+
+The JSON file used to import virtual folders consists of a list of virtual
+folder definitions with the following fields:
+
+- A mandatory lowercase **name** for the virtual folder,
```
julen: Why lowercase? What happens when it's not lowercase?

pootle/apps/virtualfolder/models.py
```diff
@@ -0,0 +1,132 @@
...
+        """Return a list with all the locations this virtual folder applies.
+
+        If the virtual folder location has no {LANG} nor {PROJ} placeholders
+        then the list only contains its location. If any of the placeholders is
+        present, then they get expanded to match all the existing languages and
+        projects.
+        """
+        if "{LANG}" in self.location and "{PROJ}" in self.location:
+            locations = []
+            temp = ""
```
julen: Why do you need this?

pootle/apps/virtualfolder/models.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        return VirtualFolder.objects.filter(
+            units__store__pootle_path__startswith=pootle_path
+        ).distinct()
+
+    def __unicode__(self):
+        return ": ".join([self.name, self.location])
+
+    def save(self, *args, **kwargs):
+        if not self.priority > 0:
```
julen: This is model validation, it should go in `clean_fields()` and raise a `ValidationError`.

pootle/apps/virtualfolder/management/commands/add_vfolders.py
unho: Changed.

docs/features/virtual_folders.rst
unho: Changed.

docs/features/virtual_folders.rst
unho: Fixed.

docs/features/virtual_folders.rst
unho: Changed.

docs/features/virtual_folders.rst
unho: Changed.

docs/features/virtual_folders.rst
unho: It is converted to lowercase.

Telling people that it must be lowercase reduces the chance of specifying two vfolders with the same lowercased name, thus failing the addition of the second one.

pootle/apps/virtualfolder/models.py
unho: When you save a vfolder, if its location contains `{LANG}` or `{PROJ}` placeholders they have to be replaced by all the existing language and project codes. This is necessary to have all the locations the vfolder applies to, so they can be concatenated with the file paths and thus retrieve the stores (if they exist) so the vfolder can create relationships with their units.

julen: Let me rephrase that: you always destroy the value as soon as you enter the loop, so what's the point of having this value set here?

docs/features/virtual_folders.rst
julen: Would it fail? I see name + location are unique together, not name.

pootle/apps/virtualfolder/models.py
```diff
@@ -0,0 +1,132 @@
...
+    @classmethod
+    def get_matching_for(cls, pootle_path):
+        """Return the matching virtual folders in the given pootle path.
+
+        Not all the applicable virtual folders have matching filtering rules.
+        This method further restricts the list of applicable virtual folders to
+        retrieve only those with filtering rules that actually match.
+        """
+        return VirtualFolder.objects.filter(
+            units__store__pootle_path__startswith=pootle_path
```
julen: I believe this wouldn't give the desired results when passing a `pootle_path` such as `/projects/<my_project>/<my_directory>/`; would it?

pootle/apps/virtualfolder/models.py
julen: Since there's no form/`is_valid()` involved, the location is fine, just raise `ValidationError`.

julen:
Did some initial testing, here are some observations in no particular order:

1. Is there a reason why passing the file via `--filename` is required? `add_vfolders myfile.json` looks natural, especially taking into account a file is the only way to provide input to the command.
2. Providing an invalid JSON file produces:  
  ```
CommandError: Some error occurred while parsing the file
  ```  
  Not very descriptive, plus not sure if this really fails into the `CommandError` category — it's a parsing error after all, and the user wants to know what needs fixing.
3. The `Succesfully parsed '<myfile>.json'` message is superfluous.
4. Some `location` paths seem to be unsupported, yet they are added as `VirtualFolder`s and sorting doesn't work (`/`, `/projects/<project_name>/`).

julen:
In the JSON file on the other hand, I believe any fields affecting filtering should go as children of a `filter` key as defined in the spec — it's explicit and easy to understand which filtering rules will be applied, and especially because more filtering options are foreseen.

docs/features/virtual_folders.rst
unho: @julen It will fail if the location is the same and the names lowercased are the same.

pootle/apps/virtualfolder/models.py
unho: Instead added a `clean_fields()` that raises `ValidationError` and added an explicit call at the beginning of `save()`. I think it helps with keeping `save()` clean.

docs/features/virtual_folders.rst
julen: Therefore it won't fail to add *two vfolders with the same lowercased name*.

pootle/apps/virtualfolder/models.py
unho: No. This is something that will need to be adjusted when this code is actually used to retrieve the vfolders for such paths. Right now we are only dealing with vfolders within TPs.

pootle/apps/virtualfolder/models.py
julen: Both are clean, in any case it'd help if in the future there's a `ModelForm` of any sort :+1:

docs/features/virtual_folders.rst
unho: Only would fail if they also share the same location.

pootle/apps/virtualfolder/models.py
unho: I got what you meant. It is not necessary. Removing it.

unho:
> Did some initial testing, here are some observations in no particular order:
> 
> 1. Is there a reason why passing the file via `--filename` is required? `add_vfolders myfile.json` looks natural, especially taking into account a file is the only way to provide input to the command.

Fixed.

> 2. Providing an invalid JSON file produces:  
>   ```
> CommandError: Some error occurred while parsing the file
>   ```  
>   Not very descriptive, plus not sure if this really fails into the `CommandError` category — it's a parsing error after all, and the user wants to know what needs fixing.

Altered error handling to raise a `CommandError` with the original message and a hint for failed JSON parsing.

> 3. The `Succesfully parsed '<myfile>.json'` message is superfluous.

Removed.

> 4. Some `location` paths seem to be unsupported, yet they are added as `VirtualFolder`s and sorting doesn't work (`/`, `/projects/<project_name>/`).

Now `/projects/<project_name>/` is transparently converted to `/{LANG}/<project_name>/` when associating the vfolder with the units. If `/` is provided a `ValidationError` is raised asking to provide instead `/{LANG}/{PROJ}/` since it might not be safe to transparently replace the location and also the specified file paths.

pootle/apps/virtualfolder/models.py
julen: So why is it restricted to TPs for now? FWIW going through `/projects/<my_project>/[<my_directory>/]` is a very valid use-case, especially for l10n managers.

unho:
> In the JSON file on the other hand, I believe any fields affecting filtering should go as children of a `filter` key as defined in the spec — it's explicit and easy to understand which filtering rules will be applied, and especially because more filtering options are foreseen.

Then the JSON file will need to be altered accordingly when such filtering options are added.

docs/features/virtual_folders.rst
unho: Fixed.

pootle/apps/virtualfolder/models.py
unho: @julen It is restricted to TPs for now because TP overview is the only place where we currently use vfolders. When we support other overviews as well, we will have to alter this accordingly.

As yes, that is the translation managers key use case.

unho:
Updated PR addressing most of the issues found. Also removed most of the comments to not overwhelm reviewers.

@julen @dwaynebailey @jleclanche @iafan @ta2-1 Any other issue on this PR?

unho:
A fix for the wrong priority ordering on the editor is still when crafted.

unho:
@julen @dwaynebailey @jleclanche @iafan @ta2-1 PR ready for final review. Priority sorting on editor now works (despite the ugly hack needed).

julen:
> Then the JSON file will need to be altered accordingly when such filtering options are added.

So why not do it right from the beginning and avoid future issues?

pootle/apps/virtualfolder/management/commands/add_vfolders.py
```diff
@@ -0,0 +1,135 @@
...
+class Command(BaseCommand):
+    help = "Add virtual folders from file."
+
+    def handle(self, *args, **options):
+        """Add virtual folders from file."""
+
+        if not args:
+            raise CommandError("You forgot to provide the mandatory filename.")
+        elif len(args) > 1:
+            logging.warning("You have provided more than one argument. Only "
```
julen: Is this really useful?

pootle/core/helpers.py
```diff
@@ -154,9 +155,17 @@ def get_overview_context(request):
     resource_obj = request.resource_obj
     resource_path = getattr(request, 'resource_path', '')
 
-    url_action_continue = resource_obj.get_translate_url(state='incomplete')
-    url_action_fixcritical = resource_obj.get_critical_url()
-    url_action_review = resource_obj.get_translate_url(state='suggestions')
+    filters = {}
+
+    if not isinstance(resource_obj, Store):
+        if VirtualFolder.get_matching_for(request.pootle_path).count():
```
julen: you can combine these into a single `if` expression.

pootle/core/browser.py
```diff
@@ -85,14 +87,14 @@ def get_table_headings(choices):
     return filter(lambda x: x['id'] in choices, HEADING_CHOICES)
 
 
-def make_generic_item(path_obj):
+def make_generic_item(path_obj, **kwargs):
     """Template variables for each row in the table."""
     return {
         'href': path_obj.get_absolute_url(),
         'href_all': path_obj.get_translate_url(),
```
julen: This should also receive `**kwargs`, I'd say

unho: @julen No. The only kwarg we are passing now is the priority sorting, and we don't want priority sorting when translating all the units.

pootle/core/helpers.py
unho: Fixed.

pootle/apps/virtualfolder/management/commands/add_vfolders.py
unho: No. Will remove it.

unho:
@julen @dwaynebailey @jleclanche @iafan @ta2-1 Any final review for this PR? Or is it gtm?

dwaynebailey:
@unho gtm from my side.

Only one thing though is @julen's comment about filter rules being children which sounds like a sensible suggestion that would allow any other filtering rules to be added without having to change the JSON.  I'm not clear why you're rejecting that idea.

unho:
@dwaynebailey I am not rejecting the idea. My point on this is that this should be done once we have a clear idea of how the filtering is going to work. And either way the JSON will be changed.

dwaynebailey:
@unho I think you might be missing my point.  Without the children implemented everyone **will** have to change their JSON (as distinct from *the* JSON).  Doing children means that the current file based and whatever new one can live side by side without a change.  Just saving all of us some pain and allowing the two to live together if needed.

That's the reasoning which I'm not seeing being challenged here.

unho:
> I think you might be missing my point.  Without the children implemented everyone **will** have to change their JSON (as distinct from *the* JSON).  Doing children means that the current file based and whatever new one can live side by side without a change.  Just saving all of us some pain and allowing the two to live together if needed.

Lets talk with examples. This is the current status:

```json
[
    {
        "name": "current-status",
        "location": "/{LANG}/proj/dir/",
        "files": [
            "fileA.po",
            "subdir/fileB.po"
        ]
    }
]
```

This was @julen proposal:

```json
[
    {
        "name": "julen-proposal",
        "filters": {
            "location": "/{LANG}/proj/dir/",
            "files": [
                "fileA.po",
                "subdir/fileB.po"
            ]
        }
    }
]
```

And this was my proposal for moving the "files" into the "filters" (but keeping "location" as a property of the vfolder):

```json
[
    {
        "name": "my-proposal",
        "location": "/{LANG}/proj/dir/",
        "filters": {
            "files": [
                "fileA.po",
                "subdir/fileB.po"
            ]
        }
    }
]
```

The reason why I am reluctant to move the "files" into the "filters" is that we are not yet sure if when a proper filtering is in place we will have any "files" anywhere. So if it is going to be dropped in the future, why bother now about moving it there?

So until we add more filtering rules, and we are sure we will keep the "files" around (at least in this way), I suggest to keep the current JSON format since it provides the same data with a simpler syntax.

unho:
@dwaynebailey @julen Altered the JSON syntax to the following. Does this address your concerns?

```json
[
    {
        "name": "my-proposal",
        "location": "/{LANG}/proj/dir/",
        "filters": {
            "files": [
                "fileA.po",
                "subdir/fileB.po"
            ]
        }
    }
]
```

dwaynebailey:
@unho yes that new JSON layout addresses my concerns

unho:
@iafan A review from your side would be much appreciated.

unho:
@translate/evernote @translate/house People seems to be ok with the new JSON syntax, which was the last blocker. Any other issue or gtm?

iafan:
I did a first round of review of this PR on our staging server, which wasn't quite successful. In short, I was unable to successfully add any v-folder description that would show up in the UI. I don't know if this is related to caching or not, but nothing changed from the UI standpoint when I added v-folders via add_vfolders command. I tried simplest scenarios when there's a v-folder inside a folder A, which combines two files (B.po and C.po) located in the same folder A.

Also, the current v-folder functionality is very limited (my assumptions here are based on the documentation and samples alone): one can only provide the set of files to group together. In our case, we can only imagine real use of this feature when one can specify any arbitrary folder (starting with project folders).

It is not obvious how to delete v-folders. When I delete vfolder definitions from the .json file and run add_vfolders, nothing happens. Does add_vfolders (or any other command) support deletion at all?

It is not obvious whether one needs to include leading and/or trailing slash in `location` path. The command gives no errors.

Side notes: I noticed these regressions:
1) the sidebar is always visible (can be expanded/collapsed), even it doesn't have any content in it. In the master branch code, it is not available if it has nothing to display.
2) the sidebar height depends on the size of the content (the number of rows in the browsing table). It should take the full height of the content area.

To sum it up, this definitely needs another round of testing and bugfixing before the feature goes live.

julen:
> Side notes: I noticed these regressions:
> 1) the sidebar is always visible (can be expanded/collapsed), even it doesn't have any content in it. In the master branch code, it is not available if it has nothing to display.
> 2) the sidebar height depends on the size of the content (the number of rows in the browsing table). It should take the full height of the content area.

These are very likely remnants of a previous state of the PR branch so it's safe to ignore it. I recreated assets just in case and cannot see this happening on our staging server.

iafan:
Thanks, @julen. Are you able to setup v-folders on your side? I think I tried to closely follow the sample JSON provided, but have no luck seeing the v-folders in the browser

iafan:
@julen,

I can actaully see the empty but open side panel if I go to https://translate.stage.evernote.com/ru/website3_evernote/home/ (again, this might be some side effect of caching, but worth checking on your side)

dwaynebailey:
@iafan this first stage commit has no UI.  The change forces editing based on v-folder priority.  The second stage, not yet PR'd, adds the UI for the TP level.

iafan:
@dwaynebailey, so how one is supposed to test this? It makes sense to add UI to this PR as well, and test the feature as a whole, IMO.

julen:
> @julen,
> 
> I can actaully see the empty but open side panel if I go to https://translate.stage.evernote.com/ru/website3_evernote/home/ (again, this might be some side effect of caching, but worth checking on your side)

@iafan I realize some setting changes to disable the imports/exports were uncommitted on myside, and that's fixed now. In any case this displays that imports/exports in master is not in good shape, and I already reported this on IRC.

pootle/apps/virtualfolder/models.py
```diff
@@ -0,0 +1,146 @@
...
+        if "{LANG}" in self.location and "{PROJ}" in self.location:
+            locations = []
+            for lang in Language.objects.all():
+                temp = self.location.replace("{LANG}", lang.code)
+                for proj in Project.objects.all():
+                    locations.append(temp.replace("{PROJ}", proj.code))
+            return locations
+        elif "{LANG}" in self.location:
+            return [self.location.replace("{LANG}", lang.code)
+                    for lang in Language.objects.all()]
```
unho: This should use only current project languages.

pootle/apps/virtualfolder/models.py
```diff
@@ -0,0 +1,146 @@
...
+        elif "{PROJ}" in self.location:
+            return [self.location.replace("{PROJ}", proj.code)
+                    for proj in Project.objects.all()]
```
unho: This should use only current language projects.

unho:
It seems that I missed the party.

> I did a first round of review of this PR on our staging server, which wasn't quite successful. In short, I was unable to successfully add any v-folder description that would show up in the UI. I don't know if this is related to caching or not, but nothing changed from the UI standpoint when I added v-folders via add_vfolders command. I tried simplest scenarios when there's a v-folder inside a folder A, which combines two files (B.po and C.po) located in the same folder A.

No descriptions are being displayed as it is not yet clear to me where does should be displayed.

> Also, the current v-folder functionality is very limited (my assumptions here are based on the documentation and samples alone): one can only provide the set of files to group together. In our case, we can only imagine real use of this feature when one can specify any arbitrary folder (starting with project folders).

This is just the first step. In the future you can expand this as necessary to support more filtering methods. For now using stores just recovers our lost goals functionality.

> It is not obvious how to delete v-folders. When I delete vfolder definitions from the .json file and run add_vfolders, nothing happens. Does add_vfolders (or any other command) support deletion at all?

No. This is not yet implemented.

You can instead hide vfolders by setting them to not being browsable. Changing vfolders is supported through the `add_vfolders` command by just providing a JSON file altered accordingly.

> It is not obvious whether one needs to include leading and/or trailing slash in `location` path. The command gives no errors.

I was assuming that people would check the docs and the sample JSON in case of doubt. As you can see the sample JSON always includes the trailing slashes. Just checked the code and it never complains, but in some cases the absence of trailing slash might result on not getting the desired results on filtering.

> Side notes: I noticed these regressions:
> 1) the sidebar is always visible (can be expanded/collapsed), even it doesn't have any content in it. In the master branch code, it is not available if it has nothing to display.
> 2) the sidebar height depends on the size of the content (the number of rows in the browsing table). It should take the full height of the content area.

No change is done in this PR to the sidebar, so these should be issues already present on master.

> To sum it up, this definitely needs another round of testing and bugfixing before the feature goes live.

I don't think this PR has bugs. Re testing, see below.

> @dwaynebailey, so how one is supposed to test this? It makes sense to add UI to this PR as well, and test the feature as a whole, IMO.

Please see the testing steps provided on comment https://github.com/translate/pootle/pull/3600#issuecomment-76679565 I am aware that lack of UI difficults too much the testing, but the steps in that comment should be enough. If they are not enough we can talk about appending here some of the commits already finished that add UI bits, despite that I would prefer to land the current set of commits since has been tested and reworked several times instead of adding more commits that might need to be reworked more times thus delaying this PR even more.

unho:
@iafan I agree we need a way to remove a vfolder.

IMHO considering that a vfolder that is not on the JSON means that it should be removed is a bad idea. I mean that if you have 100 vfolders you usually going to use the JSON to just alter some vfolders (lets say 5 of them) and having to deal with a JSON with 100 vfolder is not comfortable. Besides that having a `add_vfolders` that allows to remove them is kind of confusing.

Another option is to add a new management command `del_vfolder` which accepts some data identifying a vfolder and removes it. I think it is better for it to remove only a single vfolder at a time. Since vfolders have no unique identifiers (languages and projects have their codes) we will have to provide a name and a location, so we might end with something like:

```shell
$ ./manage.py del_vfolder my-vfolder /{LANG}/proj/dir/subdir/subsubdir/
```

which I am not sure if can be considered simple and easy.

Anyway whichever way we take re deleting, IMHO it is less priority than providing a working UI for translators to work with vfolders. Despite it is an ugly workaround, you can always hide vfolders by setting them to not being browsable.

dwaynebailey:
@unho can you create an issue to track the del_vfolder command.  I think we can delay it with the 1.0 and not browesable hack.  But we do need to fix it to complete the feature.

dwaynebailey:
@iafan the way I tested was as following.  I created 3 simple 3 unit PO files (translated, untranslated, fuzzy) a.po, b.po, c.po.  I created vfolder that set 2.0 for b.po and 0.5 priorities for a.po.  Used the editor to ensure that the files appear in the order specified which would be b.po, c.po, a.po.

This first implementation is designed simply to ensure the backend for storing goals is good and that priority sorting of gaols works correctly.

unho:
@dwaynebailey Done: https://github.com/translate/pootle/issues/3634

iafan:
> This first implementation is designed simply to ensure the backend for storing goals is good and that priority sorting of gaols works correctly.

Then I can delegate the review to @julen. If this feature is not visible and has not reached the level it can be of real use to us (Evernote), I'm ok to have it merged if Julen confirms that it looks sane from the backend side. I'll be happy to review this once the UI part has landed, and once we can start testing this on some real-life scenarios (for us the first most obvious case will be to group projects together, which needs the filters to support paths, not individual .po files).

iafan:
> MHO considering that a vfolder that is not on the JSON means that it should be removed is a bad idea. I mean that if you have 100 vfolders you usually going to use the JSON to just alter some vfolders (lets say 5 of them) and having to deal with a JSON with 100 vfolder is not comfortable. Besides that having a add_vfolders that allows to remove them is kind of confusing.

@unho, I think I disagree on this. From my perspective, it would be much easier to have one JSON file which fully represents the config for all v-folders, and maintain it, adding or deleting entries from it. Then a single command, say, `update_vfolders` would synchronize this JSON with the database, creating, updating or deleting v-folders accordingly.

Otherwise, you're going to have two commands (`add_vfolders` and `del_vfolders`) which you have to remember about and maintain data for, and deletion of vfolders will be a much less intuitive task than just deleting relevant entries from a single JSON.

unho:
>> IMHO considering that a vfolder that is not on the JSON means that it should be removed is a bad idea. I mean that if you have 100 vfolders you usually going to use the JSON to just alter some vfolders (lets say 5 of them) and having to deal with a JSON with 100 vfolder is not comfortable. Besides that having a add_vfolders that allows to remove them is kind of confusing.

> @unho, I think I disagree on this. From my perspective, it would be much easier to have one JSON file which fully represents the config for all v-folders, and maintain it, adding or deleting entries from it. Then a single command, say, `update_vfolders` would synchronize this JSON with the database, creating, updating or deleting v-folders accordingly.

> Otherwise, you're going to have two commands (`add_vfolders` and `del_vfolders`) which you have to remember about and maintain data for, and deletion of vfolders will be a much less intuitive task than just deleting relevant entries from a single JSON.

I see your point, but I was thinking on a several admins scenario. Maybe yours is better.

@dwaynebailey any input on this? 

> (for us the first most obvious case will be to group projects together, which needs the filters to support paths, not individual .po files).

@iafan Can you explain this a bit more?

iafan:
> I see your point, but I was thinking on a several admins scenario.

I think this is pretty much similar to regular Pootle configs. We have several Pootle admins, and they have an access to the Pootle configuration file, which we keep in Git. The same thing is applicable to this JSON config — especially if there are multiple people working with it — it makes sensу to keep it in version control (and this is where its history will be stored), and then apply it as a whole (e.g. as a part of deployment script) to the server.

iafan:
> (for us the first most obvious case will be to group projects together, which needs the filters to support paths, not individual .po files).
> @iafan Can you explain this a bit more?

This is pretty straightforward: imagine you can set up any path prefix in your `filters > files` section of the v-folder JSON entry. This way you would include all files down that tree, and this should work right from any point in the tree. Consider this example:
```
    {
        "name": "ios_skitch_evernote_bundle",
        "location": "/{LANG}/",
        "priority": 1,
        "is_browsable": true,
        "description": "Skitch (iOS) — all related projects",
        "filters": {
            "files": [
                "ios_skitch_evernote",
                "common_skitch_evernote"
            ]
        }
    },
```

This is a v-folder that lives inside each language folder, and which groups two translation projects together, identified by their relative paths: `ios_skitch_evernote` and `common_skitch_evernote`.

unho:
@iafan Implementing filtering by path besides filtering by specific filenames looks easy to implement. Added issue #3643 to track it.

unho:
@iafan Lets move discussion re deleting vfolders to issue #3634.

PR #3603, opened 2015-02-23
Fixing regular expression for locale codes

trejkaz:
I had tried to enter es_419 (Spanish (Latin America)) as a locale code,
but pootle rejected it as non-standard even though AFAIK this is the
standard code. Pootle’s regex seemingly doesn’t know about numeric
region codes being valid in locale codes.

### DIFF
--- a/pootle/apps/pootle_app/forms.py
+++ b/pootle/apps/pootle_app/forms.py
@@ -31,7 +31,7 @@ from pootle_project.models import Project, RESERVED_PROJECT_CODES
 from pootle_store.models import Store
 
 
-LANGCODE_RE = re.compile("^[a-z]{2,}([_-][a-z]{2,})*(@[a-z0-9]+)?$",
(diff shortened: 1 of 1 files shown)


jleclanche:
Are numeric region codes always 3 digits?

unho:
@jleclanche It looks like they are http://www.iso.org/iso/country_codes.htm

@trejkaz Despite it is possible to use such numeric codes, the standard seems to recommend using the two letters codes, and that is why Pootle didn't support any other method.

unho:
lgtm

trejkaz:
3-digit codes were a shock to me as well, but it seems to be the ICU way to encode that region. What I didn't expect at the time was that there are region codes for all kinds of overlapping regions.

dwaynebailey:
@unho no they aren't ISO country codes (those are e.g. ZA, GB, US, etc). They are it seems [UN_M.49](https://en.wikipedia.org/wiki/UN_M.49) region codes.

Our regex is very aligned with Gettext, and with other use cases we should accept those.  We could/should accept anything.  I think the reason we limited these things is that remote TM would not be able to retrieve any suggestions if we can't process the code.

lgtm

PR #3616, opened 2015-03-03
License cleanup

dwaynebailey:
Ready for review and input.

This PR cleans up the current license situation, which while not a disaster has some potential issues and missed many files that should have headers.

The document which starts in  02a2193 highlights the why.  The choice of header is explained in  9977897 and the actual header is in  e885fde

The rest of the PR covers adding the header to the Python files.  But soon realising that it should be in JavaScript, CSS, Gettext PO and Django templates files.  Adding headers to Django
bfb11be7bb87cc9e57c646ed53d074e02af20a33 seem to break some tests.

Some notes on the PR:

- The AUTHORS file - I wanted some review of the direction before spending too much time on that
- The LICENSE_CLEANUP_RATIONALE.rst will be deleted when we're done with the PR and moved to the wiki, it just makes more sense to keep it in the PR for now and to land the code with the document.
- I have not covered the migration to GPL3, it isn't worth it until we get agreement on what I have here.

Process

1. I'm looking for input from core contributor, then
2. I'll ask for input from the wider community

@iafan I'd really like some input from the Evernote perspective.


jleclanche:
Can you land the translations updates on master directly? They are making it very hard to review.

jleclanche:
The amount of hashbangs we have in unnecessary places is unnerving. Anyway, probably best leaving it for next PR, but if you want to tackle this too i think it's fine to do it here.

julen:
> Can you land the translations updates on master directly? They are making it very hard to review.

:+1:

unho:
I also agree that landing the translations updates (those that don't depend on licensing changes) should be merged to master directly to reduce noise.

unho:
@dwaynebailey I think that the commit that blanks the `__init__.py` files can be safely landed.

dwaynebailey:
@julen @unho @jleclanche OK so despite asking everybody to not bitch about l10n the general consensus is that they should be out of the PR.  I agree actually it was just painful to remove them. But I took my pain meds and edited various commits and landed those on master.

@jleclanche lets leave hashbangs to another PR. Remember our one aim is to make sure we don't actually have to think.

@jleclanche on the issues of utf-8 encoding parts of the headers.  Again we don't want to think. And I don't want to have to think if something is going to be UTF-8 in a file or just plain Larin1.  That would mean you have to check each localisable string and comment or make sure they are Latin1, silly use of our time.  And since this issue goes away on Py3 it isn't worth thinking about.  So I don't want that changed.

dwaynebailey:
@unho, @jleclanche had a similar comment about blank __init__.py.  While I agree that they can be landed safely on their own.  They are licence related and I want license related things to land together, no matter how frivolous this one appears.

dwaynebailey:
OK, Django templates are still broken.  @jleclanche @julen @unho any idea how to fix them.

@jleclanche didn't like this change.  Neither do I actually, but those files need a license and I'm not sure how else to do it. Any suggestions?

unho:
Travis is already failing on master branch with the same error as in this PR, so we can't be sure if the problem is on the templates being broken.

Also, is it really necessary for the comment to start with `{% comment "Copyright and License" %}`? I mean that it is obvious from the comment content that it is a copyright and license notice.

julen:
> OK, Django templates are still broken. @jleclanche @julen @unho any idea how to fix them?

Can you please point us to where you see they are breaking?

dwaynebailey:
@julen Django templates are broken in Travis.

dwaynebailey:
@unho no the template comment theoretically doesn't need the extra noise of "Copyright and License".  But I think there are bigger fish to fry as they're ugly as it is.

unho:
@dwaynebailey Maybe I looking at the wrong place, but Travis seems to fail when running `test_pending_agreements` test. This failure also happens on master branch.

dwaynebailey:
@julen I've adjusted the JavaScript header to align with your suggestions and convention.

unho:
lgtm

iafan:
lgtm

PR #3625, opened 2015-03-10
Sidebar: fixed when should it be displayed

julen:
Fixes the issues with the sidebar and is only rendered when there's something to be displayed.

I have tested with announcements and uploads/downloads, looks good on my end.

@unho please test and report back.


unho:
Tested and reviewed. gtm

PR #3626, opened 2015-03-10
include_raw: don't limit to `FileSystemLoader`

julen:
This ensures any available template loader will be considered, therefore not blindly relying on the `FileSystemLoader`.

@unho please test and report back.


unho:
@julen Tested and it works. Despite I feel the code can be simplified I am gtm'ing this PR.

Issue #3597, opened 2015-02-19
Drop south_migrationhistory DB table after upgrading to 2.6.0


unho:
Like those already being dropped in https://github.com/translate/pootle/blob/stable/2.6.0/pootle/apps/pootle_app/management/commands/upgrade.py#L141-L149

South migration history is no longer necessary after upgrading to 2.6.0, since in Pootle 2.7.0 South is dropped in favor of Django migrations.

unho:
It was decided to postpone the South migration history removal in order to ensure we can recover from failing upgrades. See the discussion on https://github.com/unho/pootle/commit/0650f4f0#comments for more information.

jleclanche:
This was done.

dwaynebailey:
@jleclanche could you please link to commits that achieve this, its a PITA to later find these type of comments but no link to the commit.

jleclanche:
I don't really know when it was done, but it was part of our 2.6/2.7 db wipe.

unho:
@dwaynebailey @jleclanche After final upgrade to 2.7 the `south_migrationhistory` DB table is still present. It was considered to drop it as the last step of the upgrade, but it was deemed helpful to keep it for further reference after upgrade.

So reopening.

jleclanche:
Huh.

julen:
I wonder what's the real benefit of this. The same way as no tables have been cleaned up for other scenarios (registration, django-taggit etc.), I don't think Pootle should be doing anything here. Just let the DBA do this if needed, otherwise it doesn't hurt.

PR #3628, opened 2015-03-11
Upgrade remnants for 2.6 branch

unho:


unho:
@jleclanche @dwaynebailey Please review and test if possible.

unho:
@dwaynebailey @jleclanche This PR is ready for final review as well.

unho:
@dwaynebailey Dropped the commit that removed the South migration history and postponed issue #3597.

unho:
@jleclanche @dwaynebailey Is there any issue? If not, please gtm.

dwaynebailey:
@unho lgtm but I must admit I really don't understand the two migrations.  I'm assuming the 'submissions without submitters' relate to the issues I reported previously and for which you asked for counts on mlo and LO.  Revision model, I'm in the dark.

unho:
@dwaynebailey Previously master had a Revision model that persisted the last revision number on the DB, but now it is using Redis for that. So since master dropped the DB model, there is no reason to keep it on stable/2.6.0 since it was only backported to provide a working upgrade.

About 'submissions without submitters'. In some legacy installs (mlo and libo for example) there might exist those. But a submission without submitter is useless for us, and even more, can throw tracebacks. So since we can't recover the lost submitter in most cases, it is preferable to just drop them, so we don't get tracebacks in the future.

PR #3631, opened 2015-03-11
Initdb: Add default project descriptions

### LINKED ISSUE #3416: Add default project descriptions
In order to get ``initdb`` running on ``core`` we had to drop the project [descriptions] since these move to staticpages on core.  This bug is to add those back.

To achieve that initdb should create static pages for these default projects that contain the content previously held in the [descriptions](https://github.com/translate/pootle/blob/96edf539/pootle/core/initdb.py#L300-L306).

unho:
Fixes #3416.

### DIFF
--- a/pootle/core/initdb.py
+++ b/pootle/core/initdb.py
@@ -29,6 +29,7 @@ from pootle_app.models import Directory
 from pootle_app.models.permissions import PermissionSet, get_pootle_permission
 from pootle_language.models import Language
 from pootle_project.models import Project
+from staticpages.models import StaticPage as Announcement
 
 
 def initdb():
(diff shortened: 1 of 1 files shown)


dwaynebailey:
@unho yes drop those.  Pootle localisations should be centralised.

unho:
@jleclanche @dwaynebailey Ready for final review.

unho:
@jleclanche @dwaynebailey The concerns raised have been fixed. AFAICT these changes don't affect Evernote. So as soon as I get a gtm I will land this.

unho:
@jleclanche @dwaynebailey Is there any issue? If not, please gtm.

dwaynebailey:
@unho its gtm but I don't see a default description for the Terminology project.

unho:
@dwaynebailey The default description is now being added as an announcement.

PR #3639, opened 2015-03-17
User model move

jleclanche:
Stripped from allauth branch.


julen:
Right, migration's missing.

jleclanche:
If I commit a migration, I'll have to delete it right away because it's not a migration compatible with 2.6. The migration would *just* be for evernote. Can't you guys run a simple `rename table pootle_user to accounts_user;`?

dwaynebailey:
@julen if it really is as simple as @jleclanche says then I'm not sure what having a migration will achieve here.  We run the risk of making it and its still wrong.  The only place to test it is on EN data and you guys know best the state.

We're just going to have to remove it anyway.

Help me out here, it really doesn't make sense.

dwaynebailey:
Slept on this as something about it bothered me.

@jleclanche so I did as your outlined above and it doesn't work. Somethings else needs changing.  So this needs migrations.  Please add them.

unho:
@dwaynebailey @jleclanche The migration is actually not necessary for upgrading from 2.6.0, neither for installing from scratch, but the DB table renaming is necessary for anyone that is using current master. I think it is safe to add a migration that renames ``pootle_user`` table if it exists despite it is useless for all the other people. We did this in some migrations in 2.6.0.

dwaynebailey:
@julen OK did a full check and all things are OK with this recipe so no need for any migrations if you follow this:

1. git checkout master
2. ./manage.py migrate
3. ./manage.py initdb

You are at Evernote level now

4. git checkout pr/3639
4. in SQL: alter table pootle_user rename to accounts_user;
5. rm $(find . -name "*.pyc")
5. ./manage.py migrate
6. ./manage.py runserver

All works correctly.

julen:
Thanks for the instructions @dwaynebailey. However and FTR, the alternative source languages table needs to be renamed too:
```mysql
ALTER TABLE pootle_user_alt_src_langs RENAME TO accounts_user_alt_src_langs;
ALTER TABLE pootle_user RENAME TO accounts_user;
```

GTM.

jleclanche:
Landed in c9b586fa5293bdf7d2893292d594bce24e596fe7

PR #3642, opened 2015-03-17
Second stage on vfolders

unho:
@translate/house @translate/evernote 

Instructions for setup:

- You might need to make the assets
- Run `./manage.py shell_plus`
- Run `VirtualFolder.objects.all().delete()` and exit the shell
- Get https://drive.google.com/file/d/0B-8KVyCPnkZRck1IUkZKWklCeHM and uncompress it on your `PODIRECTORY`
- Add a new `test_vfolders` project.
- Run `./manage.py update_stores --project=test_vfolders`
- Get http://dpaste.com/0NHWD4W and save it as `vfolders_test_directory.json`
- Run `./manage.py add_vfolders vfolders_test_directory.json`

Instructions for testing:

- Run `./manage.py runserver`
- Log in as a regular user (not admin)
- Go to `http://127.0.0.1:8000/af/test_vfolders/browser/chrome/`
- Check that:
  - the vfolders stats table is displayed
  - the listed vfolders are all the vfolders that should be displayed. The following must not be displayed:
    - not browsable vfolders,
    - completely translated vfolders,
    - priority below 1 vfolders.
  - the counts match. Note that each vfolder counts must only include the files they match in the current location or below it.
  -  the translate links are correct. Just are like regular translate links for this location, but with the vfolder name included just after its location, for example a vfolder with location `/af/test_vfolders/browser/` and name `test-overlapping-vfolder` will have a needs work translate link pointing to `http://127.0.0.1:8000/af/test_vfolders/translate/browser/test-overlapping-vfolder/chrome/#filter=incomplete`
- Log in as an admin user
- Check that in `http://127.0.0.1:8000/af/test_vfolders/browser/chrome/` now you can also see:
  - not browsable vfolders styled as disabled (grayed out),
  - completely translated vfolders,
  - vfolders with priority below 1 styled as disabled (grayed out).
- Click on the `need translation` translate link for vfolder `test-overrides` on the vfolders stats table and ensure that:
  - you see `browser/test-overrides/chrome` in the navbar dropdown,
  - the URL in the browser location bar is `http://127.0.0.1:8000/af/test_vfolders/translate/browser/test-overrides/chrome/#filter=incomplete`,
  - the units displayed only belong to files matched by the vfolder filter in `/af/test_vfolders/browser/chrome/`
  - it is not possible to sort by priority.
- Select the same path to be translated in the editor, but not restricted to a vfolder: `http://127.0.0.1:8000/af/test_vfolders/translate/browser/chrome/#filter=incomplete`. Check that:
  - the unit count is bigger
  - you can sort by priority.
- Open the navbar dropdown and ensure that all the vfolders that match within current project are displayed:
  - Should be included (even if they have low priority or are completely translated): "test-overlapping-vfolder", "test-overrides", "test-low-priority", "test-vfolder1", "test-vfolder2", "test-vfolder3"
  - Are not browsable and thus should not be included: "test-not-browsable", "test-completely-translated"
  - Don't apply to any file, so they should not be included: "test-directory", "test-newfile"
- Open `http://127.0.0.1:8000/af/test_vfolders/translate/chat/#filter=incomplete` and ensure that:
  - no priority is being displayed for any of the units listed,
  - it is not possible to select priority sorting.
- Open `http://127.0.0.1:8000/af/test_vfolders/translate/dom/chrome/#filter=incomplete` and ensure that:
  - the top priority is being displayed for all the units shown (the units in this location belong to the vfolders whose name starts with `test-vfolder`, which eases looking in the JSON). Remember to ensure that the top priority is shown for a unit that is included in several vfolders.
  - you can sort by priority. Remember top priorities are above and lower at the bottom.

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -18,7 +18,7 @@ from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
 from django.db import connection, models
 from django.db.models import Q
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
 from django.utils.encoding import iri_to_uri
 from django.utils.functional import cached_property
@@ -36,6 +36,7 @@ from pootle_app.models.directory import Directory
 from pootle_app.models.permissions import PermissionSet
 from pootle_store.filetypes import filetype_choices, factory_classes
 from pootle_store.util import absolute_real_path
+from virtualfolder.signals import vfolder_post_save
 
 
 RESERVED_PROJECT_CODES = ('admin', 'translate', 'settings')
@@ -283,6 +284,18 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         """Returns ``True`` if this project is a terminology project."""
         return self.checkstyle == 'terminology'
 
+    @property
+    def vfolders(self):
+        """Return the browsable virtual folders for this project."""
+        # This import must be here to avoid circular import issues.
+        from virtualfolder.models import VirtualFolder
+
+        return [vf.tp_relative_path
+                for vf in VirtualFolder.objects.filter(
+                    units__store__translation_project__project__code=self.code,
+                    is_browsable=True
+                ).distinct()]
+
     @cached_property
     def languages(self):
         """Returns a list of active :cls:`~pootle_languages.models.Language`
@@ -324,7 +337,8 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         results = cursor.fetchall()
 
         # Calculate TP-relative paths and sort them
-        resources = sorted({to_tp_relative_path(result[0]) for result in results},
+        resources_set = {to_tp_relative_path(result[0]) for result in results}
+        resources = sorted(resources_set | set(self.vfolders),
                            key=get_path_sortkey)
 
         cache.set(cache_key, resources, settings.OBJECT_CACHE_TIMEOUT)
@@ -508,6 +522,24 @@ class ProjectSet(VirtualResource, ProjectURLMixin):
         return project.code
 
     ### /TreeItem
+
+
+@receiver([vfolder_post_save, pre_delete])
+def invalidate_resources_cache_for_vfolders(sender, instance, **kwargs):
+    if instance.__class__.__name__ == 'VirtualFolder':
+        try:
+            # In case this is vfolder_post_save.
+            affected_projects = kwargs['projects']
+        except KeyError:
+            # In case this is pre_delete.
+            affected_projects = Project.objects.filter(
+                translationproject__stores__unit__vfolders=instance
+            ).distinct().values_list('code', flat=True)
+
+        cache.delete_many([
+            make_method_key('Project', 'resources', proj)
+            for proj in affected_projects
+        ])
 
 
 @receiver([post_delete, post_save])
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -38,6 +38,7 @@ from pootle_misc.forms import make_search_form
 from pootle_misc.util import ajax_required, to_int, get_date_interval
 from pootle_statistics.models import (Submission, SubmissionFields,
                                       SubmissionTypes)
+from virtualfolder.helpers import extract_vfolder_from_path
 from virtualfolder.models import VirtualFolder
 
 from .decorators import get_unit_context
@@ -459,7 +460,13 @@ def get_units(request):
     request.profile = User.get(request.user)
     limit = request.profile.get_unit_rows()
 
+    vfolder, pootle_path = extract_vfolder_from_path(pootle_path)
+
     units_qs = Unit.objects.get_for_path(pootle_path, request.profile)
+
+    if vfolder is not None:
+        units_qs = units_qs.filter(vfolders=vfolder)
+
     units_qs = units_qs.select_related(
         'store__translation_project__project',
         'store__translation_project__language',
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -811,7 +811,10 @@ def get_overview_stats(request, *args, **kwargs):
     stats = request.resource_obj.get_stats()
 
     if isinstance(request.resource_obj, Directory):
-        stats['vfolders'] = VirtualFolder.get_stats_for(request.resource_obj.pootle_path)
+        stats['vfolders'] = VirtualFolder.get_stats_for(
+            request.resource_obj.pootle_path,
+            request.user.is_superuser
+        )
 
     return JsonResponse(stats)
 
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -747,8 +747,21 @@ def get_edit_unit(request, unit):
     alt_src_langs = get_alt_src_langs(request, user, translation_project)
     project = translation_project.project
 
-    # Retrieve the unit top priority, if any.
-    priority = unit.vfolders.aggregate(priority=Max('priority'))['priority']
+    priority = None
+    vfolder_pk = request.GET.get('vfolder', '')
+
+    if vfolder_pk:
+        try:
+            # If we are translating a virtual folder, then display its priority.
+            # Note that the passed virtual folder pk might be invalid.
+            priority = VirtualFolder.objects.get(pk=vfolder_pk).priority
+        except VirtualFolder.DoesNotExist:
+            pass
+
+    if priority is None:
+        # Retrieve the unit top priority, if any. This can happen if we are not
+        # in a virtual folder or if the passed virtual folder pk is invalid.
+        priority = unit.vfolders.aggregate(priority=Max('priority'))['priority']
 
     template_vars = {
         'unit': unit,
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -31,12 +31,14 @@ from pootle.core.decorators import (get_path_obj, get_resource,
                                     permission_required)
 from pootle.core.exceptions import Http400
 from pootle.core.http import JsonResponse, JsonResponseBadRequest
+from pootle_app.models.directory import Directory
 from pootle_app.models.permissions import check_user_permission
 from pootle_misc.checks import check_names
 from pootle_misc.forms import make_search_form
 from pootle_misc.util import ajax_required, to_int, get_date_interval
 from pootle_statistics.models import (Submission, SubmissionFields,
                                       SubmissionTypes)
+from virtualfolder.models import VirtualFolder
 
 from .decorators import get_unit_context
 from .fields import to_python
@@ -807,6 +809,10 @@ def get_qualitycheck_stats(request, *args, **kwargs):
 @get_resource
 def get_overview_stats(request, *args, **kwargs):
     stats = request.resource_obj.get_stats()
+
+    if isinstance(request.resource_obj, Directory):
+        stats['vfolders'] = VirtualFolder.get_stats_for(request.resource_obj.pootle_path)
+
     return JsonResponse(stats)
 
 
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -103,11 +103,20 @@ class VirtualFolder(models.Model):
                                                         priority__gte=1)
 
     @classmethod
-    def get_stats_for(cls, pootle_path):
-        """Get stats for all the virtual folders in the given path."""
-        stats = {}
+    def get_stats_for(cls, pootle_path, all_vfolders=False):
+        """Get stats for all the virtual folders in the given path.
 
-        for vf in cls.get_visible_for(pootle_path):
+        If ``all_vfolders`` is True then all virtual folders in the passed
+        pootle_path are returned, independently of their priority of
+        browsability.
+        """
+        stats = {}
+        if all_vfolders:
+            vfolders = cls.get_matching_for(pootle_path)
+        else:
+            vfolders = cls.get_visible_for(pootle_path)
+
+        for vf in vfolders:
             units = vf.units.filter(store__pootle_path__startswith=pootle_path)
             stores = Store.objects.filter(
                 pootle_path__startswith=pootle_path,
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -7,18 +7,28 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from translate.filters.decorators import Category
+
 from django.core.exceptions import ValidationError
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
+from django.utils import dateformat
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 
 from pootle.core.markup import get_markup_filter_name, MarkupField
+from pootle.core.url_helpers import get_editor_filter, split_pootle_path
 from pootle_app.models import Directory
 from pootle_language.models import Language
+from pootle_misc.checks import get_qualitychecks_by_category
 from pootle_project.models import Project
-from pootle_store.models import Store, Unit
-from pootle_store.util import OBSOLETE
+from pootle_statistics.models import Submission
+from pootle_store.models import (QualityCheck, Store, Suggestion,
+                                 SuggestionStates, Unit)
+from pootle_store.util import (calc_total_wordcount, calc_translated_wordcount,
+                               calc_fuzzy_wordcount, OBSOLETE, UNTRANSLATED)
 
 
 class VirtualFolder(models.Model):
@@ -64,6 +74,10 @@ class VirtualFolder(models.Model):
         unique_together = ('name', 'location')
         ordering = ['-priority', 'name']
 
+    @cached_property
+    def code(self):
+        return self.pk
+
     @classmethod
     def get_matching_for(cls, pootle_path):
         """Return the matching virtual folders in the given pootle path.
@@ -75,6 +89,53 @@ class VirtualFolder(models.Model):
         return VirtualFolder.objects.filter(
             units__store__pootle_path__startswith=pootle_path
         ).distinct()
+
+    @classmethod
+    def get_visible_for(cls, pootle_path):
+        """Return the visible virtual folders in the given pootle path.
+
+        Not all the applicable virtual folders have matching filtering rules.
+        This method further restricts the list of applicable virtual folders to
+        retrieve only those with filtering rules that actually match, and that
+        are visible.
+        """
+        return cls.get_matching_for(pootle_path).filter(is_browsable=True,
+                                                        priority__gte=1)
+
+    @classmethod
+    def get_stats_for(cls, pootle_path):
+        """Get stats for all the virtual folders in the given path."""
+        stats = {}
+
+        for vf in cls.get_visible_for(pootle_path):
+            units = vf.units.filter(store__pootle_path__startswith=pootle_path)
+            stores = Store.objects.filter(
+                pootle_path__startswith=pootle_path,
+                unit__vfolders=vf
+            ).distinct()
+
+            stats[vf.code] = {
+                'total': calc_total_wordcount(units),
+                'translated': calc_translated_wordcount(units),
+                'fuzzy': calc_fuzzy_wordcount(units),
+                'suggestions': Suggestion.objects.filter(
+                    unit__vfolders=vf,
+                    unit__store__pootle_path__startswith=pootle_path,
+                    unit__state__gt=OBSOLETE,
+                    state=SuggestionStates.PENDING,
+                ).count(),
+                'critical': QualityCheck.objects.filter(
+                    unit__vfolders=vf,
+                    unit__store__pootle_path__startswith=pootle_path,
+                    unit__state__gt=UNTRANSLATED,
+                    category=Category.CRITICAL,
+                    false_positive=False,
+                ).values('unit').distinct().count(),
+                'lastaction': vf.get_last_action_for(pootle_path),
+                'is_dirty': any(map(lambda x: x.is_dirty(), stores)),
+            }
+
+        return stats
 
     def __unicode__(self):
         return ": ".join([self.name, self.location])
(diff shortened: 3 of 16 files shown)


unho:
@translate/evernote @translate/house Please review.

unho:
First of all a brief description of what is each commit: 

* VFolder: Accept specifying filter by path
  * This is allowing to specify filter by directories (relative to vfolder location) in the JSON file used to add the vfolders. Right now we only are supporting specific file paths (relative to vfolder location). This is unrelated to the rest of the commits and can be safely landed on its own.
* VFolder: Relate new units with existing matching virtual folders
  * This just adds a postsave signal handler for Unit so each time a new unit is created (new units in a file, or units in new files, or units in new language, or even units in a completely newly added project) it is included on all the existing vfolders that it matches. This can also be safely landed on its own.
* Editor: Display priority for current unit
  * This adds the ability to display the top priority for the current unit being edited in the vfolder. Top priority means the higher priority for all the vfolders the unit belongs to. If the unit is not in any vfolder then no priority is being displayed. This can be safely landed on its own.
* Cleanup: Relocate code
  * This just relocates some code in a separate commit to reduce the changeset on the next commit making it easier to review commit by commit.
* VFolder: Display virtual folders stats table on TP overview
  * This adds the ability to display a stats table for the browsable and uncompleted vfolders with priority higher or equal to 1. The table is displayed on the TP overview just above the regular stats table. The vfolders do have any unit on the current location or below it (this means that those vfolders location can be in current location, below it, or above it). Translate links provided in the table point to the editor to translate that specific vfolder restricted to current location, despite the ability to do that is implemented in a later commit.
* VFolder: List all virtual folders on TP overview for admins
  * This just displays more vfolders on the stats table added by the previous commit. This allows admin users to see completed vfolders and also vfolders with any priority, and also not browsable vfolders are displayed.
* VFolder: Display virtual folders in the navbar dropdown
  * This adds the ability to display and select vfolders in the navbar dropdown (the one that lists files and directories). This is only done at TP level, both for overview and editor. The endpoint for the URLs that are set after selecting a vfolder in the editor is implemented in a later commit. The endpoint for overview is not yet implemented, but cleanly falls back to display the vfolders location (adjusted to current TP if necessary), thus not providing any data restricted to a specific vfolder (this will be fixed in a later commit).
* Editor: Filter by vfolder
  * This provides the ability to restrict the units displayed in the editor to those belonging to a specific vfolder. This provides the endpoint to the translate URLs in the TP overview vfolders stats table and also for when chosing a vfolder in the navbar dropdown in the editor.
* Editor: Display priority when we are translating a vfolder
  * This is a followup for previous "Editor: Display priority for current unit" above. It forces to display the priority for the current vfolder being translated in the editor independently of if the current unit has a vfolder with a higher priority.
* Editor: Display priority sorting only when applies
  * As pointed out by @julen it is confusing to display the priority sorting when there are no vfolders in the current location being translated. Priority sorting is also not displayed if we are translating a specific vfolder because all the units have the same priority.

unho:
@translate/house @translate/evernote 

In order to test this locally you will have to follow these steps:

- Ensure you have a project with several files on it (per language). It is recommended to have enough files to include some of them in a vfolder while leaving some of them out
- Craft a JSON file to add some sample vfolders. It is recommended to have two directories. One directory with 4 files at least, and include 3 of them on vfolders (one vfolder with 3 files, and the other two vfolders with 2 files, sharing only one of them). The other directory can be used to specify a single filter to include all its contents in a vfolder. You can add some vfolder using a `{LANG}` placeholder so it matches new languages added after importing the vfolder in the JSON. Make sure you have a file that is matched by several vfolders with different priorities. Ensure that you add not browsable vfolders, and vfolders with priority lower than 1, and vfolders that only match completely translated files.
- Run `add_vfolders` management command to import such vfolders: `$ pootle add_vfolders virtual_folders.json`
- Run the server and go to a path where we know there are some of the files matched by a vfolder
- Check that the vfolders stats table is displayed, that the listed vfolders are all the vfolders that should be displayed, that the numbers match and the translate links are correct. Note that each vfolder counts must only include the files they match in the current location or below it. Remember that not browsable, or completed, or priority below 1 vfolders shouldn't be displayed.
- Switch to admin and check that now you can see not browsable, and uncompleted and vfolders with priority below 1 in the same overview. You might now want to switch back to a regular user in case you are displaying too many vfolders.
- Click on a translate link on the vfolders stats table and ensure that you see in the editor the right value in the navbar dropdown, that the units displayed are only those included in the vfolder in that path. Select the same path to be translated in the editor, but not restricted to a vfolder, and check that the unit number is greater if not all the units in that place are included in the vfolder. Check that it is not possible to sort by priority.
- Open the navbar dropdown and ensure that all the vfolders that match within current project are displayed. Note that not browsable vfolders are not included.
- Open the editor on a location with no vfolders and ensure no priority is being displayed for any of the units shown. Also check that is not possible to select priority sorting.
- Open the editor on a location with several vfolders and ensure the top priority is being displayed for all the units shown. Different priorities should be displayed. Remember top priorities are above and lower at the bottom, so easy way to check is open the editor and check the first units and then jump to the last ones and also check those. Remember to ensure that the top priority is shown for a unit that is included in several vfolders with different priorities. Also check that you can sort by priority.
- Open the overview for the directory included as filter in the JSON and ensure that the numbers in the above table are the sums for all the numbers in the regular table. This should be that way because all the units in the directory must be included in the vfolder.
- Add a new file that matches any of the already existing vfolders, preferably on a directory that didn't have any matching vfolder before. Visit the overview for the directory that includes that file and ensure the vfolder is correctly listed in the vfolders stats table.

dwaynebailey:
@unho can we just land the commits that are unrelated to this PR.  That's: "JSON path" 1ebe9b5 and "Only display priority sorting when we have a vfolder" 5563f2d (that was a regression tracked in its own issue #3652).

Im not sure why we need to cloud this review with those items.

unho:
@dwaynebailey No problem on landing those as soon as I get a gtm on them.

unho:
@translate/evernote @translate/house Moved the first three commits in here to https://github.com/translate/pootle/pull/3669 (Note that we still kept one of them in this PR for now since another commit on this PR depends on it). You might want to review and test PR https://github.com/translate/pootle/pull/3669 first in order to land those commits in master and ease reviewing this PR.

unho:
@dwaynebailey Just for the record "Only display priority sorting when we have a vfolder" fixes a regression but depends on other commits on this PR, so it is not possible to land as is without reworking several commits in the PR in order to make it not depend on others.

So it is going to stay here for now. I moved some of the commits to a separate PR, and I am finishing new more detailed instructions for this PR, so it should be much easier to get everything landed ASAP.

unho:
@translate/evernote @translate/house Updated the testing instructions: https://github.com/translate/pootle/pull/3642#issuecomment-90841310 Now they include a set of files to set up a new project and a JSON file to set the vfolders.

julen:
All breaks when virtualfolder is not installed:
```python
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/core/handlers/base.py, line 111, in get_response
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/transaction.py, line 394, in inner
    return func(*args, **kwargs)
  File ~/dev/translate/pootle/pootle/core/decorators.py, line 107, in wrapped
    return func(request, path_obj, *args, **kwargs)
  File ~/dev/translate/pootle/pootle/core/decorators.py, line 313, in _wrapped
    return func(request, *args, **kwargs)
  File ~/dev/translate/pootle/pootle/core/decorators.py, line 276, in wrapped
    return func(request, path_obj, dir_path, filename)
  File ~/dev/translate/pootle/pootle/apps/pootle_translationproject/views.py, line 143, in overview
    all_vfolders=is_admin
  File ~/dev/translate/pootle/pootle/apps/virtualfolder/models.py, line 133, in get_stats_for
    units = vf.units.filter(store__pootle_path__startswith=pootle_path)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/manager.py, line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/fields/related.py, line 885, in get_queryset
    return qs._next_is_sticky().filter(**self.core_filters)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/query.py, line 691, in filter
    return self._filter_or_exclude(False, *args, **kwargs)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/query.py, line 709, in _filter_or_exclude
    clone.query.add_q(Q(*args, **kwargs))
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1331, in add_q
    clause, require_inner = self._add_q(where_part, self.used_aliases)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1358, in _add_q
    current_negated=current_negated, connector=connector)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1182, in build_filter
    lookups, parts, reffed_aggregate = self.solve_lookup_type(arg)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1120, in solve_lookup_type
    _, field, _, lookup_parts = self.names_to_path(lookup_splitted, self.get_meta())
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1427, in names_to_path
    self.raise_field_error(opts, name)
  File ~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/models/sql/query.py, line 1433, in raise_field_error
    Choices are: %s % (name, , .join(available)))
FieldFieldFieldFieldFieldFieldFieldFieldFieldFieldFieldFieldFieldFieldFicommented_by, commented_by_id, commented_on, context, creation_time, developer_comment, id, index, locations, mtime, qualitycheck, reviewed_by, reviewed_by_id, reviewed_on, revision, source_f, source_hash, source_length, source_wordcount, state, store, store_id, submission, submitted_by, submitted_by_id, submitted_on, suggestion, target_f, target_length, target_wordcount, translator_comment, unitid, unitid_hash
```

dwaynebailey:
@unho @julen and I had a discussion about the tight coupling.

The vfolders code is tightly coupled at the moment and can't be uncoupled without quite a lot of low level work.  Thus [disabling the app](https://github.com/translate/pootle/pull/3642#issuecomment-93690017), as julen did, will fail.  Julen has raised this in other places that at such time of tight coupling with a real use case it presents a good opportunity to work out how to break our coupling.

Since at the moment vfolders don't appear when they are not being used, and since Evernote doesn't use this though might in future, we think its fine to keep this tight coupling for now.  Thus we don't need to go solve this problem at the moment.

So I think as a general rule we have:

1. When we have real use cases its the opportunity to work to uncouple things
2. If its too complex at least implement things such that the default Pootle appears as if the feature is disabled.

julen:
Exactly, specific situations like these should encourage and help everyone think more in specific terms when it comes to allow to inject third party code.

@dwaynebailey I believe we went even further and we are OK to consider this a core feature (realize of the coupling in this specific case: only **2 out of 15** modified files are from virtual folders), providing the feature doesn't show up UI-wise if no virtual folders have been imported.

unho:
@julen According to my reading, the last comments mean that this can be landed as soon as I get some gtm. Correct me if I am wrong, and if not I am waiting for that gtm or some sort of review.

julen:
Just gave this a shot. Note the setup instructions are missing the step to create the project. Here are some issues that I've currently found:

* The numbers I get when drilling down into directories are inconsistent. For example, let's say */af/test_vfolders/browser/chrome/* displays N words needing translation for the *override* directory (and also for its *test-overrides* virtual folder). If I drill down into */af/test_vfolders/browser/chrome/overrides/*, the number of words needing translation for the *test-overrides* folder is M, where M is higher than N.
  * Going back to */af/test_vfolders/browser/chrome/* displays N words needing translation for the *overrides* directory, but M words needing translation for the *test-overrides* vfolder.
* The resources drop-down doesn't display any vfolders as advertised.
* It doesn't look like the latest test you mention should be done in */af/test_vfolders/translate/dom/chrome/#filter=incomplete*. Can you double-check please?
* It might be worth styling the non-browsable virtual folders and those who have *P<1* differently, the same way we do with disabled projects (`is-disabled` class).

unho:
> Just gave this a shot. Note the setup instructions are missing the step to create the project. Here are some issues that I've currently found:

Thanks for testing. That create the project step is on the setup instructions.

> * The numbers I get when drilling down into directories are inconsistent. For example, let's say */af/test_vfolders/browser/chrome/* displays N words needing translation for the *override* directory (and also for its *test-overrides* virtual folder). If I drill down into */af/test_vfolders/browser/chrome/overrides/*, the number of words needing translation for the *test-overrides* folder is M, where M is higher than N.

Yes, this is because of a bug when relating the units if a directory filter is specified. The bugfix has already being landed on master. Now the numbers match. If you are going to test again, please run `VirtualFolder.objects.all().delete()` on the shell and import the JSON again before testing. 

>   * Going back to */af/test_vfolders/browser/chrome/* displays N words needing translation for the *overrides* directory, but M words needing translation for the *test-overrides* vfolder.

Should be fixed now. See above for explanation.

> * The resources drop-down doesn't display any vfolders as advertised.

It works fine here. Can you please test again. Maybe some the value is cached somewhere.

> * It doesn't look like the latest test you mention should be done in */af/test_vfolders/translate/dom/chrome/#filter=incomplete*. Can you double-check please?

The URL `http://127.0.0.1:8000/af/test_vfolders/translate/dom/chrome/#filter=incomplete` is correct.

> * It might be worth styling the non-browsable virtual folders and those who have *P<1* differently, the same way we do with disabled projects (`is-disabled` class).

Done. Also changed it to style as disabled the not browsable projects as well.

dwaynebailey:
@unho is it possible to not confuse things with ``is-disabled`` when the vfolder is not actually disabled ``'is_disabled': (not virtual_folder.is_browsable or virtual_folder.priority < 1)``.  We just want the styling not the confusion.  Because in this case the vfolder is not disabled it is just styled like disabled.

pootle/apps/virtualfolder/models.py
```diff
@@ -75,6 +103,59 @@ def get_matching_for(cls, pootle_path):
...
+                'total': calc_total_wordcount(units),
+                'translated': calc_translated_wordcount(units),
+                'fuzzy': calc_fuzzy_wordcount(units),
+                'suggestions': Suggestion.objects.filter(
+                    unit__vfolders=vf,
+                    unit__store__pootle_path__startswith=pootle_path,
+                    unit__state__gt=OBSOLETE,
+                    state=SuggestionStates.PENDING,
+                ).count(),
+                'critical': QualityCheck.objects.filter(
```
ta2-1: Should we add category=Category.CRITICAL here?

unho: @ta2-1 Yes, and probably also `false_positive=False` as well.

unho:
@dwaynebailey We set `is_disabled` on the dictionary so when rendering the table we set the `is-disabled` styling for the row. See https://github.com/translate/pootle/blob/master/pootle/templates/browser/_table.html#L26 The `is-disabled` class is reused to avoid renaming it or duplicating the definition in the CSS stylesheet.

iafan:
I agree with Dwayne — we'd better not reuse any flags that semantically shouldn't be there. Adding an extra rule in CSS is not a problem. Let's be more explicit.

unho:
@iafan @dwaynebailey Ok, done.

@translate/evernote @translate/house Any other issue? Nits? gtms?

pootle/apps/pootle_project/models.py
```diff
@@ -512,6 +525,13 @@ def _get_code(self, project):
 
 @receiver([post_delete, post_save])
 def invalidate_resources_cache(sender, instance, **kwargs):
+    if instance.__class__.__name__ == 'VirtualFolder':
+        cache.delete_many([
```
julen: This looks wrong, not sure if you tested it: `cache.delete_many()` receives a list of keys to be removed from the cache, but here you have a list comprehension which calls `cache.delete()` (!!!).

Since we already enforce *django-redis*, you can `cache.delete_pattern(make_method_key('Project', 'resources', '*'))`. But please test it.

pootle/apps/pootle_store/views.py
```diff
@@ -753,8 +761,20 @@ def get_edit_unit(request, unit):
...
-    priority = unit.vfolders.aggregate(priority=Max('priority'))['priority']
+    priority = None
+    vfolder_pk = request.GET.get('vfolder', '')
+
+    if vfolder_pk:
+        # If we are translating a virtual folder, then display its priority.
+        try:
+            vfolder = VirtualFolder.objects.get(pk=vfolder_pk)
+            priority = vfolder.priority
+        except Exception:
```
julen: `Exception` is too broad, did you perhaps mean `VirtualFolder.DoesNotExist`?

pootle/apps/pootle_store/views.py
```diff
@@ -753,8 +761,20 @@ def get_edit_unit(request, unit):
...
+            pass
+
+    if priority is None:
+        # If we have no priority, then retrieve the unit top priority, if any.
+        priority = unit.vfolders.aggregate(priority=Max('priority'))['priority']
```
julen: Depending on the meaning of the exception handler above, you might be able to move this up to the `except` block.

pootle/apps/virtualfolder/models.py
```diff
@@ -118,6 +202,35 @@ def clean_fields(self):
...
+        """Return the virtual folder location adjusted to the given path.
+
+        The virtual folder location might have placeholders, which affect the
+        actual filenames since those have to be concatenated to the virtual
+        folder location.
+        """
+        count = self.location.count("/")
+
+        if pootle_path.count("/") < count:
+            raise Exception("%s is not applicable in %s" % (self, pootle_path))
```
julen: This is a `ValueError` to my eyes. Same with the exceptions raised below. Thoughts @jleclanche?

pootle/core/decorators.py
```diff
@@ -127,15 +128,19 @@ def set_resource(request, path_obj, dir_path, filename):
 
     is_404 = False
 
+    # Get a clean pootle path for retrieving the directory or store.
```
julen: What is a *clean* pootle path?

pootle/static/css/style.css
```diff
@@ -2098,6 +2099,10 @@ table.stats .item.dirty .last-updated
     color: #606047;
 }
 
+.vfolders{
```
julen: `{` should go in its own line.

pootle/templates/browser/_table.html
```diff
@@ -23,11 +23,19 @@
...
+    <tr class="item{% if item.is_disabled %} is-disabled{% endif %}{% if item.is_grayed %} is-grayed{% endif %}">
 
       {% if 'name' in table.fields %}
       <td class="stats-name {{ item.icon }}"{% if item.description %} title="{{ item.description|striptags }}"{% endif %}>
-        <a href="{{ item.href }}"><i class="icon-{{ item.icon }}"></i> <span>{{ item.title }}</span></a>
+        {% if item.href %}<a href="{{ item.href }}">{% endif %}
+          <i class="icon-{{ item.icon }}"></i> <span>{{ item.title }}</span>
+        {% if item.href %}</a>{% endif %}
+      </td>
+      {% endif %}
```
julen: This block is irrelevant cleanup, feel free to push it to master directly.

pootle/templates/editor/_toolbar.html
```diff
@@ -4,7 +4,9 @@
   <div class="unit-sort">
     <select id="js-filter-sort">
       <option selected value="default">{% trans "Default order" %}</option>
+      {% if display_priority %}
       <option value="priority">{% trans "Decreasing priority" %}</option>
```
julen: I still believe this wording is not clear enough, and something like *Important first* is easier to understand and aligns with the existing wording for other sorting criteria. Thoughts @iafan @dwaynebailey?

pootle/templates/browser/_table.html
unho: No, it is not. Keeping it.

pootle/apps/virtualfolder/models.py
unho: Yes, changed this.

pootle/templates/editor/_toolbar.html
unho: I agree that is clearer. Will wait for input anyway. That simple change can be landed independently of this PR since it affects code unchanged by this PR.

pootle/apps/pootle_project/models.py
unho: This is obviously wrong.

pootle/templates/editor/_toolbar.html
dwaynebailey: @unho I agree with @julen on this one.  I've struggled with what would be better, @julen's suggestion is the best I've seen yet.

pootle/core/decorators.py
unho: Added a description of this.

unho:
@translate/house @translate/evernote Addressed all the issues. And tested several times the new handler to flush the projects resources cache. Please more issues/nits or gtms.

dwaynebailey:
@unho Testing, not code review.  These issues were spotted.

1. Sorting columns is confusing since we sort the tables together.  Thus you lose priority sorting if you fiddle with name sorting on the files table.  I think they should simply be kept apart in determining sort order.
2. While logged in as admin I noticed that 'test-completely-translated' while complete at the translation level had a critical check failure.  Completed vfolders with critical checks and/or suggestions should be displayed and not hidden.  Since our idea is to prioritise work these should get top attention since they are top priority and they have issues or contributions.
3. In a project marked as disabled it was not possible to browse into the critical checks to try to fix them.  That didn't work in 'test-not-browsable' and 'test-completely-translated'.  I might expect it for not-browsable, but it makes no sense for 'completed'

I'd like to see these fixed but I don't see these as blockers to the feature and could be fixed in master.  gtm from me.

unho:
@dwaynebailey Added issues for those three items.

unho:
@translate/evernote A gtm from your side is all what is needed for this PR to be landed. This has been open for weeks, has been altered several times to address all the raised issues, so I would like to merge it asap and forget about this particular nightmare. *Insert picture of sad cats here* gtm please!

pootle/apps/pootle_project/models.py
```diff
@@ -510,6 +524,24 @@ def _get_code(self, project):
...
+            affected_projects = kwargs['projects']
+        except KeyError:
+            # In case this is pre_delete.
+            affected_projects = Project.objects.filter(
+                translationproject__stores__unit__vfolders=instance
+            ).distinct().values_list('code', flat=True)
+
+        cache.delete_many([
+            make_method_key('Project', 'resources', proj)
+            for proj in affected_projects
```
julen: Honestly I don't think it's worth the complexity (and duplication) of calculating the affected projects: the invalidation is triggered for every vfolder save/delete, it might perfectly be going through several projects already, and calculating the resources is not expensive. Besides, the `vfolder_post_save` would be irrelevant here (there's a duck typing check in place already).

So I'd appreciate if you can use the approach I already proposed in this PR.

unho: Ok, I am going to change this to do as you said on https://github.com/translate/pootle/pull/3642#discussion_r29574110

julen: > and calculating the resources is not expensive.

Correcting my own words: this indeed now has become super expensive compared to the previous way resources were retrieved (half a second vs a few ms)

julen:
Tested this again against a real world dataset, and performance of the overview pages has substantially degraded. For a relatively large TP with two vfolders, this produces ~170 SQL queries which take ~1.3s — among other things, they retrieve/calculate stats from the DB directly. There's also the aforementioned expensiveness of retrieving project resources, which now has gone up to 0.5s from a few ms.

Having said that, we're fine to merge this PR at its current state, providing performance will be reviewed and addressed in a follow-up PR soon, as this is still not ready to be used in a production environment.

pootle/apps/pootle_project/models.py
unho: Then I am keeping this code since hitting all projects would be even more expensive.

unho:
On my local install (numbers after several visits to the same URL), with a Firefox-like project and seven vfolders it takes ~100 SQL queries and the total navigation time (since request to page completely rendered) is 2.76 seconds. On master in the same URL it takes 17 SQL queries and 1.94 seconds. The number of SQL queries can be reduced if we introduce stats caching for vfolders (right now we are always hitting the DB).

Retrieving the project resources takes time but it is being cached, so despite the first visit can take more time (since we are now also adding the vfolders) it should be as fast as before when it comes to retrieve already cached resources. Anyway current code for retrieving the relative path to TP for each vfolder can most likely be improved.

I am merging this PR now since for a no-vfolders scenario these changes shouldn't impact the performance too much.

julen:
> I am merging this PR now since for a no-vfolders scenario these changes shouldn't impact the performance too much.

The impact should be (nearly) zero, otherwise it's a bug.

Issue #3640, opened 2015-03-17
Sidebar layout screwed with imports_exports


julen:
The imports_exports app makes some visual customizations, which break the sidebar's layout:
* the sidebar's height is not adjusted to the height of the content
* an empty sidebar being displayed has been observed under some circumstances too, although I'm not able to trigger it right now.

Check out https://core-translate.stage.evernote.com/ru/clearly_evernote/ as an example of such layout.

dwaynebailey:
@julen in point 1) in your initial bug comment is the white space at the bottom, under the gray import/export section, what you mean when you say "the sidebar's height is not adjusted to the height of the content"?

![screen shot 2015-03-17 at 14 08 19](https://cloud.githubusercontent.com/assets/647438/6688953/22eb747a-ccaf-11e4-9d24-8e91cb91f42b.png)

julen:
It should fit the whole page area, not be trimmed after the sidebar content. Check https://translate.stage.evernote.com/ru/web_evernote/ for the expected layout.

unho:
Picked from an email. Indented comments are from @iafan and the replies are from @dwaynebailey.

> I gave a first look at import/export UI — here are my impressions (I
> might be reiterating something that Julen or you already pointed out):
>
> 1) Side bar should take the full height of the content area despite the
> absence of the project description.
> Example: https://translate.stage.evernote.com/ru/web_evernote/ — here the
> sidebar is rendered correctly, probably because it has extra content
> And here: https://translate.stage.evernote.com/ru/clearly_evernote/ the
> sidebar doesn't take the full height, thus looking odd.

I think that's the item already highlighted in this issue.

> 2) Let's not use any background color in the sidebar. The background
> should always be white.

I thought some separation of purpose might be a good thing here.  But
happy to do white.

> 3) The padding on `#import-export-sidebar` should be 2em (basically, we
> want actions and arbitrary descriptions underneath share the same padding).
>
> So the CSS to fix (2) and (3) will be like this:
>
> ```css
> #import-export-sidebar {
>   padding: 2em;
> }
> ```
>
> 4) I'd recommend being more verbose on the actions, since we seem to have
> plenty of space there. Instead of 'Download', I'd go for 'Download contents
> for offline translation'

Verbose +1.  Though ''Download for offline translation' maybe is better.

> 5) the current upload action UI definitely looks odd. A text input which
> allows you to type something in (?!), but which simply hides the file upload
> form for no good reason, since it looks and behaves even worse than the
[...]

PR #3663, opened 2015-04-08
Fix sidebar look after import_export

unho:
This fixes issue #3640

- Added a new #sidebar-content div to enclose sidebar content in order
  to correctly hide the whole sidebar except the toggle handle
- Moved padding and margin from specific content divs (#staticpage and
  #import-export-sidebar) to that new div
- Changed some texts to provide more useful hints to the users
- Use placeholder instead of input value
- Removed grey background for import-export
- Added some JavaScript to avoid user typing on the fake upload input

### DIFF
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -433,6 +433,13 @@ td.stats-name a:hover > i + span,
     transition: all 0.3s ease 0s;
 }
 
+#sidebar-content
+{
+    margin: 2em auto;
+    padding: 0 2em;
+    max-width: 960px;
+}
+
 #sidebar.sidebar-open
 {
     width: 30%;
@@ -446,12 +453,12 @@ html[dir="rtl"] #sidebar
     box-shadow: 1px 0px 2px rgba(0, 0, 0, 0.2);
 }
 
-#sidebar #staticpage
+#sidebar #sidebar-content
 {
     display: none;
 }
 
-#sidebar.sidebar-open #staticpage
+#sidebar.sidebar-open #sidebar-content
 {
     display: block;
 }
@@ -492,6 +499,12 @@ html[dir="rtl"] #sidebar-toggle
 
     box-shadow: 2px 0px 2px rgba(0, 0, 0, 0.2);
     border-radius: 0 2px 2px 0;
+}
+
+div.sidebar-group
+{
+    margin: 0;
+    padding: 0;
 }
 
 
@@ -940,6 +953,12 @@ ul.errorlist
     padding: 0 2em;
     font-size: 120%;
     max-width: 960px;
+}
+
+div.sidebar-group #staticpage
+{
+    margin: 0;
+    padding: 0;
 }
 
 #staticpage h1
@@ -2122,11 +2141,6 @@ span.email .i
 /*
  * Import-export styling
  */
-
-#import-export-sidebar {
-    background-color: #e5e5e5;
-    padding: 2em 1em;
-}
 
 .upload-form
 {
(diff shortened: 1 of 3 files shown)


unho:
@translate/house @translate/evernote Please review. Tests are also welcome. Original issue is https://github.com/translate/pootle/issues/3640

jleclanche:
lgtm, didnt test

pootle/apps/pootle_translationproject/views.py
```diff
@@ -25,7 +25,7 @@
 from staticpages.models import StaticPage
 
 
-ANN_COOKIE_NAME = 'project-announcements'
+ANN_COOKIE_NAME = 'sidebar-cookie'
```
julen: If renaming, mind doing it to something like `pootle-overview-sidebar`? Feel free to push that to master directly.

julen: And at the same time `ANN_COOKIE_NAME` should be renamed as well; `SIDEBAR_COOKIE_NAME` or similar.

julen:
FTR, the layout was fixed in 8f407a91, this PR makes some minor visual adjustments. I have added some related comments in #3640.

pootle/apps/pootle_translationproject/views.py
unho: @julen Ok. I think this can safely go in a separate commit.

dwaynebailey:
@unho from what I understand this doesn't address the concerns in #3640 about the upload widget.

unho:
@iafan @julen @dwaynebailey Updated PR to address your concerns. Please review and test again.

dwaynebailey:
@unho I tested specifically the changed upload behaviour and it's gtm

pootle/static/css/style.css
```diff
@@ -2121,15 +2127,19 @@ span.email .i
...
+{
+    margin: 0;
+    padding: 0;
 }
 
-#upload-form {
-    margin-top: .5em;
+.action-link
+{
+    display: block;
```
julen: why not make it more intuitive and wrap the inline element within a block element such as `<ul><li>`, `<p>`, or `<div>`?

julen:
* The link underline goes beyond the text boundaries, so the initial whitespace also displays an underline which is not nice. You can check how we work around that in the overview file/dir listing icons.
* When there are other hooked up elements in the sidebar, it'd be good to display some separator to make it clear the contents are unrelated.
screenshot

 
These two small issues aside looks good to merge

unho:
@julen Updated to address your latest concerns. I would appreciate if you can quickly review and test and maybe gtm the PR.

pootle/templates/browser/overview.html
```diff
@@ -78,13 +78,21 @@ <h4 class="clear">{{ item.title }}</h4>
...
+  <div id="sidebar-content">
+    {% block sidebar_content %}
 
-  {% if announcement %}
-  {% include "staticpages/_body.html" with page=announcement %}
-  {% endif %}
+    {% if announcement %}
+    {% comment %}Dirty hack to place a separator while we don't have a way to
+    tell if after an existing item we have another one.{% endcomment %}
+    {% if display_download or upload_form %}
```
julen: I'd rather not couple the sidebar to 3rd party specifics and live without a separator instead.

julen: Having said that, it's still possible to add the separator by having the content for each content type within a block, and using CSS have all those blocks, except the last one, to render a `border-bottom`.

Issue #2219, opened 2012-06-15
Use locale aware number throughout Pootle


dwaynebailey:
When we have larger numbers it is hard to read them so using locale aware numbers like 63,571 or 63.571 makes it easier for people to read.

unho:
@dwaynebailey Is this a duplicate of #3410 ?

julen:
Closing as dupe.

PR #3664, opened 2015-04-08
Add locale aware formatting to stats numbers

### LINKED ISSUE #3410: Add locale aware formatting to numbers
On some of our pages numbers are reaching into the thousand and million count which makes the numbers harder to read without locale specific formatting, as can be seen in this navbar on the front page of the staging server:

screen shot 2014-12-06 at 04 07 20

We should add locale aware formatting that would change e.g.

1000 -> 1,000 or 1 000 or 1.000
1000000 -> 1,000,000 or 1 000 000 or 1.000.000

In a locale aware away so that it is easy to read.

unho:
This relies on browser support, so not always works. It defaults
to English locale so no more long numbers without separators are
displayed.

Fixes #3410.

### DIFF
--- a/pootle/static/js/stats.js
+++ b/pootle/static/js/stats.js
@@ -40,6 +40,7 @@ var stats = {
   init: function (options) {
     this.retries = 0;
     this.pootlePath = options.pootlePath;
+    this.localeCode = options.localeCode;
     this.processLoadedData(options.data, undefined, true);
 
     $('td.stats-name').filter(':not([dir])').bidi();
@@ -77,7 +78,7 @@ var stats = {
   },
 
   updateTranslationStats: function ($tr, total, value, noTotalDefault) {
-    $tr.find('.stats-number a').html(value);
+    $tr.find('.stats-number a').html(value.toLocaleString(this.localeCode));
     $tr.find('.stats-percentage span').html(
       nicePercentage(value, total, noTotalDefault)
     );
@@ -86,14 +87,14 @@ var stats = {
 
   updateAction: function ($action, count) {
     $action.css('display', count > 0 ? 'inline-block' : 'none');
-    $action.find('.counter').text(count);
+    $action.find('.counter').text(count.toLocaleString(this.localeCode));
   },
 
   updateItemStats: function ($td, count) {
     if (count) {
       $td.removeClass('zero');
       $td.addClass('non-zero');
-      $td.find('a').html(count);
+      $td.find('a').html(count.toLocaleString(this.localeCode));
     } else {
       $td.find('a').html('');
       $td.addClass('zero');
--- a/pootle/templates/browser/overview.html
+++ b/pootle/templates/browser/overview.html
@@ -124,6 +124,7 @@ $(function () {
   PTL.search.init();
   PTL.stats.init({
     pootlePath: "{{ pootle_path }}",
+    localeCode: "{{ LANGUAGE_CODE }}",
     data: {{ stats|safe }}
   });
 


unho:
@translate/house @translate/evernote Please review and test.

dwaynebailey:
Generally gtm with some concerns:

1. We use numbers in other places e.g. the dashboard, the store drilldown and those don't seem to be handled by this patch.  We really should fix this globally if its as easy as this and we can then drop some crufty Python code that does the same.
2. What happens in the case of zero?
3. This ``toLocaleString`` has some odd behaviour if we don't supply e.g. ``locale='ar'`` (see [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) where it seems it could return OS locale or English if its undefined).  If we supply the locale, which we should know already since we can present localised UI, then we are in a safer position and we are also in a position to allow user override of the localised UI in future.

unho:
> Generally gtm with some concerns:
>
> 1. We use numbers in other places e.g. the dashboard, the store drilldown and those don't seem to be handled by this patch.  We really should fix this globally if its as easy as this and we can then drop some crufty Python code that does the same.

Since https://github.com/translate/pootle/issues/3410 is not clear about scope, and since I am not sure this is a completely acceptable fix, I decided to only try to fix the most obvious place where this happens: the overview. Also we have https://github.com/translate/pootle/pull/3664 which seems to have a much broader scope.

Which crufty Python code?

> 2. What happens in the case of zero?

0 is displayed.

> 3. This ``toLocaleString`` has some odd behaviour if we don't supply e.g. ``locale='ar'`` (see [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) where it seems it could return OS locale or English if its undefined).  If we supply the locale, which we should know already since we can present localised UI, then we are in a safer position and we are also in a position to allow user override of the localised UI in future.

To my understanding it tries the defined locales in browser, then defaults to OS, then English.

You are right that we need to first try the locale being used for Pootle, which can be different from all of those. But in the case we ask for a locale that is not present on the client computer this is going to default to any locale present in client computer. I have been able to replicate this problem with Chromium where I couldn't use Galician locale, where for Firefox or Opera it worked fine.

dwaynebailey:
@unho

1. https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/views/admin/dashboard.py#L33-L41

2. Try 0 in a non-Latin locale please like Arabic, that was what I was hinting at
3. Can you actually check what the browser does

unho:
> 1. https://github.com/translate/pootle/blob/2b5fde47/pootle/apps/pootle_app/views/admin/dashboard.py#L33-L41

Thanks. As I said the issue this PR is trying to fix is focused on overview (as I understand it). So will leave these bits for a later PR focused on the more broader scoped issue #3664 .

> 2. Try 0 in a non-Latin locale please like Arabic, that was what I was hinting at

I have altered the code to pass the language code to `toLocaleString`. Now I am getting real arabic numbers (or some characters I don't understand but that I suppose that are arabic numbers). This is both for 0 or numbers like 1243.76 or 1450035.

Also I have been able to replicate again that it falls back to english if the language is not supported by the browser (the case for Galician, despite it only fails on Chromium 41, which I am using over Chrome because the later provides Galician UI, and also happens in Opera 28).

Maybe not important, but I couldn't change the UI language on Pootle since that is now gone, so tried by changing the browser UI language.

> 3. Can you actually check what the browser does

See reply above.

unho:
For the record, another reason why this PR doesn't has a broader scope is to test if this approach is valid and works fine, and thus can be later applied to all Pootle when dealing with #2219.

unho:
@translate/evernote @translate/house Any further critic, or review or gtm is appreaciated. If your are not ok with this approach we can just drop this PR. Waiting for more input.

iafan:
Instead of writing `foo.toLocaleString(this.localeCode)`, can we probably create some wrapper function, like 'nloc(foo)'? I guess we will be using it a lot throughout the UI. For future static use, it makes sense to add something like `data-type="number"` to the wrapper containers and then apply `nloc()` function to all elements on the page.

julen:
Under the hood, [`toLocaleString()` creates instances of `Intl.NumberFormat()` objects](http://www.ecma-international.org/ecma-402/1.0/#sec-13.2.1), which are [known to be expensive](https://github.com/yahoo/react-intl/issues/27#issuecomment-61148808), and our overview pages are full of numbers.

I'd favor discarding this PR for the time being, and tackle the issue generally for the FE code using [already-established](https://github.com/yahoo/intl-format-cache) [tools](https://github.com/yahoo/intl-relativeformat).

unho:
Closing this.

@julen Can you please look into fixing this issue with your proposed approach? Since people is asking for a non ad-hoc solution I think both #2219 and #3410 could be addressed at the same time by such approach.

Issue #3518, opened 2015-01-20
Implement UI for new authentication workflow


julen:
We need to implement the UI to make the new authentication workflow (#3517) a breeze to use. See the attached mockups for a detailed overview of where we want to reach.

![sign-in-flow-v2_gh](https://cloud.githubusercontent.com/assets/16768/5070768/8bf8305a-6e6a-11e4-8673-08968bf6be20.png)
![sign-in-flow-v2_gh_2](https://cloud.githubusercontent.com/assets/16768/5070770/8f25765c-6e6a-11e4-9b87-26b4418ebe2c.png)

julen:
As we see in the mockups, the new UI can be triggered from any page, so it's not tied to an specific context. At the same time, it's a requirement being able to trigger the UI from external URLs, because:

1. they will be present in verification/password reset emails, and
2. in any situation it allows to point users in the right direction

For these scenarios where someone is navigating into Pootle using an externally-provided URL, I'd propose using the welcome page as a context.

The caveat would be if someone closes the popup when accessing one of the auth-specific URLs, the user is left off with a URL in the `/accounts/` namespace but the displayed content refers to `/`, the welcome page.
In the ideal situation, this can be addressed by wrapping auth URLs in a client-side router, so for example with a router mounted at `/`, a URL such as `/signup/` would trigger the UI and it would be a façade that knows how to talk to `/accounts/signup/` behind the scenes, and also knows it needs to restore the URL to `/` when the view has been unmounted.

The ideal situation provides the best UX but at the current state of things it would take longer to implement too. Having said this, personally I'd live with the caveat for now and propose the ideal scenario as a future improvement.

Any other suggestions or thoughts on the matter? @iafan @dwaynebailey

iafan:
What I was envisioning here is to use hash for that, since sign-in popups are shown by the client code. So if you are on any page, for example, `https://translate.evernote.com/ru/kb_evernote/translate/articles/#filter=incomplete` and you click on 'Sign In', the popup will appear and the URL (it's hash part) will be rewritten like this: `https://translate.evernote.com/ru/kb_evernote/translate/articles/#filter=incomplete&sign-in`

Similarly, if one goes to the above URL from any external link, they will see the page underneath + the sign-in popup. This will allow us to send out account activation links pointing to the same URL the user was on, so that the flow is not interrupted.

This means that we won't need any real auth-specific URLs. If the sign-in UI can be triggered from any page, why do we need any dedicated URLs?

[...]

Issue #3517, opened 2015-01-20
Implement new authentication workflow


julen:
UI aside, we need to be able to support the following login, registration and password reset workflow:

![sign-in-flow-v2_gh](https://cloud.githubusercontent.com/assets/16768/5070768/8bf8305a-6e6a-11e4-8673-08968bf6be20.png)
![sign-in-flow-v2_gh_2](https://cloud.githubusercontent.com/assets/16768/5070770/8f25765c-6e6a-11e4-9b87-26b4418ebe2c.png)

It's unclear if allauth supports the following, which are needed:
* Linking and unlinking of social providers from users. We currently provide the option to link/unlink a user from an EN account in the user settings page, and this needs to be kept. (not covered in the mockups @iafan, it probably should).
* Talking to the backend via XHR requests ([it doesn't seem to be supported](http://stackoverflow.com/questions/16162108/implementing-ajax-requests-response-with-django-allauth)).

jleclanche:
I filed an issue on django-allauth asking for confirmation on how to unlink social accounts. i suspect it's a matter of deleting the SocialAccount row but there may be more to it.

julen:
Just checked again the state of XHR and [according to the tests](https://github.com/pennersr/django-allauth/blob/master/allauth/account/tests.py) they implemented this in some views (this happened after the previously-linked SO question), so let's make sure all the workflow can be covered.

julen:
There are more workflow-related questions to be answered though.

On the one hand, when someone tries to log in with a social service for the **first time**, it's not clear if it needs to:
1. Display an error if there's no user linked to the social account. This involves first creating a user, then manually linking to it.
2. Automatically create a new user based on the social service data. I wonder if some providers may not be suitable for this step as they might not offer all of the necessary information to create a Pootle user?
3. Offer to link the social service to an existing user

The current approach at EN is a combination of 3 & 2: we want actual EN users to be the translators, so they can only register using EN, and on the first login we *offer* them to link the social service to an actual Pootle user (3). If this step is skipped, a new Pootle user is created automatically using the data provided by the social service (2).

On the other hand, the mockups don't describe how the interaction with the external auth will happen. This always implies visiting an external page to perform authentication, and it can either redirect the current page to the social service or it can open a popup with the auth request of the social service. An example of a clean and non-disruptive workflow of the latter can be experienced at stampsy.com.

iafan:
[...]

Issue #3515, opened 2015-01-20
Adjust Evernote auth backend to the allauth branch


ta2-1:
We need to take the Evernote auth backend out of core and integrate it in the allauth branch. The natural way for doing so would be to write an allauth social provider, but the steps to do so are unclear (no docs, just some provider examples) and have other effects.

Implement the backend as an allauth provider
---

Pros:
+ It uses the same mechanisms as core, better integration

Cons:
- Allauth uses its own models/DB tables for keeping the link between social providers and users
- Need for data migrations 

Keep evernote auth backend
---
Pros:
+ Already implemented, tested and deployed
+ No need for data migrations

Cons:
- It *might* distance itself a little bit from the auth approach core has taken
- Need to hijack URL patterns.

julen:
Also note that whatever approach we decide to take, we need to take the existing code away from the master branch, as well as adjust the requirements that this has imposed until now.

unho:
@julen Can't your code stay? I mean, is it not possible to alter the build scripts to exclude that code and get clean packages?

julen:
@unho I don't see much value in keeping our code there as it's only valid for us.

julen:
We had an attempt to enable our auth backend side-by-side with allauth, and it works, but measuring the risks and potential issues we have decided to take the way to implement an allauth provider.

In any case, the documentation in allauth in this regards is basically null, and since our backend is not similar to any available out there, we'd appreciate any type of help to move this forwards.

jleclanche:
@julen Keep in mind that when I originally suggested allauth, I thought your evernote auth was oauth2-based.

I was having this discussion the other day actually and I don't understand what the benefit for you is not to use the public evernote oauth system.

ta2-1:
@jleclanche, is there ready allauth provider for Evernote?

jleclanche:
@ta2-1 Not unless they added one recently, but seeing as EN uses oauth2 as a provider, it would be a mere few lines of code to write one (basically just settings).

ta2-1:
I haven't got (added or not). 
We found this issue https://github.com/pennersr/django-allauth/issues/556
It's closed and there no Evernote provider here https://github.com/pennersr/django-allauth/tree/master/allauth/socialaccount/providers.

julen:
Sorry hit the button accidentally.

dwaynebailey:
[...]

Issue #3622, opened 2015-03-05
Retrieve CSRF token from cookie


julen:
On the client-side for every XHR request the CSRF token is retrieved from the DOM. This is troublesome, because:
* It relies on having the token available in the current document's DOM structure, therefore tightly coupling it.
* When the token is refreshed, DOMs from other tabs have no clue about the change. This results in the need for some solution like #3620.

If we retrieve the token directly from the cookie we avoid these problems, and actually [it's suggested](https://docs.djangoproject.com/en/1.7/ref/contrib/csrf/#ajax) to deal with the CSRF token this way over relying on the DOM.

PR #3667, opened 2015-04-09
Allauth-based revamped auth workflow

### LINKED ISSUE #3343: Core: Reimplement allauth

julen:
This PR introduces the allauth-based authentication backend, and implements the new workflow and UI for it.

Fixes #3343, #3515, #3517, #3518, #3728, #3737, #3774.

**IMPORTANT NOTES**
* It's **highly suggested** to migrate existing users to Allauth's own way to manage email addresses.
* The commit messages have issue references with stripped `#`s to avoid spam. They'll be brought back to normal immediately before merging happens.
* The PR doesn't cover:
  * Email management (add/change/remove addresses)
  * Explicit account linking/unlinking features

## Workflow testing

Before testing, ensure you are up-to-date with npm dependencies (`npm install` within *pootle/static/js*) and make sure to build JS modules (`manage.py webpack --dev`).

An outline of possible tests of the workflow follows. Added to this, triggering form validation errors anywhere in the process can help too.

1. Sign In
  * Confirmed user (signed up, verified email address)
     * signs user in
        * redirects to
          * previous URL in case `next` was specified
          * `settings.LOGIN_URL` otherwise
  * Unconfirmed user (signed up, didn't verify email address yet)
    * re-sends activation email
    * re-check this after a cooling down period of a few minutes; prevents impatient users from being bombarded with emails
  * Unconfirmed user (existed prior to allauth, didn't verify email address in allauth) — *this should ideally be avoided by migrating users' data*
    * sends activation email
    * click email link to verify email and confirm user account
  * Inactive user (admin deactivated user, `is_active = False`)
    * trying to sign in will display the inactive screen
2. Sign Up
  * Provide new user data
    * sends activation email
  * Click email link to verify email and confirm user account
    * signs user automatically in
    * redirects to `settings.LOGIN_REDIRECT_URL`
3. Password Reset
  * Request password reset
    * sends reset link in an email
  * Click reset link in the email
    * reset password
      * resets password and signs user in, redirecting to `/` (unfortunately cannot redirect to `settings.LOGIN_REDIRECT_URL` due to allauth's limitation)
    * sign out
    * sign in with new password
  * Open invalid/outdated/already-reset password reset link
    * displays message of invalid reset link

If you feel brave you can also test some social authentication provider (just enable a provider by adding it to `INSTALLED_APPS`; it'll show up), but note that that area hasn't been fully addressed yet so YMMV.


dwaynebailey:
Did a test install and tested, didn't review the actual code.  Bullets in no particular order:

* "Sign In With..." header on Sign in popup - it is odd to see 'With...' when you only have one option. It makes sense when there are multiple signup methods.  But when there is one it just seems odd.
* ``initdb`` - default admin user is blocked as it needs email verification, we should by default have a working admin user after initdb.  Admin has no email address so that could be a problem with allauth.
* Reset password.  If passwords are different the button is enabled.  If you reset with different passwords then we stay on the page.
* Login/Logout - the process is long, and no indication that we're waiting for it complete.
* Forgotten password - sticks on the email page, we don't get to p6.5 and p7
  * Email and link arrive and is clicked on
  * Resetting password got stuck on the password reset page
  * This causes the whole site to lock
  * Forcing the reload of another page and we seem to be working again and the password is reset.

jleclanche:
> Reset password. If passwords are different the button is enabled. If you reset with different passwords then we stay on the page.

I don't understand?

dwaynebailey:
@julen finished my review

julen:
Thanks for the extensive review, @dwaynebailey!

I have pushed an update to the PR, which include:
* Rebased against master, which has fixes for #3607 and #3622. These two were affecting your testing as confirmed on IRC. If you further experience either a screen that does nothing/freezes, or some page being displayed briefly when following a link, that's a bug.
* The *Sign In With...* header now adapts to *Sign In*.
* Visual adjustments for email addresses and text.
* A message is displayed once sign in has been successful but we are waiting to be redirected.

Besides that, and reading your feedback, it's worth noting that generally speaking,
* It might not be obvious at first, but allauth differentiates between unverified and inactive users. The former is allauth's own way to tell whether users have verified their email address or not, whereas the latter is `User` model's flag to disable a user entirely — its semantics are not altered so admins can permanently block users. I've got the impression you confused these two.
* Resending activation buttons are not necessary [as outlined in a previous comment](https://github.com/translate/pootle/issues/3517#issuecomment-74875092).

Re. `initdb`, feel free to propose any necessary code changes; it's essentially up to you whether you want to force to input a valid email via `createsuperuser` or use a fake email address (note in this case email-based workflows will be *broken*). In any scenario you'll need to create an entry for allauth's `EmailAddress` model with `verified=True`, as it keeps track of them as separate entities.

dwaynebailey:
@julen started to test socialauth.  The difficult part is that we need Django Admin accessible to setup the Client ID and secret key. How would I do that?

julen:
We didn't have the need to do so because of the nature of our own provider.

While django admin can be added by any custom install, it still requires to hook in the urlconf. Personally I wouldn't be comfortable to pull django admin because of this... but I'm afraid I don't have a specific answer on the matter.

dwaynebailey:
@translate/core if you want to test socialauth the best is to follow these [instructions](http://www.sarahhagstrom.com/2013/09/the-missing-django-allauth-tutorial/).  They use SQL instead of Django Admin to set things up.  I guess that is what we should advise admins.  I used GitHub which was pretty easy, just [register](https://github.com/settings/applications/new) your app against localhost:8000.

**Github fields**
Application name: whatever you want
Homepage URL: http://localhost:8000
Application description: whatever you want
Authorization callback URL: http://127.0.0.1:8000/accounts/github/login/callback/

**SQL for the above**
```sql
UPDATE django_site SET DOMAIN = '127.0.0.1:8000', name = 'Pootle' WHERE id=1;
INSERT INTO socialaccount_socialapp (provider, name, secret, client_id, `key`)
 VALUES ("github", "GitHub", "---Client-Secret-from-above---",
 "---Client-ID-from-abive---", "");
INSERT INTO socialaccount_socialapp_sites (socialapp_id, site_id) VALUES (1,1);
```

I'm sure you can follow similar things with other Social Auth backends.

julen:
This is now complete, waiting for :bug:s, there have to be many :smiley:

julen:
Forgot to mention that I still want to look at polishing some of the UI interactions, especially when it comes to providing feedback to the user.

ta2-1:
does behavior on `/accounts/password/change/`page relate to this PR?
It has changed my password without any notification.

ta2-1:
I didn't received any notification as "password has been changed successfully" after I changed my password by the reset-password link. Is it ok?

julen:
> I didn't received any notification as "password has been changed successfully" after I changed my password by the reset-password link. Is it ok?

If you're referring to an email notification no, there's none. Other than that, UI-wise we haven't invested in this feature, and there are no regressions AFAICT.

ta2-1:
@julen I meant here https://github.com/translate/pootle/pull/3667#issuecomment-103007533 that old password change page is still avalable in http://pootle.server/accounts/password/change/. We discussed that a password can be changed via reset-link only.

julen:
> We discussed that a password can be changed via reset-link only.

Can you give me a pointer? I'm not aware of anything like this.

Also, password change is not covered in this PR besides simply replacing `django.contrib.auth`'s password change views with Allauth's owns.

dwaynebailey:
@julen complete a high level review again of normal allauth i.e. not socialauth.

## General issues
I don't see any as critical to fix.

- If email backend is on SMTP and it doesn't work then we block on login i.e no
  failure.  Not sure what correct failure would be for a user.
- Shouldn't we just use the change password popup instead of what we have on
  Settings->Password. Or just drop that and make all password changes need to happen on the login popup.
- Can't create a user without a password - future feature to be able to create
  a user that the user needs to verify and set a password.  Its a bit odd being
  an admin and setting the password.

# Worked

- Created a superuser - worked
- Got verification email and then logged in - worked
- Changed password from users settings page -worked
- Signout - worked
- Reset password via email - worked
- Created a new user through the UI - worked

dwaynebailey:
@julen here are my notes from social auth testing

## Social auth

- It's still a struggle to setup easily - but my instructions helped
- I used Github auth
- This [traceback](https://dpaste.de/zFw9) -- MultipleObjectsReturned: get() returned more than one User -- it returned 4! -- was fixed and verified.

- Tried normal login after failure above
  - Got a weird mangled error over the popup but logged in - I wouldn't worry about that.

- Faked around the 4 blank emails by giving them all addresses

- Logged in with social auth
  - http://localhost:8000/accounts/social/signup/ - this page is unstyle, seems its not handled by our workflow
    - I guess then we need to be clear in the release notes that deployments will need to style pages themselves if they're using social auth.
  - Do we really need the "Are you human page" after signin with Github?  I'd say no since we have already authenticated via the social auth.
  - http://localhost:8000/accounts/password/set/ - now also unstyled.  Is this setting a password on Pootle or Github?

- If I log in to Github but already have a user with that email on Pootle
  - I get a rejection that says I should login first then link accounts
  - I login
  - But then have no way to link the account
  - Works fine if I associate it with a new email, don't need to validate either,

- If I'm on the admin page there is now way to know that a user is linked to  social auth in any way.

## Issues

1. Difficulty in setup/setup instructions needed
2. Unstyled pages need styling or release note/docs mention.
3. Linking account isn't present
4. No indication on user admin page of any links between accounts and social auth.

dwaynebailey:
@julen from a normal auth perspective this is :+1: 

For social auth there are some issues which we handle in ``master`` and/or documenting the limitation in the docs and release notes.

If  social auth is :+1: for Evernote then I think lets land this baby.

julen:
> * If email backend is on SMTP and it doesn't work then we block on login i.e no failure. Not sure what correct failure would be for a user.

Does it result in an unhandled traceback?

> * Shouldn't we just use the change password popup instead of what we have on Settings->Password. Or just drop that and make all password changes need to happen on the login popup.

I'm fine with whatever provided there's agreement.

> * Can't create a user without a password - future feature to be able to create a user that the user needs to verify and set a password. Its a bit odd being an admin and setting the password

While somehow valid, this is admin/users responsibility and it has always been the case, hence it's unrelated to this PR.

julen:
>     1. Difficulty in setup/setup instructions needed

I reference here [a previous comment of mine](https://github.com/translate/pootle/pull/3667#issuecomment-99758994). Feel free to propose any patches on allauth to offer an alternative (CL) interface to manage/add social app instances.

>    2. Unstyled pages need styling or release note/docs mention.

The workflow that we aim to support and cover is completely styled — the pages you mention don't fall in there. If you want/plan to support extra workflows, you can work on that separately (unless you think it's a blocker).

>    3. Linking account isn't present

In the workflow we have implemented, linking happens when someone tries to sign in using an account that reports an email that already exists on the system.

There's #3553 for providing an explicit UI for such, although @iafan might argue this is not necessary. Let any related discussion happen over there.

>     No indication on user admin page of any links between accounts and social auth.

Also, while nice to have, I believe this falls out of the scope of what we're trying to provide here.

dwaynebailey:
>
>
>    - Shouldn't we just use the change password popup instead of what we
>    have on Settings->Password. Or just drop that and make all password changes
>    need to happen on the login popup.
>
>  I'm fine with whatever provided there's agreement.
>
I'm trying not to scope creep here.  So I'm happy to leave this out of it.

My choices would be to expose the JS allauth password change popup that is
part of the existing workflow.

I'd like to hear @iafan's views on this item.


>
>    - Can't create a user without a password - future feature to be able
>    to create a user that the user needs to verify and set a password. Its a
>    bit odd being an admin and setting the password
>
>  While somehow valid, this is admin/users responsibility and it has
> always been the case, hence it's unrelated to this PR.
>

Agreed, not in scope


-- 
Dwayne

*Translate*
+27 12 460 1095 (work)

dwaynebailey:
> 1. Difficulty in setup/setup instructions needed
>
>  I reference here a previous comment of mine
> <https://github.com/translate/pootle/pull/3667#issuecomment-99758994>.
> Feel free to propose any patches on allauth to offer an alternative (CL)
> interface to manage/add social app instances.
>

My current concern is not social auth so we won't be looking at simpler
models to solve that now.  My concern is that there are at least some notes
to guide someone who wants to use it.  Those are dramatically missing at
the moment.


>
>    1. Unstyled pages need styling or release note/docs mention.
>
>  The workflow that we aim to support and cover is completely styled — the
> pages you mention don't fall in there. If you want/plan to support extra
> workflows, you can work on that separately (unless you think it's a
> blocker).
>
Social auth isn't a blocker for me.  I would appreciate some pointers as to
where and how that would be styled in the same way that it is a concern
around setup.

We can slowly add those missing pages as required later.



>
>    1. Linking account isn't present
>
>  In the workflow we have implemented, linking happens when someone tries
> to sign in using an account that reports an email that already exists on
> the system.
>
> There's #3553 <https://github.com/translate/pootle/issues/3553> for
> providing an explicit UI for such, although @iafan
> <https://github.com/iafan> might argue this is not necessary. Let any
> related discussion happen over there.
>
> No indication on user admin page of any links between accounts and social auth.
>
>  Also, while nice to have, I believe this falls out of the scope of what
> we're trying to provide here.
>
I'm happy to agree this is not in this scope.  And can be dealt with
elsewhere.


-- 
Dwayne

*Translate*
+27 12 460 1095 (work)

julen:
>>> 1. Unstyled pages need styling or release note/docs mention.
>>
>>  The workflow that we aim to support and cover is completely styled — the
>>  pages you mention don't fall in there. If you want/plan to support extra
>>  workflows, you can work on that separately (unless you think it's a
>>  blocker).
>
> Social auth isn't a blocker for me.  I would appreciate some pointers as to
> where and how that would be styled in the same way that it is a concern
> around setup.

Not sure I get your point, but I'm afraid you cannot set some flag and have that workflow *automagically* supported in the code unless you actually implement the views client-side. Of course, you can create templates for the workflow views that this PR doesn't intend to support, therefore accommodating to Pootle's layout/look, but the offered UX will be average.

dwaynebailey:
@julen so if I hear you right.  Its not a matter of missing templates (although they are missing) but rather that that part of e.g. a Github workflow would need to be implemented in JS code?

julen:
Let's put it this way: the client-side code in the PR doesn't cover anything that falls out of the screens you see in the mockups (I can only think of a situation where it'd be possible to hit such scenario, which is using a provider that doesn't provide an email address; or not requiring an email address at all).

If you guys are willing to support alternative workflow steps while preserving the UX, this needs to be implemented client-side obviously. The alternative option is just including templates, but that results in an inconsistent UX.

ta2-1:
@julen I'm not able to login with existing user and password when Chrome incognito mode is enabled.
I got this:
```
Failed to load resource: https://translate.stage.evernote.com/accounts/login/?next=%2F 
the server responded with a status of 403 (FORBIDDEN)
```

julen:
Right, @jleclanche already reported that on IRC a while ago which I forgot about — filed #3774.

ta2-1:
It's allowed to change user email in admin UI.
I set the same email for two different users. Then I set the same email to EN user. Then I tried to sign in with EN, then reset password and... got two different reset password links.

julen:
Could you explain the steps to reproduce that? It's unclear if you signed in at all and therefore in which context you requested to reset your password.

Also, what is specifically an *EN user* and its email?
Accidents aside, why would you want to set the same email to two different users?

ta2-1:
@julen it doesn't make sense to set the same email to two users. 
But it's possible.
How is a linked row in `account_emailaddress` changed if I change user email? I suspect that we don't touch it. Do we have unique constraint on email field there?
I tested how social auth works with such duplicated emails if some uncovered case can be caught.
If we address this constraint I believe the case I described can't be reproduced.

jleclanche:
`ACCOUNT_UNIQUE_EMAIL` is set to True by default: https://django-allauth.readthedocs.org/en/latest/configuration.html

julen:
We quickly discussed this and everything boils down to #3737.

PR #3684, opened 2015-04-22
Bring quality checks documentation within Pootle

unho:
This staticpage also have to be created on the upgrade to 2.6. Working on that.

@translate/evernote @translate/house Please review.

Ideally Evernote checks would use the same approach it is used for TTK checks: retrieve description from each check docstring.

Also note that this doesn't handle the scenario where TTK is upgraded and new checks are added, or removed or changed.

### DIFF
--- a/pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
+++ b/pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import codecs
+import os
+
+from translate.filters.checks import TeeChecker
+
+# This must be run before importing Django.
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from django.conf import settings
+from django.core.management.base import CommandError, NoArgsCommand
+
+from pootle_misc.checks import check_names, excluded_filters
+
+
+class Command(NoArgsCommand):
+    help = 'Regenerate the Translate Toolkit quality checks descriptions.'
+
+    def handle_noargs(self, **options):
+        """Regenerate all the Translate Toolkit quality checks descriptions.
+
+        The resulting HTML is save on the 'checks/_ttk_descriptions.html'
+        template so it can be rendered when Pootle displays the quality checks
+        descriptions page.
+        """
+        try:
+            from docutils.core import publish_parts
+        except ImportError:
+            raise CommandError("Please install missing 'docutils' dependency.")
+
+        def get_check_description(name, filterfunc):
+            """Get a HTML snippet for a specific quality check description.
+
+            The quality check description is extracted from the check function
+            docstring (which uses reStructuredText) and rendered using docutils
+            to get the HTML snippet.
+            """
+            # Provide a header with an anchor to refer to.
+            description = ('\n<h3 id="%s">%s</h3>\n\n' %
+                           (name, unicode(check_names[name])))
+
+            # Clean the leading whitespace on each docstring line so it gets
+            # properly rendered.
+            docstring = '\n'.join([line.strip()
+                                   for line in filterfunc.__doc__.split('\n')])
+
+            # Render the reStructuredText in the docstring into HTML.
+            description += publish_parts(docstring, writer_name='html')['body']
+            return description
+
+        self.stdout.write('Regenerating Translate Toolkit quality checks '
+                          'descriptions.')
+
+        # Get a checker with the Translate Toolkit checks. Note that filters
+        # that are not used in Pootle are excluded.
+        filterdict = TeeChecker().getfilters(excludefilters=excluded_filters)
+
+        filterdocs = [get_check_description(name, filterfunc)
+                      for (name, filterfunc) in filterdict.iteritems()]
+
+        filterdocs.sort()
+
+        body = u'\n'.join(filterdocs)
+
+        # Output the quality checks descriptions to the HTML file.
+        filename = os.path.join(settings.WORKING_DIR,
+                                'templates/help/_ttk_quality_checks.html')
+
+        with codecs.open(filename, 'w', 'utf-8') as f:
+            f.write(body)
+
+        self.stdout.write('Done.')
--- a/pootle/templates/help/quality_checks.html
+++ b/pootle/templates/help/quality_checks.html
@@ -1,0 +1,23 @@
+{% extends "layout.html" %}
+{% block title %}Quality Checks | {{ block.super }}{% endblock title %}
+
+{% block header_block %}{% endblock %}
+
+{% block content %}
+<div id="staticpage" class="checks-descriptions" dir="ltr">
+  <h1 class="title">Quality Checks</h1>
+
+  <p>Pootle has an ability to check for common translation mistakes. Once you submit a translation, it will compare its certain features with the original string and identify potential problems. Сhecks are displayed in the upper right corner, just above the submit button. Once there are checks in red, you will stay on the same unit, until you have each check resolved or muted. Note that you should only mute checks if you know what you're doing. Muted checks will be reviewed periodically by the managers.</p>
+
+  {% if settings.CAN_CONTACT %}
+  <p>If you are not sure what a particular error means and how to fix it properly, use <i>"Report a problem with this string"</i> link at the top of the unit. The managers will try to assist you as soon as possible.</p>
+  {% endif %}
+
+  <p>Below is the description of each quality check.</p>
+
+  {% include "help/_pootle_quality_checks.html" %}
+
+  {% include "help/_ttk_quality_checks.html" %}
+
+</div>
+{% endblock content %}
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,9 @@
 # /pootle/settings/
 /pootle/settings/*-local.conf
 
+# TTK checks descriptions template
+/pootle/templates/help/_ttk_quality_checks.html
+
 # Possible customizations
 /pootle/templates/custom
 /pootle/static/*/custom
--- a/docs/developers/releasing.rst
+++ b/docs/developers/releasing.rst
@@ -139,6 +139,18 @@ that needs fixing.
     $ git grep 2013  # Should pick up anything that should be examined
 
 
+Update checks descriptions
+--------------------------
+
+The quality checks descriptions are kept as a static HTML page that has to be
+regenerated in order to ensure the descriptions match the currently available
+quality checks.
+
+.. code-block:: bash
+
+    $ pootle regenerate_checks_descriptions
+
+
 Update translations
 -------------------
 
--- a/docs/releases/2.7.0.rst
+++ b/docs/releases/2.7.0.rst
@@ -211,6 +211,8 @@ Command changes
   commands.
 - Added command and store action logging.
 - Added ``test_checks`` management command.
+- Added ``regenerate_checks_descriptions`` command to recreate the checks
+  descriptions.
 - Removed ``--directory`` and ``--path-prefix`` parameters from management
   commands. ``--project`` and ``--language`` should be used instead to
   reduce the scope of commands.
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -345,6 +345,19 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
 against.
 
 
+.. _commands#regenerate-checks-descriptions:
+
+regenerate_checks_descriptions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.7
+
+This command regenerates the static page holding the quality checks
+descriptions. If new checks are added, or existing checks are updated or
+removed it is advisable to run this command in order to keep the descriptions
+updated.
+
+
 .. _commands#translation-memory:
 
 Translation Memory
--- a/docs/server/upgrading.rst
+++ b/docs/server/upgrading.rst
@@ -144,6 +144,20 @@ To perform the upgrade follow the next steps:
     (env)$ pootle setup
 
 
+* If you are upgrading from Pootle 2.7.0 or later you also have to regenerate
+  the quality checks descriptions.
+
+  .. code-block:: bash
+
+    (env)$ pootle regenerate_checks_descriptions
+
+
+  .. note::
+
+    It is necessary to install the ``docutils`` package to be able to run this
+    command.
+
+
 * Reapply your custom changes to Pootle code, templates or styling. Check about
   the :doc:`customization of style sheets and templates
   </developers/customization>` to move your customizations to the right
--- a/pootle/apps/pootle_statistics/models.py
+++ b/pootle/apps/pootle_statistics/models.py
@@ -169,8 +169,7 @@ class Submission(models.Model):
                 unit.update({
                     'check_name': check_name,
                     'check_display_name': check_names[check_name],
-                    'checks_url': reverse('pootle-staticpages-display',
-                                          args=['help/quality-checks']),
+                    'checks_url': reverse('pootle-checks-descriptions'),
                 })
                 source.update({
                     'check_name': '<a href="%(checks_url)s#%(check_name)s">'
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -604,8 +604,7 @@ def timeline(request, unit):
                 entry.update({
                     'check_name': check_name,
                     'check_display_name': check_names[check_name],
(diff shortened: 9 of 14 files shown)


dwaynebailey:
lgtm didn't test.  I'm not sure about the unescaped & that appear in the reference EN HTML file.

iafan:
Please note that we added some new quality checks and descriptions recently (couple of weeks ago). Is this based on most recent edits? After this PR is merged, who's going to maintain this file?

iafan:
> Yes. It could be put as &amp; but works the same.

My personal preference to have all HTML files actually XHTML (so that they can be parsed/validated as a part of the test/build process). So I'd suggest using '&amp;' everywhere (including URLs).

unho:
@iafan All Evernote quality checks have description except https://github.com/translate/pootle/blob/master/pootle/apps/pootle_misc/checks.py#L455-L463 which doesn't look like a quality check. Or at least it seems not to be used anywhere.

About maintenance. I suppose it is up to the people that adds or removes checks. IMHO it might be better to put each quality check description in the docstring, like we just did with TTK ones: https://github.com/translate/translate/commit/602b9286

I am replacing  the `&amp;`.

pootle/core/initdb.py
```diff
@@ -268,7 +274,7 @@ def create_default_projects():
                  'guide</a>.</div>'),
         'virtual_path': "announcements/projects/"+tutorial.code,
     }
-    ann = Announcement(**criteria)
+    ann = StaticPage(**criteria)
```
julen: Is this an error?

julen: Hmm just saw the change in the import. Still, I don't believe it's necessary the rename, and it's completely irrelevant in the context of this PR.

pootle/core/initdb.py
```diff
@@ -315,3 +321,42 @@ def create_default_admin():
...
+    # Now get Translate Toolkit quality checks descriptions.
+
+    def get_check_description(name, filterfunc):
+        # Provide a header with an anchor to refer to.
+        description = ('\n<h3 id="%s">%s</h3>\n\n' %
+                       (name, unicode(check_names[name])))
+
+        # Clean the leading whitespace on each docstring line so it gets
+        # properly rendered.
+        docstring = filterfunc.__doc__.replace('        ', '')
```
julen: Relying on a specific number of whitespace characters is error-prone. You should `strip()` whitespace instead. Something like `' '.join([line.strip() for line in filterfunc.__doc__.split('\n')])`

julen:
* Not sure how the *requirements/base.txt* changes are relevant here.
* Can we have the HTML file somewhere else please? *templates/checks/descriptions.html* is a good candidate.
* Also I wonder how are descriptions supposed to be updated whenever new checks appear or disappear.

pootle/core/initdb.py
unho: When I first imported `StaticPage` was for using it on announcement creation, so I imported it as `Announcement`. Now that we are creating regular static pages it feels cleaner to use instead `StaticPage` since using `Announcement` can be confusing.

pootle/core/initdb.py
unho: Will use `'\n'.join([line.strip() for line in filterfunc.__doc__.split('\n')])` instead in order to get exactly the same result as previously. This keeps the paragraphs, resulting in a more readable description.

unho:
* Not sure how the *requirements/base.txt* changes are relevant here.

We require latest available Toolkit code, because of the changes we recently merged there (each check function now keeps its description on its docstring for easier maintenance and to ease programmatic regeneration), so I would say it is necessary.

* Can we have the HTML file somewhere else please? *templates/checks/descriptions.html* is a good candidate.

I don't like the idea, but I changed it that way.

In fact I think the best approach to keep the descriptions updated would be to move them to each quality check docstring (like in TTK), but since that would take time I would leave it for future cleanups. Also I am not sure if you are even ok with this approach.

* Also I wonder how are descriptions supposed to be updated whenever new checks appear or disappear.

There is no way to do that in this PR. Any suggestion? Maybe some management command to regenerate the whole checks descriptions staticpage?

dwaynebailey:
@unho a management command does sound like the easiest to recreate the pages, buys us time to not have to solve the further for a bit of time.

unho:
@translate/evernote @translate/house Added a mixing requirement and a new management command to recreate the checks descriptions on demand.

julen:
The new dependency on `docutils` for the base requirements made me think if the docstring extraction + rst2html conversion is not something better worth doing as a some external/build step. After all check docstrings rarely change, and this step (and the dependency on docutils) can easily be avoided for the rest of humans. So `initdb` would just have to read a pre-generated HTML file and import it as an static page.

unho:
@julen So you suggest to include a static HTML page that has to be recreated on each release? Do you suggest us to handle the descriptions for the TTK checks in the same way?

unho:
@julen Let me try to explain my point a bit more clearly. Having to maintain static HTML files both in Pootle and TTK to avoid having a new dependency feels like a high toll. I see that as a maintenance nightmare, to be clear. In fact I was looking more towards even removing the existing HTML file in Pootle for the Evernote checks description, specially after @iafan asked about who is going to maintain that file (which I understood as that he sees that as maintenance burden). On the other hand I don't like the idea of pulling new dependencies.

iafan:
I personally don't quite like what @julen have proposed: if we change the description in the bundled html file, it won't get propagated to static files in the database. We'd better simply serve this as a static page (and outside the /pages/ prefix) without having to copy it anywhere.

julen:
> I personally don't quite like what @julen have proposed: if we change the description in the bundled html file, it won't get propagated to static files in the database.

This is not any new proposal, but how the code in the PR operates.

iafan:
Ok, I was probably confused by this comment of yours:

> So initdb would just have to read a pre-generated HTML file and import it as an static page.

This is what I would like to avoid. We don't have to copy template into a 'static page' entity in the database just to be able to show static content. We can render it straight from the template.

julen:
> Having to maintain static HTML files both in Pootle and TTK to avoid having a new dependency feels like a high toll.

Forbid me if my comment wasn't clear enough, but I'm not suggesting to maintain any HTML files on TTK. The explanation/help page is on the Pootle side, so any HTML file belong there.

Also, here a static HTML string is being generated, which is then fed to an `StaticPage` object. Note that at the current state of things generating the output HTML string is necessary in any case, but creating the `StaticPage` object adds an extra layer which can be avoided by serving the HTML directly.

dwaynebailey:
@julen @iafan OK let me weigh in here.  Firstly, this PR was originally done to try quickly get TTK check descriptions into Pootle and also EN check descriptions into a default install.  Secondly, to make it possible to keep those well maintained without duplication.  So lets keep that in mind.

I also think there is a better way to support check descriptions and that's close to the check and not in a static page.  But we didn't want to go that route for this, at that would be a long fix... which ironically this quick fix is now turning into.

@julen from what I understand your issues and suggestions are these:

1. We don't actually need a Static Page object we could just have a static HTML page (I think @iafan agrees) that Pootle is serving.
2. We can remove the docutils dependency by simply creating the HTML file as part of the Pootle release process, instead of the build process.  header+encheck+ttkchecks = static HTML file

@unho raises the following concern with this approach:

1. TTK can be anything and thus any new TTK release would require you to generate the static page again to pick up any new checks or changed descriptions.  Thus there is no one static page for TTK

My suggestion for going forward:

1. Drop StaticPage
2. Generate the TTK part as part of the release process
3. Keep the management command to allow creation of the static/ page
4. Move docutils dependency into dev.txt.  Thus anyone wanting to update the checks page can do it.

julen:
Your understanding is correct @dwaynebailey and the suggestion is reasonable.

Also note the current `StaticPage` is available at `/pages/help/quality-checks/`. @iafan was suggesting to keep the new HTML file out of the `/pages/` namespace, although it would be possible to retain the old URL if deemed necessary. Thoughts @iafan?

unho:
@julen @iafan I see no problem on having the descriptions on a static HTML that is served directly. But note that that requires to adjust the links on the editor and to agree in which URL is the static HTML going to be served.

@dwaynebailey Keeping the command to regenerate the descriptions (necessary since a given Pootle version can support multiple TTK releases with possibly different checks) pulls in the `docutils` dependency. If that dependency is not in base requirements we will have to warn people on docs about the need to install an extra dependency and enclose the import on a try block to avoid tracebacks in case of docutils not being available. And that is fine, despite IMHO it is not precisely user friendly.

All of this just reminded me that we have to include checks description regeneration in the upgrade instructions.

iafan:
Adjusting the links in the editor is not a problem. Let's just do this as a part of this PR. As for the URL for the page, it can be simply `/help/quality-checks/`.

julen:
@unho instead of having the command break when `docutils` is missing, you can provide a hint to install it.

unho:
@julen that is the plan.

dwaynebailey:
@unho the command can actually use the Django checks infrastructure to bail and ask for docutils to be installed.

unho:
@translate/evernote @translate/house Reworked the whole PR. Now quality checks descriptions are being served in `/help/quality-checks/` (so no more static pages). The management command now only regenerates the Translate Toolkit checks descriptions, which are saved in the `checks/_ttk_descriptions.html` template.

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
```diff
@@ -0,0 +1,83 @@
...
+
+        filterdocs = [get_check_description(name, filterfunc)
+                      for (name, filterfunc) in filterdict.iteritems()]
+
+        filterdocs.sort()
+
+        body = u"\n".join(filterdocs)
+
+        # Output the quality checks descriptions to the HTML file.
+        filename = os.path.join(
```
julen: Better:
```python
from django.conf import settings
filename = os.path.join(settings.WORKING_DIR,
                        'templates/checks/_ttk_descriptions.html')
```

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
```diff
@@ -0,0 +1,83 @@
...
+            os.path.abspath(os.path.dirname(__file__)),
+            '../../../../templates/checks/_ttk_descriptions.html'
+        )
+
+        with codecs.open(filename, 'w', 'utf-8') as f:
+            f.write(body)
+
+        logging.info('Successfully regenerated Translate Toolkit quality '
```
julen: `self.stdout.write()` suffices here, and same above. Also a simple message such as *Regenerating descriptions... done.* looks good enough TBH.

julen:
Thanks @unho, this looks good now.

I don't believe an empty *_ttk_descriptions.html* needs to be under version control, and even more, it should probably be in the *.gitignore* file as it should be generated when creating a distribution package.

You might as well want to align file names with URLs (*checks/descriptions.html* -> *help/quality_checks.html*), but up to you.

The only concern I have is that when the command hasn't been run (or when there's no plan to use TTK checks), it still displays an empty header (*Translate Toolkit Checks*) and a paragraph mentioning two types of checks. 
From our side we can customize that template to remove these irrelevant references and be set with that; not sure about other potential scenarios though. @iafan your thoughts on this?

GTM from my side but let's wait to hear thoughts on the previous point before merging.

unho:
> I don't believe an empty *_ttk_descriptions.html* needs to be under version control, and even more, it should probably be in the *.gitignore* file as it should be generated when creating a distribution package.

This template now contains a placeholder text saying TTK checks are not enabled.

> You might as well want to align file names with URLs (*checks/descriptions.html* -> *help/quality_checks.html*), but up to you.

Done.

> The only concern I have is that when the command hasn't been run (or when there's no plan to use TTK checks), it still displays an empty header (*Translate Toolkit Checks*) and a paragraph mentioning two types of checks. 
> From our side we can customize that template to remove these irrelevant references and be set with that; not sure about other potential scenarios though. @iafan your thoughts on this?

Now the *Translate Toolkit Checks* part displays a placeholder saying that those are not enabled. I think it is not clean enough, and I have been thinking on having a third include for the problematic paragraph while having the TTK include blank. But I would like to hear from you.

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
```diff
@@ -0,0 +1,81 @@
...
+        body += u'\n'.join(filterdocs)
+
+        # Output the quality checks descriptions to the HTML file.
+        filename = os.path.join(settings.WORKING_DIR,
+                                'templates/help/_ttk_quality_checks.html')
+
+        with codecs.open(filename, 'w', 'utf-8') as f:
+            f.write(body)
+
+        self.stdout.write('Regenerating descriptions... done.')
```
julen: Realize you're repeating *Regenerating descriptions* here, it's already written above.

julen:
>> I don't believe an empty _ttk_descriptions.html needs to be under version control, and even more, it should probably be in the .gitignore file as it should be generated when creating a distribution package.
> 
> This template now contains a placeholder text saying TTK checks are not enabled.

The core issue I was trying to highlight wasn't the fact of having an *empty* file. It was about the fact of having a target file which can be generated and therefore possibly end up accidentally modified in version control, potentially creating conflicts or causing other trouble that can initially be avoided.

pootle/templates/help/_ttk_quality_checks.html
```diff
@@ -0,0 +1,4 @@
+<h2>Translate Toolkit Checks</h2>
+
+
+<p>This server has not enabled the Translate Toolkit quality checks.</p>
```
dwaynebailey: Why have this at all if the server is not using TTK checks?

dwaynebailey:
@unho did a review and the generation works.

One concern.  The docs that we're replacing in TTK have [full description](http://docs.translatehouse.org/projects/translate-toolkit/en/stable-1.12.0/commands/pofilter_tests.html#accelerators) of checks, much like Evernote.  Yet in the autogenerated ones we only have the first line of the content.

Is this a bug, the code looks like it wants to give the full description.  Or was it intentional?

unho:
> One concern.  The docs that we're replacing in TTK have [full description](http://docs.translatehouse.org/projects/translate-toolkit/en/stable-1.12.0/commands/pofilter_tests.html#accelerators) of checks, much like Evernote.  Yet in the autogenerated ones we only have the first line of the content.
>
> Is this a bug, the code looks like it wants to give the full description.  Or was it intentional?

@dwaynebailey That might be you not having upgraded translate-toolkit dependency. Note that we require the latest version on git. Once we release 1.13 final we will require that.

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
unho: @julen It is intentional. Lets not force users to guess what is done.

dwaynebailey:
@unho spot on it was an older toolkit instead of the one from Git.  With a new version from github this works as expected.

unho:
@translate/house @translate/evernote All issues have been addressed. Now the `help/_ttk_quality_checks.html` template is not being added and is on the gitignore. Also all the references to toolkit or Evernote have been removed and the CSS styles are now in the stylesheets. Any other issue or gtms?

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
julen: I'd understand it if it was displayed as *Regenerating descriptions...* at the beginning, and once it's done it was overwritten with *Regenerating descriptions... done*. There's no guesswork involved.

pootle/templates/help/quality_checks.html
```diff
@@ -0,0 +1,21 @@
...
+
+{% block header_block %}{% endblock %}
+
+{% block content %}
+<div id="staticpage" class="checks-descriptions" dir="ltr">
+  <h1 class="title">Quality Checks</h1>
+
+  <p>Pootle has an ability to check for common translation mistakes. Once you submit a translation, it will compare its certain features with the original string and identify potential problems. Сhecks are displayed in the upper right corner, just above the submit button. Once there are checks in red, you will stay on the same unit, until you have each check resolved or muted. Note that you should only mute checks if you know what you're doing. Muted checks will be reviewed periodically by the managers.</p>
+
+  <p>If you are not sure what a particular error means and how to fix it properly, use <i>"Report a problem with this string"</i> link at the top of the unit. The managers will try to assist you as soon as possible.</p>
```
julen: Realize this only applies if `CAN_CONTACT = True`

pootle/apps/pootle_app/management/commands/regenerate_checks_descriptions.py
unho: @julen Ok, changing this so you can gtm.

pootle/templates/help/quality_checks.html
unho: Fixed that, despite the string errors report form is displayed indepedently of the value of `CAN_CONTACT`.

unho:
@translate/house @translate/evernote Any new issue?

dwaynebailey:
@unho lgtm

unho:
Merged since it seems I have gtms from both sides. If any other issue arises it can be fixed on master.

PR #3696, opened 2015-04-29
Use logging.debug to remove migration noise

dwaynebailey:
Just want a quick check and gtm @translate/house 

I assumed I didn't need to setup a logger.  But would like some feedback as this looks like the only migration with logging messages in it.


unho:
lgtm

Issue #3377, opened 2014-11-26
Make sure `POOTLE_LOG_DIRECTORY` is respected


julen:
After #3366 being merged, the situation of logging has regressed: setting `POOTLE_LOG_DIRECTORY` won't have any effects unless `LOGGING` is redefined as well to take the new value into account. This makes no sense.

Last week we discussed on IRC ways to address this and here's my suggestion:

1. For clarity, move logging-specific default settings to its own file, say *25-logging.conf*
2. Rename default logging settings from `LOGGING` to `POOTLE_DEFAULT_LOGGING`. Here we'd remove any references to `POOTLE_LOG_DIRECTORY` because we'll inject them later. For clarity, a comment indicating so is desirable.
3. If no logging has been configured as part of any *90-local.conf* file, in *95-outro.conf* make sure `POOTLE_LOG_DIRECTORY` is used in the locations where it's supposed to go. Also copy the resulting dictionary to `LOGGING` so that Django sees the final results.

Here's a snippet for the last part:
```python
if not 'LOGGING' in globals():
    POOTLE_DEFAULT_LOGGING['handlers']['log_action']['filename'] = \
        os.path.join(POOTLE_LOG_DIRECTORY, 'pootle-activity.log')
    LOGGING = POOTLE_DEFAULT_LOGGING.copy()
```

PR #3701, opened 2015-04-30
Cleanup logging config

dwaynebailey:
I've tried to improve the state of our default logging here.  The idea is that in general you shouldn't need to alter the logging.


pootle/settings/20-backends.conf
```diff
@@ -81,11 +78,25 @@ RQ_QUEUES = {
...
+# The directory where Pootle writes its logs
 POOTLE_LOG_DIRECTORY = working_path("log")
 
+# Useful references:
+#
+# Logging configuration:
+# Django: https://docs.djangoproject.com/en/1.7/topics/logging/#configuring-logging
+# RQ: https://github.com/ui/django-rq#configuring-logging
+# Sentry: http://raven.readthedocs.org/en/latest/integrations/django.html#integration-with-logging
+#
```
julen: ES: http://elasticsearch-py.readthedocs.org/en/master/#logging

...and ideally Pootle should also need its own reference, but that requires reviewing existing loggers and naming them consistently.

pootle/settings/20-backends.conf
```diff
@@ -81,11 +78,25 @@ RQ_QUEUES = {
...
+# Format attreibutes: https://docs.python.org/2/library/logging.html#logrecord-attributes
```
julen: *attributes*

julen:
:+1: from my side, also fixes #3377.

julen:
Oh, at the same time, and as proposed in #3377, I'd move the logging settings to a *25-logging.conf* file for clarity — these settings have nothing to do with backends after all.

PR #3702, opened 2015-05-01
Language list: use simplified names

### LINKED ISSUE #3700: Overview dropdown is using ISO639 names and not simplified toolkit names
E.g. nso is using "Pedi; Sepedi; Nothern Sotho" instead of simply 'Northern Sotho'

julen:
At the same time this enables localized language names where applicable.

Please test with different UI locales — I couldn't do it since iso-codes seems broken on my end.

Fixes #3700.

### DIFF
--- a/pootle/apps/pootle_language/models.py
+++ b/pootle/apps/pootle_language/models.py
@@ -7,6 +7,7 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import locale
 from collections import OrderedDict
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


unho:
@julen lgtm. Couldn't test because iso-codes is also not working on my local environment.

julen:
@dwaynebailey I'd be happy if you could test this on localized UIs.

dwaynebailey:
@julen first part works, we're getting the properly simplified names.  Still checking localisation of language names.

dwaynebailey:
@julen gtm nice fix

screen shot 2015-05-05 at 21 40 34

I installed the iso-codes package into my virtualenv and tested on German.

Issue #3530, opened 2015-01-21
Differentiate between disabled and hidden projects


julen:
Projects have the `disabled` field (40da23079 & a6b5dd9d), but it has resulted to be confusing — since it has effects on the UI, lots of times it's unclear the purpose of setting that bit on. In other words, it's difficult to know if a project is only *hidden* from the UI, or actually *disabled* from any sort of processing.

By adding a new `hidden` field for projects and translation projects, it'd be easy to distinguish between both scenarios, and the code would result to be much more explicit.

This can be summarized as:
- *enabled* and *visible*: standard situation, accessible for whoever has access rights to it
- *enabled* and *hidden*: stats still kept, admins can access and browse it, other users get 404 and cannot see them in listings
- *disabled*: no stats calculated, used for gone projects and archival purposes, implies being hidden for everyone when browsing (even admins), resulting in 404. It's only listed in the admin UI.

iafan:
As per our today's talk, I think we'd better keep this as is: disabled project should simply mean that it should be hidden from anybody expect admins, but otherwise it should be a regular project, with stats being kept, displayed and updated.

The idea behind having a disabled project was to hide the project from translation for anybody (especially at project creation time); do the maintenance, setup permissions, update stats, then be able to open it with the flip of the switch, without messing with permissions and permission overrides. Or keep it disabled forever for archived projects (but for admins still to be able to peek into its details/history).

Regarding terminology, I don't like the term _hidden_ (or _visible_), because projects can also be hidden due to permission restrictions.

So the project is _hidden_ because either user has no permissions, or because it was explicitly _disabled_ by admins. "_disabled_" is a project property, and "_hidden_" is an end result.

ta2-1:
> but for admins still to be able to peek into its details/history

@iafan what does "history" means? Timeline?
It seems we got exactly what you asked.
Can you look at my [comment](https://github.com/translate/pootle/pull/4008#issuecomment-134645914)?

iafan:
history = how the project looked like before it was disabled. Tree structure, stats, stores, units, their translations and their timeline.

ta2-1:
Then admins will be able to translate units in disabled projects. Is it ok?

iafan:
I think that is ok. "Disabled" == "globally hidden for anybody expect admins, but otherwise functional."

julen:
Closing per the decision already made.

PR #3705, opened 2015-05-02
Local TM: robustness, documentation, refactoring, etc

dwaynebailey:
This improves on the current local tm implementation:

- Local TM is on and available by default
- Documentation has been added
- update_tmserver manage.py command as been added based on the original Perl script
- tmserver.py was replace by:
  - A base search class
  - An Elasticsearch backend
  - A broker that allows the possibility of other backends and multiple TM servers 
- Tested and made more robust:
  - We don't show anything if local tm is not configured (though its configured by default)
  - We don't crash if we can't find the Elasticsearch server (thus you can take that down and Pootle will still run)

Note that the notes in 45344e0 will be removed and turned into issues before this lands.

Not yet, and likely in future updates:
- Weighting of results from multiple server setup
- Any improvements to the display of TM results
- Investigation into odd fuzzy results we seem to be getting
- Any effort to bring Amagama into the available TM server ENGINES.


docs/releases/2.7.0.rst
```diff
@@ -184,7 +184,8 @@ Sysadmins
   <https://readthedocs.org/projects/django-allauth/>`_.  This gives Pootle
   implicit support for OpenID, OAuth, OAuth2 and Persona sign-in protocols.
 - Integrated ElasticSearch-based local TM server into Pootle. Unit submissions
-  update the index instantly.
+  update the index instantly.  To configure adjust :setting:`POOTLE_TM_SERVER`
+  and to load the TM use the ``update_tmserver`` management command.
```
julen: Does `:command:'update_tmserver'` work? (Replace `'` with backticks)

docs/server/commands.rst
```diff
@@ -345,6 +345,28 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
 against.
 
 
+.. _commands#translation-memory:
+
+Translation Memory
+------------------
+
+These command allow you to setup and manage :doc:`Translation Memory
```
julen: *commands*

docs/server/commands.rst
```diff
@@ -345,6 +345,28 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
...
+</features/translation_memory>`.
+
+.. _commands#update_tmserver:
+
+update_tmserver
+^^^^^^^^^^^^^^^
+
```
julen: ++ `.. versionadded:: 2.7`

docs/server/settings.rst
```diff
@@ -340,6 +340,22 @@ Translation environment configuration settings.
   Pootle user to be able to read them.
 
 
+.. setting:: POOTLE_TM_SERVER
```
julen: ++ `.. versionadded:: 2.7`

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                    action="store_true",
+                    dest='dry_run',
+                    default=False,
+                    help='Report only the number of translations to index '
+                         'and quit'),
+    )
+
+    def handle(self, *args, **options):
+        if not getattr(settings, 'POOTLE_TM_SERVER', False):
+            raise CommandError("POOTLE_TM_SERVER is missing from your settings.")
```
julen: A blank line after this one wouldn't hurt.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+        },
+        ])
+
+        last_indexed_revision = -1
+
+        if options['rebuild'] and not options['dry_run']:
+            if es.indices.exists(INDEX_NAME):
+                es.indices.delete(index=INDEX_NAME)
+
+        if not options['rebuild'] and not options['overwrite'] and es.indices.exists(INDEX_NAME):
```
julen: Nit: line too long

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                        }
+                    }
+                }
+            )
+            last_indexed_revision = result['facets']['stat1']['max']
+
+        self.stdout.write("Last indexed revision = %s" % last_indexed_revision)
+
+
+        units_qs = Unit.simple_objects \
```
julen: Realize this query's results:
* will be massive for relatively big servers when run for the first time (no last index optimization available yet)
* also includes obsolete units
* also includes units from disabled projects

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                'target_f',
+                'submitted_by__username',
+                'submitted_by__full_name',
+                'submitted_by__email',
+                'store__translation_project__project__fullname',
+                'store__pootle_path',
+                'store__translation_project__language__code'
+            ).order_by()
+
+        total = len(units_qs)
```
julen: Note this will evaluate the whole queryset and load it in memory. Just to know the result count, use `units_qs.count()`.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if not total:
```
julen: `total` being a number, `if total == 0:`

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+            self.stdout.write("No translations to index")
+            sys.exit()
+
+        self.stdout.write("%s translations to index" % total)
+
+        if options['dry_run']:
+            sys.exit()
+
+        i = 0
+        for unit in units_qs:
```
julen: `for index, unit in enumerate(units_qs):` is the Pythonic way do express this ;)
Also, use `units_qs.iterator()` to avoid loading the entire result set in memory.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                    'path': unit['store__pootle_path'],
+                    'username': unit['submitted_by__username'],
+                    'fullname': fullname,
+                    'email_md5': email_md5,
+                    'source': unit['source_f'],
+                    'target': unit['target_f'],
+                }
+            )
+
+            if ((i % 1000 == 0) or (i == total)):
```
julen: Redundant outer parens.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                percent = "%.1f" % (i * 100.0 / total)
+                self.stdout.write("%s (%s%%)" % (i, percent))
+
+        self.stdout.write("")
+
+        if (i != total):
```
julen: Unnecessary parens.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+            self.stdout.write("Oops.  Stopped at i = %s" % i)
```
julen: I see the original script included this message, but if it's considered worth keeping it, we might wanna be more descriptive towards the user who invoked the command.

pootle/core/search/backends/__init__.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from .elastic_search import *
```
julen: Were there any import-related issues if the module was named `elasticsearch`?

pootle/core/search/backends/elastic_search.py
```diff
@@ -0,0 +1,91 @@
...
+        self._es = self._get_es_server()
+        self._create_index_if_missing()
+
+    def _server_setup_and_alive(self):
+        return self._es is not None and self._es.ping()
+
+    def _get_es_server(self):
+        if self._settings is None and Elasticsearch is None:
+            return None
+        return Elasticsearch([{'host': self._settings['HOST'], 'port': self._settings['PORT']}, ])
```
julen: Nit: line too long.

pootle/core/search/backends/elastic_search.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _create_index_if_missing(self):
+        if self._server_setup_and_alive():
+            if not self._es.indices.exists(self._settings['INDEX_NAME']):
+                self._es.indices.create(self._settings['INDEX_NAME'])
+
+    def _is_valuable_hit(self, unit, hit):
+        if str(unit.id) == hit['_id']:
```
julen: What about `return str(unit.id) != hit['_id']`?

pootle/core/search/backends/elastic_search.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def search(self, unit):
+        if not self._server_setup_and_alive():
+            return []
+
+        counter = {}
+        res = []
+        language = unit.store.translation_project.language.code
+        es_res = self._es.search(index=self._settings['INDEX_NAME'],
+                                doc_type=language,
```
julen: Arguments are not vertically aligned. Due to the search body's nesting, you may want to place each arg in a separate line.

pootle/core/search/backends/elastic_search.py
```diff
@@ -0,0 +1,91 @@
...
+                                body={
+                                    "query": {
+                                        "match": {
+                                            'source': {
```
julen: Knowing that ES's query DSL is based in JSON, I'd rather use double quotes for search bodies. Do not mix them, at least.

pootle/core/search/backends/elastic_search.py
```diff
@@ -0,0 +1,91 @@
...
+                    counter[translation_pair] += 1
+
+        for item in res:
+            item['count'] = counter[item['source']+item['target']]
+
+        return res
+
+    def update(self, language, obj):
+        if self._server_setup_and_alive():
+            self._es.index(index=self._settings['INDEX_NAME'],
```
julen: Either vertically align the arguments or place them all in new lines properly aligned.

pootle/core/search/base.py
```diff
@@ -0,0 +1,35 @@
...
+SERVER_SETTINGS_NAME = 'POOTLE_TM_SERVER'
+
+
+class SearchBackend(object):
+    def __init__(self, config_name):
+        pass
+
+    def _setup_settings(self, config_name=None):
+        self._settings = getattr(settings, SERVER_SETTINGS_NAME, None)
+        if config_name:
```
julen: `if config_name is not None:`

pootle/core/search/base.py
```diff
@@ -0,0 +1,35 @@
...
+            self._settings = self._settings[config_name]
+
+    def search(self, unit):
+        """Search for TM results.
+
+        :param unit: :cls:`~pootle_store.models.Unit`
+        :return: list of results or [] for no results or offline
```
julen: I guess this docstring is obsolete.

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+
+from . import SearchBackend
+
+import importlib
+
+from django.conf import settings
+
+
+class SearchBroker(SearchBackend):
+    def __init__(self, config_name=[]):
```
julen: Never use a mutable type as a default arg (`config_name=[]`). Rather use `None`, and set a value based on that.

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+        self._setup_settings()
```
julen: I see all subclasses of `SearchBackend` are calling its `_setup_settings()` method when initting. Why not just do it in `SearchBackend`'s `__init__()` and let subclasses call `super()`? It's a *private* method of the super class after all, children classes shouldn't be dealing with it.

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+        self._servers = {}
+        if self._settings is None:
+            return
```
julen: Blank lines surrounding this block wouldn't hurt.

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+        for server in self._settings:
+            if config_name == [] or server in config_name:
```
julen: When checking if a sequence has elements, use the fact that empty sequences are false (PEP8): `if not config_name or server in config_name:`

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+                _module = ('.').join(self._settings[server]['ENGINE'].split('.')[:-1])
```
julen: Unnecessary parens in `('.').join...`

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+                _search_class = self._settings[server]['ENGINE'].split('.')[-1]
+                module = importlib.import_module(_module)
+                self._servers[server] = getattr(module, _search_class)(server)
+
+    def search(self, unit):
```
julen: Shouldn't the param be `obj` as in `update()`?

pootle/core/search/broker.py
```diff
@@ -0,0 +1,51 @@
...
+        results = []
+        counter = {}
+        if not self._servers:
+            return []
```
julen: Blank lines surrounding this block wouldn't hurt. Also, `return results` or make the check early on.

julen:
I wanna give this a shot before green lighting, although it looks good, nice work @dwaynebailey!

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+                    'fullname': fullname,
+                    'email_md5': email_md5,
+                    'source': unit['source_f'],
+                    'target': unit['target_f'],
+                }
+            )
+
+            if ((i % 1000 == 0) or (i == total)):
+                percent = "%.1f" % (i * 100.0 / total)
+                self.stdout.write("%s (%s%%)" % (i, percent))
```
julen: For a nicer progress indicator:
```python
                self.stdout.write("%s (%s%%)" % (i, percent), ending='\r')
                self.stdout.flush()
```

pootle/core/search/backends/__init__.py
dwaynebailey: There where when I had them at the same level, haven't checked again.  The Elasticsearch module provides and Elasticsearch() class which was causing problems.

julen: Adding `__all__ = ('ElasticSearchBackend',)` might be enough to overcome that potential issue.

pootle/core/search/base.py
dwaynebailey: No that one is still relevant, I didn't change the supplied data in my refactoring.  Unless I'm missing something here.

pootle/core/search/broker.py
dwaynebailey: I tried to avoid any tinkering with the ``search()`` and ``update()`` parameters.  Likely they could be done better but I decided to just keep at the tmserver.py level for now.

julen: :+1:

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: I ran this with the whole of LO, so relatively big query, took 25 minutes.  So yeah potentially big but I think once off usually.

I thinks its OK to bring in obsolete units.  Its a TM and we want everything we can get.

Not so sure what we should do about disabled projects. Thoughts?

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: Interestingly we went with ``len()`` instead of ``.count()`` to prevent a second query, but maybe we should leave that optimisation to the server.

julen: Not always more queries means slower, and here `count()` is fast & cheap, whereas loading the whole result set in memory... combined with `iterator()` proposed below you'll notice the difference.

dwaynebailey:
@julen thanks that was a brilliant review.  Polishing up my rusty Python.

I've landed the changes you suggested and posed questions as needed.  I haven't tested this again so likely I've introduced some things that break things.

pootle/apps/pootle_app/management/commands/update_tmserver.py
julen: Re. projects, it really depends: the thing is currently there's no way to know exactly what does `disabled` mean, as it's being used in different contexts (#3530 should address that lack of clarity). In case it means *disabled* for archival purposes, it's fine for units to go into the TM, but if it means *hidden* for privacy or similar purposes, then we should prevent that.

My suggestion, let's bear this in mind when working on #3530 and adjust the command only then.

pootle/core/search/backends/__init__.py
julen: hm, also `from __future__ import absolute_import` and you should be able to import `elasticsearch` from a module named `elasticsearch`.

docs/server/commands.rst
```diff
@@ -345,6 +345,30 @@ By default, :ref:`commands#test_checks` tests all existing checks. When
...
+.. versionadded:: 2.7
+
+Updates the ``default`` server in :setting:`POOTLE_TM_SERVER`.  The command
+reads translations from the current Pootle install and builds the TM resources
+in the TM server.
+
+By default the command will only add new translations to the server.  To
+rebuild the server from scratch use ``--rebuild``, this will completely remove
+the TM and rebuild it.  To ensure that the TM server remains available when you
+rebuild you can add ``--overwrite``.
```
unho: @dwaynebailey I suggest using `:option:` to highlight ``--rebuild`` and ``--overwrite``.

docs/server/settings.rst
```diff
@@ -340,6 +340,24 @@ Translation environment configuration settings.
...
+.. versionadded:: 2.7
+
+``POOTLE_TM_SERVER``
+  Default: Set to ``http://localhost:9200/translations``
+
+  This is configured to access a standard ElasticSearch setup.  Change the
+  settings for any non-standard setup.  Change ``HOST`` and ``PORT`` settings
+  as required.
+
+  Use ``MIN_SCORE`` to set the Leveshtein Distance score.  Set it to ``AUTO``
```
unho: @dwaynebailey Typo. It is **Levenshtein**.

docs/server/settings.rst
```diff
@@ -340,6 +340,24 @@ Translation environment configuration settings.
...
+  so that EslastiSearch will adjust the required score depending on the length
```
unho: @dwaynebailey Typo: ElasticSearch.

docs/server/settings.rst
```diff
@@ -340,6 +340,24 @@ Translation environment configuration settings.
...
+  of the string being translated. ElasticSearch documentation provides further
```
unho: @dwaynebailey Nit, it is Elasticsearch.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from __future__ import print_function
```
unho: @dwaynebailey I think this is no longer necessary.

docs/server/commands.rst
unho: @dwaynebailey There is nothing about the `--dry-run` option.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
...
+    def handle(self, *args, **options):
+        if not getattr(settings, 'POOTLE_TM_SERVER', False):
+            raise CommandError("POOTLE_TM_SERVER is missing from your settings.")
+
+        INDEX_NAME = settings.POOTLE_TM_SERVER['default']['INDEX_NAME']
+        es = Elasticsearch([{
+            'host': settings.POOTLE_TM_SERVER['default']['HOST'],
+            'port': settings.POOTLE_TM_SERVER['default']['PORT']
+        },
+        ])
```
unho: @dwaynebailey Nit: remove comma from previous line and move this to the previous line.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if options['rebuild'] and not options['dry_run']:
+            if es.indices.exists(INDEX_NAME):
+                es.indices.delete(index=INDEX_NAME)
+
+        if (not options['rebuild'] and
+            not options['overwrite'] and
+            es.indices.exists(INDEX_NAME)):
+            result = es.search(
+                index=INDEX_NAME,
```
unho: @dwaynebailey Wouldn't be cleaner to allow setting the index on es object creation?

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
...
+
+        last_indexed_revision = -1
+
+        if options['rebuild'] and not options['dry_run']:
+            if es.indices.exists(INDEX_NAME):
+                es.indices.delete(index=INDEX_NAME)
+
+        if (not options['rebuild'] and
+            not options['overwrite'] and
+            es.indices.exists(INDEX_NAME)):
```
unho: @dwaynebailey `es.indices.exists(INDEX_NAME)` is being used twice. If this costs to being run maybe it is better to cache its value on a variable, which can also help with code clarity.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
...
+        INDEX_NAME = settings.POOTLE_TM_SERVER['default']['INDEX_NAME']
+        es = Elasticsearch([{
+            'host': settings.POOTLE_TM_SERVER['default']['HOST'],
+            'port': settings.POOTLE_TM_SERVER['default']['PORT']
+        },
+        ])
+
+        last_indexed_revision = -1
+
+        if options['rebuild'] and not options['dry_run']:
```
unho: @dwaynebailey We are accessing several times to these options. I wonder if keeping the value on variables can help a bit with keeping code readable.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,161 @@
...
+            if es.indices.exists(INDEX_NAME):
```
unho: @dwaynebailey I think this if should be merged into its parent if.

pootle/apps/pootle_store/models.py
```diff
@@ -211,6 +210,9 @@ def stringcount(string):
         return 1
 
 
+TMServer = SearchBroker()
```
unho: @dwaynebailey I had to read this line to know that `SearchBroker` is actually for translation memory. I suggest renaming it to a more descriptive name like `TMBroker`.

pootle/core/search/__init__.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
```
unho: @dwaynebailey I think it might be better to name the module `pootle.core.tm` instead.

pootle/apps/pootle_store/models.py
dwaynebailey: The thinking is that TM is just a specialised search and we want a general search abstraction.  So I'd prefer to leave it as this for now.

pootle/core/search/__init__.py
dwaynebailey: For the same reason as previous, this isn't just about TM.  TM just happens to be the current use case.

docs/server/settings.rst
```diff
@@ -340,6 +340,24 @@ Translation environment configuration settings.
...
+
+``POOTLE_TM_SERVER``
+  Default: Set to ``http://localhost:9200/translations``
+
+  This is configured to access a standard Elasticsearch setup.  Change the
+  settings for any non-standard setup.  Change ``HOST`` and ``PORT`` settings
+  as required.
+
+  Use ``MIN_SCORE`` to set the Levenshtein Distance score.  Set it to ``AUTO``
+  so that Eslasticsearch will adjust the required score depending on the length
```
unho: @dwaynebailey Typo: Elasticsearch.

unho:
@dwaynebailey Just finished a second review and found only one typo. lgtm after that typo is fixed.

pootle/core/search/base.py
```diff
@@ -0,0 +1,35 @@
...
+# AUTHORS file for copyright and authorship information.
+
+from django.conf import settings
+
+
+SERVER_SETTINGS_NAME = 'POOTLE_TM_SERVER'
+
+
+class SearchBackend(object):
+    def __init__(self, config_name=None):
```
julen: It might be a good idea to accept `**kwargs` for flexibility.

pootle/core/search/broker.py
```diff
@@ -0,0 +1,54 @@
...
+from . import SearchBackend
+
+import importlib
+
+from django.conf import settings
+
+
+class SearchBroker(SearchBackend):
+    def __init__(self, config_name=None):
+        super(SearchBroker, self).__init__(config_name)
```
julen: Better to pass `config_name` as kwargs here and elsewhere; you never know how the parent class could change.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+
+            if (i % 1000 == 0) or (i == total):
+                percent = "%.1f" % (i * 100.0 / total)
+                self.stdout.write("%s (%s%%)" % (i, percent), ending='\r')
+                self.stdout.flush()
+
+        self.stdout.write("")
+
+        if i != total:
```
julen: Realize `enumerate()` provides a 0-based index, and the previous code was incrementing the index at the beginning of the loop, making that a 1-based index effectively. Therefore, here you need to `if i+1 != total:`, otherwise it'll always report it didn't load all of the expected translations.

julen:
It'd be great if you could name the ES module as `elasticsearch.py` and fix the latest small issues outlined. Other than that, this is good to be merged.

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: That doesn't seem to be possible I'm afraid.

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: These are variables, so no I don't think this will make anything clearer here but rather introduce temporary variables that I hate.

dwaynebailey:
This was fixed in LocalTM changes.

PR #3706, opened 2015-05-04
Cleanup: use `JsonResponse` for XHR requests

### LINKED ISSUE #3529: Use `JsonResponse` for XHR responses
[Django 1.7 introduced](https://docs.djangoproject.com/en/1.7/releases/1.7/#requests-and-responses) a convenience [`JsonResponse` class](https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.JsonResponse), which reduces boilerplate when returning XHR responses.

Pootle can greatly benefit from this but we must ensure the previous behavior is kept, so we'll need to review if using the [custom `PootleJSONDecoder`](https://github.com/translate/pootle/blob/0cd9f9d2134f741756dbd09fa748771630bcee71/pootle/apps/pootle_misc/util.py#L65-L77) is still necessary, otherwise it'll be necessary to create a custom class subclassing `JsonResponse` to fulfill our custom needs.

julen:
This PR introduces `JsonResponse` (and friends), therefore reducing a lot of boilerplate when returning JSON responses. It also adjusts the existing `PootleJSONEncoder` to reuse bits from Django's own `DjangoJSONEncoder`.

Please test as much views as you can and report back if you experience any failures.

Fixes #3529.

### DIFF
--- a/pootle/apps/accounts/models.py
+++ b/pootle/apps/accounts/models.py
@@ -28,8 +28,8 @@ from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 
 from pootle.core.cache import make_method_key
+from pootle.core.utils.json import jsonify
 from pootle_language.models import Language
-from pootle_misc.util import jsonify
 from pootle_statistics.models import Submission, SubmissionTypes
 from pootle_store.models import SuggestionStates
 
--- a/pootle/apps/evernote_reports/views.py
+++ b/pootle/apps/evernote_reports/views.py
@@ -15,7 +15,6 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured
 from django.core.urlresolvers import reverse
-from django.http import HttpResponse, HttpResponseNotFound, HttpResponseBadRequest
 from django.shortcuts import render_to_response
 from django.template import RequestContext
 from django.utils import timezone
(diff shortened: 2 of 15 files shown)


unho:
@julen Perhaps it is beyond the scope of this PR, but maybe the following can be simplified: https://github.com/julen/pootle/blob/eccda2b7550362ada35f09ceae774657bd28dd46/pootle/core/views.py#L340-L350

jleclanche:
What's left in PootleJSONEncoder which we need on top of django's builtin?

julen:
> What's left in PootleJSONEncoder which we need on top of django's builtin?

Django's encoder doesn't know about serializing [`Markup` objects](https://github.com/translate/pootle/blob/master/pootle/core/markup/fields.py) and [lazy translation objects](https://docs.djangoproject.com/en/1.7/topics/serialization/#serialization-formats-json).

julen:
> @julen Perhaps it is beyond the scope of this PR, but maybe the following can be simplified: https://github.com/julen/pootle/blob/eccda2b7550362ada35f09ceae774657bd28dd46/pootle/core/views.py#L340-L350

Indeed, I just want to tackle the referenced issue here.

julen:
Thanks for having a look at this — the messed up imports are adjusted now.

jleclanche:
Good cleanup, thank you for taking care of it. lgtm (untested)

PR #3709, opened 2015-05-05
Search: follow-up on ES imports

julen:
This is actually something I raised during code review, let me know what you think.


dwaynebailey:
gtm for me, though I didn't test it, I assume you did @julen ;)

PR #3710, opened 2015-05-06
Display again missing contact links

unho:
They are conditionally displayed based on `CAN_CONTACT` setting and are
included in blocks that can be easily overwritten in custom installs.

### DIFF
--- a/pootle/middleware/errorpages.py
+++ b/pootle/middleware/errorpages.py
@@ -29,7 +29,6 @@ except ImportError:
 from pootle.core.exceptions import Http400
 from pootle.core.http import (JsonResponseBadRequest, JsonResponseForbidden,
                               JsonResponseNotFound, JsonResponseServerError)
-from pootle.core.utils.json import jsonify
 from pootle_misc.baseurl import get_next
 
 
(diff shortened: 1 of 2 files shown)


dwaynebailey:
@unho can you explain what the issue is that you are addressing here.

unho:
@dwaynebailey Right now it is impossible to display the contact form. These changes conditionally render links that allow to display the contact form.

iafan:
@unho, can we make the contact form appear in the same way we do it on https://translate.evernote.com/ (just in the header, next to sign in button?)

unho:
@iafan It is already displayed there. The differences you might be seeing are basically styling. If you meant display it only on the top (not in the footer as well) I am ok with that. My only concern is to display it in some highly visible place all over the UI if `CAN_CONTACT` is set to `True`.

iafan:
> If you meant display it only on the top (not in the footer as well) I am ok with that. My only concern is to display it in some highly visible place all over the UI if CAN_CONTACT is set to True.

Yes, this is what I was talking about (show only in the header). My primary incentive here is to make as little template overrides on our side as possible (and keep the design consistent).

unho:
@iafan Now the contact link is only displayed on the header. A gtm would be appreciated.

iafan:
@julen, please review and see if this aligns with how we display that link in our overridden template. Probably we just want to share our template with Leandro and get rid of some code in our templates. Note that we still want to show an extra 'Getting Started' link in the top menu, and the core template needs a clean way to extend this top menu with extra items for better reusability.

julen:
Good to merge @unho.

PR #3719, opened 2015-05-09
MarkupField: do not consider `help_text` for migrations

### LINKED ISSUE #3687: Running migrate twice gives an error about changed models
If you run ``migrate`` a second time directly after an initial migration you will get the following error.
```
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

``makemigrations`` produces this file:

```py
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import models, migrations
import pootle.core.markup.fields


class Migration(migrations.Migration):

    dependencies = [
        ('virtualfolder', '0001_initial'),
    ]

    operations = [
        migrations.AlterField(
            model_name='virtualfolder',
            name='description',
            field=pootle.core.markup.fields.MarkupField(help_text='Use this to provide more information or instructions. Allowed markup: HTML', verbose_name='Description', blank=True),
            preserve_default=True,
        ),
    ]
```

@unho Why are virtualfolders doing this?

julen:
While this might cause issues when redefining fields, it's safe in this
case because it's a custom Pootle field used for a very specific scenario.

Refs. https://code.djangoproject.com/ticket/21498#comment:6

Fixes #3687.


jleclanche:
gtm

PR #3740, opened 2015-05-15
Display TP and Language announcements on sidebar

unho:


unho:
@translate/core Ready to review.

dwaynebailey:
@unho sorry but your rebase lost all my comments so I'm not sure if anything I reviewed has been fixed or not.  I did see some pushback re refactoring that made zero sense especially since some of the changes are simple copy and pastes which are always highly error prone.

unho:
@translate/core Fixed all reported issues. Also refactored announcement retrieval to reduce code and get/set the announcements mtimes on cookie using the announcement virtual_path as key.

unho:
@translate/core I am not quite sure about changes in the refactor commit. Extensive testing will be much appreciated.

dwaynebailey:
@unho gtm I tested with lang, proj and tp.  The three combine as expected.  I don't think it's an ideal result but it does mean that existing announcements are now available i.e. lossless upgrade.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -50,6 +50,51 @@ def admin_permissions(request, translation_project):
...
+            announcements.append(announcement)
+
+            ann_mtime = dateformat.format(announcement.modified_on, 'U')
+            stored_mtime = cookie_data.get(virtual_path, None)
+
+            if ann_mtime != stored_mtime:
+                is_sidebar_open = True
+                new_cookie_data[virtual_path] = ann_mtime
+
+    get_announcement('announcements/projects/' + project.code)
```
julen: This is very unintuitive, with a function named `get_X()` programmers expect to have some value in return, but the function mutates an object which lives out of the function and returns nothing. Even more, also alters `is_sidebar_open` and `new_cookie_data`. I'd suggest to `get_announcement()` to do just what its name says, no more.

Also rather than building and passing a string, I'd rather pass `project=<project_code>, language=<language_code>` and build the string internally (use interpolation, no need for concatenation), that way the callers don't have to know or deal with the internal details of announcements.

unho: Renamed the function to `_set_announcement_context_and_cookie` so it does what its name says. Also added a docstring which explains the code a bit more.

Changed to use interpolation. If with "build the string internally" you mean having a if..elif..else to build the virtual path, I disagree.

unho:
@translate/core Addressed some of the concerns raised. Also added an extra change to provide a useful hint when creating new announcements on Admin.

pootle/apps/staticpages/views.py
```diff
@@ -124,7 +124,10 @@ def get_form(self, form_class):
         if self.page_type == ANN_TYPE:
             del form.fields['url']
             form.fields['virtual_path'] \
-                .widget.attrs['placeholder'] = u'projects/<project_code>'
+                .widget.attrs['placeholder'] = (u'projects/<project_code> or '
```
julen: Now you need i18n

pootle/apps/pootle_translationproject/views.py
julen: It looks like you misunderstood or my explanation wasn't clear enough: the idea I wanted highlight was to keep things easy to reason about and not have functions that have many side-effects (single responsibility principle).

With adjusting `get_announcement()` I meant something like:

```python
def get_announcement(project_code=None, language_code=None):
    if language_code is None and project_code is None:
        return None

    if language_code is None and project_code is not None:
        path = u'/projects/%s' % (project_code,)
    else:
        path = u'/'.join(filter(None, [language_code, project_code]))

    virtual_path = u'/announcements/%s/' % (path, )
    try:
        return StaticPage.objects.live(request.user).get(
            virtual_path=virtual_path,
        )
    except StaticPage.DoesNotExist:
        return None    
```

`get_announcement()` does one thing and does it well (the path construction could even be factored out into an utility function), and callers don't need to know about internal implementation details such as what a `virtual_path` is and how it's constructed specifically for announcements.

pootle/apps/staticpages/views.py
unho: Addressed.

unho:
@julen Updated with your latest recommendations. Please review.

julen:
Clicking to collapse the sidebar produces:
```
SyntaxError: JSON.parse: unexpected non-whitespace character after JSON data at line 1 column 4 of the JSON data
```

unho:
@julen Tested it all the ways I can thought of, and basically the problem is that the browser receives a cookie whose value is a string that begins and ends with double quotes, like `""value""`, so `JSON.parse()` in https://github.com/unho/pootle/blob/3c077899b7702141f04680f7bfc1fdd594c22fda/pootle/static/js/common.js#L163 fails to parse it because it expects `"value"`. I honestly have no idea why the new code is not working as before.

unho:
Did some further testing. The problem is in the last commit, the one that refactors the sidebar announcements retrieval.

unho:
@translate/core Fixed the issue raised by @julen. Please review so we can hopefully land this soon.

julen:
Feel free to merge it.

PR #3751, opened 2015-05-19
Drop some EN specific customisations

dwaynebailey:
Some things I picked up in LO deployment.  PR'ing so that @translate/evernote are aware of changes needed to templates on your deployment.

- First drops the 'English' contact requirements - deployment specific
- Second - I know this should go with allauth - while allauth hasn't landed this functionality should be gone and this allows us to extend this as needed in future.

### DIFF
--- a/pootle/templates/contact_form/xhr_contact_form.html
+++ b/pootle/templates/contact_form/xhr_contact_form.html
@@ -6,7 +6,7 @@
     {% block contact_preamble %}
     <p>
       {% blocktrans trimmed %}
-      You have to fill all the fields, and please <b>write in English</b>.
+      Please fill in all the fields.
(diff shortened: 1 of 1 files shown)


julen:
The first one is ok, leave out the second one please, it's gone in the allauth branch.

unho:
I don't think anyone managing a Pootle server with more than a few languages is capable of  understand feedback written in any language. That is the reason behind the request to write in English. I think we must keep that as the default, and allow to overextend to accept other languages in custom installs.

jleclanche:
Let's not impose "write in english" on everyone, the condition doesn't need to be mentioned either way.

jleclanche:
gtm per julen's instructions

Issue #3735, opened 2015-05-14
 Awkward URL for accessing multiple failing checks


julen:
*(Original report: evernote/pootle#95)*

When clicking on the number of critical failing checks to access the editor from overview pages, a long and awkward URL is used which takes the following form: `http://server.tld/path/to/ctx/translate/#filter=checks&checks=java_format,template_format,percent_sign_closure_placeholders,non_printable,changed_attributes,mustache_placeholder_pairs,whitespace,unbalanced_tag_braces,uppercase_placeholders,mustache_placeholders,dollar_sign_placeholders,broken_entities,unescaped_ampersands,unbalanced_curly_braces,c_format,test_check,unbalanced_tags,objective_c_format,javaencoded_unicode,date_format,android_format,percent_sign_placeholders,accelerators`.

This is insane. We should rather aim at using named aliases in URLs, resulting in URLs like `http://server.tld/path/to/ctx/translate/#filter=checks&category=critical`.

Since we already name quality check failure categories (*critical*, *functional*, etc.), the natural thing would be reusing those names. In the end, the above-mentioned long URL is just a list of checks contained within the *critical* category.
(*Note: We can keep the ability to specify multiple individual checks in the long form, but using the named categories should already cover our main use-case*).

Apart from that, currently the UI doesn't reflect the multiple checks that are being displayed for the current view. When switching to named categories, the category should be selected in the drop-down. This should also ensure users are able to select check categories from the drop-down.

julen:
Considering this a bug because it couples the critical check links with the project checker, and [the generated URLs are `ENChecker`-specific](https://botbot.me/freenode/pootle-dev/2015-05-14/?msg=39165660&page=2).

unho:
Agreed that it is insane. Filtering by categories is saner, or pointing to a specific check.

PR #3753, opened 2015-05-19
Drop hardcoded dependence on ENChecker()

dwaynebailey:
This PR lands a few things:

1. A change by @ta2-1 to request check categories instead of lists of checks
2. Correct setup of QUALITY_CHECKER
3. Documentation of the same
4. Dropping other instances of hardcoding ENChecker()

I haven't looked at Vfolders that also need to drop the use of ``get_qualitychecks_by_category`` as in 57634d50d73c5fa5585e6f7625695d8357839a5c I'll leave that to @unho, see [virtualfolder/models.py](https://github.com/dwaynebailey/pootle/blob/master/pootle/apps/virtualfolder/models.py#L354-L356)

Likely there are much cleaner ways to excise ``ENChecker()``.  But this gets part of the way there for now and is what I need for a LibreOffice deployment.

@ta2-1 would appreciate your review as I think you have a better understanding of some of the bit of code I've hacked at.

### DIFF
--- a/pootle/apps/pootle_misc/checks.py
+++ b/pootle/apps/pootle_misc/checks.py
@@ -7,6 +7,7 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import logging
 import re
 re._MAXCACHE = 2000
 
@@ -23,6 +24,9 @@ from pootle_misc.util import import_func
 category_names = {
     Category.CRITICAL: _("Critical"),
     Category.COSMETIC: _("Cosmetic"),
+    Category.FUNCTIONAL: _("Functional"),
+    Category.EXTRACTION: _("Extraction"),
+    Category.NO_CATEGORY: _("Other"),
 }
 
 
@@ -973,14 +977,30 @@ def run_given_filters(checker, unit, check_names=[]):
 
 
 def get_qualitychecks():
-    sc = ENChecker()
-    for filt in sc.defaultfilters:
-        if filt not in excluded_filters:
-            # don't use an empty string because of
-            # http://bugs.python.org/issue18190
-            getattr(sc, filt)(u'_', u'_')
+    available_checks = {}
 
-    return sc.categories
+    if settings.QUALITY_CHECKER:
+        checkers = [import_func(settings.QUALITY_CHECKER)()]
+    else:
+        checkers = [checker() for checker in checks.projectcheckers.values()]
+
+    for checker in checkers:
+        for filt in checker.defaultfilters:
+            if filt not in excluded_filters:
+                # don't use an empty string because of
+                # http://bugs.python.org/issue18190
+                try:
+                    getattr(checker, filt)(u'_', u'_')
+                except Exception as e:
+                    # FIXME there must be a better way to get a list of
+                    # available checks.  Some error because we're not actually
+                    # using them on real units.
+                    logging.error("Problem with check filter '%s': %s", filt, e)
+                    continue
+
+        available_checks.update(checker.categories)
+
+    return available_checks
 
 
 def get_qualitycheck_schema(path_obj=None):


docs/server/settings.rst
```diff
@@ -414,6 +414,19 @@ Translation environment configuration settings.
...
+
+``QUALITY_CHECKER``
+  Default: ``''``
+
+  The import path to a funtion that provides alternate quality checks to
+  Pootle.  When left unset then the checking functions available from the
+  Translate Toolkit, and set in the project Admin page, are used instead.
+
+  .. note:: If set only that checker function is used and the checker funtions
+     from Translate Toolkit are not used, the two cannot be selectively applied.
```
unho: This note looks confusing to me.

docs/server/settings.rst
```diff
@@ -414,6 +414,19 @@ Translation environment configuration settings.
...
 
 
+.. setting:: QUALITY_CHECKER
+
+``QUALITY_CHECKER``
+  Default: ``''``
+
+  The import path to a funtion that provides alternate quality checks to
+  Pootle.  When left unset then the checking functions available from the
+  Translate Toolkit, and set in the project Admin page, are used instead.
```
unho: I think it is more readable the following:

```
If it is unset then the Translate Toolkit checking functions are used,
unless the project has overridden this.
```

pootle/static/js/editor/app.js
```diff
@@ -1006,6 +1010,10 @@ PTL.editor = {
     if (this.filter === 'checks' && this.checks.length) {
       reqData.checks = this.checks.join(",");
     }
+    if (this.filter === 'checks' && this.categories.length) {
+      reqData.categories = this.categories.join(",");
+    }
+
```
unho: Nit: unnecessary blank line.

unho:
lgtm after addressing raised concerns.

dwaynebailey:
@unho And the vfolder stuff?

dwaynebailey:
Generally comment after a good night's sleep.

The changes I've made here work.  But I actually think the better approach is to rework QUALITY_CHECKER such that a list of alternate checkers can be added and these are actually made available in the project setup.  ``ENChecker`` becomes ``PootleChecker`` (or similar) and it just available like all the rest when you setup the project.  So we get rid of all the ``if setting.QUALITY_CHECKER`` logic and make this work generically for whatever checkers are defined.

I'm just raising this here in case anyone in review decides that this is the correct approach.  I'd like to land what I have here and look at this work as some future cleanup.

docs/server/settings.rst
julen: I believe you mean: *.. note:: If set, only the checker function defined here is used instead of the Translate Toolkit counterparts. Both cannot be selectively applied.*

julen:
One thing I notice is the category links contain the actual values of the categories in URLs, which are not user-friendly, e.g. `#filter=checks&categories=100`. Let's make this use natural human-readable names: *critical*, *functional* etc. Also, since this accepts a single category at a time, the query string parameter should be `category`.

pootle/apps/pootle_misc/checks.py
```diff
@@ -967,14 +970,30 @@ def run_given_filters(checker, unit, check_names=[]):
...
-        if filt not in excluded_filters:
-            # don't use an empty string because of
-            # http://bugs.python.org/issue18190
-            getattr(sc, filt)(u'_', u'_')
-
-    return sc.categories
+    available_checks = {}
+
+    if settings.QUALITY_CHECKER:
+        checkers = import_func(settings.QUALITY_CHECKER)()
```
julen: As we checked on IRC, this has to be a list: `checkers = [import_func(settings.QUALITY_CHECKER)()]`

julen:
> I actually think the better approach is to rework QUALITY_CHECKER such that a list of alternate checkers can be added and these are actually made available in the project setup. `ENChecker` becomes `PootleChecker` (or similar) and it just available like all the rest when you setup the project. So we get rid of all the if setting.QUALITY_CHECKER logic and make this work generically for whatever checkers are defined.

@dwaynebailey so that would expand whatever comes from TTK, right? Something like:
```python
POOTLE_QUALITY_CHECKER = {
    'evernote': 'module.implementing.checks.CheckerClass',
}
```
The checker style drop-down in the projects admin page would include all checkers from TTK + 'evernote' in this example.

--

Leaving your suggestion aside for a moment, in our current scenario we could take `ENChecker` away from Pootle and integrate it into TTK as mentioned earlier on IRC. That was actually the initial idea on our side (correct me if I'm wrong @iafan), and it would make code considerably simpler, as this new `QUALITY_CHECKER` setting and its logic added in this PR would render useless.

To make it clear, @iafan from our side the implications would be:
* We'd need to ensure our own quality checker type is assigned to every single project — it might make sense though to provide means to specify what the default quality checker is for a server.
* We'd need to ensure we update TTK from git more frequently

--

Whatever approach is considered better, I believe we can target to make such a change before the 2.7 release, as it incurs no major complexity or risks.

Understanding that having checks readily available is P1 for you @dwaynebailey, we can merge this PR now as you suggest (provided #3735 is properly fixed) and switch to any other approach sooner than later in a separate step.

iafan:
I would prefer not to have to set our checker as an override for every project we set up. So it looks like we still need a notion of either 'default checker' or 'replacement checker' setting in Pootle config. I guess the last option is what this PR does?

dwaynebailey:
@iafan the PR mostly cleans up the code to do what QUALITY_CHECKER claims to do, that is allow an override or TTK checks.  Before the PR there are a number of instances that hardcode the dependence on ENChecker.  I'll move this discussion to its own issue once this lands.

PR #3754, opened 2015-05-20
Doc fixes

julen:
These are some minor changes and cleanups I came across while reading the commands page.


dwaynebailey:
@julen gtm.  Since we'll likely drop 2.5 support with 2.7 release, simply because it will be too painful, I'm OK with the ``versionadded`` cleanup.

PR #3762, opened 2015-05-26
Cleanup: Drop handle_store() method from PootleCommand

unho:
@translate/core Please review

### DIFF
--- a/pootle/apps/pootle_app/management/commands/__init__.py
+++ b/pootle/apps/pootle_app/management/commands/__init__.py
@@ -56,16 +56,6 @@ class PootleCommand(NoArgsCommand):
                 logging.exception(u"Failed to run %s over %s's files",
                                   self.name, tp)
                 return
-        elif hasattr(self, "handle_store"):
-            store_query = tp.stores.live()
-            for store in store_query.iterator():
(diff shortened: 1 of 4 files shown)


julen:
Reasoning for this being?

While part of the code seems dead code, when in `update_stores` the `handle_translation_project()` method returns `True`, `handle_store()` is currently being called. This PR removes it for no apparent reason.

@ta2-1 could you double-check and confirm what I read is right?

unho:
@julen You are right. I will adjust this to just drop `handle_store()` on the `flush_cache` and `sync_stores` management commands.

ta2-1:
Few thoughts: 
I thought different command require different handlers and as I understood, there exist cases when it's enough to implement `handle_store()` only. It will allow to avoid some duplication of a code. Then `handle_store()` can leave in `__init__py`.
But we don't have such commands now and duplication of one for loop is not a big deal probably.

`update_stores` and `sync_stores` commands are addressed in #3583.

unho:
I am going to directly merge the `handle_store()` method removal from `flush_cache`. As @ta2-1 already said #3583 handles the removal from `update_stores` and `sync_stores` so we can wait for it to be merged.

PR #3767, opened 2015-05-28
Directory: make pootle_path unique

julen:
It doesn't make sense to have directories with duplicated `pootle_path`s; in fact, [it causes trouble to people](http://markmail.org/thread/54p23sjq3d22v675).


unho:
lgtm.

PR #3773, opened 2015-05-30
Use PEP440 versioning

dwaynebailey:
The brings PEP440 versioning in Pootle.

* We maintain a tuple ``VERSION`` that is used for versioning throughout Pootle
* The uses of ``sver`` are removed to use ``__version__``
* Docs and POT files use the correct version
* It introduces a number of helpers originally from Django but now quite liberally hacked.  So much so that I wonder if I should have turned this into a Class.

The change also include Git helpers to get changeset date, hash and branch.  I haven't looked at using those in e.g. banners when ``POOTLE_INSTANCE_ID = 'development'``


docs/developers/releasing.rst
```diff
@@ -242,31 +242,20 @@ Up version numbers
...
-
-    $MAJOR-$MINOR-$MICRO[-$EXTRA]
+    (major, minor, micro, candidate, extra)
 
 E.g. ::
 
-    1.10.0
-    0.9.1-rc1
+    (1, 10, 0, 'final', 0)
+    (2, 7, 0, 1, 'alpha', 1)
```
julen: There's an extra item here; rather `(2, 7, 0, 'alpha', 1)`

docs/developers/releasing.rst
```diff
@@ -242,31 +242,20 @@ Up version numbers
...
 
-``$EXTRA`` is optional but all the three others are required.  The first
-release of a ``$MINOR`` version will always have a ``$MICRO`` of ``.0``. So
-``2.6.0`` and never just ``2.6``.
+When in developmemt we use 'alpha' with ``extra`` of 0.  The first release of a
```
julen: *development*

pootle/apps/pootle_misc/version.py
```diff
@@ -0,0 +1,150 @@
...
+import subprocess
+
+from django.utils.lru_cache import lru_cache
+
+
+CANDIDATE_MARKERS = ('alpha', 'beta', 'rc', 'final')
+
+
+def get_version(version=None):
+    "Returns a PEP 440-compliant version number from VERSION."
```
julen: Erm, triple quotes for docstrings. Same in the other functions below.

pootle/apps/pootle_misc/version.py
```diff
@@ -0,0 +1,150 @@
...
+def get_docs_version(version=None, positions=2):
+    """Return the version used in documentation."""
+    version = get_complete_version(version)
+    candidate = _get_candidate(version)
+    candidate_pos = _get_candidate_pos(version)
+    if positions > candidate_pos:
+        positions = candidate_pos
+    if _is_development_candidate(version):
+        return 'dev'
+    else:
```
julen: `else:` redundant

julen:
Please let's move away from `pootle_misc`. Something like `pootle.core.utils.version` is a better home for these version-related utils.

pootle/apps/pootle_misc/version.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    return command_subprocess.communicate()[0]
+
+
+def get_git_changeset():
+    """Returns a numeric identifier of the latest git changeset.
+
+    The result is the UTC timestamp of the changeset in YYYYMMDDHHMMSS format.
+    This value isn't guaranteed to be unique, but collisions are very unlikely,
+    so it's sufficient for generating the development version numbers.
```
unho: I would be more in favor of using the first eight chars of the commit SHA, but I see this has been copied from Django.

dwaynebailey: The hash function appears later.  I'd prefer to not use that in any version output as its not sortable.

dwaynebailey:
All the key issue have been addressed.

* Moved to ``pootle.core.utils.version`` - I knew there was a better place
* Documented ``version.py`` till you get sick of it.

Can I get some GTM please?

unho:
lgtm

pootle/core/utils/version.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    timestamp = _shell_command('git log --pretty=format:%ct --quiet -1 HEAD')
+    try:
+        timestamp = datetime.datetime.utcfromtimestamp(int(timestamp))
+    except ValueError:
+        return None
+    return timestamp.strftime('%Y%m%d%H%M%S')
+
+
+def get_git_branch():
```
jleclanche: Why are you introducing this function?

dwaynebailey: It will allow us to expose git branch on a dev machine.  Its as useful as the hash function at the moment.

jleclanche: If you want to keep it then change it to return None for no branch. But i don't see the use; a hash should be enough.

pootle/core/utils/version.py
```diff
@@ -0,0 +1,214 @@
...
+def _shell_command(command):
+    """Return the first result of a shell ``command``"""
+    repo_dir = os.path.dirname(os.path.abspath(__file__))
+
+    command_subprocess = subprocess.Popen(
+        command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True,
+        cwd=repo_dir, universal_newlines=True
+    )
+
+    return command_subprocess.communicate()[0]
```
jleclanche: Popen with shell=True is a security risk. Not something we want to have in version. Please split the command to use a list of arguments instead. See docs:

https://docs.python.org/3/library/subprocess.html

pootle/core/utils/version.py
```diff
@@ -0,0 +1,214 @@
...
+CANDIDATE_MARKERS = ('alpha', 'beta', 'rc', 'final')
+
+
+def get_version(version=None):
+    """Returns a PEP 440-compliant version number from VERSION.
+
+    The following examples show a progression from development through
+    pre-release to release and the resultant versions generated:
+
+    >>> get_version((2, 7, 1, 'alpha', 0))
```
julen: Is this and the output right?

pootle/core/utils/version.py
```diff
@@ -0,0 +1,214 @@
...
+    candidate_pos = _get_candidate_pos(version)
+    candidate = version[candidate_pos]
+    candidate_extra = version[candidate_pos+1]
+
+    sub = ''
+    if _is_development_candidate(version):
+        git_changeset = get_git_changeset()
+        if git_changeset:
+            sub = '.dev%s' % git_changeset
+
```
julen: Nit: redundant blank line.

pootle/core/utils/version.py
```diff
@@ -0,0 +1,214 @@
...
+
+def get_git_changeset():
+    """Returns a numeric identifier of the latest git changeset.
+
+    The result is the UTC timestamp of the changeset in YYYYMMDDHHMMSS format.
+    This value isn't guaranteed to be unique, but collisions are very unlikely,
+    so it's sufficient for generating the development version numbers.
+
+    >>> get_git_changset()
```
julen: *changeset*

julen:
GTM from my side once the latest concerns have been addressed.

pootle/tools/createpootlepot
```diff
@@ -17,7 +17,7 @@ POOTLE_JS_POT="$POT_DIR/pootle_js.pot"
 
 # Header settings
 PACKAGE="Pootle"
-VERSION=`egrep "^ver" $SRC_DIR/__version__.py | sed "s,.*(\(.*\)),\1,g" | sed "s,[ ],,g" | sed "s/,/./g"`
+VERSION=$(python -c "from pootle.core.utils.version import *; print(get_main_version())")
```
jleclanche: A thought comes to mind, if you want to use it like this you could create an `if __name__ == "__main__"` block in the file which prints the version, and run it with `python -mpootle.core.utils.version`. Cleaner, and doesn't break once we start using relative imports more.

Otherwise, gtm.

pootle/core/utils/version.py
dwaynebailey: Yes, double checked with doctest

pootle/core/utils/version.py
dwaynebailey: Good spot.  First I thought it remote.  Then realised we'd rather not have code that can become a hole.  Slightly inelegant now but command specificed in the [] notation and lru_cache moved back to the calling functions, so we still can cache the results.

pootle/tools/createpootlepot
dwaynebailey: Had to make it slightly convoluted as we're just using the main version to avoid needless header noise.  But managed to craft a __main__ that allows you to ask for alternative forms as needed.

pootle/core/utils/version.py
```diff
@@ -0,0 +1,232 @@
...
+    from sys import argv
+    if len(argv) == 2:
+        if argv[1] == "main":
+            print get_main_version()
+        elif argv[1] == "major_minor":
+            print get_major_minor_version()
+        elif argv[1] == "docs":
+            print get_docs_version()
+    else:
+        print get_version()
```
jleclanche: For future reference, don't forget to add parentheses to print statements, even if you don't import print_function. It's free and it's better. :)

PR #3778, opened 2015-06-01
Changes in flush cache

unho:
@translate/core Should be easy and quick to review/reject.

Note: If this gets merged issue #3500 will need to be updated accordingly.


jleclanche:
lgtm I guess

julen:
-1 on `flush_` as I [mentioned a while ago](https://github.com/translate/pootle/issues/3500#issuecomment-73143799).

unho:
@julen Would be `clear_stats` an acceptable compromise for now?

jleclanche:
No, clear_stats has a completely different meaning.

unho:
@julen @jleclanche So one of you says no `flush_` and the other says `clear_stats` has a different meaning. Can we please settle on one of them?

julen:
I'd say go for `clear_`, — the underlying methods used by the command use the term *clear* too.

unho:
I am merging this despite not everybody agrees with the naming. Renaming should be quite simple if we eventually agree on a new name.

PR #3780, opened 2015-06-02
Legal pages: fix pending agreements

### LINKED ISSUE #3769: Legal pages don't remember your acceptance of the agreemenet
This is likely only a problem on the [LibreOffice dev server](https://translations-test.documentfoundation.org/projects/) but tracking it here so we don't forget.

julen:
Django's ORM seems to be playing with us, hence dropping to raw SQL
instead of shooting ourselves in the foot.

Fixes #3769.


jleclanche:
Syntax looks ok from a postgres perspective, but didn't test.

dwaynebailey:
Tested on LO new server.  It works! Thanks a million on this one, my sanity is restored.

PR #3801, opened 2015-06-05
Development ribbon

dwaynebailey:
This lifts the styling for the development ribbon on the EN staging server.

The ribbon display if ``POOTLE_INSTANCE_ID`` is set.  It does styling colour magic if it is set to stage or development.

The Github URL is hardcoded, ideally it should point to the correct remote. 

Would like some review and am completely open to styling fixes.

Not I'm not sure what role the CSS selector 'is-default-branch' plays in this, but it's set to on for now.


julen:
This is pulling development-specific bits into production, and honestly I don't think this should be in Pootle (a plugin perhaps?).

jleclanche:
I'm +0 on it. Don't mind it, don't find it particularly necessary.

unho:
The cleanup looks to be ok to be merged.

What is the purpose of the rest?

dwaynebailey:
@julen to be honest not sure I even would know how to make this a plugin.  Though I had your concerns about dev in prod, my view was that in production the dev parts just disappear.

@jleclanche I guess because you don't have three live servers, with three test instances and a local instance or two that you are also testing on. Doing testing on the wrong instance or a live instance is a very real problem that this would avoid.

@unho see https://translate.stage.evernote.com/projects/ the ribbon at the right hand side explains it all.

julen:
Dev parts do not disappear in production as you're including a CSS file in a bundle which is used in production.

Also there's no way to make plugins until real use-cases appear. We raised a few times in the past that specific situations like this will enable one to think and implement such extension points.

jleclanche:
I'm changing my +0 to a +1. I don't see how this PR can be made more modular tbh, it looks fine as is and it introduces an insignificant amount of code.

pootle/static/css/ribbon.css
```diff
@@ -0,0 +1,61 @@
...
+  display: block;
+  font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+  font-weight: bold;
+  font-size: 85%;
+  margin: 1px 0;
+  padding: 10px 50px;
+  min-width: 100px;
+  text-align: center;
+  text-decoration: none;
+  text-shadow: 0 0 5px #444;
```
unho: 2 spaces indent here, but below it is 4 spaces.

unho:
As previously pointed out, I wanted the cleanup to be landed, and since people is here I directly landed it in 3118d744. Perhaps it was not polite, but just wanted us to keep moving forward.

@dwaynebailey Re the ribbon, I like the idea, but I think it should be disabled by default. I don't see how we can make the CSS stay out of the bundle. A setting `POOTLE_RIBBON_ENABLED` can allow to conditionally include the CSS in the assets and pass the needed context to the template. It looks saner than adding a new app just for overextending the `layout.html` template.

PR #3804, opened 2015-06-07
Cleanup: removed some custom signals

### LINKED ISSUE #3397: Review sent and connected signals
There are some signals that are being sent but no one is listening. There are other listeners being registered for no practical use. Let's review and clean this up.

* The signals defined in `pootle_store/signals.py` are being sent in `pootle_store.fields`, `pootle_store.models` and `pootle_store.views`, but no one listens to them.
* `pootle_autonotices/__init__.py` connects to `post_save` to register new languages and projects, but with the new admin UI it makes no sense to keep these around. Besides unregistering the handlers and removing the dead code, it makes sense to cleanup the related DB tables too, as the `pootle_autonotices` app would die.

julen:
Reasoning being they're not listened anywhere internally nor documented, therefore unsupported.

Fixes #3397.


unho:
lgtm

PR #3807, opened 2015-06-09
Renamed 'overview' to 'browse'

julen:
As discussed on IRC, PRing for a good to go. Also replaces internal references besides user-facing strings.


pootle/apps/pootle_language/views.py
```diff
@@ -13,7 +13,7 @@
                                  get_table_headings)
 from pootle.core.decorators import get_path_obj, permission_required
 from pootle.core.helpers import (get_export_view_context,
-                                 get_overview_context,
+                                 get_browser_context,
```
unho: `browser`?

pootle/apps/import_export/templates/browser/index.html
```diff
@@ -1,4 +1,4 @@
-{% overextends "browser/overview.html" %}
+{% overextends "browser/index.html" %}
```
unho: If you are renaming perhaps it should be `browse/index.html`.

pootle/templates/editor/_actions.html
```diff
@@ -2,7 +2,7 @@
 <ul>
   <li class="back-to-browser">
     <a class="js-back-to-browser" href="{{ previous_url }}">
-      <span class="caption">{% trans 'Go back to overview' %}</span>
+      <span class="caption">{% trans 'Go back to browser' %}</span>
```
unho: Didn't someone raised the concern that browser can be misleading?

pootle/apps/pootle_store/urls.py
```diff
@@ -19,9 +19,9 @@
     url(r'^xhr/stats/checks/?$',
         'get_qualitycheck_stats',
         name='pootle-xhr-stats-checks'),
-    url(r'^xhr/stats/overview/?$',
-        'get_overview_stats',
-        name='pootle-xhr-stats-overview'),
+    url(r'^xhr/stats/?$',
```
unho: Just wondering why we keep the `?`. These are XHR calls so we can enforce these URLs to always end on `/` on our JS code.

unho:
Finished review. Just noticed some bits that I am not 100% ok with, but lgtm anyway.

dwaynebailey:
Similar concerns to @unho when we use browser instead of browse.  As it's a potential confusion between the webbrowser and our project browser.

Otherwise gtm

pootle/apps/pootle_language/views.py
julen: yes, in line with `get_translation_context`.

pootle/apps/import_export/templates/browser/index.html
julen: This is in line with *templates/editor*.

pootle/templates/editor/_actions.html
julen: Yes, and we are fine with that.

pootle/apps/pootle_store/urls.py
julen: Irrelevant here.

pootle/templates/editor/_actions.html
julen: I mean, what is the file/dir browser? Can you name it otherwise? cc @iafan.

julen: I can think of *Go back to browsing* or *Go back to browse view* (also, *mode* instead of *view*).

unho: Any of those two (*Go back to browsing* or *Go back to browse view*) is fine for me.

iafan: +1 for _Go back to browsing_

pootle/apps/import_export/templates/browser/index.html
jleclanche: I agree the "browser" wording is a bit confusing.

jleclanche:
gtm but i share @unho and @dwaynebailey's concerns.

jleclanche:
Woops I'm too late.

Issue #3726, opened 2015-05-12
Create the system user when upgrading from 2.6.0 to 2.7.0


dwaynebailey:
I get the following error when trying to view the timeline on this unit:
http://linode2.locamotion.org/ach/firefox-beta/translate/#filter=all&unit=17545118

This is a post mlo migration, so it could be related to bad migration.  But I guess we either have users or the code should handle the case where a user was deleted.

```pytb
Environment:


Request Method: GET
Request URL: http://linode2.locamotion.org/xhr/units/17545118/timeline/

Django Version: 1.7.8
Python Version: 2.7.6
Installed Applications:
['django.contrib.sessions',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.humanize',
 'django.contrib.sites',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'accounts',
 'contact',
 'import_export',
 'pootle',
 'pootle_app',
 'pootle_misc',
 'pootle_store',
 'pootle_language',
 'pootle_project',
 'pootle_translationproject',
 'pootle_profile',
 'pootle_statistics',
 'pootle_notifications',
 'pootle_autonotices',
 'pootle_terminology',
 'staticpages',
 'virtualfolder',
 'django_assets',
 'django_rq',
 'overextends',
 'evernote_auth',
 'evernote_reports',
 'django_extensions']
Installed Middleware:
['pootle.middleware.baseurl.BaseUrlMiddleware',
 'django.middleware.cache.UpdateCacheMiddleware',
 'pootle.middleware.cache.CacheAnonymousOnly',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.http.ConditionalGetMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'pootle.middleware.setlocale.SetLocale',
 'django.contrib.messages.middleware.MessageMiddleware',
 'pootle.middleware.errorpages.ErrorPagesMiddleware',
 'django.middleware.common.CommonMiddleware',
 'pootle.middleware.captcha.CaptchaMiddleware',
 'django.middleware.cache.FetchFromCacheMiddleware']


Traceback:
[...]

Issue #3651, opened 2015-03-23
'NoneType' object has no attribute 'pk' in /user/<user>/stats/activity


jleclanche:
See: http://pootle.lxde.org/user/jleclanche/stats/activity?month=2015-03&username=jleclanche

```pytb
Traceback:
File "/var/www/pootle/env/local/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/django/views/generic/base.py" in view
  69.             return self.dispatch(request, *args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/django/utils/decorators.py" in _wrapper
  29.             return bound_func(*args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/Pootle-2.7.0a1-py2.7.egg/pootle/apps/pootle_misc/util.py" in wrapper
  107.         return f(request, *args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/django/utils/decorators.py" in bound_func
  25.                 return func.__get__(self, type(self))(*args2, **kwargs2)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/Pootle-2.7.0a1-py2.7.egg/pootle/apps/evernote_reports/views.py" in dispatch
  99.         return super(UserActivityView, self).dispatch(request, *args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/Pootle-2.7.0a1-py2.7.egg/pootle/core/views.py" in dispatch
  91.             .dispatch(request, *args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/django/views/generic/base.py" in dispatch
  87.         return handler(request, *args, **kwargs)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/Pootle-2.7.0a1-py2.7.egg/pootle/apps/evernote_reports/views.py" in get
  102.         data = get_activity_data(self.request, self.get_object(), self.month)
File "/var/www/pootle/env/local/lib/python2.7/site-packages/Pootle-2.7.0a1-py2.7.egg/pootle/apps/evernote_reports/views.py" in get_activity_data
  430.         json['grouped'] = get_grouped_paid_words(scores, user, month)
[...]

PR #3808, opened 2015-06-09
Backports, fixes and missing migrations for 2.6

unho:


unho:
Landed two of the commits (the ones that fixed #3726 and #3651). Held the rest until there is a PR for squashing the equivalent migrations on master branch.

Also it is necessary to check if there are duplicate `pootle_path` for `Directory` before running the migration, and abort with a warning so the user can check that. Duplicate dirs have been reported in the past.

All the migrations run ok both on a fresh 2.5.1 DB dump and in 2.6 DB.

dwaynebailey:
Backporting evernote_reports doesn't make sense considering #3822 and for an app that only actually appears in 2.7.0, i.e. there is no real migration, I'm still at a complete loss as to why we need to backport that.  @jleclanche will have to explain this to me again.  But I've just done a migration to 2.7.0 and apart from where things are actually broken the migrations just seamlessly link between 2.6.0 and 2.7.0 so putting effort this seems like a spectacular waste of time.

dwaynebailey:
I'd like to test these on a LO and mlo migrations.  But general review they look good.

On mlo I notice I have waffle tables.  We had that at one time, I'll check if there are any others that should be gone.

unho:
Addressed the raised concerns.

unho:
@dwaynebailey You said that you would test again but you didn't tell how it went. I tested again yesterday with no problem, so unless some problem is raised I will merge this today.

Issue #3809, opened 2015-06-09
Vfolders name clashes with directory name


dwaynebailey:
After an extensive day of binary testing I managed to reduce the vfolder producing the error to this:

```json
[
    {
        "filters": {
            "files": [
                "dom/chrome/security/security.properties.po"
            ]
        },
        "is_browsable": true,
        "location": "/{LANG}/firefox/",
        "name": "security",
        "priority": 0.4
    }
]
```

The clash of names between ``security`` in the vfolder name and the directory seem to produce as dirty count of 6.  This doesn't seem to change with the addition of other files in that folder.

The problem disappears in a vfolder with 

```json
[
    {
        "filters": {
            "files": [
                "security/manager/chrome/pippki/pippki.properties.po"
            ]
        },
        "is_browsable": true,
        "location": "/{LANG}/firefox/",
        "name": "security",
        "priority": 0.4
    }
]
```

The same top level name of ``security``

There must be some component that I'm still missing as using the first example and calling the vfolder ``chrome`` doesn't seem to produce the same errors but by my theory they should since the names clash.

unho:
Just as an explanation of what it is going on.

> ```json
> [
>     {
>         "filters": {
>             "files": [
>                 "dom/chrome/security/security.properties.po"
>             ]
>         },
>         "is_browsable": true,
>         "location": "/{LANG}/firefox/",
>         "name": "security",
>         "priority": 0.4
>     }
> ]
> ```

The previous JSON creates the following `VFolderCachedTreeItem`s for
`/af/firefox/`:

- `/af/firefox/security/dom/chrome/security/`
- `/af/firefox/security/dom/chrome/`
- `/af/firefox/security/dom/`
- `/af/firefox/security/`

The last `VFolderCachedTreeItem` is the problematic one, since it also exists
a regular directory with the same `pootle_path`.

> ```json
> [
>     {
>         "filters": {
>             "files": [
>                 "security/manager/chrome/pippki/pippki.properties.po"
>             ]
>         },
>         "is_browsable": true,
>         "location": "/{LANG}/firefox/",
>         "name": "security",
>         "priority": 0.4
>     }
> ]
> ```

The previous JSON creates the following `VFolderCachedTreeItem`s for
`/af/firefox/`:

- `/af/firefox/security/security/manager/chrome/pippki/`
- `/af/firefox/security/security/manager/chrome/`
- `/af/firefox/security/security/manager/`
- `/af/firefox/security/security/`
[...]

Issue #3534, opened 2015-01-22
VFolders: Alter VirtualFolder.save() to to append `-virtual` to the name if necessary


unho:
If the vfolder name conflicts with any the name for any existing directory or file in the vfolder location it is necessary to rename it. Have in mind that the location can use any combination of `{LANG}` and `{PROJ}` placeholders, so it would be necessary to check all the matching TPs.

unho:
@iafan What about raising an error when we try to add a vfolder named like a regular folder?

iafan:
@unho, I'm not opposed to this approach. How will we cover the situation when the physical folder is added on update_stores, and such a virtual folder already exists? Also raise an exception? How will one rename the virtual folder to resolve the conflict?

dwaynebailey:
@iafan My view is this:

1. If there is a name clash we disable the vfolder (even if that clash happens by the appearance of a new folder that matches the vfolder)
2. Report this as part of update_stores
3. I'd like Pootle to run fine even with an unresolved clash so no exception raising that would disable anything.  An exception might be useful for someone doing some remote monitoring.
4. To be able to rename the vfolder we'd need to have the ability to delete vfolders
5.  I'd style a clashing vfolder differently on the admin view of the vfolders and disable all information.

iafan:
Thanks, @dwaynebailey. I agree with your vision here. It might be easier to simply disable the vfolder than to implicitly rename it (this would change its URL and break inbound links, if any).

As for item 4, I think at some point I shared an idea to replace all the vfolders based on the vfolder description .json file. So deleting or renaming the vfodler would mean adjusting the same master .json file that contains all the descriptions and running the `add_vfolders` command (also it might make sense to rename this command to `update_vfolders` for clarity). Such a command would complain on any folder naming conflicts.

unho:
Related #3809.

unho:
See #3905 for temporary workaround.

phlax:
this was fixed by #5221

PR #3816, opened 2015-06-11
Cache virtual folder stats using CachedTreeItem

unho:
@translate/core  Ready to review.

The purpose of these changes is to use CachedTreeItem in order to keep stats in cache, resulting in a faster retrieval. In order to do so VFolderCachedTreeItem trees are created besides regular tree, see an example in https://imgrush.com/leOJqX9pAlUj. Unfortunately creating those tree structures results in worsening the vfolder and regular directory name clash (#3534) which was reported as #3809. Stats can only be calculated using `refresh_stats_rq`.

Instructions for setup:

- You might need to make the assets to ensure you use the latest JS.
- Run `./manage.py shell_plus`
- Run `VirtualFolder.objects.all().delete()` and exit the shell
- If you have a `test_vfolders` project remove it from Pootle and from `POOTLE_TRANSLATION_DIRECTORY`
- Get https://docs.google.com/file/d/0B-8KVyCPnkZRSHAzbUFRYmV3LWM and uncompress it on your `POOTLE_TRANSLATION_DIRECTORY`
- Add a new `test_vfolders` project.
- Run `./manage.py update_stores --project=test_vfolders`
- Get http://dpaste.com/15NCKRS and save it as `vfolders_test_directory.json`
- Run `./manage.py add_vfolders vfolders_test_directory.json`
- Run `./manage.py refresh_stats_rq --project=test_vfolders`

Instructions for testing:

- Run `./manage.py runserver`
- Log in as a regular user (not admin)
- Go to `http://127.0.0.1:8000/af/test_vfolders/browser/chrome/`
- Check that:
  - the vfolders stats table is displayed
  - the listed vfolders are all the vfolders that should be displayed. The following must not be displayed:
    - not browsable vfolders,
    - completely translated vfolders,
    - priority below 1 vfolders,
    - vfolders that don't match any file in that path.
  - you can see completely translated vfolders that have failing checks.
  - the counts match. Note that each vfolder counts must only include the files they match in the current location or below it.
  -  the translate links are correct. Just are like regular translate links for this location, but with the vfolder name included just after its location, for example a vfolder with location `/af/test_vfolders/browser/` and name `test-overlapping-vfolder` will have a needs work translate link pointing to `http://127.0.0.1:8000/af/test_vfolders/translate/browser/test-overlapping-vfolder/chrome/#filter=incomplete`
- Log in as an admin user
- Check that in `http://127.0.0.1:8000/af/test_vfolders/browser/chrome/` now you can also see styled as disabled (grayed out):
  - not browsable vfolders ,
  - completely translated vfolders,
  - vfolders with priority below 1.
- Click on the `need translation` translate link for vfolder `test-overrides` on the vfolders stats table and ensure that:
  - you see `browser/test-overrides/chrome` in the navbar dropdown,
  - the URL in the browser location bar is `http://127.0.0.1:8000/af/test_vfolders/translate/browser/test-overrides/chrome/#filter=incomplete`,
  - the units displayed only belong to files matched by the vfolder filter in `/af/test_vfolders/browser/chrome/`
  - it is not possible to sort by priority.
- Select the same path to be translated in the editor, but not restricted to a vfolder: `http://127.0.0.1:8000/af/test_vfolders/translate/browser/chrome/#filter=incomplete`. Check that:
  - the unit count is bigger
  - you can sort by priority.
- Open the navbar dropdown and ensure that all the vfolders that match within current project are displayed:
  - Should be included (even if they have low priority or are completely translated):
    - test-overlapping-vfolder
    - test-overrides
    - test-low-priority
    - test-vfolder1
    - test-vfolder2
    - test-vfolder3
    - test-completely-translated-with-failing-checks
  - Are not browsable and thus should not be included:
    - test-not-browsable
    - test-completely-translated
  - Don't apply to any file, so they should not be included:
    - test-directory
    - test-newfile
- Open `http://127.0.0.1:8000/af/test_vfolders/translate/chat/#filter=incomplete` and ensure that:
  - no priority is being displayed for any of the units listed,
  - it is not possible to select priority sorting.
- Open `http://127.0.0.1:8000/af/test_vfolders/translate/dom/chrome/#filter=incomplete` and ensure that:
  - the top priority is being displayed for all the units shown (the units in this location belong to the vfolders whose name starts with `test-vfolder`, which eases looking in the JSON). Remember to ensure that the top priority is shown for a unit that is included in several vfolders.
  - you can sort by priority. Remember top priorities are above and lower at the bottom.

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -36,7 +36,6 @@ from pootle_app.models.directory import Directory
 from pootle_app.models.permissions import PermissionSet
 from pootle_store.filetypes import filetype_choices, factory_classes
 from pootle_store.util import absolute_real_path
-from virtualfolder.signals import vfolder_post_save
 
 
 RESERVED_PROJECT_CODES = ('admin', 'translate', 'settings')
@@ -287,14 +286,17 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
     @property
     def vfolders(self):
         """Return the public virtual folders for this project."""
-        # This import must be here to avoid circular import issues.
-        from virtualfolder.models import VirtualFolder
+        if 'virtualfolder' in settings.INSTALLED_APPS:
+            # This import must be here to avoid circular import issues.
+            from virtualfolder.models import VirtualFolder
 
-        return [vf.tp_relative_path
-                for vf in VirtualFolder.objects.filter(
-                    units__store__translation_project__project__code=self.code,
-                    is_public=True
-                ).distinct()]
+            return [vf.tp_relative_path
+                    for vf in VirtualFolder.objects.filter(
+                        units__store__translation_project__project__code=self.code,
+                        is_public=True
+                    ).distinct()]
+
+        return []
 
     @cached_property
     def languages(self):
@@ -524,22 +526,25 @@ class ProjectSet(VirtualResource, ProjectURLMixin):
     ### /TreeItem
 
 
-@receiver([vfolder_post_save, pre_delete])
-def invalidate_resources_cache_for_vfolders(sender, instance, **kwargs):
-    if instance.__class__.__name__ == 'VirtualFolder':
-        try:
-            # In case this is vfolder_post_save.
-            affected_projects = kwargs['projects']
-        except KeyError:
-            # In case this is pre_delete.
-            affected_projects = Project.objects.filter(
-                translationproject__stores__unit__vfolders=instance
-            ).distinct().values_list('code', flat=True)
+if 'virtualfolder' in settings.INSTALLED_APPS:
+    from virtualfolder.signals import vfolder_post_save
 
-        cache.delete_many([
-            make_method_key('Project', 'resources', proj)
-            for proj in affected_projects
-        ])
+    @receiver([vfolder_post_save, pre_delete])
+    def invalidate_resources_cache_for_vfolders(sender, instance, **kwargs):
+        if instance.__class__.__name__ == 'VirtualFolder':
+            try:
+                # In case this is vfolder_post_save.
+                affected_projects = kwargs['projects']
+            except KeyError:
+                # In case this is pre_delete.
+                affected_projects = Project.objects.filter(
+                    translationproject__stores__unit__vfolders=instance
+                ).distinct().values_list('code', flat=True)
+
+            cache.delete_many([
+                make_method_key('Project', 'resources', proj)
+                for proj in affected_projects
+            ])
 
 
 @receiver([post_delete, post_save])
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -286,14 +286,14 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
 
     @property
     def vfolders(self):
-        """Return the browsable virtual folders for this project."""
+        """Return the public virtual folders for this project."""
         # This import must be here to avoid circular import issues.
         from virtualfolder.models import VirtualFolder
 
         return [vf.tp_relative_path
                 for vf in VirtualFolder.objects.filter(
                     units__store__translation_project__project__code=self.code,
-                    is_browsable=True
+                    is_public=True
                 ).distinct()]
 
     @cached_property
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1367,6 +1367,13 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         return self.name.startswith('pootle-terminology')
 
     @property
+    def parent_vfolder_treeitems(self):
+        if 'virtualfolder' in settings.INSTALLED_APPS:
+            return self.parent_vf_treeitems.all()
+
+        return []
+
+    @property
     def units(self):
         if hasattr(self, '_units'):
             return self._units
@@ -1426,6 +1433,7 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     def delete(self, *args, **kwargs):
         parent = self.get_parent()
+        vfolder_treeitems = self.parent_vfolder_treeitems
 
         store_log(user='system', action=STORE_DELETED,
                   path=self.pootle_path, store=self.id)
@@ -1440,6 +1448,9 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         self.clear_all_cache(parents=False, children=False)
         if parent is not None:
             parent.update_all_cache()
+
+        for vfolder_treeitem in vfolder_treeitems:
+            vfolder_treeitem.update_all_cache()
 
     def makeobsolete(self):
         """Make this store and all its units obsolete."""
@@ -2139,6 +2150,17 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         """This TreeItem method is used on directories, translation projects,
         languages and projects. For stores do nothing"""
         return
+
+    def all_pootle_paths(self):
+        """Get cache_key for all parents (to the Language and Project)
+        of current TreeItem
+        """
+        pootle_paths = super(Store, self).all_pootle_paths()
+
+        for vfolder_treeitem in self.parent_vfolder_treeitems:
+            pootle_paths.extend(vfolder_treeitem.all_pootle_paths())
+
+        return pootle_paths
 
     ### /TreeItem
 
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1432,7 +1432,7 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
             self.update_dirty_cache()
 
     def delete(self, *args, **kwargs):
-        parent = self.get_parent()
+        parents = self.get_parents()
         vfolder_treeitems = self.parent_vfolder_treeitems
 
         store_log(user='system', action=STORE_DELETED,
@@ -1446,8 +1446,8 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         super(Store, self).delete(*args, **kwargs)
 
         self.clear_all_cache(parents=False, children=False)
-        if parent is not None:
-            parent.update_all_cache()
+        for p in parents:
+            p.update_all_cache()
 
         for vfolder_treeitem in vfolder_treeitems:
             vfolder_treeitem.update_all_cache()
@@ -2045,11 +2045,11 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
     def can_be_updated(self):
         return not self.obsolete
 
-    def get_parent(self):
+    def get_parents(self):
         if self.parent.is_translationproject():
-            return self.translation_project
+            return [self.translation_project]
         else:
-            return self.parent
+            return [self.parent]
 
     def get_cachekey(self):
         return self.pootle_path
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1433,7 +1433,6 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     def delete(self, *args, **kwargs):
         parents = self.get_parents()
-        vfolder_treeitems = self.parent_vfolder_treeitems
 
         store_log(user='system', action=STORE_DELETED,
                   path=self.pootle_path, store=self.id)
@@ -1448,9 +1447,6 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         self.clear_all_cache(parents=False, children=False)
         for p in parents:
             p.update_all_cache()
-
-        for vfolder_treeitem in vfolder_treeitems:
-            vfolder_treeitem.update_all_cache()
 
     def makeobsolete(self):
         """Make this store and all its units obsolete."""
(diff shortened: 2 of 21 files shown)


dwaynebailey:
I've tested this extensively on mlo2 - linode.locamotion.org.  We picked up a few of the miscount and dirty count issues that seem to relate to name clashes.  Those I think should be fixed outside of this as without clashes this is working well.

In terms of speed we're slower than a project without vfolders.  But my stopwatch based test don't show major slowdown due to the vfolders.

Loading of vfolders is time consuming, but not massively so.

pootle/apps/pootle_store/models.py
```diff
@@ -1437,8 +1439,11 @@ def delete(self, *args, **kwargs):
         super(Store, self).delete(*args, **kwargs)
 
         self.clear_all_cache(parents=False, children=False)
-        if parent is not None:
-            parent.update_all_cache()
+        for p in parents:
+            p.update_all_cache()
+
+            for vfcti in vfctis:
```
julen: Any reader will need some help to decipher any variables named like this — let's have some human-readable name and reduce the cognitive load.

unho: Any suggestion?

julen: I don't even know what it means :stuck_out_tongue:

unho: `vfcti` is short for `VFolderCachedTreeItem`.

unho:
@translate/core Please review and test. Lets see if we can land this sooner than later.

dwaynebailey:
Sorry wasn't clear in my first comments its GTM but I haven't done any code review.

pootle/apps/pootle_store/models.py
dwaynebailey: I think the point being that you have to explain what it means here.  Good luck to anyone reviewing this code at a later date.  If you where treading with fear that you'd go over 80 chars, remember that its more important that the code id readable before you worry that the line is 100 characters long.

pootle/apps/pootle_store/models.py
```diff
@@ -2033,11 +2038,15 @@ def get_max_unit_revision(self):
...
-    def get_parent(self):
+    def get_parents(self):
         if self.parent.is_translationproject():
-            return self.translation_project
+            parents = [self.translation_project]
         else:
-            return self.parent
+            parents = [self.parent]
+
+        parents.extend(self.parent_vf_treeitems.all())
```
ta2-1: I think we should skip DB querying for `parent_vf_treeitems.all()` if we don't use virtual folders.

pootle/apps/pootle_store/models.py
```diff
@@ -2139,6 +2148,17 @@ def refresh_stats(self, include_children=True, cached_methods=None):
         languages and projects. For stores do nothing"""
         return
 
+    def all_pootle_paths(self):
+        """Get cache_key for all parents (to the Language and Project)
+        of current TreeItem
+        """
+        pootle_paths = super(Store, self).all_pootle_paths()
+
+        for vfcti in self.parent_vf_treeitems.iterator():
```
ta2-1: There is the same the I mentioned in the comment above. Extra DB query if we don't use VF.

pootle/apps/pootle_store/views.py
```diff
@@ -845,10 +845,11 @@ def get_stats(request, *args, **kwargs):
     stats = request.resource_obj.get_stats()
 
     if isinstance(request.resource_obj, Directory):
-        stats['vfolders'] = VirtualFolder.get_stats_for(
-            request.resource_obj.pootle_path,
-            request.user.is_superuser
-        )
+        stats['vfolders'] = {}
+
+        for vfcti in request.resource_obj.vf_treeitems.iterator():
```
ta2-1: Extra DB query if we don't use VF.

pootle/apps/pootle_store/models.py
```diff
@@ -1424,7 +1425,8 @@ def save(self, *args, **kwargs):
             self.update_dirty_cache()
 
     def delete(self, *args, **kwargs):
-        parent = self.get_parent()
+        parents = self.get_parents()
+        vfctis = self.parent_vf_treeitems.all()
```
ta2-1: Extra DB query if we don't use VF.

pootle/apps/pootle_translationproject/models.py
```diff
@@ -324,8 +325,20 @@ def get_children(self):
...
+    def clear_all_cache(self, children=True, parents=True):
+        super(TranslationProject, self).clear_all_cache(children=children,
+                                                        parents=parents)
+
+        # VFolderCachedTreeItem can only have VFolderCachedTreeItem parents so
+        # it is necessary to flush their cache by calling them one by one.
+        tp_vfctis = VFolderCachedTreeItem.objects.filter(
+            pootle_path__startswith=self.pootle_path
+        )
+        for vfcti in tp_vfctis.iterator():
```
ta2-1: Extra DB query if we don't use VF.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -171,10 +171,11 @@ def browse(request, translation_project, dir_path, filename=None):
             })
 
             #FIXME: set vfolders stats in the resource, don't inject them here.
-            stats['vfolders'] = VirtualFolder.get_stats_for(
-                directory.pootle_path,
-                all_vfolders=is_admin
-            )
+            stats['vfolders'] = {}
+
+            for vfcti in directory.vf_treeitems.iterator():
```
ta2-1: Extra DB query if we don't use VF.

pootle/core/helpers.py
```diff
@@ -155,8 +157,10 @@ def get_browser_context(request):
 
     filters = {}
 
-    if (not isinstance(resource_obj, Store) and
-        VirtualFolder.get_matching_for(request.pootle_path).count()):
+    if ((isinstance(resource_obj, Directory) and
+         resource_obj.vf_treeitems.count()) or
```
ta2-1: Extra DB query if we don't use VF?

pootle/apps/pootle_store/models.py
unho: This is a global reply to all the equivalent comments.

We can check if `virtualfolder` app is in `INSTALLED_APPS` in every single place we query the DB for vfolders. I am going to add another commit to the PR in order to do that.

pootle/apps/pootle_store/models.py
jleclanche: http://www.slideshare.net/pirhilton/how-to-name-things-the-hardest-problem-in-programming

pootle/apps/pootle_store/models.py
ta2-1: That would be great)

unho:
@translate/core Did several minor adjustments:

- Rebased on top of latest master.
- Renamed `vfcti` to `vfolder_treeitem`. If anyone is not ok with this name please let us know.
- Added checks for `virtualfolder` app in `INSTALLED_APPS` to avoid extra DB hits. This includes adding some convenience properties in `Directory` and `Store`.
- Some minor cleanups.

unho:
Just added to this PR a commit that renames `is_browsable` to `is_public`. The JSON provided in the instructions to test this PR has been updated accordingly.

pootle/apps/virtualfolder/models.py
```diff
@@ -54,8 +56,8 @@ class VirtualFolder(models.Model):
         help_text=_('Number specifying importance. Greater priority means it '
                     'is more important.'),
     )
-    is_browsable = models.BooleanField(
-        _('Is browsable?'),
+    is_public = models.BooleanField(
+        _('Is public?'),
         default=True,
         help_text=_('Whether this virtual folder is active or not.'),
```
julen: Based on a previous IRC discussion, I guess *active* sort of contradicts with *public* here, no?

pootle/apps/virtualfolder/models.py
```diff
@@ -347,19 +275,156 @@ def get_adjusted_pootle_path(self, pootle_path):
...
 
+class VFolderCachedTreeItemManager(models.Manager):
+    use_for_related_fields = True
+
+    def get_queryset(self):
+        return super(VFolderCachedTreeItemManager, self) \
+            .get_queryset().select_related('vfolder')
+
+
+class VFolderCachedTreeItem(models.Model, CachedTreeItem):
```
julen: As I was discussing with @ta2-1 the other day, we are very confused to have a model named after *CachedTreeItem*, giving the false impression this object comes from Redis directly, when it's a DB model.

We are not either sure of the need for a separate model here (there's also `VirtualFolder`, which isn't cached).

julen:
Just checking: is this PR supposed to considerably reduce the amount of SQL queries needed in vfolders? In previous testing there were circa 172 queries in a browsing page (cold start), vs. circa 167 now. Did you verify this on your end?

On a separate note, would it make sense to name migration files after what they actually do? Auto-generated filenames with datetimes give no clue what the migrations are about. FWIW we already have named migrations in the repo.

I'm delegating a deeper stats/cache review on @ta2-1. Consider a GTM from him as a +1 from my side.

pootle/apps/virtualfolder/models.py
unho: Will rename *active* to *public* as well.

unho:
> Just checking: is this PR supposed to considerably reduce the amount of SQL queries needed in vfolders? In previous testing there were circa 172 queries in a browsing page (cold start), vs. circa 167 now. Did you verify this on your end?

No, the purpose of this PR is to improve the performance. I haven't checked if the number of queries has changed.

> On a separate note, would it make sense to name migration files after what they actually do? Auto-generated filenames with datetimes give no clue what the migrations are about. FWIW we already have named migrations in the repo.

Just renamed the migrations.

pootle/apps/virtualfolder/models.py
unho: We can rename it to *VirtualFolderTreeItem* to keep it descriptive while dropping the *Cached* part. But it can still be confusing.

On the need for a separate model:

Lets suppose we add a `VirtualFolder` with location `/af/firefox/browser/` with a filter `chrome/browser/overview/one.po, chrome/browser/migrations/two.po`. But we want to have specific stats for the virtual folder in each path between the vfolder files and the root location. Specifically the vfolder will have different stats for path `/af/firefox/browser/chrome/browser/overview/` and path `/af/firefox/browser/chrome/browser/migrations/` and also for their parent path `/af/firefox/browser/chrome/browser/`.

So since the TreeItem class provides the stats caching infrastructure provided that a tree structure is used it felt more natural to have a separate class to keep those treeitems for the vfolders.

Having `VFolderCachedTreeItem` also allows to easily retrieve all the vfolders for a given `Directory` without need to filter the `VirtualFolder` instances based on if they have units below that directory. See https://github.com/translate/pootle/pull/3816/files#diff-2595359c0c549cb2373dc7ec23dc8cb2L133 compared to https://github.com/translate/pootle/pull/3816/files#diff-2595359c0c549cb2373dc7ec23dc8cb2R289

I must say I don't see how `VirtualFolder` and `VFolderCachedTreeItem` can be merged together without either:

- Going back to having methods that convert provided directory pootle_path and vfolder to the equivalent `VFolderCachedTreeItem` pootle_path or
- Replicating the `VirtualFolder` specific data across all the `VFolderCachedTreeItem` instances for each vfolder.

I am not sure if this clarifies this to you. Please ask again for further explanations if necessary.

ta2-1: @unho, let's don't speak about merging together `VirtualFolder` and `VirtualFolderCachedTreeItem`. 
Actually we have one `RootVirtualFolder` and its children.
What do you think if we replace `VirtualFolder` with `RootVirtualFolder(models.Model, CachedTreeItem)` and then `VirtualFolderCachedTreeItem` can become `VirtualFolder(models.Model, CachedTreeItem)`? 
And then even if we need to override `CachedTreeItem` that class can be called `VirtualFolderCachedTreeItem`.

unho: I find `RootVirtualFolder` to be more confusing, since actually that is the virtual folder.

Also in the current stage there is no need for `CachedTreeItem` to be applied to that `RootVirtualFolder` model since we don't yet display whole stats for a virtual folder. But in case we do we still have to decide what `get_cachekey` is going to return for such items. Anyway this feels like it belongs to further developments to display virtual folders outside a specific TP.

unho: Just for the record. After discussion with Taras on IRC we decided to rename `VFolderCachedTreeItem` to `VirtualFolderTreeItem`. It has the same number of characters but it is easier to understand. `VirtualFolder` is kept unchanged.

unho:
@translate/core Renamed `VFolderCachedTreeItem` to `VirtualFolderTreeItem`. That required changing migrations, so you will have to:

- Remove the migrations from `django_migration_history` DB table,
- Remove the created DB tables (the ones starting with `virtualfolder`), and
- Run the migrations again.

Please let me know if there is any other blocker. We would like to release Pootle 2.7 by this week and this PR is blocking it.

pootle/apps/pootle_project/models.py
```diff
@@ -286,14 +286,14 @@ def is_terminology(self):
 
     @property
     def vfolders(self):
```
ta2-1: @unho, this property should also check if `virtualfolder` app is used. It was added before this PR changes so I don't know if we can do it here.

unho: @ta2-1 Yes, adding it to the PR. Any other problem with the PR?

ta2-1:
@unho, we have wrong vfolder stats after any store file is marked as obsolete. 
`parent.update_all_cache()` refreshes stats for regular tree items (see
https://github.com/translate/pootle/blob/bbf14fed9a63bf690d98a8ca63bc6b1fac8d9039/pootle/apps/pootle_app/project_tree.py#L159). We need to check if this parent has virtual folder tree items and refresh them too.

ta2-1:
@unho, GTM after addressing previous comment.

PR #3821, opened 2015-06-11
Fix #3752 and assorted cleanups

### LINKED ISSUE #3752: Generate help/_ttk_quality_checks.html on setup.py build
The template is missing and causing a TemplateDoesNotExist on /help/quality-checks/.

According to @dwaynebailey there is a manage.py command to do that. Whatever that command does, it should be built upon setup.py build instead.

jleclanche:
The cleanups were needed to be able to import those parts of pootle from setup.py.


dwaynebailey:
Question.  We've removed all these docs on fixing the checks.  What happens when you upgrade ttk and get new checks.  In real life you'd need to built those again, now you can't because you'll not see any notes.  This does need to be instructions to update the page poiting to the setup.py command.  Making a Pootle package is not the only time this rebuild is needed.

docs/developers/releasing.rst
```diff
@@ -148,7 +148,7 @@ quality checks.
 
 .. code-block:: bash
 
-    $ pootle regenerate_checks_descriptions
+    $ DJANGO_SETTINGS_MODULE=pootle.settings ./setup.py build_checks_templates
```
unho: Is this ok?

unho: I mean, can't it be split into two commands, one for setting the envvar and another to run the command?

unho: Have you tried by exporting that envvar within the command?

unho:
How does this cope with the scenario where you installed Pootle using pip and you upgraded to a newer TTK that requires rebuilding the checks descriptions?

docs/developers/releasing.rst
jleclanche: what?

jleclanche:
There is no reason you should upgrade ttk without upgrading pootle, the use case is far too remote to care about.

docs/server/upgrading.rst
```diff
@@ -144,20 +144,6 @@ To perform the upgrade follow the next steps:
     (env)$ pootle setup
 
 
-* If you are upgrading from Pootle 2.7.0 or later you also have to regenerate
-  the quality checks descriptions.
```
unho: Are the quality checks descriptions regenerated by simply installing the package when upgrading to 2.7 or later?

unho:
The assorted cleanups are lgtm.

docs/developers/releasing.rst
unho: I asked if it is possible to include the exporting of `DJANGO_SETTINGS_MODULE` within `setup.py` instead of prepending it to every call of this command. We have `os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'` in several management commands, including the one this PR replaces.

jleclanche: This is irrelevant.

docs/server/upgrading.rst
dwaynebailey: The quality checks are part of the package.  You should never have to regenerate them on upgrade.

dwaynebailey:
gtm from me. The use case of a ttk upgrade is real. But I now agree with @jleclanche so unusual that the effort isn't worth it.

unho:
This is basically moving the shit around. Not sure how this is better than the previous approach.

jleclanche:
I'm not really in the mood to re-explain why setup.py is a more appropriate place for this.

PR #3827, opened 2015-06-15
German translation...

pfennig59:
...continued.


dwaynebailey:
Looks like there is a messup in the PR.  Can you try rebase.  Alternatively you can simply upload the PO file to Pootle.

dwaynebailey:
Landed in d2c99dddbde7f278cfe2dcffa432bf60e1974e80 - thanks again.  I'll upload this also to http://pootle.locamotion.org/projects/pootle/

@pfennig59 we don't like merge commits in Pootle, so careful when you land your work.  Easiest way around this is to use ``git rebase``.

pfennig59:
I'm sorry. Seems I misunderstood you.

Issue #3819, opened 2015-06-11
Traceback when extracting terminology


julen:
If one tries to generate a terminology using the extraction feature, the result is:

```python
Traceback (most recent call last):
  File "~/.virtualenvs/pootle/lib/python2.7/site-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "~/.virtualenvs/pootle/lib/python2.7/site-packages/django/db/transaction.py", line 394, in inner
    return func(*args, **kwargs)
  File "~/dev/translate/pootle/pootle/core/decorators.py", line 107, in wrapped
    return func(request, path_obj, *args, **kwargs)
  File "~/dev/translate/pootle/pootle/core/decorators.py", line 316, in _wrapped
    return func(request, *args, **kwargs)
  File "~/dev/translate/pootle/pootle/apps/pootle_terminology/views.py", line 111, in extract
    for store in translation_project.stores.live()
  File "~/dev/translate/pootle/pootle/apps/pootle_terminology/views.py", line 111, in <genexpr>
    for store in translation_project.stores.live()
AttributeError: 'Store' object has no attribute '_get_total_wordcount'
```

Guess what, it's been broken quite a bit of time.

Instead of fixing this, I'm more in favor of dropping terminology auto-generation as people already mentioned in #3656.

unho:
I am ok with dropping terminology auto-generation if we provide some fallback.

dwaynebailey:
@unho can you explain what you mean by fallback.  People can already use terminology if there is a pootle-terminology.po file or get it from the Terminology project.  So it isn't a case of fallback, the usage exists just people need to make their terminology lists outside of Pootle

gterzian:
In the meantime, here is a quick fix for the missing method: https://github.com/translate/pootle/pull/3832

unho:
By fallback I meant provide some hints on docs about how to workaround the absence of the feature. For example "Use this tool to create PO files with terminology and added them to your project as `pootle-terminology.po for each TP in the project.`"

PR #3832, opened 2015-06-17
Fix missing _get_total_wordcount on Store

gterzian:
WIP for https://github.com/translate/pootle/issues/3819

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -2044,7 +2044,7 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         return self.pootle_path
       
     def _get_total_wordcount(self):
-        pass
+        return self._get_wordcount_stats()['total']
 
     def _get_wordcount_stats(self):
         """calculate full wordcount statistics"""
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -2042,7 +2042,7 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     def get_cachekey(self):
         return self.pootle_path
-  
+
     def _get_total_wordcount(self):
         return self._get_wordcount_stats()['total']
 
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -2042,6 +2042,9 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     def get_cachekey(self):
         return self.pootle_path
+      
+    def _get_total_wordcount(self):
+        pass
 
     def _get_wordcount_stats(self):
         """calculate full wordcount statistics"""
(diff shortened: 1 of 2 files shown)


unho:
Hi.
I think the right fix is to instead alter https://github.com/translate/pootle/blob/master/pootle/apps/pootle_terminology/views.py#L110 to be:

```python
        source_words = sum(store. _get_wordcount_stats()['total']
```

since the `_get_total_wordcount` method was dropped as part of 2c31ad9c.

gterzian:
@unho Thanks for the suggestion! I think it's a good idea, and I couldn't quickly find an existing test for that view or a similar one which I could use as an example to write such a test myself. Could you perhaps point me in the right direction?

unho:
I myself am a bit lost when it comes to testing. Perhaps @julen can provide some hints.

gterzian:
@unho, OK I understand. @julen, please advise on this. One of the reason I just re-added the method was so I wouldn't have to write a test for the whole view, especially since there are already model unit-tests I could quickly use as a template.

dwaynebailey:
@gterzian this doesn't really address the issue as raised in #3819, this is a bandaid.  At the end of the report you can see the real proposed solution which is to remove auto generation of terminology from Pootle.

I'd suggest doing that or picking another problem to address.  I like that you've added tests and such but this doesn't make much sense to land if a) we'll rip it out later and b) it clutters the code we're trying to cleanup with methods that we'll want to remove anyway.

gterzian:
@dwaynebailey  Thanks Dwayne. You obviously have a great team and done an equally great job at steering this project over the years, I take your feedback very seriously.

I'm sure you're aware I'm doing this as part of your interview process, and at this point we might do best to focus on figuring out whether we're going to be a good fit at all. I might have totally different ideas than you when it comes to writing software, and I'de be very interested in hearing about those differences. So I'm going to go ahead and tell you a bit more about how I like to work, in the context of the current discussion: 

* Bandaid is a honking great idea, and we need more of that :smiley:  
* Pootle is a huge project, and although it was obviously written by Python experts, the inevitable complexity has crept up over the years. That one module alone is 2000 lines. I think it could be great to have hundreds of pieces of "bandaid" (meaning small changes, with tests) applied steadily to the project. 
* Perhaps in this particular case, it would be better to remove this terminology app altogether. Still, there's always a quick and small fix you can apply today, and a larger more difficult task you can do, well, let's say "some other day". 
* One of my favorite past-times is to focus on making it easy for people to write tests. Have you considered switching to using `unittest`, perhaps combined with a tool like https://github.com/Fluxx/exam ? 

So that's it. I guess I'm just a simple "bandaid" type of guy. I'd be very interested in hearing your and your team's thoughts on the above points. It could be really helpful to find out if we're a good fit :smile:

julen:
@gterzian thanks for your effort on trying to fix the feature. However, as @dwaynebailey says the intention of #3819 is to get rid of the feature no more.

Seeing that this was resulting in unrelated discussions and with the will to keep us focused and avoid non-productive efforts, I've already wiped out the feature.

Feel free to continue out of this PR with any other discussions you might wanna bring up.

PR #3839, opened 2015-06-17
Some l10n fixes

unho:
I cam across these when translating Pootle.


dwaynebailey:
This close to release I'd prefer not to fiddle with quotes which is a pure stylistic thing.  Every single translator who translated that now has to fix their translations because of this, there is no value to them and little value to us.

GTM despite that as it's buried within good changes.

julen:
GTM.

PR #3840, opened 2015-06-18
Upgrade and installation docs adjustments

unho:
There are still some unhandled concerns in f27b4f8c and 98511c41.

### DIFF
--- a/docs/server/installation.rst
+++ b/docs/server/installation.rst
@@ -118,7 +118,7 @@ Now create a virtual environment on your location of choice by issuing the
   $ virtualenv env
 
 
-To activate the virtual environment run the ``activate`` script:
+To activate the virtual environment run the :command:`activate` script:
 
 .. code-block:: bash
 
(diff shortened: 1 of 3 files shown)


docs/server/installation.rst
```diff
@@ -312,19 +312,6 @@ the database and populate it with initial data. This is done by executing the
   (env) $ pootle initdb
 
 
-.. _installation#admin_user:
-
-Creating an admin user
-----------------------
```
dwaynebailey: This is required.

Allauth is setup to require an email.  When you first log in as admin/admin it will claim to be sending you an email but can't because there is no admin user.

Forcing the creation of a superuser ensures that the email is created.

dwaynebailey:
@unho GTM except the bit about createsuperuser

docs/server/installation.rst
unho: This can probably be circumvented by setting an email when creating the user on `initdb`. But that will happen in the future so I prefer to land this and focus on more pressing tasks.

PR #3844, opened 2015-06-19
Prevent raise EmptyResultSet

cedk:
When there are no ids matching in the indexer, the filter is against an empty list for which django raise an exception.
So we initialize the list with a fake id -1.

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -292,7 +292,7 @@ def get_search_step_query(request, form, units_queryset):
                           u" using database search")
             return get_non_indexed_search_step_query(form, units_queryset)
         else:
-            alldbids = [-1]
+            alldbids = []
             for translation_project in translation_projects:
                 logging.debug(u"Found %s indexer for %s, using indexed search",
                               translation_project.indexer.INDEX_DIRECTORY_NAME,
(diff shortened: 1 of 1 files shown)


jleclanche:
That's crazy - just check whether `alldbids` is empty before returning.

jleclanche:
BTW what are you fixing here exactly?

cedk:
I fix the raise of the exception which makes search on the web-page never ending.
I'm not familiar with the django ORM so I don't know if I can return units_queryset without the filter call.

jleclanche:
Sure you can.

cedk:
OK but I discover the .none method that does what I need.

unho:
Can you please squash the commits?

cedk:
I don't know how to do that.

unho:
@cedk First make sure you have some reference to previous status (the commits that are currently on this PR). Then read "Squashing commits" in https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History and http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html in case you have some trouble understanding. Then apply to your case and force push your changes so this PR gets updated.

cedk:
This is too much work for a simple patch. Feel free to apply it.

unho:
Merged in 5cc0e4f6. Thanks for your contribution.

PR #3845, opened 2015-06-19
Rename evernote_reports app to just reports

unho:
Tested and works both for fresh install and upgrade.


jleclanche:
gtm

julen:
I'd strongly prefer to be consistent with the entire codebase in respect to URL pattern naming and stick to the `pootle-reports-` prefix.

unho:
@julen I don't disagree. Are you going to do it, or should I take the renaming?

julen:
Feel free to go ahead and PR it.

PR #3849, opened 2015-06-20
Fix plural string

SafaAlfulaij:
Try to fix a plural string.
Even if it don't make since to have a singular form of the string, It doesn't matter. Some languages like Arabic has 6 plural forms 4 of them are if count>2.


unho:
Merged in 8737d585. Thanks for your contribution!

PR #3883, opened 2015-07-05
Reintroduce support for --modified-since in list_projects/list_languages

### LINKED ISSUE #3864: list_language no longer has an idea of a changeid
We dropped change_id concept with 2.7.0 and now we have the idea of ``revision``.  Previously I could get a list of active languages since a `change_id` using ``list_languages``.  Now however I cannot.  I used `change_id` in the past to allow me to decide which languages to sync.  I also use it to record in a log which languages have been active since the last changeid.

So this is to either

1. Bring back ``--revision`` to ``list_languages`` so we can see languages active since some revision
2. Explain what other process could be used to gather this info

jleclanche:
Was dropped in 71fddc04d00aa3ee13e8fa2eb9448a6e19b3c805

Fixes #3864

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -265,6 +265,9 @@ list_languages
 Lists all the language codes for languages hosted on the server. This can be
 useful for automation.
 
+Accepts the :option:`--modified-since` parameter to list only those languages
+modified since the revision given by :djadmin:`revision`.
+
 
 .. django-admin:: list_projects
 
@@ -273,6 +276,9 @@ list_projects
 
 Lists all the project codes on the server. This might can be useful for
 automation.
+
+Accepts the :option:`--modified-since` parameter to list only those projects
+modified since the revision given by :djadmin:`revision`.
 
 
 .. django-admin:: revision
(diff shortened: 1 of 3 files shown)


jleclanche:
cc @dwaynebailey

jleclanche:
I don't know what's failing in travis...

unho:
@jleclanche This is what is failing: https://travis-ci.org/translate/pootle/jobs/69608166#L784

unho:
@jleclanche Perhaps ``:djadmin:`revision` ``  is what you need.

docs/server/commands.rst
```diff
@@ -274,6 +277,9 @@ list_projects
 Lists all the project codes on the server. This might can be useful for
 automation.
 
+Accepts the ``--modified-since`` parameter to list only those projects
```
unho: Nit: Use `` :option:`--modified-since` ``. The same for the above appearance.

pootle/apps/pootle_app/management/commands/list_languages.py
```diff
@@ -19,6 +19,9 @@ class Command(NoArgsCommand):
     option_list = NoArgsCommand.option_list + (
             make_option('--project', action='append', dest='projects',
                         help='Limit to PROJECTS'),
+            make_option("--modified-since", action="store", dest="modified_since",
+                        type=int,
+                        help="Only process translations newer than revision"),
```
unho: Nit: Only process translations newer than **specified** revision

unho:
Some nits raised, but lgtm.

PR #3885, opened 2015-07-07
Rename reports app url patterns

unho:


julen:
GTM.

julen:
:+1: thanks for taking care of this!

PR #3910, opened 2015-07-21
Display email on public profile

### LINKED ISSUE #3901: Show email of users in Pootle
Right now in many cases, it's impossible to know who is who on Pootle. In many occasions, I need to contact localizers in regards to an urgent project, but am unable to because people use nick names that do not correspond to any existing Mozillians profile and that do not appear on the l10n team page. Not having emails accessible prevents us from contacting these localizers in a timely manner. Requesting a fix for this :)

unho:
Picked from ac2b602b.

Fixes #3901.

### DIFF
--- a/pootle/templates/user/profile.html
+++ b/pootle/templates/user/profile.html
@@ -18,6 +18,9 @@
         {{ object.display_name }}
       {% endif %}
     </h1>
+    {% if user.is_superuser or request.permissions.administrate %}
+    <p><a href="mailto:{{ object.email }}">{{ object.email }}</a></p>
+    {% endif %}
   </div>
 
 {% if not object.is_meta %}


julen:
Even if this is admin-only, @iafan would appreciate your input on this.

pootle/templates/user/profile.html
```diff
@@ -18,6 +18,15 @@ <h1 class="user-name">
         {{ object.display_name }}
       {% endif %}
     </h1>
+    {% if user.is_superuser or request.permissions.administrate %}
+    <p>
+      {% if object.email %}
+      <a href="mailto:{{ object.email }}">{{ object.email }}</a>
+      {% else %}
+      {% trans 'No email address.' %}
```
julen: Since we require email addresses, this is redundant.

unho: Thanks. Removing it.

unho:
@iafan Can you please quickly review? Thanks in advance.

julen:
Just to make things easier, @iafan here's how this looks like when admins visit any user profile page.
john_doe_profile

iafan:
@unho, @julen, what happens when one clicks on that email?

unho:
@iafan It is a regular `mailto` URL, so it should trigger the default handling for those URLs, which usually is opening the mail program (Thunderbird and such).

iafan:
Ok, GTM.

Issue #3802, opened 2015-06-06
Improve performance of inserting/deleting units in the store


julen:
*(Original issue and discussion at evernote/pootle#145)*

Currently each unit stores its position (index) in the .po file. Consider a .po file with 1000 units. If a new unit is added to the very top of the .po file, and then `update_stores` is run, the new unit will be created in the database, but in addition to that, 1000 units' position values will be updated (this is another 1000 DB updates). Inserting one unit into a large store can take several seconds. And this happens for each unit that is being inserted or deleted, which slows down the sync cycle.

We'll try to overcome this by implementing the following:

* Update the store updating algorithm, so that when we're trying to insert `N` units after unit with index `X`, and the gap between `X` and next index `Y` is smaller than `N`, then run a query that shifts all indexes starting with `Y` at once by the value of `N-(Y-X)` (which is cheap). Then do an insert.
* Deleting a unit (or making it an obsolete one) shouldn't affect other units' indexes. We just introduce a natural gap in numeration that can be reused later. Deleting units should be done **before** we insert new units, because the most common scenario is when the string is simply altered, which leads to deleting the old unit and creating a new one at exactly the same index. In this case, no index-shifting queries are necessary.

The code resulting out of this issue must come with unit tests.

Also let's measure the old & new methods over the same operations to see the performance gains we get out of this.

PR #3912, opened 2015-07-22
Cleanup store.update()

julen:
These are follow-up cleanups on #3802. There's more TBD here (especially the difflib part), but focusing on these parts for now. 

PRing to have the code checked by others and get a green light.

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1796,7 +1796,7 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
                         sub.save()
 
                     # Set unit fields if target was updated
-                    if SubmissionFields.TARGET in create_subs:
+                    if unit._target_updated:
                         unit.submitted_by = system
                         unit.submitted_on = current_time
                         unit.reviewed_on = None
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1586,6 +1586,53 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
             index=op(F('index'), delta)
         )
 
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -1704,105 +1850,26 @@ def insert(i1, j1, j2):
...
-                if self.last_sync_revision is not None:
-                    filter_by.update({'revision__gt': self.last_sync_revision})
-
-                # If a dbunit is obsolete
-                # then the dbunit should be resurrected in any case
-                modified_units = set(
-                    Unit.objects.filter(**filter_by).exclude(state=OBSOLETE)
-                                .values_list('id', flat=True).distinct()
-                )
+            changes['obsolete'] = obsoleted
```
unho: Is there some reason why this is not like the following?

```python
            changes['obsolete'] = self.mark_units_obsolete(obsolete_dbids)
```

It would avoid having an unnecessary variable and reduce another line of code.

julen: No specific reason

jleclanche:
Looks like a really good cleanup.

pootle/apps/pootle_store/models.py
julen: Ah yeah, just recalled there was a reason: I wanted to get rid of the dictionary altogether, but didn't reach that point as that requires cleaning up the difflib-related code too — will avoid temporary vars for the time being.

unho:
lgtm, after nit fixed. Since issue raised on https://github.com/julen/pootle/commit/8cb354c591f04ea3e661638e86cd8b4c0d1285d3#commitcomment-12295727 is already present on master I am ok with it to be fixed on a separate PR.

ta2-1:
Let's merge it.

Issue #3911, opened 2015-07-21
SQL error when running update_stores on an XLIFF project


dwaynebailey:
I'm assuming that the central issue may be that this related to XLIFF support on Pootle 2.7.0.  But I'm not sure if the issue is just the XLIFF file.

```pytb
2015-07-21 16:12:29,073 ERROR Failed to run update_stores over /cy/ios/firefox-ios.xliffTraceback (most recent call last):
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_app/management/commands/__init__.py", line 65, in do_translation_project
    self.handle_store(store, **options)
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_app/management/commands/update_stores.py", line 50, in handle_store
    store.update(overwrite=overwrite, only_newer=not force)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/transaction.py", line 394, in inner
    return func(*args, **kwargs)
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_store/models.py", line 1746, in update
    self.save()
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_store/models.py", line 1421, in save
    super(Store, self).save(*args, **kwargs)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/base.py", line 589, in save
    force_update=force_update, update_fields=update_fields)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/base.py", line 617, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/base.py", line 679, in _save_table
    forced_update)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/base.py", line 723, in _do_update
    return filtered._update(values) > 0
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/query.py", line 600, in _update
    return query.get_compiler(self.db).execute_sql(CURSOR)
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 1005, in execute_sql
    cursor = super(SQLUpdateCompiler, self).execute_sql(result_type)
[...]

PR #3915, opened 2015-07-23
Drop alttrans support in convert/update

jleclanche:
It's causing more problems than it's solving.

Refs #3911


julen:
:+1: ship it.

PR #3917, opened 2015-07-24
Should fix 3707

odonnellryan:


julen:
Hey @odonnellryan, thanks for looking into this.

Can you elaborate how you think this would fix the issue? FWIW using `nav` elements already imply `display: block;` (Chrome's default). Unfortunately even with this patch I've been able to reproduce the original issue :disappointed:

PR #3918, opened 2015-07-27
Initialize from templates newly added TPs

### LINKED ISSUE #3877: Adding a TP does not initialise its stores from Templates
If you have a template language and add a new TP then the project is not initialised i.e. it does not create stores based on the Template language.  This likely relates to our loss of ``update_against_templates``.  Pootle is currently relying on initialisation of a language happening via ``update_stores`` which will pick up languages from the filesystem. But this then doesn't allow admins to quickly add a language to a Pootle project without having access to initialise the project from the filesystem.

unho:
Fixes #3877.

I am still unable to get stats automatically calculated after TP creation from UI, but @ta2-1 said that it worked for him. It would be great if someone else can quickly test this. Instructions follow:

- Get test project files from https://drive.google.com/open?id=0B-8KVyCPnkZRbW1qMzVObk0xWW8
- Uncompress it on `translations` directory
- Open the browser and go to Admin -> Projects
- Create a new `test` project
- Run `update_stores --project=test` to get the files into the DB
- Go to http://127.0.0.1:8000/projects/test/admin/languages/ and add a new language (for example `gl`)
- Go to http://127.0.0.1:8000/gl/test/ and check that the TP is there and stats are calculated
- Check on `translations` directory that the files for the new TP are present there

### DIFF
--- a/pootle/apps/pootle_translationproject/models.py
+++ b/pootle/apps/pootle_translationproject/models.py
@@ -245,6 +245,19 @@ class TranslationProject(models.Model, CachedTreeItem):
     def save(self, *args, **kwargs):
         created = self.id is None
 
+        if created:
+            from pootle_app.project_tree import translation_project_should_exist
+
+            template_tp = self.project.get_template_translationproject()
+            initialize_from_templates = False
+
+            if (not self.is_template_project and
+                template_tp is not None and
+                not translation_project_should_exist(self.language,
+                                                     self.project)):
+
+                initialize_from_templates = True
+
         self.directory = self.language.directory \
                                       .get_or_make_subdir(self.project.code)
         self.pootle_path = self.directory.pootle_path
@@ -258,7 +271,21 @@ class TranslationProject(models.Model, CachedTreeItem):
         super(TranslationProject, self).save(*args, **kwargs)
 
         if created:
+            if initialize_from_templates:
+                # We are adding a new TP and there are no files to import from
+                # disk, so initialize the TP files using the templates TP ones.
+                from pootle_app.project_tree import init_store_from_template
+
+                for template_store in template_tp.stores.iterator():
+                    init_store_from_template(self, template_store)
+
             self.scan_files()
+
+            if initialize_from_templates:
+                # Trigger stats refresh for TP added from UI.
+                # FIXME: This won't be necessary once #3547 is fixed.
+                for store in self.stores.live().iterator():
+                    store.update()
 
     def delete(self, *args, **kwargs):
         directory = self.directory
--- a/pootle/apps/pootle_app/project_tree.py
+++ b/pootle/apps/pootle_app/project_tree.py
@@ -313,6 +313,26 @@ def translation_project_should_exist(language, project):
     return False
 
 
+def init_store_from_template(translation_project, template_store):
+    """Initialize a new file for `translation_project` using `template_store`.
+    """
+    if translation_project.file_style == 'gnu':
+        target_pootle_path, target_path = get_translated_name_gnu(translation_project,
+                                                                  template_store)
+    else:
+        target_pootle_path, target_path = get_translated_name(translation_project,
+                                                              template_store)
+
+    # Create the missing directories for the new TP.
+    target_dir = os.path.dirname(target_path)
+    if not os.path.exists(target_dir):
+        os.makedirs(target_dir)
+
+    output_file = template_store.file.store
+    output_file.settargetlanguage(translation_project.language.code)
+    output_file.savefile(target_path)
+
+
 def get_translated_name_gnu(translation_project, store):
     """Given a template :param:`store` and a :param:`translation_project` return
     target filename.


unho:
This basically picks the code for `update_against_templates` from 2.6 branch and drops all the stuff is no longer necessary for 2.7 codebase (for example monolingual, hooks, signals, version control, support for merging) and does some further adjustments.

On 2.6 basically `update_against_templates` calls https://github.com/translate/pootle/blob/stable/2.6.0/pootle/apps/pootle_translationproject/models.py#L400-L488 which relies on https://github.com/translate/pootle/blob/stable/2.6.0/pootle/apps/pootle_app/project_tree.py#L347-L409.

Now we don't have `update_against_templates`, but we want to get TP initialized automatically in some cases, so instead of getting back `TP.update_against_templates()` we expand `TP.save()` to deal this specific scenario.

The ported logic for converting a template file is now kept on `pootle_app.project_tree.init_store_from_template`, which is some kind of merge of:

- `TP.update_against_templates()`
- `pootle_app.project_tree.convert_template`
- `pootle_app.project_tree.ensure_target_dir_exists`

from 2.6 (after dropping all what is no longer necessary). 

`TP.save()` now has some code picked from 2.6 `TP.update_against_templates()` and also some new code to trigger stats refresh. This new code is basically a try to trigger the same stats refresh that 2.7 `update_stores` management command performs: https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/management/commands/update_stores.py

julen:
I have a simple question: if the initial assumption is there are templates, why is any sort of *conversion* needed?

I can think this was needed in the past because of monolingual, but that doesn't apply anymore. In bilingual formats you can just copy the template file as-is, and probably adjust some meta-data such as language codes.

unho:
@julen You mean calling `pot2po.convert_stores`?

julen:
`pot2po` yeah (you also mentioned it: *The ported logic for converting a template*)

dwaynebailey:
@julen in some projects there are name changes .pot -> .po while others .xliff and .ts wouldn't have any name changes.  But apart from that that's about all that is needed.  In PO there is no meta data prep until the file actually first gets edited.  I'm not sure about others.

unho:
With *The ported logic for converting a template* I meant the ported code from `update_against_templates`.

`pot2po.convert_stores` is called because it seems to run some specific code for PO https://github.com/translate/translate/blob/master/translate/convert/pot2po.py#L188-L257 Actually I didn't check in detail what `pot2po` was for since it was code outside of Pootle. After a deeper review it seems that it is only necessary to adjust the headers for PO, since all the other stuff is either skipped or unnecessary. But I am still reluctant to drop it since we know that it handles all the scenarios.

unho:
Updated to get rid of `pot2po`. Also further simplified a bit the code.

pootle/apps/pootle_translationproject/models.py
```diff
@@ -258,8 +271,24 @@ def save(self, *args, **kwargs):
...
+                    init_store_from_template(self, template_store)
+
             self.scan_files()
 
+            if initialize_from_templates:
+                # Trigger stats refresh for TP added from UI.
+                for store in self.stores.live().iterator():
+                    try:
+                        store.update()
+                    except Exception:
```
unho: > What kind of exception can you get here?

No idea, but whatever comes from https://github.com/translate/pootle/blob/878a1418/pootle/apps/pootle_store/models.py#L1547 for example.

jleclanche:
gtm

pootle/apps/pootle_translationproject/models.py
```diff
@@ -258,8 +271,21 @@ def save(self, *args, **kwargs):
...
+                # We are adding a new TP and there are no files to import from
+                # disk, so initialize the TP files using the templates TP ones.
+                from pootle_app.project_tree import init_store_from_template
+
+                for template_store in template_tp.stores.iterator():
+                    init_store_from_template(self, template_store)
+
             self.scan_files()
 
+            if initialize_from_templates:
```
julen: Let's add a comment here indicating that this won't be necessary once the work for #3547 has been completed (more specifically [this will supersede it](https://github.com/translate/pootle/pull/3583/files#diff-f684a8a253f640ccaace73ac502ad438R68)).

julen:
Thank you @unho, works fine and given the constraints I believe it's good to be merged.

There's one caveat worth mentioning though: if someone creates, then deletes a TP, the FS is left untouched. Therefore, whenever someone tries to re-create a TP which was created-then-deleted via the web UI, nothing will be done.

On a related note, at the current state of things I believe it makes sense to disallow adding new languages via the web UI if no templates are available for a TP, otherwise users end up with unexpected/broken behavior .

unho:
> Thank you @unho, works fine and given the constraints I believe it's good to be merged.

Merging it then.

> There's one caveat worth mentioning though: if someone creates, then deletes a TP, the FS is left untouched. Therefore, whenever someone tries to re-create a TP which was created-then-deleted via the web UI, nothing will be done.

Yes, that is what happens. I am not sure if deleting the files in the file system will affect in any way resurrecting the stores. Perhaps @ta2-1 can provide some insight.

> On a related note, at the current state of things I believe it makes sense to disallow adding new languages via the web UI if no templates are available for a TP, otherwise users end up with unexpected/broken behavior.

I agree. Created #3942 for this.

Issue #3693, opened 2015-04-28
Stats: increase efficiency of `refresh_stats` command (with multiple workers)


ta2-1:
We have two `refresh_stats` commands since #3615 has been closed. 
- The old `refresh_stats` command that requires single rqworker. It calculates stats for all `Store` objects in `O(1)` DB-queries. Then, starting on every `Project`, it updates stats recursively for the entire tree structure.
- The new `refresh_stats_rq` command uses the same approach from #3615 and creates **one RQ job for every `Store` object**. When a RQ job for a `Store` object finishes it creates a new RQ job for its parent. 
As we know from [previous testing](https://github.com/translate/pootle/pull/3621#issuecomment-79028033), the old command (with a single rqworker) is twice as fast compared tothe new one (with 6 rqworkers).

The main achievement of #3615 is the absence of race conditions: it's impossible to run two RQ jobs for the same object at the same point in time. But the new command's efficiency needs to be improved.

I see two ways how we can increase the efficiency of the new `refresh_stats` command:
- Try to **avoid long queries** for getting `last_action`, `last_updated` and `last_mtime` values for every `Store` object. (For instance, we can save the last submission id as a field of the `Store` object).
- Try to **create jobs for objects higher up in the tree hierarchy** (`TranslationProject` or `Directory`, instead of `Store`), as already suggested by @julen and @jleclanche. This means that in order to add this *bigger* RQ job we need to make sure that no jobs are already running, queued or deferred for any TP child object. At the same time we need to ensure no jobs are created for any children objects of a TP while the *"bigger"* TP job is not finished.

My personal feeling is that blocking mechanism will make things more complicated here. But I'll be glad if I'm wrong.

unho:
Command is gone, thus the closing.

PR #3930, opened 2015-07-29
Removing pot files from stats

### LINKED ISSUE #3245: POT files counted into Need Translation
In statistics, string from POT files appears as "Need Translation" and therefore breaks stats about translation completeness

Pootle 2.6.0-rc1 (Django 1.12.0-rc1, Translate Toolkit 1.6.5)

phlax:
When calculating wordcount stats don't include .pot template files

fixes #3245

as requested by @jleclanche - the process i followed was:

- looked at main project models to get better understanding of db architecture
- used grep to find out where stats where coming from in ui (which was mostly pointless)
- on realising that they are cached by refresh_stats used pdb to step through until i found what looked like a good place to add a filter
- checked that stats are correctly updated when further translations are completed

### DIFF
--- a/pootle/apps/pootle_app/management/commands/refresh_stats.py
+++ b/pootle/apps/pootle_app/management/commands/refresh_stats.py
@@ -275,6 +275,9 @@ class Command(PootleCommand):
         if unit_filter:
             units = units.filter(**unit_filter)
 
+        # exclude template files from wordcount
+        units = units.exclude(store__file__endswith="pot")
+
         res = units.values('store', 'state') \
                    .annotate(wordcount=Sum('source_wordcount')) \
                    .order_by('store', 'state')
--- a/pootle/apps/pootle_app/management/commands/refresh_stats.py
+++ b/pootle/apps/pootle_app/management/commands/refresh_stats.py
@@ -275,9 +275,9 @@ class Command(PootleCommand):
         if unit_filter:
             units = units.filter(**unit_filter)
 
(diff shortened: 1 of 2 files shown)


jleclanche:
I don't think filtering on just endswith "pot" is the appropriate way to exclude template files. @julen @unho do we have a way of checking whether a file is a template?

phlax:
yep that was a concern, but .po files and .pot files share the same file_class, so i figured it was the only way

phlax:
...and afaict there is no content/mimetype stored on the filefield, or file classes

julen:
A couple ways I can see ([1](https://github.com/translate/pootle/blob/d6c4809fe881dac706908c22d304a24c24c93052/pootle/apps/pootle_project/models.py#L429-L437), [2](https://github.com/translate/pootle/blob/d6c4809fe881dac706908c22d304a24c24c93052/pootle/apps/pootle_app/project_tree.py#L48-L64)) — not definitely filtering by a hard-coded *pot*.

ta2-1:
Old `refresh_stats` command is going away (https://github.com/translate/pootle/issues/3693).
It's worth to fix https://github.com/translate/pootle/issues/3245 for `refresh_stats_rq` command.

unho:
This was moved to #3934. Not sure about the reason.

phlax:
i put it on a new branch/pr as i didn't know to force push my rebased repo

PR #3934, opened 2015-07-30
Restrict wordcount stats to project files

### LINKED ISSUE #3245: POT files counted into Need Translation

phlax:
when calculating wordcount stats only use files with extentions matching project.localfiletype

this removes .pot or other template files from the wordcount

fixes #3245

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -2162,9 +2162,14 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
             'translated': 0,
             'fuzzy': 0
         }
+
+        # only count files with an ext matching the project ext
+        proj_ext = "." + self.translation_project.project.localfiletype
+
         # XXX: `order_by()` here is important as it removes the default
         # ordering for units. See #3897 for reference.
-        res = self.units.order_by().values('state') \
+        res = self.units.filter(store__name__endswith=proj_ext) \
+                        .order_by().values('state') \
                         .annotate(wordcount=models.Sum('source_wordcount'))
         for item in res:
             ret['total'] += item['wordcount']


unho:
Just raising again comment https://github.com/translate/pootle/pull/3930#issuecomment-125950390

pootle/apps/pootle_store/models.py
```diff
@@ -2138,9 +2138,14 @@ def _get_wordcount_stats(self):
...
             'fuzzy': 0
         }
+
+        # only count files with an ext matching the project ext
+        proj_ext = self.translation_project.project.localfiletype
+
         # XXX: `order_by()` here is important as it removes the default
         # ordering for units. See #3897 for reference.
-        res = self.units.order_by().values('state') \
+        res = self.units.filter(store__file__endswith=proj_ext) \
```
unho: Most likely using `store__name__endswith=proj_ext` might be slightly faster since name is shorter than file, e.g. `tutorial/templates/tutorial.pot` vs just `tutorial.pot`.

phlax:
cool - i changed it to use store_name

pootle/apps/pootle_app/management/commands/refresh_stats.py
```diff
@@ -285,6 +288,12 @@ def _set_wordcount_stats(self, unit_filter):
         key = None
 
         for item in res.iterator():
+            # only include files with ext matching project ext
+            file_ext = os.path.splitext(item['store__file'])[1][1:]
+            proj_ext = item['store__translation_project__project__localfiletype']
+            if not file_ext == proj_ext:
```
unho: `if file_ext != proj_ext:` is simpler.

phlax:
k, changed

pootle/apps/pootle_app/management/commands/refresh_stats.py
```diff
@@ -275,7 +275,10 @@ def _set_wordcount_stats(self, unit_filter):
         if unit_filter:
             units = units.filter(**unit_filter)
 
-        res = units.values('store', 'state') \
+        local_filetype = 'store__translation_project__project__localfiletype'
+        res = units.values('store', 'state',
+                           local_filetype,
+                           'store__name') \
```
unho: All this can go on the same line:

```python
        res = units.values('store', 'state', local_filetype, 'store__name') \
```

phlax:
k, done

pootle/apps/pootle_app/management/commands/refresh_stats.py
```diff
@@ -285,6 +286,12 @@ def _set_wordcount_stats(self, unit_filter):
         key = None
 
         for item in res.iterator():
+            # only include files with ext matching project ext
+            file_ext = os.path.splitext(item['store__name'])[1][1:]
+            proj_ext = item['store__translation_project__project__localfiletype']
+            if file_ext != proj_ext:
+                continue
```
unho: I would prefer this filtering to be done in the queryset itself, but when I tried to use an `F` expression it didn't work. Just FYI I was trying the following in the above queryset:

```python
        res = units.filter(store__name__endswith=F('store__translation_project__project__localfiletype')) \
                   .values('store', 'state') \
                   .annotate(wordcount=Sum('source_wordcount')) \
                   .order_by('store', 'state')
```

But I was obviously doing something wrong there.

phlax:
I'm not sure but i don't think you can use F() with endswith, just with numerical operators or equivalence

phlax:
although it seems to have been fixed about 8 months ago:

https://code.djangoproject.com/ticket/16731

unho:
We are working with Django 1.7, and that change only arrives on Django 1.8, so we should drop the idea of using an `F` expression here.

unho:
@phlax As pointed out in comment https://github.com/translate/pootle/pull/3930#issuecomment-125950390 it is worth addressing the issue on `refresh_stats_rq` since `refresh_stats` is going to be dropped.

phlax:
@unho i believed that i had by adding the filter to Store._get_wordcount_stats - does that not fix it for refresh_stats_rq?

unho:
I think you are right.

lgtm on my side. I would like to hear from others.

unho:
@julen @ta2-1 Any input from you?

julen:
Looks good, however I'd take the `refresh_stats` code out as it's going away.

Issue #3746, opened 2015-05-18
Uploading file that changes units does not seem to invalidate cache so subsequent downloads are broken


dwaynebailey:
1. Download a file
2. Make changes to the file
3. Upload the file
4. Inspect and units are correctly changed
5. Download file again - we have the old file without any of the updates that we got via upload

jleclanche:
I've been looking at this all evening, I cannot figure out why the store isn't getting updated. It's a problem in the Pootle/TTK binding. @julen @unho feedback appreciated.

dwaynebailey:
My gut feel is that its nothing to do with any 'binding'.  I'm simply getting old data that does not include recently submitted or recently uploaded values.  So my gut feeling is that somehow when we export we're not requesting a brand new serialization but are getting a cached version.

julen:
In `Store` instances, `file` references the on-disk file, and its `store` property holds a reference to an on-disk parsed file.

In the new export code, `serialize()` always returns [the string representation of the on-disk file](https://github.com/translate/pootle/blob/ac782c6ed3eda1c9fba1acbbe1b87c562c461b6f/pootle/apps/pootle_store/models.py#L1801), so the behavior described in the OP aligns with what I read in the code.

In other words, `serialize()` does zero work. I believe `serialize()` has good intentions, but it ends up trying to reinvent the wheel, as [the serialization step is already implemented in `convert()`](https://github.com/translate/pootle/blob/ac782c6ed3eda1c9fba1acbbe1b87c562c461b6f/pootle/apps/pootle_store/models.py#L1939-L1950).

I highly advise to write some test cases besides fixing the issue.

jleclanche:
That convert code is just nasty... I'm not even sure how to clean this up properly.

jordis:
Note that this issue exists even if the strings are changed/translated inside the Pootle IU, that is, string changes do not need to happen offline and no file download/upload is involved.
See: https://groups.google.com/d/topic/mozilla.dev.l10n/QUoBfpfr4xU/discussion

PR #3935, opened 2015-07-30
Force a sync before serializing a non-cached file

jleclanche:


pootle/apps/pootle_store/models.py
```diff
@@ -1902,6 +1902,7 @@ def serialize(self):
 
         ret = cache.get(path, version=rev)
         if not ret:
+            self.sync()
```
unho: Is it not better to sync after updating file header?

unho:
Why is the sync necessary?

unho:
Does this address any specific issue? I wonder if #3746 is related to this.

dwaynebailey:
This is completely the wrong way to do this.  Its essentially doing a ``sync_stores`` which will mess with the actual operation of ``sync_stores``.

jleclanche:
Updated to a different approach. @dwaynebailey @julen 1 gtm to land please.

unho:
lgtm

julen:
:+1: fine to merge, thank you @jleclanche.

jleclanche:
Landed

PR #3939, opened 2015-07-31
admin/admin cleanup

dwaynebailey:
This PR:

1. Adds a check for admin/admin and warns if the default password is still set (this should catch bad setup on old installs)
2. Removes the creation of admin from initdb

I did look at docs cleanup but it moved so quickly from a simple lets check if we need ``createadminuser`` instructions to a cleanup of the upgrade page.  So lets leave that out of this PR please.

I don't want to automate the default admin user creation but rather make that part of the setup of the server.  Installation should be fines, its documented.  Upgrade should also be fine as you'll have an admin user, you might however not have one with an email address but we can deal with that in the upgrade docs.

### DIFF
--- a/docs/server/upgrading.rst
+++ b/docs/server/upgrading.rst
@@ -179,6 +179,15 @@ To perform the upgrade follow the next steps:
 Post-upgrade adjustments
 ------------------------
 
+* Check that you can login as an admin user.  If your existing adin user has no
+  email address then you will not be able to login.  In order the fix this use
+  :djadmin:`createsuperuser` to create a new superuser, or to create a
+  temporary superuser to fix your existing one.
+
+  .. code-block:: bash
+
+     (env) $ pootle createsuperuser
+
 After a succesful upgrade you can now consider:
 
 * Implementing some :doc:`optimizations <optimization>` to your setup.
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -235,3 +235,28 @@ def check_settings(app_configs=None, **kwargs):
                 ))
 
     return errors
+
+
+@checks.register()
+def check_users(app_configs=None, **kwargs):
+    from django.contrib.auth import get_user_model
+    from django.db import ProgrammingError
+    from django.db.utils import OperationalError
+
+    errors = []
+
+    User = get_user_model()
+    try:
+        admin_user = User.objects.get(username='admin')
+    except (User.DoesNotExist, OperationalError, ProgrammingError):
+        pass
+    else:
+        if admin_user.check_password('admin'):
+            errors.append(checks.Warning(
+                _("The default 'admin' user still has a password set to "
+                  "'admin'."),
+                hint=_("Remove the 'admin' user or change its password."),
+                id="pootle.W016",
+            ))
+
+    return errors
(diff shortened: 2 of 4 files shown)


pootle/checks.py
```diff
@@ -235,3 +235,27 @@ def check_settings(app_configs=None, **kwargs):
...
+def check_users(app_configs=None, **kwargs):
+    from django.contrib.auth import get_user_model
+    from django.contrib.auth import authenticate
+
+    errors = []
+
+    try:
+        admin_user = get_user_model().objects.get(username='admin')
+        if admin_user is not None:
+            admin_authenticated = authenticate(username='admin', password='admin')
```
julen: No need to `authenticate()`, just do `if admin_user.check_password('admin'):`

pootle/checks.py
```diff
@@ -235,3 +235,27 @@ def check_settings(app_configs=None, **kwargs):
...
+            if admin_authenticated is not None:
+                errors.append(checks.Warning(
+                    _("The default 'admin' user still has a password set to "
+                      "'admin'."),
+                    hint=_("Remove the 'admin' user or change its password."),
+                    id="pootle.W016",
+                ))
+    except:
```
julen: Better:
```python
User = get_user_model()
try:
    admin_user = User.objects.get(username='admin')
except User.DoesNotExist:
    pass
else:
    if admin_user.check_password('admin'):
        errors.append(...)
```

pootle/checks.py
dwaynebailey: Nice

dwaynebailey:
@julen @jleclanche addressed your comments.

I've landed docs changes now also.  And added the warning post initdb to assist seasoned Pootle users so that they don't get caught out and we're not answering questions on IRC.

jleclanche:
gtm

docs/server/upgrading.rst
```diff
@@ -179,6 +179,15 @@ To perform the upgrade follow the next steps:
 Post-upgrade adjustments
 ------------------------
 
+* Check that you can login as an admin user.  If you existing adin user has no
```
julen: *your existing admin*

docs/server/upgrading.rst
```diff
@@ -179,6 +179,15 @@ To perform the upgrade follow the next steps:
...
+  email address then you will not be able to login.  In order the fix this use
+  :djadmin:`createsuperuser` to create a new superusers, or to create a
```
julen: *superuser* singular

julen:
Tests broke I'm afraid; good to go provided that's fixed.

docs/server/upgrading.rst
```diff
@@ -179,6 +179,15 @@ To perform the upgrade follow the next steps:
...
+  temporary superuser to fix your existing one.
+
+  .. code-block:: bash
+
+     (env) $ pootle createsuperuser
```
unho: This is not post-upgrade adjustment. It is a step of the upgrade, so this must be before "Post-upgrade adjustments" section in the upgrade instructions.

dwaynebailey: I didn't want to fiddle those docs anymore.  I have some upgrade cleanups and will deal with that with those.

docs/server/upgrading.rst
```diff
@@ -179,6 +179,15 @@ To perform the upgrade follow the next steps:
 Post-upgrade adjustments
 ------------------------
 
+* Check that you can login as an admin user.  If your existing adin user has no
```
unho: Typo: adin

PR #3954, opened 2015-08-10
Redis version number

saily:
Latest stable release just says nothing about required version. I ran into this issue while installing new Pootle on my CentOS 6.6 server where latest redis is just 2.4. Please use a version number instead.


saily:
Used this version information: https://github.com/translate/pootle/blob/2.7.0/pootle/checks.py#L29

julen:
Thank you for clearing this up. Mind adding the same in *docs/releases/dev.rst*?

julen:
Thank you @saily, I've pushed your change in 985b183 and a follow-up in ﻿7685859.

Issue #3867, opened 2015-06-26
Manage command to get a list of contributors


dwaynebailey:
We need to be able to credit translators for version control commits.  For now we're not interested in more than knowing the information and leaving it to tools to use that information in commit messages.

The command should do the following:

1. Allow us to get a count ``--count`` so that we'd know if we can attribute this commit to one person or a group of people.
2. Specify a project and language - so that we can limit the information to one "domain" and commit
3. Give a range.  Which is likely either a time period of a revision count.  So ``--from-revision`` as we'll likely know what was the last revision we looked at before this commit.

Ideally data output is just text that we can use in our outside automation:

```
$ ./manage.py attribution --count --project=firefox --language=af
1
```

```
$ ./manage.py attribution --project=firefox --language=af
Dwayne Bailey <dwayne@translate.org.za>
```

Multiple contributors in a period would do the following
```
$ ./manage.py attribution --project=firefox --language=af
Dwayne Bailey <dwayne@translate.org.za>
A Nother <another@example.com>
```

We leave it to the outside scripts to use this information to make attribution,

codeadict:
I can take this one, please @dwaynebailey let me know

jleclanche:
@codeadict Go ahead.

PR #3958, opened 2015-08-11
Add manage.py contributors

jleclanche:
#3867

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -287,6 +287,18 @@ Accepts the :option:`--modified-since` parameter to list only those projects
 modified since the revision given by :djadmin:`revision`.
 
 
+.. django-admin:: contributors
+
+contributors
+^^^^^^^^^^^^
+
+Lists the contributors to a language, project or overall and the amount
+of contributions they have.
+
+Accepts the :option:`--from-revision` parameter to only take into account
+contributions newer than the revision given by :djadmin:`revision`.
+
+
 .. django-admin:: revision
 
 revision
--- a/pootle/apps/pootle_app/management/commands/contributors.py
+++ b/pootle/apps/pootle_app/management/commands/contributors.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+os.environ["DJANGO_SETTINGS_MODULE"] = "pootle.settings"
+
+from collections import Counter
+from optparse import make_option
+from django.contrib.auth import get_user_model
+from pootle_app.management.commands import PootleCommand
+from pootle_store.models import Unit
+
+
+User = get_user_model()
+
+
+class Command(PootleCommand):
+    option_list = PootleCommand.option_list + (
+        make_option("--from-revision", type=int, default=0, dest="revision",
+                    help="Only count contributions newer than this revision"),
+    )
+
+    help = "Print a list of contributors."
+
+    def handle_all(self, **options):
+        system_user = User.objects.get_system_user()
+        units = Unit.objects.exclude(submitted_by=system_user).exclude(submitted_by=None)
+
+        if options["revision"]:
+            units = units.filter(revision__gte=options["revision"])
+
+        if self.projects:
+            units = units.filter(store__translation_project__project__code__in=self.projects)
+
+        if self.languages:
+            units = units.filter(store__translation_project__language__code__in=self.languages)
+
+        contribs = Counter()
+        for v in units.values("submitted_by"):
+            contribs.update((v["submitted_by"], ))
+
+        self.list_contributions(contribs)
+
+    def list_contributions(self, contribs):
+        for id, count in contribs.items():
+            user = User.objects.get(id=id)
+            name = user.display_name
+            if user.email:
+                name += " <%s>" % (user.email)
+            self.stdout.write("%s (%i contributions)" % (name, count))


unho:
Please squash both commits together.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,53 @@
...
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from collections import Counter
+from django.contrib.auth import get_user_model
+from pootle_app.management.commands import PootleCommand
+from pootle_store.models import Unit
```
unho: Please use the style guidelines for imports here:

```python
import os
from collections import Counter

from django.contrib.auth import get_user_model

from pootle_app.management.commands import PootleCommand
from pootle_store.models import Unit
```

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+User = get_user_model()
+
+
+class Command(PootleCommand):
+    help = "Update database stores from files."
```
unho: This help message doesn't match the command purpose.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,53 @@
...
+        for v in units:
+            contribs.update((v["submitted_by"], ))
+
+        self.list_contributions(contribs)
+
+    def list_contributions(self, contribs):
+        for id, count in contribs.items():
+            user = User.objects.get(id=id)
+            if user.is_system():
+                continue
```
unho: Perhaps it is better to exclude `system` user in the `units` queryset above instead of having this `if` here.

jleclanche: That's harder. No sense doing it.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,62 @@
...
+class Command(PootleCommand):
+    option_list = PootleCommand.option_list + (
+        make_option("--from-revision", type=int, default=0, dest="revision",
+                    help="Only count contributions newer than this revision"),
+    )
+
+    help = "Print a list of contributors."
+
+    def handle_all(self, **options):
+        units = Unit.objects.values("submitted_by").exclude(submitted_by=None)
```
unho: I assume using `Unit` instead of `ScoreLog` is because we can get the list of contributors for currently active translations, and not get some history of contributors including those whose contributions have been dropped. Am I right?

jleclanche: Yes, it's cheap and more reliable.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,62 @@
...
+# AUTHORS file for copyright and authorship information.
+
+import os
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from collections import Counter
+from optparse import make_option
+from django.contrib.auth import get_user_model
+from pootle_app.management.commands import PootleCommand
+from pootle_store.models import Unit
```
unho: Please use the style guidelines for imports here:

```python
import os
from collections import Counter
from optparse import make_option

# This must be run before importing Django.

from django.contrib.auth import get_user_model

from pootle_app.management.commands import PootleCommand
from pootle_store.models import Unit
```

jleclanche: The style guidelines don't respect pep8 and add nothing of value here.

pootle/apps/pootle_app/management/commands/contributors.py
unho: I don't see how it is harder. You are already filtering `units` above, so excluding `system` should reduce the number of units that must be iterated:

```python
system = User.objects.get_system_user()
units = Unit.objects.exclude(submitted_by=None).exclude(submitted_by=system).values("submitted_by")
```

jleclanche: Fair enough, I'll change it.

pootle/apps/pootle_app/management/commands/contributors.py
unho: How are the guidelines not respecting pep8 here?

jleclanche: We already discussed it. To refresh your memory: Unnecessary newlines and special casing for no good reason of django and ttk.

unho:
Please squash both commits together.

docs/server/commands.rst
```diff
@@ -287,6 +287,15 @@ Accepts the :option:`--modified-since` parameter to list only those projects
 modified since the revision given by :djadmin:`revision`.
 
 
+.. django-admin:: contributors
+
+contributors
+^^^^^^^^^^^^
+
+Lists the contributors to a language, project or overall and the amount
+of contributions they have.
```
unho: Noticed that you added `--from-revision` which is not explained here.

Also more detailed explanation on what result you will get when passing this or other options will most likely be appreciated by people reading the docs.

pootle/apps/pootle_app/management/commands/contributors.py
dwaynebailey: @jleclanche we have a style guide, followup it.  THe styleguide is very clear that we're extending PEP.  This is just a waste of time for everyone reviewing or commenting in that they have to read this again and again from you.  Follow the project style guide or the PR will hang.  You yourself have written countless times about wanting people to repsect your own projects styleguide, how about taking your own advice here.  You have an option of PR'ing a change to the style guide, but you seem to not want to choose that option but instead drag this up whenever anyone has an issue about your import block.  Which doesn't even follow PEP8.  But can you please stop wasting everyones time on this issue, this is so seriously demotivating for anyone having to review your work.

unho:
Tested locally and works ok. gtm after the docs have been completed.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -0,0 +1,59 @@
...
+        contribs = Counter()
+        for v in units.values("submitted_by"):
+            contribs.update((v["submitted_by"], ))
+
+        self.list_contributions(contribs)
+
+    def list_contributions(self, contribs):
+        for id, count in contribs.items():
+            user = User.objects.get(id=id)
+            if user.full_name:
```
julen: No need for `if/else`, just `name = user.display_name`.

julen:
Just tested this and the output is completely random for the user. Without going fancy with sorting options, I'd minimally expect the output to be sorted either alphabetically or by the number of contributions.

Also, the styling concerns haven't been addressed.

unho:
A suggestion, the following returns a list of `user_id`-`contribution_number` pairs, ordered by descending number of contributions, and also allows to drop the for loop that uses `Counter`:

```python
contribs = units.values_list('submitted_by') \
                .annotate(contributions=Count('submitted_by')) \
                .order_by('-contributions')
```

jleclanche:
I'm going to sort alphabetically, as sorting by contributions still presents the same issue in case there are multiple users with the same amount of contribs (which will happen a lot)

jleclanche:
b4c7a07

julen:
Please do fix the styling issues.

Issue #3812, opened 2015-06-10
Open sidebar doesn't push footer down


unho:
Seen in http://linode2.locamotion.org/af/firefox/

Screenshot below in case server disappears:

![screenshot](https://cloud.githubusercontent.com/assets/939880/8085567/216054e8-0f91-11e5-8bae-d219508e047e.png)

julen:
I cannot reproduce this neither locally nor in the link you provided nor in [our production server](https://translate.evernote.com/zh_CN/web_evernote/) (Firefox, Chromium, Safari).

julen:
You probably had improperly built assets and this has been a transient issue? Would be great to confirm.

unho:
I can confirm that I refreshed the server several times to force retrieval of last version from server. I also asked @dwaynebailey to rebuild assets and the problem persists.

julen:
Last (and supported) version is 43.

unho:
Just updated to Chromium 43 and the problem persists after several force refreshes and wiping all cache.

dwaynebailey:
Testing on mlo migration this seems to happen as follows:

1. Not a problem when the list of items in the browser is less than the size of the browser screen.  In this case the footer appears correctly at the bottom of the sidebase
2. A problem when the list of items in the browser occupies more than one browser screen, i.e. you need to scroll down to see further files and directories.  In this case the footer will appear directly below the last item in the browser and cut across the sidebar.

julen:
FWIW I can only see this on mlo – do you guys have any customizations on the server? The code that fixes the sidebar height doesn't kick in.

dwaynebailey:
@julen @unho got it.

This only occurs in this instance:

1. We have a project announcements, and
2. We have a language announcement.

When we have that we get the stripe through the middle.

phlax:
im working on this...

phlax:
this appears to have been fixed in commit 6a0d4d6

however it does not calculate the sidebar height correctly - so the footer is pushed much further down than it should be...

PR #3963, opened 2015-08-11
Set sidebar width when measuring height

### LINKED ISSUE #3969: Sidebar gets wrong height when hidden on page load
to reproduce:

- put a lot of content in the sidebar
- load a page with the sidebar closed
- expand the sidebar and witness that the sidebar is ~2x the height of the contained elements

this is fixed by #3963 and is related to #3812

phlax:
When taking a measurement of the hidden sidebar, it needs to have the correct width

touch #3812

fixes #3969

### DIFF
--- a/pootle/static/js/helpers.js
+++ b/pootle/static/js/helpers.js
@@ -37,10 +37,17 @@ var helpers = {
         bodyPadding = parseInt($body.css('padding-bottom'), 10),
         contentAreaHeight = $('#wrapper').height() - $body.offset().top -
                             bodyPadding,
-        sidebarHeight = $('#sidebar #sidebar-content').height() +
-                        $('#footer').height() + bodyPadding,
-        newHeight = Math.max(contentAreaHeight, sidebarHeight);
+        sidebarHeight,
+        newHeight;
 
+    // Set sidebar width before measuring height of content
+    $('#sidebar').css('width', '30%');
+    sidebarHeight = $('#sidebar #sidebar-content').height() +
+                    $('#footer').height() + bodyPadding;
+    newHeight = Math.max(contentAreaHeight, sidebarHeight);
+
+    // Remove sidebar width setting - allow CSS to set width
+    $('#sidebar').css('width', '');
     if (bodyHeight < newHeight) {
       $body.css('height', newHeight);
     }


pootle/static/js/helpers.js
```diff
@@ -37,10 +37,13 @@ var helpers = {
         bodyPadding = parseInt($body.css('padding-bottom'), 10),
         contentAreaHeight = $('#wrapper').height() - $body.offset().top -
                             bodyPadding,
-        sidebarHeight = $('#sidebar #sidebar-content').height() +
-                        $('#footer').height() + bodyPadding,
-        newHeight = Math.max(contentAreaHeight, sidebarHeight);
+        sidebarHeight, newHeight;
 
+    $('#sidebar').css("width", "30%");
```
unho: Can you explain why do you set this width and not other?

unho:
Just tested and this seems not to be fixing #3812.

julen:
Could you elaborate on how this affects height and therefore why is it necessary? I see no correlation between both http://jsfiddle.net/27dbcxpk/1/

phlax:
If you load a page *with the sidebar closed* and *a lot of content in sidebar* the measurement taken for the height of the sidebar is wrong (tested iceweasel 39.0, chromium 44.0.2403.89). This is due to the #sidebar-content element being squashed widthways when hidden, and therefore having more height.

so to be clear this is *not* #3812 - its almost the opposite.

to reproduce:
- put a lot of content in the sidebar
- load a page with the sidebar closed
- expand the sidebar and witness that the sidebar is ~2x the height of the contained elements

When the sidebar is expanded it uses 30% of viewport. As #sidebar has "layout" setting the contained elements width to a % will use a %  of the parent element not the viewports - @unho this is why the width is temporarily set on #sidebar and not #sidebar-content

@julen your jsfiddle example does not have hidden elements with layout and contained elements - so im not sure its that relevant.

Re: #3812 - i havent been able to reproduce and have had difficulty getting access to linode and the referenced page. I am going to try to resolve this now

unho:
> so to be clear this is not #3812 - its almost the opposite.

Let me suggest you to create an issue for the bug this PR is trying to fix and reference only to that new issue instead of unrelated #3812. Thanks!

phlax:
As suggested i have put the problem that this addresses into a separate ticket - #3969

julen:
@phlax this is basically the same code as the sidebar http://jsfiddle.net/27dbcxpk/3/

phlax:
@julen im struggling to reproduce as a jsfiddle.

Could you please try to reproduce bug #3969

pootle/static/js/helpers.js
julen: It's [the width of the sidebar](https://github.com/translate/pootle/blob/a328894179ff7c7a4ae6f88bf15087f979590865/pootle/static/css/style.css#L531) when it's open.

Issue #3876, opened 2015-07-02
Hardcoding path to git is fragile (and breaks e.g. on Windows)


iafan:
Because of https://github.com/translate/pootle/commit/7c9a5d7d640124cc85fa1e09e296b8823345e1fe I'm getting a traceback on Windows ("File not found" error). But even putting Windows aside, we'd better avoid hardcoding paths. 'git' should be in the path, otherwise a server can be considered underconfigured.

Also, I guess Pootle should not be crashing when git is not available.

dwaynebailey:
@iafan if you have a solution that works around the issue on servers then go for it, I agree with you but simply couldn't find a solution that actually worked on a server.

unho:
Note PR #3965 proposes a fix.

unho:
@phlax Check previous comment for a proposed fix.

phlax:
if we are going to use git in standard pootle (not such a good idea imho) then we should depend on gitpython and use that, and not worry about paths to binaries etc

PR #3965, opened 2015-08-11
draft 3876

jleclanche:
Fix for #3876. Also lets us set `POOTLE_GIT_PATH` environment variable to let us override the path to the git binary.

### DIFF
--- a/pootle/core/utils/version.py
+++ b/pootle/core/utils/version.py
@@ -183,6 +183,29 @@ def _shell_command(command):
     return command_subprocess.communicate()[0]
 
 
+def _is_executable_file(path):
+    return os.path.isfile(path) and os.access(path, os.X_OK)
+
+
+def _git_binary():
+    """
+    Tries to find the path to the `git` program.
+    Can be overridden by setting POOTLE_GIT_PATH environment variable.
+    See #3876.
+    """
+    setting = os.environ.get("POOTLE_GIT_PATH")
+    if setting:
+        return setting
+
+    for path in os.environ["PATH"].split(os.pathsep):
+        path = path.strip('"')
+        trypath = os.path.join(path, "git")
+        if _is_executable_file(trypath):
+            return trypath
+
+    return "/usr/bin/git"
+
+
 @lru_cache()
 def get_git_changeset():
     """Returns a numeric identifier of the latest git changeset.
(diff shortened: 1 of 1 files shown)


jleclanche:
@iafan does this resolve your issue?

pootle/core/utils/version.py
```diff
@@ -183,6 +183,29 @@ def _shell_command(command):
...
+    return os.path.isfile(path) and os.access(path, os.X_OK)
+
+
+def _git_binary():
+    """
+    Tries to find the path to the `git` program.
+    Can be overridden by setting POOTLE_GIT_PATH environment variable.
+    See #3876.
+    """
+    setting = os.environ.get("POOTLE_GIT_PATH")
```
unho: Is this really necessary to add a new custom `POOTLE_GIT_PATH` envvar? IMHO it is good enough to use the logic below.

unho:
Please use a better commit message, one that includes a reference to the issue it tries to fix.

pootle/core/utils/version.py
jleclanche: It's free, honestly.

iafan:
> @iafan does this resolve your issue?

I still get this traceback:

```
  File "manage.py", line 15, in <module>
    from pootle import syspath_override
  File "C:\Storage\Work\GitHub\pootle\pootle\__init__.py", line 14, in <module>
    __version__ = get_version(VERSION)
  File "C:\Storage\Work\GitHub\pootle\pootle\core\utils\version.py", line 64, in get_version
    git_changeset = get_git_changeset()
  File "C:\Program Files (x86)\Python\lib\site-packages\django\utils\lru_cache.py", line 125, in wrapper
    result = user_function(*args, **kwds)
  File "C:\Storage\Work\GitHub\pootle\pootle\core\utils\version.py", line 221, in get_git_changeset
    [_git_binary(), 'log', '--pretty=format:%ct', '--quiet', '-1', 'HEAD']
  File "C:\Storage\Work\GitHub\pootle\pootle\core\utils\version.py", line 180, in _shell_command
    universal_newlines=True
  File "C:\Program Files (x86)\Python\lib\subprocess.py", line 709, in __init__
    errread, errwrite)
  File "C:\Program Files (x86)\Python\lib\subprocess.py", line 957, in _execute_child
    startupinfo)
WindowsError: [Error 2]
```

iafan:
@jleclanche, and the reason your code doesn't work on Windows is simply because on windows the executable file name is `git.exe`, not `git`.

I think we shouldn't be checking for the file existence for a dozen of times each time we try to get the git executable path. This is a more simple code that works:

```Python
def _git_binary():
    """
    Tries to find the path to the `git` program.
    Can be overridden by setting POOTLE_GIT_PATH environment variable.
    See #3876.
    """
    setting = os.environ.get("POOTLE_GIT_PATH")
    if setting:
        return setting

    trypath = "/usr/bin/git"
    if _is_executable_file(trypath):
        return trypath

    return "git"
```

PR #3967, opened 2015-08-12
Updated JavaScript styleguide

julen:
While standing on top of the Airbnb style guide, which is pretty detailed and it
already covers the semantics we had, it also adds specific mentions to ES2015
and JSX code.

PRing just to raise awareness of the change.


jleclanche:
gtm, the airbnb styleguide is excellent. (Not that I'm a fan of 2 spaces but... you know. :) )

unho:
gtm.

Issue #3908, opened 2015-07-20
Allow language managers to edit their language's special characters


gueroJeff:
Essentially, it used to be that project admins could add their own special characters to the virtual keyboards for their project. There was an option for it in the project description portion of the project landing page. It appears that the field for adding special characters via that workflow has been removed.

Use case: Julio Gomes, who has been mentor for many indigenous Latin American locales, is project admin for each of these locales. After they were created in Pootle, he would do additional research into the writing systems for each locale and add those special character keys on his own within the project's description editing dialogue. Julio's request is to have it restored.

dwaynebailey:
Some added info on this.  The ability to add a language description page also had the ability to edit the special characters.  Currently on ``master`` these changes can only be made at the global admin level.

I think we should look at exposing more of these config options on the Browse/Translate dropdown for language admin.

That would include:

1. Special characters
2. Language descriptions page (possibly)

gueroJeff:
I think that would be a great place for them.

unho:
Plan of action:

- Add some intermediate level of permissions between regular users and global admin, so some users can edit settings and announcements for specific TPs, Languages or Projects.
- If allowing such users to visit the admin site is deemed too dangerous or undesirable, then provide some new admin UI that just allows to edit a single TP, Language or Project.
- Consider displaying the links to access admin UIs in the actions dropdown:

![abbb](https://cloud.githubusercontent.com/assets/939880/9177109/9ee3564e-3f8f-11e5-9723-857e651ef378.png)

iafan:
> Add some intermediate level of permissions between regular users and global admin, so some users can edit settings and announcements for specific TPs, Languages or Projects

@unho, we already have project-level admins, which could be a good fit for the purpose.

unho:
@iafan How do you specify a user to be a project-level admin? I have been checking existing permissions and I didn't saw any that can be checked to tell that a user is a project manager (I do prefer this term over admin, which I only use for global-admin or superuser).

iafan:
@unho, try going to e.g. https://translate.stage.evernote.com/projects/website3_evernote/admin/permissions/ — there you can add a user with 'Can administrate a translation project' permission.

unho:
> @unho, try going to e.g. https://translate.stage.evernote.com/projects/website3_evernote/admin/permissions/ — there you can add a user with 'Can administrate a translation project' permission.

[...]

PR #3968, opened 2015-08-12
Display edit link for sidebar announcements

### LINKED ISSUE #3861: Display edit link for announcements on sidebar
It has been reported that the ability to easily edit the instructions has been lost. Old project, language and TP descriptions allowed to edit in place. This probably is not a good idea, but at least some edit link must be provided for users with enough permissions to edit the announcements.

jleclanche:
Fixes #3861

Not sure it's right to get the `page_type` this way but couldn't figure out a better way, suggestions welcome.

### DIFF
--- a/pootle/templates/staticpages/_body.html
+++ b/pootle/templates/staticpages/_body.html
@@ -9,7 +9,14 @@
   {% endblocktrans %}
   </div>
   {% endif %}
-  <h1 class="title">{{ page.title }}</h1>
+  <h1 class="title">
+    {{ page.title }}
+    {% if user.is_superuser %}
+    <span class="edit-link">
+      [<a href="{% url 'pootle-staticpages-edit' pk=page.id page_type='announcements' %}">edit</a>]
+    </span>
+    {% endif %}
+  </h1>
   {% with language_code=language.code %}
   {{ page.body|rewrite_language_links:language_code }}
   {% endwith %}
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -477,6 +477,11 @@ td.stats-name a:hover > i + span,
     list-style: inherit;
 }
 
+.staticpage .edit-link
+{
+    font-size: small;
+}
+
 
 /*
  * CHECKS DESCRIPTIONS


pootle/templates/staticpages/_body.html
```diff
@@ -9,7 +9,14 @@
   {% endblocktrans %}
   </div>
   {% endif %}
-  <h1 class="title">{{ page.title }}</h1>
+  <h1 class="title">
+    {{ page.title }}
+    {% if user.is_superuser %}
```
unho: We want to display the edit link not only to global admins, but also to TP/Language/Project managers as for #3908.

pootle/templates/staticpages/_body.html
```diff
@@ -9,7 +9,14 @@
...
+    <span class="edit-link">
```
unho: Tested this on my computer, but the edit link is not displayed with smaller font. I refreshed several times and dropped cache. Perhaps it is only me, so I would like someone else to test and confirm this is displayed as expected.

I also considered this to be not in the header line since some headers can be long, but I don't think it is a major issue. In other words, this is something I can live with.

julen: The CSS class definition is invalid.

jleclanche: Oh that explains it, I thought I had a caching issue.

pootle/templates/staticpages/_body.html
jleclanche: How would I check for that?

pootle/templates/staticpages/_body.html
unho: Yes, missed that. It should be `.staticpage .edit-link`.

pootle/templates/staticpages/_body.html
unho: The following allows to check if a specific user can administrate a TP:

```python
check_user_permission(user, 'administrate', tp.directory)
```

I cannot find any specific permissions for administrating a project or a language, so perhaps you should use the `change_project` and `change_language` permissions.

unho: Or perhaps use `view` permission, but that doesn't seem related to management. @julen Any idea?

jleclanche: Which types of users are allowed to edit this? Ideally I'd like to show the link to anyone who can edit it.

unho: By type of announcement:

- Project announcement: Can be edited by global admins and project managers
- Language announcement: Can be edited by global admins and language managers
- TP announcement: Can be edited by global admins and TP managers

As commented above we don't seem to have any permission to tell if a user is a project manager or language manager.

unho: Just noticed that perhaps the admin UI for editing announcements must be also adjusted so the displayed announcements are restricted to those having only the required permissions.

unho: After discussion on #3908, and given that the announcements are only displayed on TPs, it is ok to display the edit link if the user is a superuser or has `administrate` permission in the current TP.

unho: Noticed that this is a template that is included both in the template that displays staticpages (including announcements) and in the browser template to render announcements in the sidebar. In the former we probably can't check for TP `administrate` permission since it is completely detached from any particular TP.

pootle/templates/staticpages/_body.html
unho: For the record this particular issue is now fixed.

pootle/templates/staticpages/_body.html
```diff
@@ -9,7 +9,14 @@
...
+      [<a href="{% url 'pootle-staticpages-edit' pk=page.id page_type='announcements' %}">edit</a>]
```
unho: The following concern was raised on the PR initial message:

> Not sure it's right to get the `page_type` this way but couldn't figure out a better way, suggestions welcome.

I don't like the idea of hardcoding this. I would suggest to create a new URL on `staticpages` app named `pootle-staticpages-announcement-edit` that imports `ANN_TYPE` from `staticpages.views` and uses it for the URL pattern. This way it won't be necessary to hardcode anything on the template.

@julen Thoughts on this?

unho: Forget about this. Noticed that `page.get_edit_url` should provide us the URL without need to use the `url` templatetag.

unho:
Merged in 87e0474a with some minor fixes.

Note that finally was decided not to display the link if the user has `administrate` permission since doing that would require to alter the admin for staticpages so users with such permission can edit specific announcements. Also it would require altering the view to see the staticpages (like http://127.0.0.1:8000/pages/announcements/projects/firefox/) in order to check for the `administrate` permission as well.

pootle/templates/staticpages/_body.html
julen: This lacks i18n of any sort.

unho: Missed that. Landed a fix in e7db2b13.

PR #3970, opened 2015-08-12
Set #body height if it is less than the height of sidebar or content

phlax:
Currently after calculating the max height of sidebar and content elements, #body height is only set if "contentAreaHeight" is bigger than #body height

This works in firefox at least, but doesnt work in chrome

This commit instead checks if the max height is bigger than #body - and if so it sets the #body height to the max value

### DIFF
--- a/pootle/static/js/helpers.js
+++ b/pootle/static/js/helpers.js
@@ -41,7 +41,7 @@ var helpers = {
                         $('#footer').height() + bodyPadding,
         newHeight = Math.max(contentAreaHeight, sidebarHeight);
 
-    if (bodyHeight < contentAreaHeight) {
+    if (bodyHeight < newHeight) {
       $body.css('height', newHeight);
     }
   },


unho:
Is this a fix for #3812?

If yes, it is working on Chromium, but now the content area are always has the height of its higher item. In other words, now the footer is always at the bottom, but the page height doesn't shrink when the sidebar is collapsed.

phlax:
@unho yep - ive added a reference in the commit message.

in neither browser that i have tested (chrome 44.0.2403.89 , iceweasel 39.0) does the page shrink when the sidebar is collapsed.

In fact as set out in #3969 if the sidebar is collapsed the page is expanded even further

julen:
Thank you @phlax for taking a look at this. I want to merge the change, however as said on IRC can you please fix the commit message? Thank you.

phlax:
done

julen:
If you rebase the changes on top of current master we can have the PR automagically closed after merging.

phlax:
done

julen:
Thanks again, merged.

PR #3972, opened 2015-08-13
Remember expanded details in overview page

### LINKED ISSUE #3784: Expanded overview page does not remember if it was expanded or contracted
The (+) next to the progress bar on the overview page can be expanded to show more details, such as the full list of failing checks or the status counts.

If you follow a failing check from this page and then return to the overview you will arrive with the overview contracted. This is confusing especially if you are stepping through types of failing checks from the overview.

phlax:
When user expands details, this is remembered if the user returns.

This implementation uses history.pushState

fixes #3784

### DIFF
--- a/pootle/core/url_helpers.py
+++ b/pootle/core/url_helpers.py
@@ -172,6 +172,10 @@ def get_previous_url(request):
             if '?' in referer_url:
                 referer_url = referer_url[:referer_url.index('?')]
 
+                # But append ?details if present
+                if parsed_referer.query == "details":
+                    referer_url = referer_url + "?details"
+
             return referer_url
 
     return reverse('pootle-home')
--- a/pootle/static/js/stats.js
+++ b/pootle/static/js/stats.js
@@ -61,6 +61,8 @@ const stats = {
       e.preventDefault();
       this.refreshStats();
     });
+
+    this.reopenChecks();
   },
 
   refreshStats() {
@@ -261,7 +263,7 @@ const stats = {
     } else {
       // this is a single store stats, let's expand its details
       // only on first page load, and unless it is already expanded
-      if (firstPageLoad && !this.isExpanded) {
+      if (firstPageLoad && !this.checksStateIsOpen()) {
         this.toggleChecks();
       }
 
@@ -328,6 +330,8 @@ const stats = {
     this.$expandIcon.attr('title', newText);
 
     this.$extraDetails.slideToggle('slow', 'easeOutQuad');
+
+    this.checksStatePush();
   },
 
   loadChecks() {
@@ -366,11 +370,37 @@ const stats = {
 
       $('#js-stats-checks').show();
     }
-
     this.hasChecksData = true;
     this.toggleChecksVisibility();
-  }
+  },
 
+  checksStateIsOpen() {
+    return (window.location.search === '?details');
+  },
+
+  checksStatePush() {
+    if (window.history && window.history.pushState) {
+      if (!this.isExpanded) {
+	window.history.pushState("", "", this.pootlePath);
+      } else {
+	window.history.pushState("", "", "?details");
+      }
+    }
+  },
+
+  reopenChecks() {
+    var self = this;
+    if (this.checksStateIsOpen() && !this.isExpanded) {
+      this.toggleChecks();
+    }
+    window.onpopstate = function (e) {
+      if (self.checksStateIsOpen() && !self.isExpanded) {
+        self.toggleChecks();
+      } else if (!self.checksStateIsOpen() && self.isExpanded) {
+        self.toggleChecks();
+      }
+    }
+  }
 };
 
 


unho:
I am using Chromium 43 and if I visit a browser URL with or without `?details` it works as expected. But if I first open `/af/project/` and then click on the (+) button (which effectively changes the URL to `/af/project/?details`) and then I hit the browser Back button (switching to previous `/af/project/` URL) the details are kept expanded despite they should be automatically collapsed. If you then refresh the details are collapsed, but this is because you are getting a fresh state from the browser. 

Directly visiting a `/af/project/?details` URL works as expected.

pootle/core/url_helpers.py
```diff
@@ -172,6 +172,10 @@ def get_previous_url(request):
             if '?' in referer_url:
                 referer_url = referer_url[:referer_url.index('?')]
 
+            # But append ?details if present
+            if parsed_referer.query == "details":
+                referer_url = referer_url + "?details"
```
unho: I would move this within the previous `if` to reduce the number of checks.

phlax:
@unho ive added the indentation that you suggested - thanks

re the 2nd problem im not sure how this can be resolved - afaict hitting the back button doesnt reload the page just changes the url (or moves back in state), and as such doesnt affect the page layout.

unho:
> re the 2nd problem I am not sure how this can be resolved - afaict hitting the back button doesn't reload the page just changes the url (or moves back in state), and as such doesn't affect the page layout.

I have been using `Ctrl+Left` to go back in my history.

> it might be possible to listen for state changes and then test for presence of `?details`, although it seems a bit of an edge case

It is the only solution I could think of. Perhaps other people can contribute a better approach.

iafan:
You guys seem to be mixing `#?details` (hash part of the URL) and `?details` (URL query parameters). You're using the former, and indeed, when you hit back in the browser, if it's only the hash part that changes, browser will not reload the page. However, it will trigger an event in JavaScript, which can be used to open/close the details.

Also, if we're talking about the hash part, then `?` is not necessary. @julen may provide you a more quality feedback.

phlax:
@iafan this implementation use html5 history.pushState which allows you to use query params similar to how hashes have been used in html4

@unho ive added an event handler to listen for state changes, and open/close details

phlax:
@julen i have updated the commit to work with refactoring of stats.js

pootle/static/js/stats.js
```diff
@@ -366,11 +370,37 @@ const stats = {
...
     this.toggleChecksVisibility();
-  }
+  },
+
+  checksStateIsOpen() {
+    return (window.location.search === '?details');
+  },
+
+  checksStatePush() {
+    if (window.history && window.history.pushState) {
```
julen: This is redundant code — we are good to assume these are available.

pootle/static/js/stats.js
```diff
@@ -366,11 +370,37 @@ const stats = {
...
+      if (!this.isExpanded) {
+	window.history.pushState("", "", this.pootlePath);
```
julen: This will break in non-root deployments. We wrap paths within `l()` to ensure the server's script prefix is not missed out.

pootle/static/js/stats.js
```diff
@@ -366,11 +370,37 @@ const stats = {
...
+      } else {
+	window.history.pushState("", "", "?details");
```
julen: As a precaution, it's important not to blindly `pushState()`, otherwise extra history entries are created which add no value. So it's necessary to check if the current location differs from the target location before creating unnecessary history entries.

Also note the first parameter to `pushState()` is an object, not a string ([see docs](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState)).

pootle/static/js/stats.js
```diff
@@ -366,11 +370,37 @@ const stats = {
...
+      }
+    }
+  },
 
+  reopenChecks() {
```
julen: This is unintuitive. First off, it's taking more responsibilities than opening the details — it's also overwriting the `onpopstate` handler for `window`.

Also, arguably `reopen` has misleading semantics: this is not opening anything *again*.

pootle/static/js/stats.js
```diff
@@ -366,11 +370,37 @@ const stats = {
...
+    var self = this;
+    if (this.checksStateIsOpen() && !this.isExpanded) {
+      this.toggleChecks();
+    }
+    window.onpopstate = function (e) {
```
julen: Please do not overwrite listeners that may have been set by other code.
```javascript
window.addEventListener('popstate', (e) => {
  // do something/callback
});
```

julen:
I started reviewing this PR yesterday and ended up fixing other tangential issues which affect this specific module. I'm afraid I have created merge conflicts however everything should be simpler.
(*I have still left inline comments on the conflicting changes as an exercise for the reader.*)

--

In general I had the feeling of diving into ad-hoc code, which I'd like to avoid whenever possible — we already have enough of that.

Since the main idea is to have the `?details` query string denote the state of the UI, the simplest thing is to check if such query string is present on page load (`stats.init()`), and assign the result of the test to `this.state.isExpanded`. This actually decouples the feature from a store-specific scenario, practically allowing to load an expanded UI in any browsing page by just appending `?details` to the URL.

This means we will want to trigger this behavior explicitly from browsing URLs, i.e. by having links to stores (in the browsing table or in the breadcrumb's resources drop-down) expose this query string.

At the same time, the history entry needs to be kept in sync whenever user interactions provoke app state changes. Notice `pushState()` allows to store a state object which can be read in a handler of the `popstate` event. This allows to restore the UI state when users interact with the browser back/forward buttons.

Finally a friendly reminder of the [JS conventions](http://docs.translatehouse.org/projects/pootle/en/latest/developers/styleguide.html#javascript).

julen:
A simple stub for the concept I described https://gist.github.com/julen/ba805b533b42f4a14d2a

Issue #3975, opened 2015-08-14
Uploads should use suggestions for conflicts


dwaynebailey:
Currently when uploading a file that you've worked on offline if a unit has changed upstream we will reject the upload.  This is based on the Revision number that we inject into the PO files that was downloaded from Pootle.

But since we have the revision number we could do more intelegent things that don't mess the user around so much.  Using the revision number we should be able to process each unit as follows:

1. If unit revision is older than uploaded file revision - accept the translation
2. If unit revision is newer (i.e. someone translated online while we were out) - turn that into a suggestion.
3. If unit doesn't exist in upstream - skip it (I think that's the correct thing to do and might not be impacted by this change but it needs to be checked)

Nice to haves:

1. Report the fact that we've made suggestions where there was a clash.
2. Post upload the user is taken the the suggestions that have just been made so that they can resolve them as needed.

ta2-1:
@dwaynebailey I remember you raised this question on IRC but only now I found `Store.update()` method changed.
As I understood 1,2 and 3 from description were already addressed by @phlax (probably it'd be worth to create a GH tracker for this issue).

There was a discussion here but I didn't pay attention because I was sure it was totally about import-export. But I see that  `Store.update()` method is going to grow. We discussed with @lulen that we have to avoid making of `store.update()` more hairy. And I see that it was the only place where we could add such functionality in this case.

As I understood, `X-Pootle-Revision` doesn't affect EN workflow since we don't use export (cc @iafan).

dwaynebailey:
@ta2-1 yes no place to alter this except in ``Store.update()``.  The good news is we actually have tests and I know @phlax has been addding and is well aware that we have very few to no tests in this area.

My long terms aim is that there is no real difference between ``update_stores``/``sync_stores`` and import/export.  Since they are overlapping functionality.

I'm taking the view that as we improve and add tests for this code it will be more clear how we can refactor this code and make ``Store.update()`` less complex.

But summary is that without having ``X-Pootle-Revision`` in the store the normal behaviour will continue and you are unaffected.

dwaynebailey:
Fixed in 4fbaa573f51b66d796582d85eaf4c44862dec89c

PR #3976, opened 2015-08-14
Use suggestions on upload conflicts

phlax:
If user uploads translation file with version number less than a unit to be
updated (ie unit has been edited ttw since translation file was downloaded),
then add the uploaded unit as a suggestion instead of updating.

Touch #3975


dwaynebailey:
Tested uploads.  Changes made online are kept and the conflcting translation appears as a suggestion.  All the other stuff works as expected.

I didn't check a submission with a unit that is now obsolete.

PR #3981, opened 2015-08-16
Adding django-transaction-hooks

### LINKED ISSUE #3974: When uploading translation files, async worker is triggered before transaction completes (was: cache is invalidated before changes are applied)
To reproduce:

- download translation file
- make change (change 1)
- upload file

At this point the stats/last activity does not show change 1

- download translation file again
- make change (change 2)
- upload file

At this point the stats/last activity shows change 1 and not change 2

phlax:
When async code is called from within sync code, there is no guarantee that the
transaction has been committed. This could cause serious db regressions and was
preventing the stats from being updated correctly on file upload.

Fixes #3974

### DIFF
--- a/docs/server/installation.rst
+++ b/docs/server/installation.rst
@@ -275,11 +275,26 @@ Use the :command:`psql` command to create a user and database:
    postgres@ $ createdb --encoding='utf-8' --locale=en_US.utf8 --template=template0 --owner=pootle pootledb
 
 
+.. _installation#database_backends:
+
+Database backends
+-----------------
+
+.. warning:: Pootle now requires django-transaction-hooks.
+   **You should update your database backend if migrating from a version older than 2.7.1**
+
+
 Following the database creation, you need to modify the :setting:`DATABASES`
 setting appropriately in your custom settings file, ensuring that you are using
 the correct :setting:`ENGINE <DATABASE-ENGINE>` setting for your chosen
 database backend.
 
+Pootle requires `django-transaction-hooks <https://pypi.python.org/pypi/django-transaction-hooks/>`_
+to connect to the database. The following database backends are supported:
+
+- sqlite: transaction_hooks.backends.sqlite3
+- mysql: transaction_hooks.backends.mysql
+- postgres: transaction_hooks.backends.postgresql_psycopg2
 
 .. _installation#populating_the_database:
 
--- a/pootle/core/mixins/treeitem.py
+++ b/pootle/core/mixins/treeitem.py
@@ -16,6 +16,7 @@ from functools import wraps
 
 from django.conf import settings
 from django.core.urlresolvers import set_script_prefix
+from django.db import connection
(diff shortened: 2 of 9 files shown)


unho:
I must say I am reluctant to add a new dependency.

unho:
lgtm but not sure how to test. Can you please provide detailed instructions. https://github.com/translate/pootle/issues/3974#issue-100967596 seems a bit vague to me.

dwaynebailey:
This also needs a review from @translate/evernote, actually julen requested @ta2-1 have a look.

phlax:
@unho i think i can write a test for this - ill take a look this morning

the other thing i wasn't 100% clear about is where to trigger the on_commit hook - i think we probably want it that all triggers to the async worker wait for the pending transaction commit if relevant

phlax:
> I must say I am reluctant to add a new dependency.

no, me too - but i don't really see a way around it. At least the dep is only there until django 1.9

The other thing we probably need to update is docs - to say that you must use the transaction_hooks db backends

docs/server/installation.rst
```diff
@@ -280,6 +280,13 @@ setting appropriately in your custom settings file, ensuring that you are using
 the correct :setting:`ENGINE <DATABASE-ENGINE>` setting for your chosen
 database backend.
 
+Pootle requires `django-transaction-hooks <https://pypi.python.org/pypi/django-transaction-hooks/>`_
+to connect to the database. The following database backends are supported:
+
+  - sqlite: transaction_hooks.backends.sqlite3
+  - mysql: transaction_hooks.backends.mysql
+  - postgres: transaction_hooks.backends.postgresql_psycopg2
```
unho: This belongs to release notes. Also make sure all the setting files are updated accordingly.

pootle/core/mixins/treeitem.py
```diff
@@ -17,6 +17,7 @@
 from django.conf import settings
 from django.core.urlresolvers import set_script_prefix
 from django.utils.encoding import force_unicode, iri_to_uri
+from django.db import connection
```
unho: Move this one line up. Please follow our styleguide on imports http://docs.translatehouse.org/projects/pootle/en/latest/developers/styleguide.html#imports

docs/server/installation.rst
phlax: I dont really agree

At the moment it links to the normal django db config page. If you just follow that it will create a problem so some explanation needs to be given here

unho: It is still necessary to include this warning on the release notes. Users upgrading must not miss this important change.

Since it is an important change I also would like to see some checks on https://github.com/translate/pootle/blob/master/pootle/checks.py to ensure the old values (regular Django database backends) are no longer used.

I suggest this to be converted into a warning to catch the reader's eye, move it to the beginning of the section, and highlight `transaction_hooks.backends.sqlite3` and relatives.

phlax: yep - cool - ill add these things

PR #3988, opened 2015-08-18
Differentiating non-critical checks

### LINKED ISSUE #3776: Non critical checks in Pootle editor are shown in red
We're only exposing critical errors in the overview.  And exposing all checks in the editor.  We only block submission on critical errors, we don't block on other error classes.

However, we show all errors in Red.

For non-critical errors we should choose a cooler colour from the UI so that translators will know what are the critical errors.

phlax:
Fixes #3776

### DIFF
--- a/pootle/apps/pootle_misc/checks.py
+++ b/pootle/apps/pootle_misc/checks.py
@@ -21,14 +21,21 @@ from django.utils.translation import ugettext_lazy as _
 from .util import import_func
 
 
-category_ids = {
+CATEGORY_IDS = {
     'critical': Category.CRITICAL,
     'cosmetic': Category.COSMETIC,
     'functional': Category.FUNCTIONAL,
     'extraction': Category.EXTRACTION,
     'other': Category.NO_CATEGORY,
 }
-
+CATEGORY_CODES = {v: k for k, v in CATEGORY_IDS.iteritems()}
+CATEGORY_NAMES = {
+    Category.CRITICAL: _("Critical"),
+    Category.COSMETIC: _("Cosmetic"),
+    Category.FUNCTIONAL: _("Functional"),
+    Category.EXTRACTION: _("Extraction"),
+    Category.NO_CATEGORY: _("Other"),
+}
 
 check_names = {
     'accelerators': _(u"Accelerators"),  # fixme duplicated
@@ -232,6 +239,18 @@ def get_checker(unit):
         return import_func(settings.POOTLE_QUALITY_CHECKER)()
 
     return unit.store.translation_project.checker
+
+
+def get_category_id(code):
+    return CATEGORY_IDS.get(code)
+
+
+def get_category_code(cid):
+    return CATEGORY_CODES.get(cid)
+
+
+def get_category_name(code):
+    return unicode(CATEGORY_NAMES.get(code))
 
 
 class SkipCheck(Exception):
@@ -1000,22 +1019,17 @@ def get_qualitychecks():
 
 
 def get_qualitycheck_schema(path_obj=None):
+    # TODO: add tests
+
     d = {}
     checks = get_qualitychecks()
-
-    category_names = {
-        Category.CRITICAL: _("Critical"),
-        Category.COSMETIC: _("Cosmetic"),
-        Category.FUNCTIONAL: _("Functional"),
-        Category.EXTRACTION: _("Extraction"),
-        Category.NO_CATEGORY: _("Other"),
-    }
 
     for check, cat in checks.items():
         if cat not in d:
             d[cat] = {
                 'code': cat,
-                'title': u"%s" % category_names[cat],
+                'name': get_category_code(cat),
+                'title': get_category_name(cat),
                 'checks': []
             }
         d[cat]['checks'].append({
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -832,6 +832,15 @@ class Unit(models.Model, base.TranslationUnit):
     def get_qualitychecks(self):
         return self.qualitycheck_set.all()
 
+    def get_critical_qualitychecks(self):
+        return self.get_qualitychecks().filter(category=Category.CRITICAL)
+
+    def get_active_critical_qualitychecks(self):
+        return self.get_active_qualitychecks().filter(category=Category.CRITICAL)
+
+    def get_warning_qualitychecks(self):
+        return self.get_qualitychecks().exclude(category=Category.CRITICAL)
+
     def get_active_qualitychecks(self):
         return self.qualitycheck_set.filter(false_positive=False)
 
--- a/pootle/templates/editor/units/xhr_checks.html
+++ b/pootle/templates/editor/units/xhr_checks.html
@@ -1,17 +1,35 @@
 {% load i18n locale staticpages %}
-{% if unit.get_qualitychecks.count %}
-<span class="sidetitle" lang="{{ LANGUAGE_CODE }}" title="{% trans 'Possible issues with the translation' %}">{% trans "Failing checks:" %}</span>
-<ul class="checks">
-  {% for check in unit.get_qualitychecks.iterator %}
-  <li class="check{% if check.false_positive %} false-positive{% endif %}">
-    <a href="{% url 'pootle-checks-descriptions' %}#{{ check.name }}" target="_blank">{{ check.display_name }}</a>
-    {% if canreview %}
-    <a class="js-toggle-check toggle-check" data-check-id="{{ check.id }}">
-      <i title="{% trans 'Mute quality check' %}" class="mute-check icon-block"></i>
-      <i title="{% trans 'Unmute quality check' %}" class="unmute-check icon-block-muted"></i>
-    </a>
-    {% endif %}
-  </li>
-  {% endfor %}
-</ul>
+{% with unit.get_critical_qualitychecks as checks_critical %}
+{% if checks_critical %}
+<div class="unit-checks">
+  <span class="sidetitle" lang="{{ LANGUAGE_CODE }}" title="{% trans 'Critical issues with the translation' %}">{% trans "Critical errors:" %}</span>
+  <ul class="checks checks-critical">
+    {% for check in checks_critical %}
+    <li class="check{% if check.false_positive %} false-positive{% endif %}">
+      <a href="{% url 'pootle-checks-descriptions' %}#{{ check.name }}" target="_blank">{{ check.display_name }}</a>
+      {% if canreview %}
+      <a class="js-toggle-check toggle-check" data-check-id="{{ check.id }}">
+	<i title="{% trans 'Mute quality check' %}" class="mute-check icon-block"></i>
+	<i title="{% trans 'Unmute quality check' %}" class="unmute-check icon-block-muted"></i>
+      </a>
+      {% endif %}
+    </li>
+    {% endfor %}
+  </ul>
+</div>
 {% endif %}
+{% endwith %}
+{% with unit.get_warning_qualitychecks as checks_warning %}
+{% if checks_warning %}
+<div class="unit-checks">
+  <span class="sidetitle" lang="{{ LANGUAGE_CODE }}" title="{% trans 'Possible issues with the translation' %}">{% trans "Watch out for:" %}</span>
+  <ul class="checks checks-warning">
+    {% for check in checks_warning %}
+    <li>
+      <a href="{% url 'pootle-checks-descriptions' %}#{{ check.name }}" target="_blank">{{ check.display_name }}</a>
+    </li>
+    {% endfor %}
+  </ul>
+</div>
+{% endif %}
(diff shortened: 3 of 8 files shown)


iafan:
Do you mind providing a screenshot that depicts the UI change?

phlax:
@iafan - here you go



editor-checks-non-critical-orange

iafan:
To be honest, I don't believe this color change is thought-through and desired. We're introducing a new key color for non-critical checks that is close to the color of suggestions. We still have the red 'mute suggestion' icon next to the orange suggestion name; lastly, orange on top of red background is not particularly legible.

I'd suggest not to change the color, but rather add italic style to such non-critical checks. It will not introduce any new key colors, and at the same time it will provide the additional visual clue.

unho:
I agree that color only probably won't be enough. I suggest choosing another color and use italics as suggested above. It is better to have two clues than just one.

iafan:
I really don't want to introduce any new colors. We have a pretty solid and distinct color scheme when it comes to translations/suggestions/checks, and I don't want that scheme to be diluted for not much reason.

iafan:
I did a quick test, and using simply italic is enough to tell two kinds apart.

phlax:
italics are generally best avoided as they lower accessibility - see here for brief discussion:
http://uxmovement.com/content/6-surprising-bad-practices-that-hurt-dyslexic-users/

I chose orange because i figured orange is to warning as red is to error.

I think its not so bad that the contrast is subtle given the comments in #3776

iafan:
I'm all for accessibility, but are you seriously trying to optimize _translation_ UI for people with _dyslexia_?

I would personally prefer to leave all the checks in the same color and style, but put those checks in the editor into different corresponding sections so that one can easily tell them apart:
```
Critical: aaaaaaaa bbb cccccc
Functional: ddddd ee
Cosmetic: fffffff
```

phlax:
> are you seriously trying to optimize translation UI for people with dyslexia?
no - trying to optimize for accessibility generally

in the overview they are in separate sections.

I think the spirit of #3776 is that if you make everything critical/error then nothing becomes critical - i do think some sort of differentiation is useful

unho:
> I would personally prefer to leave all the checks in the same color and style, but put those checks in the editor into different corresponding sections so that one can easily tell them apart:
```
Critical: aaaaaaaa bbb cccccc
Functional: ddddd ee
Cosmetic: fffffff
```

+1

dwaynebailey:
@iafan I'm OK with your proposal though it uses a fair bit of real estate to achieve that.

What is being missed here about the colour communication is this:  Only critical checks block the editor.  If you want to move off the editor you have to fix that check or mute it.  With checks all being the same colour (even with categorisation) it is not clear to the translator that some checks are more important than others.

Using a different colour (any colour to be honest) for failing non-critical checks allows us to communicate that these are different and non-blocking and that they are not as critical as the 'critical' checks which we keep in red.  And red hopefully for most people communicates danger/no/don't.

At the moment, everything is red.  So even with categories everything is equally dangerous and important.

julen:
I think this boils down to only using red for critical checks. This also applies to the background of the editing widget: currently it's red even if the check is not critical (I have suffered this myself on MLO, it's confusing and it provides the wrong impression of brokenness).

iafan:
Let me step back a bit and think of non-critical checks in a more broad scope than just styling.

We want to have non-critical checks, by their nature there can be many checks of this kind, and in most cases the majority of them will never be addressed (since they are hidden by default, and don't prevent user from advancing to the same unit. Such checks are just additional hints on what to pay attention at when translating the unit.

Allowing to mute and unmute non-critical checks, I suppose, is also not of a big value, since this doesn't affect the way people work with units. Given the number of non-critical checks, it is not likely that people will be regularly going through these checks to mute or fix them.

Having said that, we can probably treat non-critical checks basically as already muted ones: we can render them in black, and do not present mute/unmute icons next to them in the editor. Even if they will му visually distinct because of color, I still think it makes sense to split checks at least into two explicit categories:

```
Critical errors:   Newlines(–)   XML Tags(–)   printf(–)
Watch out for:   Double quotes   Ending whitespace
```

We would also display pink background only on units that have one or more critical errors.

phlax:
I have implemented the changes suggested above - feeback welcome

phlax:
Here are some screenshots:

editor_with_critical_errors
editor_with_non_critical_warnings
checks_overview

phlax:
I have made the labels "Error" and "Warning" - lmk if these should be changed

iafan:
Thanks, @phlax! Yes, I think `Critical errors:` and `Watch out for:` are a tad more friendly and easier to read. Let's change the labels and see how it looks.

dwaynebailey:
lgtm, agree with @iafan on the wording

iafan:
@phlax, did you also have a chance to remove the ability to mute/unmute non-critical checks? It's not clear from the commits/screenshots.

phlax:
yep

> Removes un/mute widget for non-critical checks

phlax:
wording updated

iafan:
Thanks! that's a gtm from me, functionality- and ui-wise.

pootle/apps/pootle_misc/checks.py
```diff
@@ -1004,18 +1004,19 @@ def get_qualitycheck_schema(path_obj=None):
...
+        Category.NO_CATEGORY: "Other",
     }
 
     for check, cat in checks.items():
         if cat not in d:
             d[cat] = {
                 'code': cat,
-                'title': u"%s" % category_names[cat],
+                'name': category_names[cat].lower(),
+                'title': u"%s" % _(category_names[cat]),
```
julen: This doesn't work, the localizable names above should be kept. Feel free to add an extra mapping for retrieving identifiers.

pootle/templates/editor/units/xhr_checks.html
```diff
@@ -1,8 +1,9 @@
 {% load i18n locale staticpages %}
-{% if unit.get_qualitychecks.count %}
-<span class="sidetitle" lang="{{ LANGUAGE_CODE }}" title="{% trans 'Possible issues with the translation' %}">{% trans "Failing checks:" %}</span>
-<ul class="checks">
-  {% for check in unit.get_qualitychecks.iterator %}
+{% with unit.get_qualitychecks_critical as checks_critical %}
+{% if checks_critical.count %}
```
julen: I'd say there's no need for `.count`, as you're accessing the actual qs results right away. Also, if you'd only want to check for existence, `.exists` is your friend.

pootle/templates/editor/units/xhr_checks.html
```diff
@@ -15,3 +16,17 @@
   {% endfor %}
 </ul>
 {% endif %}
+{% endwith %}
+{% with unit.get_qualitychecks_warning as checks_warning %}
+{% if checks_warning.count %}
```
julen: Ditto as above.

pootle/templates/editor/units/xhr_checks.html
```diff
@@ -15,3 +16,17 @@
...
+<ul class="checks checks-warning">
+  {% for check in checks_warning.iterator %}
+  <li class="check{% if check.false_positive %} false-positive{% endif %}">
+    <a href="{% url 'pootle-checks-descriptions' %}#{{ check.name }}" target="_blank">{{ check.display_name }}</a>
+  </li>
+  {% endfor %}
+</ul>
+{% endif %}
+{% endwith %}
+
```
julen: No need for this extra blank line.

pootle/apps/pootle_misc/checks.py
phlax: Thanks I wasn't sure about this

pootle/apps/pootle_misc/checks.py
```diff
@@ -1008,13 +1010,17 @@ def get_qualitycheck_schema(path_obj=None):
         Category.COSMETIC: _("Cosmetic"),
         Category.FUNCTIONAL: _("Functional"),
         Category.EXTRACTION: _("Extraction"),
-        Category.NO_CATEGORY: _("Other"),
+        Category.NO_CATEGORY: _("Other")
```
unho: Please do not remove the trailing comma. If any new entry is added in the future this would save us from having an unnecessary extra diff line.

pootle/templates/editor/units/xhr_checks.html
```diff
@@ -15,3 +16,16 @@
   {% endfor %}
 </ul>
 {% endif %}
+{% endwith %}
+{% with unit.get_qualitychecks_warning as checks_warning %}
+{% if checks_warning %}
+<span class="sidetitle" lang="{{ LANGUAGE_CODE }}" title="{% trans 'Possible issues with the translation' %}">{% trans "Watch out for:" %}</span>
+<ul class="checks checks-warning">
+  {% for check in checks_warning %}
+  <li class="check{% if check.false_positive %} false-positive{% endif %}">
```
unho: Is actually necessary to check for `false_positive` for warning checks that can't be muted?

pootle/apps/pootle_store/models.py
```diff
@@ -831,6 +831,15 @@ def update_qualitychecks(self, keep_false_positives=False,
     def get_qualitychecks(self):
         return self.qualitycheck_set.all()
 
+    def get_qualitychecks_critical(self):
+        return self.get_qualitychecks().filter(category=Category.CRITICAL)
+
+    def get_active_qualitychecks_critical(self):
```
unho: I think naming it `get_active_critical_qualitychecks` feels more natural. Similar to the previous method.

unho:
Nit: Can we add a bit some extra blank vertical space between both groups of checks if the two of them are present?

pootle/apps/pootle_misc/checks.py
```diff
@@ -1008,13 +1010,17 @@ def get_qualitycheck_schema(path_obj=None):
...
     }
 
+    cat_id_keys = category_ids.keys()
+    cat_id_values = category_ids.values()
+
     for check, cat in checks.items():
         if cat not in d:
             d[cat] = {
                 'code': cat,
+                'name': cat_id_keys[cat_id_values.index(cat)],
```
julen: Isn't it simpler to create an inverse mapping of `category_ids` at the module level and use it right away?
```python
category_codes = {v: k for k, v in category_ids.iteritems()}
...
    'name': category_codes[cat],
...
```

As a side note, we may want to rename `'name'` to `'code'`, and `'code'` to `'id'`; but that's a separate cleanup.

julen:
Thank you for taking this @phlax, tested and works like a charm. It's good to be merged after addressing the inline comments.

pootle/apps/pootle_misc/checks.py
phlax: yep agreed - id been thinking about this

i also figured it would be good to have a get_category_name_from_code() func or similar

pootle/apps/pootle_misc/checks.py
```diff
@@ -29,6 +29,15 @@
...
 }
 
+CHECK_CATEGORY_CODES = {v: k for k, v in category_ids.iteritems()}
+CHECK_CATEGORY_NAMES = {
+    Category.CRITICAL: _("Critical"),
+    Category.COSMETIC: _("Cosmetic"),
+    Category.FUNCTIONAL: _("Functional"),
+    Category.EXTRACTION: _("Extraction"),
+    Category.NO_CATEGORY: _("Other"),
+    }
```
unho: Please remove indent here.

pootle/apps/pootle_misc/checks.py
```diff
@@ -29,6 +29,15 @@
...
 }
 
+CHECK_CATEGORY_CODES = {v: k for k, v in category_ids.iteritems()}
+CHECK_CATEGORY_NAMES = {
+    Category.CRITICAL: "Critical",
+    Category.COSMETIC: "Cosmetic",
+    Category.FUNCTIONAL: "Functional",
+    Category.EXTRACTION: "Extraction",
+    Category.NO_CATEGORY: "Other",
+    }
```
unho: Remove indent here.

pootle/apps/pootle_misc/checks.py
```diff
@@ -1000,22 +1020,17 @@ def get_qualitychecks():
...
-        Category.NO_CATEGORY: _("Other"),
-    }
-
     for check, cat in checks.items():
         if cat not in d:
             d[cat] = {
                 'code': cat,
-                'title': u"%s" % category_names[cat],
+                'name': get_check_category_code(cat),
+                'title': u"%s" % get_check_category_name(cat),
```
unho: A bit easier to read:

```python
                'title': unicode(get_check_category_name(cat)),
```

phlax:
> Nit: Can we add a bit some extra blank vertical space between both groups of checks if the two of them are present?

Im not sure if there is enough room - atm the elements on the right are positioned absolutely so there is a risk of the elements colliding if there is too much height

unho:
Review and tested locally. gtm

PR #4000, opened 2015-08-21
Projects dropdown: visually differentiate disabled projects

### LINKED ISSUE #3996: Visually differentiate disabled projects in the navigation dropdown
In the navigation dropdown, admins can see and select disabled projects (which is ok). The problem is that they have the same style as non-disabled ones.

Let's render disabled project as italic and with .5 opacity, to match their styling everywhere else in the UI.

julen:
Note this changes the data for projects stored in the cache, so it needs to be cleared for the new structure to work correctly.

PRing so everyone is aware of the change in the cache. Would be great if someone can come up with a better way to communicate this, especially across versions.

Fixes #3996.

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -36,6 +36,16 @@ Details of changes
   is now built on the client. This requires refreshing all server stats using
   the :djadmin:`refresh_stats_rq` command (:issue:`3835`).
 
+- Disabled projects are visually differentiated in the projects drop-down
+  (:issue:`3996`).
+  Since the in-cache data structure supporting this changed, it's necessary to
+  clear the cache. Assuming your ``default`` cache lives in the DB number ``1``,
+  you can clear it as follows:
+
+  .. code-block:: bash
+
+    $ redis-cli -n 1 KEYS "*method-cache:Project:cached_dict:*" | xargs redis-cli -n 1 DEL
+
 - Pulled latest translations.
 
 
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -62,9 +62,11 @@ class ProjectManager(models.Manager):
         projects = cache.get(cache_key)
         if not projects:
             logging.debug('Cache miss for %s', cache_key)
+            projects_dict = self.for_user(user).order_by('fullname') \
+                                               .values('code', 'fullname',
+                                                       'disabled')
             projects = OrderedDict(
-                self.for_user(user).order_by('fullname')
-                                   .values_list('code', 'fullname')
+                (project.pop('code'), project) for project in projects_dict
             )
             cache.set(cache_key, projects, settings.POOTLE_CACHE_TIMEOUT)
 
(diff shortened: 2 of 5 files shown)


docs/releases/dev.rst
```diff
@@ -34,6 +34,10 @@ Details of changes
   is now built on the client. This requires refreshing all server stats using
   the :djadmin:`refresh_stats_rq` command (:issue:`3835`).
 
+- Disabled projects are visually differentiated in the projects drop-down
+  (:issue:`3996`). Since the in-cache data structure supporting this changed,
+  it's necessary to clear the cache.
```
unho: Perhaps it is a good idea to provide some hint on how to do that.

unho: Just tested this locally and works without need to clear the cache.

julen: > Perhaps it is a good idea to provide some hint on how to do that.

Any specific user-friendly suggestions? if you read the PR description I'm asking for help on that.

julen: > Just tested this locally and works without need to clear the cache.

Unless something else has triggered clearing your cache, it's absolutely required to do so. This code doesn't work with stale data.

unho: >> Perhaps it is a good idea to provide some hint on how to do that.

> Any specific user-friendly suggestions? if you read the PR description I'm asking for help on that.

No. I can only think on asking the user to open a shell and run some simple Python code. But that is definitely not friendly. On the other hand throwing away tons of calculated stats sounds really really scary. Perhaps we should expand `clear_stats` to allow to remove some specific cache by providing some identifier, like `--projectdropdowns`.

julen: This has nothing to do with the stats cache.

unho: Missed that.

Adding a specific command for this feels a bit excessive, but I can't think of a better solution.

dwaynebailey:
@julen this is stored in our default cache right? That times out after 60 seconds.  I would suggest that it requires a ``redis-cli`` command to clean things in the upgrade notes.  We hit this need for every UI related regeneration anyway so it could be a good part of instructions for new assets.

julen:
This is stored in the default cache but its timeout is set by `POOTLE_CACHE_TIMEOUT`, which defaults for a week. I rebased and added the `redis-cli` command to clear this up. This is good to go from my side.

docs/releases/dev.rst
```diff
@@ -36,6 +36,14 @@ Details of changes
   is now built on the client. This requires refreshing all server stats using
   the :djadmin:`refresh_stats_rq` command (:issue:`3835`).
 
+- Disabled projects are visually differentiated in the projects drop-down
+  (:issue:`3996`).
+  Since the in-cache data structure supporting this changed, it's necessary to
+  clear the cache. Assuming your ``default`` cache lives in the DB number ``1``,
+  you can clear it as follows::
+
+  $ redis-cli -n 1 KEYS "*method-cache:Project:cached_dict:*" | xargs redis-cli -n 1 DEL
```
unho: Please put this in a `code-block` with `shell` highlight.

unho:
gtm after the raised nit is addressed.

Issue #3797, opened 2015-06-04
Deleting users should correctly shift attribution


dwaynebailey:
From IRC:

Dwayne: @julen if I delete a user on Pootle in the shell i.e. `u = User.objects.get(id=NN)` and then `u.delete()` does that do all the work of oving and contributions to 'nobody' user?

Julen: @dwaynebailey: nope, and the message in the UI is misleading as that hasn't been implemented in any case

iafan:
When it comes to deleting users, do we always want to attribute everything to _nobody_? What's the primary use case for user deletion? I think one of the common cases for user deletion would be to clean up spam accounts, which means deleting user's suggestions (and reverting their changes) along with deleting the user itself.

dwaynebailey:
These are my use cases:

- Delete - I'm doing cleanup of duplicate emails or users with no emails, etc and I don't want to lost any of their contributions
- Merge - this is the same user with different usernames, email.  I'd like to create one user out of the two
- Spam - I've never actually needed or wanted to do this to be honest.  But yes I can imagine a case where we want to revert all contributions.  But there is probably a software case, not of spam, but objectionable user or politically motivated suggestions.  Where you want to remove all pending suggestions from that person but you might not want to revert anything that was reviewed.

iafan:
Thanks, @dwaynebailey 

So it looks like this boils down to two action items:

1. Proper user merging (because 'deleting' a user in the sense you describe actually means merging it with 'nobody', which would shift attributions). This will also give us arbitrary merging capabilities.
2. Wiping out the user (because of e.g. spam). In this case we will need to remove pending suggestions, and, probably, reverting active submissions to the prior state.

unho:
Related #3810.

phlax:
@dwaynebailey @iafan 

Currently when User.delete is called **all** related_objects are deleted - units, suggestions, submissions, etc

From the above comments I understand the changes to be:
 - On User.delete default to merging account with "nobody"
 - Have the ability to merge 2 arbitrary accounts
 - Have the ability to purge an account

I have added a commit to address this issue

It does the following:

  - Adds merge_user function to merge user's related_objects to other user
  - On User.delete() merge_user is called to merge to "nobody" account
  - Adds "purge" parameter to User.delete (default=False) to restore default behaviour
  - Adds merge_user and purge_user management commands
  - Adds tests to test User.delete, merging and purging

[...]

PR #4002, opened 2015-08-23
Set attribution on User.delete

phlax:
This commit:
  - Adds `merge_user` function to merge user's related_objects to other user
  - On `User.delete()` `merge_user` is called to merge to "nobody" account
  - Adds `purge_user` function:
      - Reverts edits
      - Reverts comments
      - Reverts reviews
      - Removes units created by user with no other activity
  - Adds "purge" parameter to `User.delete` `(default=False)
  - Adds `merge_user` and `purge_user` management commands
  - Adds tests to test `User.delete`, merging and purging

touch #3797

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -22,6 +22,7 @@ Major Changes
 
 - Updated translations.
 - Added django-transaction-hooks
+- Changed user delete behaviour
 
 
 Below we provide much more detail. These are by no means exhaustive, view the
@@ -65,12 +66,31 @@ Django transaction hooks
   - postgres: transaction_hooks.backends.postgresql_psycopg2
 
 
+Changed user delete behaviour
+-----------------------------
+
+On deleting a user account their submissions, suggestions and reviews are now
+re-assigned to the "nobody" user.
+
+If you wish to remove the user's contributions also, you can use the
+:djadmin:`purge_user` command, or call ``user.delete(purge=True)`` to delete the
+user programatically.
+
+
 Command changes and additions
 -----------------------------
 
 - Added a :djadmin:`contributors` command to get the list of contributors
   (:issue:`3867`).
 
+- Added a :djadmin:`merge_user` command to get merge submissions, comments and
+  reviews from one user account to another. This is useful for fixing users
+  that have multiple accounts and want them to be combined. No profile data
+  is merged.
+
+- Added a :djadmin:`purge_user` command to purge a user from the site and revert
+  any submissions, comments and reviews that they have made. This is useful to
+  revert spam or a malicious user.
 
 ...and lots of refactoring, cleanups to remove old Django versions specifics,
 improved documentation and of course, loads of bugs were fixed.
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -636,6 +636,52 @@ and the process will start a watchdog to track any client-side scripts for
 changes. Use this only when developing Pootle.
 
 
+.. _commands#user-management:
+
+Managing users
+--------------
+
+
+.. django-admin:: merge_user
+
+merge_user
+^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+This can be used if you have a user with two accounts and need to merge one
+account into another. This will re-assign all submissions, units and
+suggestions, but not any of the user's profile data.
+
+This command requires 2 mandatory arguments, ``src_username`` and
+``target_username``, both should be valid usernames for users of your site.
+Submissions from the first are re-assigned to the second. The users' profile
+data is not merged.
+
+.. code-block:: bash
+
+    $ pootle merge_user src_username target_username
+
+
+.. django-admin:: purge_user
+
+purge_user
+^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+This command can be used if you wish to permanently remove a user and revert
+the edits, comments and reviews that the user has made. This is useful for
+removing a spam account or other malicious user.
+
+This command requires a mandatory ``username`` argument, which should be a valid
+username for a user of your site.
+
+.. code-block:: bash
+
+    $ pootle purge_user username
+
+
 .. _commands#running:
 
 Running WSGI servers
--- a/pootle/apps/accounts/management/commands/__init__.py
+++ b/pootle/apps/accounts/management/commands/__init__.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand, CommandError
+
+
+User = get_user_model()
+
+
+class UserCommand(BaseCommand):
+    """Base class for handling user commands."""
+
+    args = "user"
+
+    def handle(self, *args, **kwargs):
+        self.check_args(*args)
+
+    def create_parser(self, prog_name, subcommand):
+        self.prog_name = prog_name
+        self.subcommand = subcommand
+        return super(UserCommand, self).create_parser(prog_name, subcommand)
+
+    def get_user(self, username):
+        try:
+            return User.objects.get(username=username)
+        except User.DoesNotExist:
+            raise CommandError("User %s does not exist" % username)
+
+    def usage_string(self):
+        return self.usage(self.subcommand).replace('%prog', self.prog_name)
+
+    def check_args(self, *args):
+        if len(args) != len(self.args.split(" ")):
+            raise CommandError("Wrong number of arguments\n\n%s"
+                               % self.usage_string())
--- a/pootle/apps/accounts/management/commands/merge_user.py
+++ b/pootle/apps/accounts/management/commands/merge_user.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import accounts
+
+from . import UserCommand
+
+
+class Command(UserCommand):
+    args = "user other_user"
+    help = "Merge user to other_user"
+
+    def handle(self, *args, **kwargs):
+        super(Command, self).handle(*args, **kwargs)
+        accounts.utils.UserMerger(self.get_user(username=args[0]),
+                                  self.get_user(username=args[1])).merge()
--- a/pootle/apps/accounts/models.py
+++ b/pootle/apps/accounts/models.py
@@ -38,6 +38,7 @@ from pootle_statistics.models import Submission, SubmissionTypes
 from pootle_store.models import SuggestionStates, Unit
 
 from .managers import UserManager
+from .utils import UserMerger, UserPurger
 
 
 CURRENCIES = (('USD', 'USD'), ('EUR', 'EUR'), ('CNY', 'CNY'), ('JPY', 'JPY'))
@@ -240,6 +241,13 @@ class User(AbstractBaseUser):
         """
         if self.is_meta:
             raise ProtectedError('Cannot remove meta user instances', None)
+
+        purge = kwargs.pop("purge", False)
+
+        if purge:
+            UserPurger(self).purge()
+        else:
+            UserMerger(self, User.objects.get_nobody_user()).merge()
 
         super(User, self).delete(*args, **kwargs)
 
--- a/pootle/apps/accounts/models.py
+++ b/pootle/apps/accounts/models.py
@@ -35,7 +35,7 @@ from pootle.core.cache import make_method_key
 from pootle.core.utils.json import jsonify
 from pootle_language.models import Language
 from pootle_statistics.models import Submission, SubmissionTypes
-from pootle_store.models import SuggestionStates
+from pootle_store.models import SuggestionStates, Unit
 
 from .managers import UserManager
 
@@ -320,8 +320,23 @@ class User(AbstractBaseUser):
         return 'https://secure.gravatar.com/avatar/%s?s=%d&d=mm' % \
             (self.email_hash, size)
 
+    def get_suggestion_reviews(self):
+        return self.submission_set.get_unit_suggestion_reviews()
+
     def get_unit_rows(self):
         return min(max(self.unit_rows, 5), 49)
+
+    def get_unit_states_changed(self):
+        return self.submission_set.get_unit_state_changes()
+
+    def get_units_created(self):
+        """Units that were created by this user.
+
+        :return: Queryset of `Unit`s that were created by this user.
+        """
+        return (Unit.objects
+                    .filter(pk__in=(self.submission_set.get_unit_creates()
+                                        .values_list("unit", flat=True))))
 
     def pending_suggestion_count(self, tp):
         """Returns the number of pending suggestions for the user in the given
--- a/pootle/apps/accounts/utils.py
+++ b/pootle/apps/accounts/utils.py
@@ -7,9 +7,19 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import functools
+import logging
+import sys
+
 from django.contrib.auth import get_user_model
 
 from allauth.account.models import EmailAddress
+
+from pootle_store.models import SuggestionStates
+from pootle_store.util import FUZZY, UNTRANSLATED
+
+
+logger = logging.getLogger(__name__)
 
 
 def get_user_by_email(email):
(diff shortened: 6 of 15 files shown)


pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,36 @@
...
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.contrib.auth import get_user_model
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import CommandError
+
+from . import UserCommand
+import accounts
```
unho: According to our styleguide this should be:

```python
import accounts

from . import UserCommand
```

pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,36 @@
...
+
+        User = get_user_model()
+        try:
+            src = User.objects.get(username=args[0])
+        except ObjectDoesNotExist:
+            raise CommandError("User %s does not exist" % args[0])
+
+        try:
+            target = User.objects.get(username=args[1])
+        except ObjectDoesNotExist:
```
unho: Is there any reason to not use `User.ObjectDoesNotExist`?

pootle/apps/accounts/management/commands/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+    def create_parser(self, prog_name, subcommand):
+        self.prog_name = prog_name
+        self.subcommand = subcommand
+        return super(UserCommand, self).create_parser(prog_name, subcommand)
+
+    def usage_string(self):
+        return self.usage(self.subcommand).replace('%prog', self.prog_name)
+
+    def check_args(self, *args):
+        if not len(args) == len(self.args.split(" ")):
```
unho: It seems `cls.args` is actually not used, so perhaps we should just use the number of args instead of splitting and counting the words.

pootle/apps/accounts/management/commands/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+            error = "Wrong number of arguments\n\n%s" % self.usage_string()
+            raise CommandError(error)
```
unho: I would prefer not to create unnecessary variables.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +30,38 @@ def get_user_by_email(email):
             return User.objects.get(email__iexact=email)
         except User.DoesNotExist:
             return None
+
+
+def merge_user(src, target):
+
```
unho: Please add a docstring. Future you will thank.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +30,38 @@ def get_user_by_email(email):
             return User.objects.get(email__iexact=email)
         except User.DoesNotExist:
             return None
+
+
+def merge_user(src, target):
```
unho: Please use `src_user` and `target_user` instead. They are more verbose and less prone to produce confussion.

pootle/apps/accounts/management/commands/__init__.py
phlax: Its used in generating self.usage()

pootle/apps/accounts/management/commands/__init__.py
phlax: hmmm that was me trying to be pep8 + style guide compliant

pootle/apps/accounts/utils.py
phlax: yep - ok

pootle/apps/accounts/utils.py
phlax: k

pootle/apps/accounts/management/commands/merge_user.py
phlax: no - ill update

pootle/apps/accounts/management/commands/__init__.py
unho: Where I can see that?

pootle/apps/accounts/management/commands/__init__.py
unho: I think the following complies with both:

```python
            raise CommandError("Wrong number of arguments\n\n%s" %
                               self.usage_string())
```

pootle/apps/accounts/management/commands/merge_user.py
unho: This happens in several places across the whole PR. Please adjust them as well.

tests/models/user.py
```diff
@@ -0,0 +1,63 @@
...
+    # Update store as user
+    _update_from_upload_file(store, file_name, user=user)
+
+    # Add a suggestion
+    unit = store.units.first()
+    unit.add_suggestion('SUGGESTION', user=user)
+    return unit
+
+
+def _test_user_merged(unit, src, target):
```
unho: Please use `src_user` and `target_user`.

tests/models/user.py
```diff
@@ -0,0 +1,63 @@
...
+    member.delete()
+    _test_user_merged(unit, member, nobody)
+
+
+@pytest.mark.django_db
+def test_purge_user(en_tutorial_po, member):
+    """Test purging user - the django default behaviour."""
+    _create_submission_and_suggestion(en_tutorial_po, member)
+    member.delete(purge=True)
+    assert en_tutorial_po.units.count() == 0
```
unho: I don't know what this assert means. I was actually expecting some asserts that the translation/suggestion/review... counts for `nobody` didn't change.

pootle/apps/accounts/management/commands/__init__.py
phlax: django.core.management.base.BaseCommand

tests/models/user.py
phlax: at this stage i didnt go further - because if you follow discussion in #3797, the functionality has not been decided

pootle/apps/accounts/management/commands/__init__.py
unho: Thanks.

pootle/apps/accounts/models.py
```diff
@@ -241,6 +242,19 @@ def delete(self, *args, **kwargs):
...
             raise ProtectedError('Cannot remove meta user instances', None)
 
+        if 'purge' in kwargs:
+            purge = kwargs.pop("purge")
+        else:
+            purge = False
+
+        if not purge:
+            try:
+                nobody = User.objects.get(username="nobody")
```
julen:
```python
nobody = User.objects.get_nobody_user()
```

docs/server/commands.rst
```diff
@@ -581,6 +581,50 @@ and the process will start a watchdog to track any client-side scripts for
...
+
+.. django-admin:: merge_user
+
+merge_user
+^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+If you have a user with two accounts or need to merge one account into another
+you can use `merge_user`. This will re-assign all submission, units and
```
unho: Release notes are mainly to leave users know about changes, so providing a quick reasoning why these commands where introduced would be really good. For example `purge_user` is to kick evil users :)

docs/releases/dev.rst
```diff
@@ -71,6 +71,10 @@ Command changes and additions
 - Added a :djadmin:`contributors` command to get the list of contributors
   (:issue:`3867`).
 
+- Added a :djadmin:`merge_user` command to get merge a user account to another
+
+- Added a :djadmin:`purge_user` command to purge a user's submissions
```
unho: Release notes are mainly to leave users know about changes, so providing a quick reasoning why these commands where introduced would be really good. For example `purge_user` is to kick evil users :)

docs/server/commands.rst
unho: I don't particularly like explanations for a command that refer to the command itself. Same below for the other command.

docs/server/commands.rst
```diff
@@ -581,6 +581,50 @@ and the process will start a watchdog to track any client-side scripts for
...
+^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+If you wish to permanently remove a user and their contributions from your site
+you can use `purge_user`. This will revert any submissions the user has made
+and delete the user object.
+
+This command requires a mandatory username argument, which should be a valid
+username for a user of your site
```
unho: End period?

pootle/apps/accounts/management/commands/__init__.py
unho: Why not use `!=` instead of `not ... ==`?

pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,32 @@
...
+        self.check_args(*args)
+        User = get_user_model()
+        try:
+            src_user = User.objects.get(username=args[0])
+        except User.DoesNotExist:
+            raise CommandError("User %s does not exist" % args[0])
+        try:
+            target_user = User.objects.get(username=args[1])
+        except User.DoesNotExist:
+            raise CommandError("User %s does not exist" % args[1])
```
unho: The following feels easier to understand (less repetitive):

```python
        def get_user(username):
            try:
                return User.objects.get(username=username)
            except User.DoesNotExist:
                raise CommandError("User %s does not exist" % username)

        src_user = get_user(args[0])
        target_user = get_user(args[1])
```

tests/data/po/tutorial/en/tutorial_update_evil.po
```diff
@@ -0,0 +1,19 @@
...
+"X-Pootle-Revision: 3\n"
+
+#: Hello, world
+msgid "Hello, world"
+msgstr "Hello, world EVIL"
+
+
+#: Goodbye, world
+msgid "Goodbye, world"
+msgstr "Goodbye, world EVIL"
```
unho: Please do not add files no end of line in the last line.

pootle/apps/accounts/management/commands/merge_user.py
unho: In fact it is better if you convert that `get_user` function into a `UserCommand` method since that way you can also simplify `purge_user` management command.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,106 @@ def get_user_by_email(email):
             return User.objects.get(email__iexact=email)
         except User.DoesNotExist:
             return None
+
+
+def merge_user(src_user, target_user):
+    """Merges one user to another
```
unho: End period missing.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,106 @@ def get_user_by_email(email):
...
+
+    The following are updated:
```
unho: The following fields and models are updated (model: list of fields):

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,106 @@ def get_user_by_email(email):
...
+    - units: submitted_by, commented_by, reviewed_by
+    - submissions: submitter
+    - suggestions: user, reviewer
+
+    :param src_user: `User` instance to merge from
+    :param target_user: `User` instance to merge to
+    """
+
+    # move units to target_user
```
unho: Please capitalize first letter of the comment. Lets write proper english since that makes non-english people lifes easier.

pootle/apps/accounts/utils.py
phlax: Titles do not have full stops

unho: Yes they do. Refer to our styleguide http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#docstrings and PEP 257 https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

tests/data/po/tutorial/en/tutorial_update_evil.po
phlax: added

phlax:
@dwaynebailey - i believe the this should take care of reviews and comments correctly now

@julen @unho - if you would be able to look at the logic in the added accounts.utils functions, that would be great. Ive added "NB:..." where i wasnt sure of the best query to use

pootle/apps/accounts/management/commands/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+    args = "user"
+
+    def create_parser(self, prog_name, subcommand):
+        self.prog_name = prog_name
+        self.subcommand = subcommand
+        return super(UserCommand, self).create_parser(prog_name, subcommand)
+
+    def get_user(self, username):
+        User = get_user_model()
```
unho: I suggest putting this line in the module (outside of `UserCommand` class), so it is assigned just once.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,180 @@ def get_user_by_email(email):
...
+    for unit in user.submitted.iterator():
+
+        # Find the last submission by different user that updated the
+        # unit.target.
+        # NB: not sure if this is the best way.
+        last_edit = (unit.submission_set
+                     .exclude(submitter=user)
+                     .exclude(new_value__isnull=True)
+                     .exclude(field=SubmissionFields.COMMENT)
+                     .filter(type__in=editing_types)
```
unho: If it is a revert it might not necessarily be a revert to a translation not done by `user`. I mean, if you have a unit with some translation by `user` and is then changed by some other user, and then reverted by some user to the translation done by `user`, you don't want to go back to that `user` translation, isn't it?

unho:
> @julen @unho - if you would be able to look at the logic in the added accounts.utils functions, that would be great. Ive added "NB:..." where i wasnt sure of the best query to use 

I only have concerns regarding the "NB:..." for `revert_units_edited_by`.

pootle/apps/accounts/utils.py
phlax: well the point is to revert anything that `user` has done

so we want to revert the edit to the last one done by `a_different_user`

Im not 100% about:
+                     .exclude(new_value__isnull=True)
+                     .exclude(field=SubmissionFields.COMMENT)
+                     .filter(type__in=editing_types)

as the best way to find the last change to target_f though - i think it works, but there's probably a cleaner way

unho: Since you are not excluding reverts to some `user` edit, you might actually end with the edit done by `user`. Please reread my previous comment.

phlax: it does - the previous line is:

.exclude(submitter=user)

phlax: sorry @unho i follow now - specifically a `revert` - let me think on that for a mo

unho: Also what happens if unit was created by `user` and then translated by other people? In that case you get `last_edit` but those should also be dropped since unit was created by `user`, isn't it?

phlax: As i understand it was decided not to do this - it would create too much risk of deleting used units.

Atm this code doesnt delete if there is **any** activity by a different user - that might not be the best test - ie maybe it should be any editing activity.

Its not easy to tell exactly what is set for each type of submission and the SubmissionType SuggestionField class variables are not easy to work with - but i did check the implementation of each of the things i was reverting - and added tests accordingly

The tests were written prior to the factoring out of the reversion functions - so could do with being separated also - but it would be good to get this landed for now

phlax: ...i also tested ttw - and tracked in pootle shell, incidentally

phlax:
> Since you are not excluding reverts to some user edit, you might actually end with the edit done by user. Please reread my previous comment.

As there is no implementation of REVERT afaict i have removed that from the `editing_types`.

If we have an implementation in the future we can consider adding this.

pootle/apps/accounts/utils.py
phlax: @unho 

> Since you are not excluding reverts to some user edit, you might actually end with the edit done by user. Please reread my previous comment.

As there is no implementation of REVERT afaict i have removed that from the `editing_types`.

If we have an implementation in the future we can consider adding this.

julen:
As already requested in other PRs, can we please avoid commits with multiple changes?

pootle/apps/accounts/models.py
```diff
@@ -241,6 +242,16 @@ def delete(self, *args, **kwargs):
...
             raise ProtectedError('Cannot remove meta user instances', None)
 
+        purge = kwargs.pop("purge", False)
+
+        if purge:
+            purge_user(self)
+        else:
+            try:
+                merge_user(self, User.objects.get_nobody_user())
+            except User.DoesNotExist:
```
julen: The `nobody` user is an integral part of Pootle, therefore this try/except block is redundant — if for some reason this exception happened, Pootle would terribly be broken.

phlax: in tests its not always present - which is why i had been concerned.

Ill add the fixture to the tests that require it - but it does mean that tests using User.delete might fail if they are missing the fixture

docs/releases/dev.rst
```diff
@@ -22,7 +22,7 @@ Major Changes
 
 - Updated translations.
 - Added django-transaction-hooks
-
+- Changed user delete behaviour
```
unho: This is pretty uninformative. Add more details for third party people to learn something useful.

unho: Also avoid having a diff that removes a blank line for no reason.

pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,21 @@
...
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from . import UserCommand
+import accounts
```
unho: These imports must be in the reverse order and with one blank line separating them.

pootle/apps/accounts/management/commands/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.core.management.base import BaseCommand, CommandError
+from django.contrib.auth import get_user_model
```
unho: These imports must be in the reverse order.

docs/releases/dev.rst
unho: Just noticed that this is expanded below. But the blank line concern still applies.

pootle/apps/accounts/utils.py
```diff
@@ -11,6 +11,10 @@
 
 from allauth.account.models import EmailAddress
 
+from pootle.core.models import Revision
+from pootle_store.models import SuggestionStates
+from pootle_statistics.models import SubmissionTypes, SubmissionFields
```
unho: `statistics` comes before `store`.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,179 @@ def get_user_by_email(email):
...
+    activity.
+
+    :param user: `User` to remove units for.
+    """
+    # Delete units created by user without submissions by others.
+    created_by_user = (user.submission_set
+                       .filter(type=SubmissionTypes.UNIT_CREATE))
+    for submission in created_by_user.iterator():
+        # Find submissions by other users on this unit.
+        # NB: not sure if this is the best way.
```
unho: Unnecessary comment.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,179 @@ def get_user_by_email(email):
...
+    just remove the comment.
+
+    :param user: `User` to remove submissions for.
+    """
+
+    # Revert unit comments where user is latest commenter.
+    for unit in user.commented.iterator():
+
+        # Find previous comment by different user.
+        # NB: not sure if this is the best way.
```
unho: Unnecessary comment.

tests/models/user.py
```diff
@@ -0,0 +1,221 @@
...
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.utils import timezone
+
+import pytest
+
+import accounts
+from pootle_store.models import SuggestionStates
+from pootle_statistics.models import (SubmissionFields,
```
unho: `statistics` comes before `store`. Please reverse import order.

unho:
lgtm on my side.

phlax:
@julen are you happy for this to land?

docs/server/commands.rst
```diff
@@ -636,6 +636,52 @@ and the process will start a watchdog to track any client-side scripts for
...
+
+.. versionadded:: 2.7.1
+
+This can be used if you have a user with two accounts and need to merge one
+account into another. This will re-assign all submissions, units and
+suggestions, but not any of the user's profile data.
+
+This command requires 2 mandatory arguments, ``src_username`` and
+``target_username``, both should be valid usernames for users of your site.
+Submissions from the first are re-assigned to the second. The users profile
```
julen: *user's*

pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,22 @@
...
+
+from . import UserCommand
+
+
+class Command(UserCommand):
+    args = "user other_user"
+    help = "Merge user to other_user"
+
+    def handle(self, *args, **kwargs):
+        self.check_args(*args)
```
julen: Since this is being called in every command inheriting from `UserCommand`, wouldn't it make sense to leave this responsibility to the base class, and call `super()` from actual commands?

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,177 @@ def get_user_by_email(email):
...
+
+def remove_units_created_by(user):
+    """Remove units created by user that have not had further
+    activity.
+
+    :param user: `User` to remove units for.
+    """
+    # Delete units created by user without submissions by others.
+    created_by_user = (user.submission_set
+                       .filter(type=SubmissionTypes.UNIT_CREATE))
```
julen: Are the parens to avoid a backslash for the line wrap? We are not using this style anywhere with querysets ([see example](https://github.com/translate/pootle/blob/56e70821d8865e441a34c4dfb546ccfcad228568/pootle/apps/pootle_store/models.py#L825-L826)). Either way, please align the `.filter()` vertically. Same applies to elsewhere in this PR.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,177 @@ def get_user_by_email(email):
...
+    The following are fields are updated (model: fields):
+    - units: submitted_by, commented_by, reviewed_by
+    - submissions: submitter
+    - suggestions: user, reviewer
+
+    :param src_user: `User` instance to merge from.
+    :param target_user: `User` instance to merge to.
+    """
+
+    # Move units to target_user.
```
julen: This doesn't *move* any units — it changes unit attribution for the denormalized unit fields.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,177 @@ def get_user_by_email(email):
...
+    # Revert unit target where user is the last submitter.
+    for unit in user.submitted.iterator():
+
+        # Find the last submission by different user that updated the
+        # unit.target.
+        # NB: not sure if this is the best way.
+        last_edit = (unit.submission_set
+                     .exclude(submitter=user)
+                     .exclude(new_value__isnull=True)
+                     .exclude(field=SubmissionFields.COMMENT)
```
julen: This can be error-prone if new fields are added in the future that should be excluded too. What about having a `.filter(field__in=[SubmissionFields.TRANSLATION_FIELDS])`, where `SubmissionFields.TRANSLATION_FIELDS = [<list of fields to consider>]`?

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,177 @@ def get_user_by_email(email):
...
+    editing_types = (SubmissionTypes.NORMAL,
+                     SubmissionTypes.SYSTEM,
+                     SubmissionTypes.UNIT_CREATE,
+                     SubmissionTypes.UPLOAD)
+
+    # Revert unit target where user is the last submitter.
+    for unit in user.submitted.iterator():
+
+        # Find the last submission by different user that updated the
+        # unit.target.
```
julen: I don't think this is about `unit.target` only: `unit.state` can change alone as well.

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +34,177 @@ def get_user_by_email(email):
...
+        # NB: not sure if this is the best way.
+        last_edit = (unit.submission_set
+                     .exclude(submitter=user)
+                     .exclude(new_value__isnull=True)
+                     .exclude(field=SubmissionFields.COMMENT)
+                     .filter(type__in=editing_types)
+                     .latest())
+        if last_edit:
+            unit.target_f = last_edit.new_value
```
julen: `last_edit.new_value` could be a change in state, so here the value of the new state is being written to the `target_f` field.

pootle/apps/accounts/management/commands/merge_user.py
phlax: The current classes have args and therefore use the check_args functionality - but i can definitely imagine other user commands that do not - so i would prefer to leave it as it is

pootle/apps/accounts/utils.py
phlax: fair point - ill change the wording

tests/core/views.py
```diff
@@ -268,7 +268,7 @@ def test_apiview_put(rf):
 
 
 @pytest.mark.django_db(transaction=True)
-def test_apiview_delete(rf):
+def test_apiview_delete(rf, trans_nobody):
```
julen: What's the purpose of this change?

tests/data/po/tutorial/en/tutorial_update_new_unit.po
```diff
@@ -7,7 +7,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: Pootle Tests\n"
 "X-Pootle-Path: /en/tutorial/tutorial.po\n"
-"X-Pootle-Revision: 1\n"
+"X-Pootle-Revision: 3\n"
```
julen: Is this change needed for this PR?

tests/core/views.py
phlax: test fails without creating a mutable nobody user

tests/models/user.py
```diff
@@ -0,0 +1,221 @@
...
+def _create_submission_and_suggestion(store, user,
+                                      filename=None, suggestion=None):
+    filename = filename or "tests/data/po/tutorial/en/tutorial_update.po"
+
+    # Update store as user
+    _update_from_upload_file(store, filename, user=user)
+
+    # Add a suggestion
+    unit = store.units.first()
+    unit.add_suggestion(suggestion or 'SUGGESTION', user=user)
```
julen: Instead of using `suggestion or 'SUGGESTION'`, setting the default parameter to the function as `suggestion='SUGGESTION'` is straightforward and natural. Same for `filename` above.

pootle/apps/accounts/utils.py
phlax: @julen can you suggest the best query for getting the last change that updated target_f?

pootle/apps/accounts/utils.py
phlax: again i really wasnt sure about this.

Reading the code for SubmissionFields would it not be better just to filter for SubmissionFields.TARGET to get the last change to target_f?

tests/models/user.py
```diff
@@ -0,0 +1,221 @@
...
+    last_revision = _test_revision_increased(store, last_revision)
+
+    # evil_member makes following changes:
+    #   - rejects member's suggestion on unit
+    #   - changes unit
+    #   - adds another suggestion on unit
+    #   - accepts their own suggestion
+    #   - adds a comment on unit
+    #   - adds another unit
+    _reject_suggestion(store.units.first().get_suggestions().first(),
```
julen: Why not use `Unit`'s `<reject|accept>_suggestion()` methods? Caching aside, the body of the functions are the same, so this is heading towards code duplication that can be avoided.

phlax: Thats a fair point - i think i was trying to establish how the review process worked when i copied out that code

pootle/apps/accounts/utils.py
phlax: maybe we need to check for last state change that was not made by user and update if different to current state?

pootle/apps/accounts/management/commands/merge_user.py
phlax: Actually on second thoughts perhaps it would be better to run that check in super - ill update

pootle/apps/accounts/utils.py
phlax: ok - ive used the alignment that you have suggested.

this is one of those formats that is hard to get used to as i have to manually type in spaces to make it look this way.

phlax: and yep - i avoid trailing backslash for line continuation as i believe its less readable and easier to make mistakes

pootle/apps/accounts/utils.py
phlax: Ive added a revert_unit_states_changed_by(user) function to take care of state changes

phlax:
> As already requested in other PRs, can we please avoid commits with multiple changes?

@julen apologies this PR/commit grew unexpectedly - i have separated it into multiple commits

phlax:
@julen @unho i have tidied this PR up and fixed tests - could you take a look when you have chance - ta.

julen:
Thank you @phlax, busy taking a look at it...

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,215 @@ def get_user_by_email(email):
...
+        # microseconds - so creation_time can be ambiguous
+        if submission != unit.state_changes.latest('pk'):
+            # If the unit has been changed more recently we don't need to
+            #  revert the unit state.
+            submission.delete()
+            return
+        submission.delete()
+        other_submissions = unit.state_changes.exclude(submitter=user)
+        if other_submissions.exists():
+            new_state = other_submissions.latest().new_value
```
phlax: this should also use latest("pk")

pootle/apps/pootle_store/models.py
```diff
@@ -333,6 +333,37 @@ class Meta:
...
+        return self.submission_set.filter(field=SubmissionFields.COMMENT)
+
+    @property
+    def suggestion_reviews(self):
+        """Submissions from this unit that review (reject/accept) suggestions.
+
+        :return: Queryset of `Submissions`s for this unit that
+            `REJECT`/`ACCEPT` `Suggestion`s
+        """
+        # reject_suggestion does not set field so we must exclude STATE reviews
```
julen: What is a state review? During reviews state can only change due to a positive review and when the unit was in fuzzy state before. Not sure if that's what you meant.

phlax: I think in my testing there was some submissions that had field set to `SubmissionFields.STATE` but had type in `SubmissionType.REVIEW_TYPES`

phlax: ...which needed to be excluded

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,215 @@ def get_user_by_email(email):
...
+        else:
+            new_state = UNTRANSLATED
+        if new_state != unit.state:
+            if unit.state == FUZZY:
+                unit.markfuzzy(False)
+            elif new_state == FUZZY:
+                unit.markfuzzy(True)
+            unit.state = new_state
+            unit._state_updated = True
```
julen: We probably want to add a note on why this is necessary. Ditto on the other occurrences of `unit._*_updated`

docs/server/commands.rst
```diff
@@ -636,6 +636,52 @@ and the process will start a watchdog to track any client-side scripts for
...
+^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+This can be used if you have a user with two accounts and need to merge one
+account into another. This will re-assign all submissions, units and
+suggestions, but not any of the user's profile data.
+
+This command requires 2 mandatory arguments, ``src_username`` and
+``target_username``, both should be valid usernames for user's of your site.
```
julen: *for users*

docs/server/commands.rst
```diff
@@ -636,6 +636,52 @@ and the process will start a watchdog to track any client-side scripts for
...
+Submissions from the first are re-assigned to the second. The user's profile
```
julen: *users'*

pootle/apps/accounts/management/commands/merge_user.py
```diff
@@ -0,0 +1,22 @@
...
+from . import UserCommand
+
+
+class Command(UserCommand):
+    args = "user other_user"
+    help = "Merge user to other_user"
+
+    def handle(self, *args, **kwargs):
+        super(Command, self).handle(*args, **kwargs)
+        accounts.utils.merge_user(self.get_user(username=args[0]),
```
julen: This errors, I think you need to adjust it to the new merger classes that have been added.

julen: Yep, in fact
```python
        accounts.utils.UserMerger(self.get_user(username=args[0]),
                                  self.get_user(username=args[1])).merge()
```

julen:
Great work overall @phlax!

Been testing this and there are some outstanding comments, in no particular order:

- The `merge_user` and `purge_user` commands give absolute no feedback. At least they should provide an informative message of the ongoing progress. Something like:
```bash
$ ./manage.py merge_user foo bar
Merging user foo into bar...
  * Merging submissions... DONE
  * Merging suggestions... DONE
...
```
As a bonus point, it might be handy to add some logging for debugging purposes too.

- The `merge_user` command takes really long time. I tried merging a user with many submissions into another user, and it took a few minutes (didn't measure it, but somewhere around 5 to 10). When we at Evernote merged a bunch of users using a snippet I shared earlier, the merges were fast compared to this command (the order of a few seconds at max).
Note this becomes especially problematic when *deleting* users via the web UI (i.e. it'll merge them with the *nobody* user), as it'll be run synchronously, and it's likely to cause timeout issues.

- When merging I think it'd be wise to merge user's properties as well. Of course, this is not desirable when merging contributions into *meta* users (mostly *nobody*), but in general I'd expect to merge user's personal information too. We did so when merging users as part of our move to Allauth. (cc @iafan)

- When deleting a user, the new default behavior will be to re-attribute its contributions to the *nobody* user. Just wanna double-check with you that this is in line with the UI message it's displayed in the admin page for users: *Note: deleting the user will make its suggestions and translations become attributed to an anonymous user (nobody).* In case it's not accurate enough, let's adjust it.

- I merged user `foo` into `bar`, and then ran `purge_user foo`. I'm not sure if my understanding is correct, but I'd expect `foo` to be unavailable at this stage; it wasn't. Running `purge_user foo` again gives the *expected* `CommandError: User foo does not exist` message.

- As requested the other day, can we please add comments for the `_*_updated` attributes so we know why they are actually necessary?

phlax:
> The merge_user and purge_user commands give absolute no feedback. 

Ill add some

> Note this becomes especially problematic when deleting users via the web UI (i.e. it'll merge them with the nobody user), as it'll be run synchronously

Yep - i can see that being a problem ttw. Perhaps we can run this as an async job in this case.

I think this approach makes far more db calls than eg just using django's delete collector with an object/attributes mapping to update - but i also think it provides more control/extensibility.

> When merging I think it'd be wise to merge user's properties as well.

I dont think that is what was specified - it may be better to rename this to merge_user_contributions to be explicit about what is being merged

> Note: deleting the user will make its suggestions and translations become attributed to an anonymous user (nobody). 

I think this is fairly accurate, but we may want to change slightly as its not just suggestions/translations that are updated

> I merged user foo into bar, and then ran purge_user foo. I'm not sure if my understanding is correct, but I'd expect foo to be unavailable at this stage

the UserPurger.purge method does not delete the user it just purges contributions - im guessing that is why this occurred.

the purge_user CLI command calls user.delete(purge=True) and so should take care of both purging and deleting the user

> As requested the other day, can we please add comments for the _*_updated attributes so we know why they are actually necessary?

I need to test again if they are - due to the mysql timestamp issue my tests were not reliable - so im not clear atm. Also even if they are necessary im also not clear why - the code is pretty messy around there - but i will endeavour to remove or explain inclusion

pootle/apps/accounts/utils.py
```diff
@@ -30,3 +33,245 @@ def get_user_by_email(email):
...
+        # Revert unit comments where self.user is latest commenter.
+        for unit in self.user.commented.iterator():
+
+            # Find comments by other self.users
+            comments = unit.get_comments().exclude(submitter=self.user)
+
+            if comments.exists():
+                # If there are previous comments by others update the
+                # translator_comment, commented_by, and commented_on
+                last_comment = comments.latest('pk')
```
phlax: @iafan - in reference to #4025:

here is an example of somewhere that ive had to use ordering by pk - when really i want to order by creation_time

i think its important to be able to order by creation time - because creation_time unlike pk can be manipulated. The fix that i made uses pk only as a secondary ordering when the creation_times match.

iafan: I fully agree with your reasoning. As for the implementation side, I'd leave that to Julen (I'm not a Python guy). Hope you can find some elegant solution to this.

pootle/apps/accounts/utils.py
```diff
@@ -7,10 +7,20 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import logging
+import functools
```
unho: `functools` after `logging`?

unho:
Just wondering if permissions should also be merged. They are obviously not contributions, but if we merge contributions and the user has no permissions to alter them perhaps we will get some support questions about this.

PR #4005, opened 2015-08-24
[RFC] Adding Dockerfile

phlax:
Adding a debian based Dockerfile to build Pootle 2.7.0 docker images from

Using docker would:

 - Allow us to create/test standardised distro
 - Allow easy testing of multiple versions and/or migrations
 - Ease installation for new/prospective users

This Dockerfile uses sqlite so is not production-focused - we could fairly
easily add mysql/postgres versions

I wasn't able to get refresh_stats to run synchronously so it has to be run at
first run. This is not ideal and we could run this step by hand if we were to
make a distro image from it.

I have not added any VOLUMES atm - but we would probably want to.

Some documentation would probably be useful too.


unho:
What is the purpose of adding RFC to the PR title? Does it mean it is a work in progress or it just means "Request For Comments" (which actually adds nothing since PRs are for that).

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+	  then \
+	       pootle refresh_stats_rq && \
+     	       mkdir var -p && \
+	       touch var/first-run; \
+	  fi) && \
+	 pootle runserver 0.0.0.0:8000 "
+
+
+EXPOSE 8000
+
```
unho: Is this blank line necessary?

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+FROM debian
+
+MAINTAINER Ryan Northey <ryan@synca.io>
+
+RUN apt-get update && \
+    apt-get install -y \
+            python-pip \
+	    libxml2-dev \
+	    libxslt-dev \
+	    python-dev \
```
unho: How do we ensure this is precisely Python 2.7?

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    useradd -m -d /home/pootle \
+            -k /etc/skel \
+            -s /bin/bash \
+            -g pootle pootle && \
+    /etc/init.d/redis-server start && \
+    sudo -u pootle bash -c "\
+    	 mkdir ~/dev/env -p && \
+	 cd ~/dev/env && \
+    	 virtualenv . && \
+    	 . bin/activate && \
```
unho: I suggest using `virtualenvwrapper` which allows to use `workon` (which IMHO is much cleaner). The less commands we run, the easier it is to understand this script and notice what is wrong.

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+CMD /etc/init.d/redis-server start && \
+    sudo -u pootle bash -c "\
+    	 cd ~/dev/env && \
+	 . bin/activate && \
+	 (pootle rqworker &) && \
+	 (if [ ! -e "var/first-run" ]; \
+	  then \
+	       pootle refresh_stats_rq && \
+     	       mkdir var -p && \
```
unho: Perhaps `mkdir -p var && \`? According to http://linux.die.net/man/1/mkdir options go before directories.

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    	 echo 'DEBUG=True' >> ~/.pootle/pootle.conf && \
+    	 pootle migrate --noinput && \
+    	 pootle initdb && \
+	 echo \"from django.contrib.auth import get_user_model; \	       
+	        from django.conf import settings; \
+	       	from django.core import management; \
+	       	from allauth.account.models import EmailAddress; \
+	       	user = get_user_model().objects.create_superuser('admin', 'admin@example.com', 'pootle'); \
+	       	EmailAddress.objects.create(user=user, email='admin@example.com', primary=True, verified=True) \" \
+	    | pootle shell"
```
unho: Is it necessary to run this code to create a superuser? Is it not possible to call https://docs.djangoproject.com/en/1.7/intro/tutorial02/#creating-an-admin-user ?

phlax:
I was adding WIP and RFC if the PR was not ready for merge, and to distinguish between something being worked on and something that needs feedback

Dockerfile
phlax: yep - either works ill - put it first as its clearer

Dockerfile
phlax: i dont believe so:
http://stackoverflow.com/questions/6244382/how-to-automate-createsuperuser-on-django

also even if you can - you still need to auto verify the email address to make it ready for login

```from django.conf import settings;``` 

does need to be removed though

File: Dockerfile (hunk 3 above)
Comment (phlax) [inline Dockerfile]: the less installed programs the better

if you were to build a dev Dockerfile then i would totally agree

File: Dockerfile (hunk 2 above)
Comment (phlax) [inline Dockerfile]: because that's what the current debian:latest/stable/jessie provides

we could  put ```FROM debian:jessie``` on line 5

File: Dockerfile (hunk 3 above)
Comment (unho) [inline Dockerfile]: I don't think it such a big deal. Another reason to use it is that we tell to use it in the docs, so the final result is closer to what we are used to, in case we need to provide support.

Comment (phlax) [inline Dockerfile]: thats a fair point

Comment (phlax) [inline Dockerfile]: actually - the docs dont use virtualenvwrapper:

http://pootle.readthedocs.org/en/stable-2.7.0/server/installation.html#setting-up-the-virtual-environment

Comment (unho) [inline Dockerfile]: ok. Sorry.

File: dockerfiles/2.7.0/Dockerfile [hunk 6]
```diff
@@ -0,0 +1,62 @@
...
+           -m \
+           -d /home/pootle \
+           -k /etc/skel \
+           -s /bin/bash \
+           -g pootle \
+         pootle \
+    && /etc/init.d/redis-server start \
+    && sudo -u pootle \
+         bash -c "\
+           mkdir ~/dev/env -p \
```
unho: Please use `mkdir -p DIR`. Check for other appearances as well.

dockerfiles/2.5.1/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+         && rm site.html"
+
+CMD sudo -u pootle \
+      bash -c "\
+        cd ~/dev/env \
+        && . bin/activate \
+        && pootle runserver --insecure 0.0.0.0:8000 "
+
+EXPOSE 8000
+
```
unho: Please remove final blank line.

dockerfiles/2.7.1-migration/data/home/pootle/.pootle/pootle.conf
```diff
@@ -0,0 +1,182 @@
...
+
+#
+# Backends
+#
+
+# Database backend settings
+DATABASES = {
+    'default': {
+        # Replace 'sqlite3' with 'postgresql_psycopg2', 'mysql' or 'oracle'.
+        'ENGINE': 'django.db.backends.sqlite3',
```
unho: No `transaction_hooks`?

dockerfiles/2.7.1-migration/data/home/pootle/.pootle/pootle.conf
```diff
@@ -0,0 +1,182 @@
...
+        # Database name or path to database file if using sqlite3.
+        'NAME': working_path('dbs/pootle.db'),
+        # Not used with sqlite3.
+        'USER': '',
+        # Not used with sqlite3.
+        'PASSWORD': '',
+        # Set to empty string for localhost. Not used with sqlite3.
+        'HOST': '',
+        # Set to empty string for default. Not used with sqlite3.
+        'PORT': '',
```
unho: On our default settings samples we append `'ATOMIC_REQUESTS': True,` as well.

dockerfiles/2.7.1-migration/data/home/pootle/.pootle/pootle.conf
phlax: no - beacause this starts as a 2.5.1 install

transaction hooks are injected during upgrade

dockerfiles/2.7.1-migration/data/home/pootle/.pootle/pootle.conf
phlax: not in 2.5 tho - ill add that in to the migration

julen:
I won't dive into the technical details of the PR, so just a general question: who will keep this file up-to-date, support any users it has and ensure it doesn't bitrot? Remember this already happened with fabric scripts.

phlax:
Fabric and docker are fundamentally different tools.

Although docker can be used for remote deployment - it is a way of defining and distributing a collection of software in a *clean* install environment - that can be run on almost any host platform - i believe even windows have added container support (albeit in a vm).

It is also useful for writing installation/migration docs - as the docs should match the steps taken in the docker recipe

For historical dockerfiles - they wont need to be maintained - they are just a build recipe that we have for reference, but are also useful if you want to test migration on a clean install from an historical build to a later one.

dwaynebailey:
@phlax lesson.  Make sure the computer you think is x86_64 is in fact that.  My issues is that it runs i686 i.e. 32 bit.  No go for docker.  Will try test on a 64bit machine tomorrow

PR #4010, opened 2015-08-25
Force suggestions to have a user

### LINKED ISSUE #3810: Suggestions should always have a user
It was proposed in #3651 to enforce `Suggestion.user` to be never `NULL`.

unho:
Fixes #3810.


julen:
Ship it.

PR #4011, opened 2015-08-26
Ensure email server is available

### LINKED ISSUE #3448: With no EMAIL_* users are still able to attempt to register
If no `EMAIL_*` settings are set then password reset will fail with a traceback.  Pootle should not issue a traceback, it should at least report on the admin interface that users can't register/reset passwords because no `EMAIL` is setup (though I thought it did this previously).

Reset password should be disabled if there is no `EMAIL_*` setup.

unho:
Fixes #3448.

### DIFF
--- a/pootle/apps/accounts/adapter.py
+++ b/pootle/apps/accounts/adapter.py
@@ -7,10 +7,16 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import logging
+
 from django.conf import settings
 from django.http import JsonResponse
+from django.utils.translation import ugettext_lazy as _
 
 from allauth.account.adapter import DefaultAccountAdapter
+
+
+logger = logging.getLogger('action')
 
 
 class PootleAccountAdapter(DefaultAccountAdapter):
@@ -47,3 +53,14 @@ class PootleAccountAdapter(DefaultAccountAdapter):
     def add_message(self, request, level, message_template, *args, **kwargs):
         """Silence messages altogether."""
         pass
+
+    def send_confirmation_mail(self, *args, **kwargs):
+        try:
+            super(PootleAccountAdapter, self).send_confirmation_mail(*args,
+                                                                     **kwargs)
+        except Exception:
+            logger.exception("ERROR: Sign up failed. Couldn't sent "
+                             "confirmation email.")
+            raise RuntimeError(_('Some problem happened when tried to send '
+                                 'the confirmation email. Please try again '
+                                 'later.'))
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -304,3 +304,26 @@ def check_db_transaction_on_commit(app_configs=None, **kwargs):
             id="pootle.C006",
         ))
     return errors
+
+
+@checks.register()
+def check_email_server_is_alive(app_configs=None, **kwargs):
+    from django.conf import settings
+
+    errors = []
+    if settings.POOTLE_SIGNUP_ENABLED or settings.POOTLE_CONTACT_ENABLED:
+        from django.core.mail import get_connection
+
+        connection = get_connection()
+        try:
+            connection.open()
+        except Exception:
+            errors.append(checks.Warning(
+                _("Email server is not available."),
+                hint=_("Review your email settings and make sure your email "
+                       "server is working."),
+                id="pootle.W004",
+            ))
+        else:
+            connection.close()
+    return errors
--- a/pootle/settings/91-travis.conf
+++ b/pootle/settings/91-travis.conf
@@ -37,6 +37,7 @@ if os.environ.get("TRAVIS"):
         DATABASES['default']['USER'] = 'travis'
 
     SILENCED_SYSTEM_CHECKS = [
+        'pootle.W004',  # Pootle requires a working mail server
         'pootle.W005',  # DEBUG = True
         'pootle.W010',  # DEFAULT_FROM_EMAIL has default setting
         'pootle.W011',  # POOTLE_CONTACT_EMAIL has default setting


pootle/checks.py
```diff
@@ -276,3 +276,21 @@ def check_db_transaction_on_commit(app_configs=None, **kwargs):
...
+    from django.core.mail import get_connection
+
+    errors = []
+    connection = get_connection()
+
+    try:
+        connection.open()
+    except Exception:
+        errors.append(checks.Critical(
+            _("Email server not available."),
```
julen: I'd add the verb to keep it consistent with other messages: *is not available*.

julen:
GTM.

phlax:
Im not sure if we want to make having an email server available a CRITICAL check

I think instead we need to disable the registration/login functionality if its not

unho:
A regular Pootle server allows new users to sign up and by default it allows any user (even anonymous) to contact the admins (this can be disabled). So forcing to have a working mail server doesn't look so excessive to me, moreover since this will avoid us having to provide support for users that inexplicably can't display sign up on their servers.

phlax:
Making this a hard requirement means that you cannot run the server in an isolated env with no access to external resources like email servers - for example when testing

It will make it harder for people to set up

It will make the site **dependent** on an external service - or rather the availability of that service at the point of the server starting

Crucially - this fix will not deal with the situation where an email server is available when the server starts - but becomes unavailable during runtime

unho:
> Making this a hard requirement means that you cannot run the server in an isolated env with no access to external resources like email servers - for example when testing

> It will make the site **dependent** on an external service - or rather the availability of that service at the point of the server starting

You can use:

```python
EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
```

> Crucially - this fix will not deal with the situation where an email server is available when the server starts - but becomes unavailable during runtime

Yes. That is right. Thinking.

phlax:
@unho generally i make anything that communicates with a system that i dont have control over, and where feasible, into an async job.

In the case of email - this usually pays off because notification emails can take a while to deliver - esp if there are a few

The other advantage of using an async job is that if it fails - it sits in the queue - not lost

dwaynebailey:
A Crtitical check would mean you can't actually start the server.  This is not in the same league as an unavailable DB or Redis being unavailable.  The server can run without email server even though its important for a live server.  It should be a warning which would allow the server to at least start.

unho:
@phlax The problem is here is that if there is no email server available, then there is no way to make to the queue.

unho:
Downgraded from critical to warning.

phlax:
this will still slow down server start up

How about we focus on the registration/login problem

julen:
> this will still slow down server start up

One thing to note is that as opposed to `runserver`, Django's [checks are not implicitly run on production](https://docs.djangoproject.com/en/1.7/topics/checks/#system-check-framework) as part of the WSGI request/response cycle.

unho:
Updated so now you see a message telling you that signup can't be completed because the email server is unavailable.

@julen I tried to adjust `PootleAccountAdapter` first since that felt the right way to do it, but I am not particularly sure about that approach.

julen:
I don't understand why this check has to be run when submitting the sign up form — it's unnecessarily complicated, the system check works just fine. Even if the mail server can go down anytime, since #4013 we already display other exceptions in the UI.

In this particular case, I assume the types of errors an unavailable mail server can generate are Greek to the user, and it'd be fine to error out with a friendlier message such as the one you have included (*Email server is currently unavailable. Please try again later*).

In order to achieve that you don't need to create any custom views, nor forms, and not even check for the availability of the mail server the same way a system check does: just override the [`send_confirmation_email()`](https://github.com/pennersr/django-allauth/blob/20cd2a6fc922972056c26a4bce894254a85a476f/allauth/account/adapter.py#L338) method from the account adapter, handle any exceptions that denote the server being down, and re-raise with any custom messages.

unho:
> I don't understand why this check has to be run when submitting the sign up form — it's unnecessarily complicated, the system check works just fine. Even if the mail server can go down anytime, since #4013 we already display other exceptions in the UI.

It seems it doesn't work since I can't see any error displayed in the form. I made sure to get the latest master and the right assets.

> In order to achieve that you don't need to create any custom views, nor forms, and not even check for the availability of the mail server the same way a system check does: just override the [`send_confirmation_email()`](https://github.com/pennersr/django-allauth/blob/20cd2a6fc922972056c26a4bce894254a85a476f/allauth/account/adapter.py#L338) method from the account adapter, handle any exceptions that denote the server being down, and re-raise with any custom messages.

I tried that unsuccessfully but for some reason the errors are not being displayed in the UI. Moreover the registration is created but no email is sent where the right behavior here would be to create the registration and send the email or do nothing.

julen:
> I tried that unsuccessfully but for some reason the errors are not being displayed in the UI

Can you share the code?

> the right behavior here would be to create the registration and send the email or do nothing.

This is something you might wanna report to allauth.

unho:
>> I tried that unsuccessfully but for some reason the errors are not being displayed in the UI

> Can you share the code?

Here it is https://github.com/translate/pootle/compare/translate:master...unho:try-account-adapter

>> the right behavior here would be to create the registration and send the email or do nothing.

> This is something you might wanna report to allauth.

Added a new comment to https://github.com/pennersr/django-allauth/issues/1110

phlax:
Ideally it would also give the user or at very least the admin the ability to see and resend failed confirmation emails.

unho:
> Ideally it would also give the user or at very least the admin the ability to see and resend failed confirmation emails.

The user is not registered so won't be able to do anything. What you say about admins is true, please leave your comment on https://github.com/pennersr/django-allauth/issues/1110

phlax:
> The user is not registered so won't be able to do anything.

No - but on failure it could give the user a button "try resending confirmation email"

IMHE email servers go down/become unavailable fairly frequently and are often available again soon after

unho:
> No - but on failure it could give the user a button "try resending confirmation email"

In the current PR status they can resend the whole signup form which will trigger the whole registration process from scratch. That button can be something that allauth people might want to provide, but I have my doubts about it.

julen:
@unho the code works, the only thing I had to change to make it show up in the UI is setting `DEBUG=False` (exceptions are logged to the console in `DEBUG` mode, and the response cycle is abruptly aborted).

Also, this logic makes more sense in `send_mail()` which is used generally, as opposed to `send_confirmation_email()`.

unho:
Updated again. Please review.

The checks is now only run if either contact or signup are enabled. I am not sure about if this restriction is ok, and that is the reason why I didn't directly land that particular commit.

pootle/apps/accounts/adapter.py
```diff
@@ -47,3 +48,11 @@ def is_open_for_signup(self, request):
...
         """Silence messages altogether."""
         pass
+
+    def send_confirmation_mail(self, *args, **kwargs):
+        try:
+            super(PootleAccountAdapter, self).send_confirmation_mail(*args,
+                                                                     **kwargs)
+        except Exception:
+            raise Exception(_('Some problem happened when tried to send the '
+                              'confirmation email. Please try again later.'))
```
phlax: Could we capture and log the exception - so sysadmins can find out that its happened

pootle/apps/accounts/adapter.py
```diff
@@ -47,3 +48,11 @@ def is_open_for_signup(self, request):
     def add_message(self, request, level, message_template, *args, **kwargs):
         """Silence messages altogether."""
         pass
+
+    def send_confirmation_mail(self, *args, **kwargs):
+        try:
+            super(PootleAccountAdapter, self).send_confirmation_mail(*args,
+                                                                     **kwargs)
+        except Exception:
+            raise Exception(_('Some problem happened when tried to send the '
```
phlax: I think make this a RuntimeError

unho: I don't see any problem catching any kind of exception here. If it works ok no exception of any kind is raised.

phlax: yep - sorry - i meant make the error that is raised a RuntimeError

phlax:
@unho i tried running the fix, but the UI still hangs, and it still errors out in the logs

pootle/apps/accounts/adapter.py
```diff
@@ -47,3 +48,12 @@ def is_open_for_signup(self, request):
...
         pass
+
+    def send_confirmation_mail(self, *args, **kwargs):
+        try:
+            super(PootleAccountAdapter, self).send_confirmation_mail(*args,
+                                                                     **kwargs)
+        except Exception:
+            raise RuntimeError(_('Some problem happened when tried to send '
+                                 'the confirmation email. Please try again '
+                                 'later.'))
```
phlax: I think we should comunicate that their signup was successful - but just not sending conf email

Otherwise they might try to sign up again - but they already have an account

unho:
To test this succesfully:

- Set `DEBUG = False`
- Make sure the email server in settings does not exist (to ensure email cannot be sent)
- `pootle runserver --insecure`
- Try to sign up a new user
- It should display an error message instead of look like nothing happened after clicking on "Sign Up" button

pootle/apps/accounts/adapter.py
phlax: Also - please add some logging or sysadmins will never know it happened

unho: Logging added.

unho: After the sign up form is successfully sent (and confirmation email is sent) the user is presented a new window telling them:

```
Your account needs activation.

We have sent an email containing the special link to
YOUR_EMAIL

Please follow that link to continue the account creation.
```

pootle/apps/accounts/adapter.py
```diff
@@ -47,3 +50,13 @@ def is_open_for_signup(self, request):
     def add_message(self, request, level, message_template, *args, **kwargs):
         """Silence messages altogether."""
         pass
+
+    def send_confirmation_mail(self, *args, **kwargs):
+        try:
+            super(PootleAccountAdapter, self).send_confirmation_mail(*args,
+                                                                     **kwargs)
+        except Exception:
+            log("\n\n\nSign up failed. Couldn't sent confirmation email.\n\n")
```
phlax: do you not want to do logger.error(e.message) - that way the sysadmin can see the exact failure

That is why i asked for it - because we swallow the raised exception and raise a different one - i think the original exception should be logged

pootle/apps/accounts/adapter.py
phlax: sorry - my bad - i was thinking that the account was still created but the email wasnt sent.

pootle/apps/accounts/adapter.py
unho: Done.

PR #4014, opened 2015-08-28
Adding --db flag to "pootle init"

phlax:
By default running pootle init a configuration for sqlite is created.

This is useful as it allows a user to get up and running quickly.

However, sqlite is "not supported" so this commit adds a "--db" flag
to allow the user to specify "mysql" or "postgresql"

### DIFF
--- a/pootle/runner.py
+++ b/pootle/runner.py
@@ -30,7 +30,7 @@ if sys.version_info[0] < 3:
     input = raw_input
 
 
-def init_settings(settings_filepath, template_filename):
+def init_settings(settings_filepath, template_filename, db="sqlite"):
     """Initializes a sample settings file for new installations.
 
     :param settings_filepath: The target file path where the initial settings
@@ -49,6 +49,20 @@ def init_settings(settings_filepath, template_filename):
     # We can't use regular python string formatting here.
     output = output.replace("${default_key}",
                             b64encode(os.urandom(KEY_LENGTH)).decode("utf-8"))
+
+    db_module = {
+        'sqlite': 'sqlite3',
+        'mysql': 'mysql',
+        'postgresql': 'postgresql_psycopg2',
+        }[db]
+
+    output = output.replace("'ENGINE': 'transaction_hooks.backends.sqlite3'",
+                            "'ENGINE': 'transaction_hooks.backends.%s'" % db_module)
+
+    if db != "sqlite":
+        output = output.replace("'NAME': working_path('dbs/pootle.db')",
+                                "'NAME': ''")
+
     fp.write(output)
     fp.close()
 
@@ -106,6 +120,8 @@ def run_app(project, default_settings_path, settings_template,
     parser.add_argument("--noinput", action="store_true", default=False,
                         help=u"Never prompt for input")
     parser.add_argument("--version", action="version", version=get_version())
+    parser.add_argument("--db", default="sqlite",
+                        help=u"Use the specified database backend")
 
     args, remainder = parser.parse_known_args(sys.argv[1:])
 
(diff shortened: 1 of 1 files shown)


phlax:
I havent added any docs because "pootle init" is currently undocumented

pootle/runner.py
```diff
@@ -124,13 +128,24 @@ def run_app(project, default_settings_path, settings_template,
                 print("File already exists, not overwriting.")
                 exit(2)
 
+        if args.db in ["mysql", "postgresql"]:
+            settings_template = "%s.%s" % (settings_template, args.db)
+        elif args.db != "sqlite":
+            raise management.CommandError("Unrecognised database '%s': should "
+                                          "be one of sqlite, mysql or "
+                                          "postgresql" % args.db)
```
unho: I would recommend enclosing the acceptable DB engines between ' quotes.

pootle/runner.py
```diff
@@ -124,13 +128,24 @@ def run_app(project, default_settings_path, settings_template,
                 print("File already exists, not overwriting.")
                 exit(2)
 
+        if args.db in ["mysql", "postgresql"]:
+            settings_template = "%s.%s" % (settings_template, args.db)
```
unho: Do we actually need to create two new sample settings files? I mean, if only the database setting is being altered why not replace it in the default template?

pootle/runner.py
```diff
@@ -124,13 +128,24 @@ def run_app(project, default_settings_path, settings_template,
...
         try:
             init_settings(config_path, settings_template)
         except (IOError, OSError) as e:
             raise e.__class__('Unable to write default settings file to %r'
                 % config_path)
 
-        print("Configuration file created at %r" % config_path)
+        if args.db in ["mysql", "postgresql"]:
+            print("Configuration file created at %r: you must now update "
+                  "the settings for %s database" % (config_path, args.db))
```
unho: Perhaps it is better to also ask for those details to be passed as arguments to `pootle init`. That way this can be completely automated.

dwaynebailey:
Maintenance nightmare.  Rather just rewrite the DATABASES section for the other DBs.  With this approach we have to maintain N sample/refernce configs instead of one, keeping them all in sync.

One way to achieve that would be to mark the reference on with a #DATABASE# section and keep other DBs mysql etc in a mysql.snippet. And do a replace when writing that file.

pootle/runner.py
phlax: I havent added this at this stage - but agree it would be a good feature

phlax:
@dwaynebailey - cool - rewritten so it overwrites the 1 template

julen:
What @dwaynebailey said, this is a nightmare to maintain.

pootle/runner.py
```diff
@@ -49,6 +49,14 @@ def init_settings(settings_filepath, template_filename):
     # We can't use regular python string formatting here.
     output = output.replace("${default_key}",
                             b64encode(os.urandom(KEY_LENGTH)).decode("utf-8"))
+    if db != "sqlite":
+        if db == "mysql":
+            db = "mysql"
+        elif db == "postgresql":
+            db = "postgresql_psycopg2"
+        output.replace("'ENGINE': 'transaction_hooks.backends.sqlite3'",
```
julen: I'd highly prefer not to replace an actual value, but a placeholder as we do with `SECRET_KEY` — less error-prone, no need to change values in multiple places. Also, `str.replace()` doesn't work in-place, it returns a modified copy.

```python
db_module = {
  'sqlite': 'sqlite3',
  'mysql': 'mysql',
  'postgresql': 'postgresql_psycopg2',
}[db]

output = output.replace('${db_module}', db_module)
```

unho:
lgtm.

PR #4020, opened 2015-08-30
Add verify_user command

phlax:
This will allow a sysadmin to verify a user account on the command line without the user having to go through the normal email verification process.

This is useful when creating a superuser account and also for migrating a site with already verified users

The command can be called with either a username or the --all flag to verify all users

Depends on PR #4002 

Tests added and docs updated

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -682,6 +682,31 @@ username for a user of your site.
     $ pootle purge_user username
 
 
+.. django-admin:: verify_user
+
+verify_user
+^^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+Verify a user without the user having to go through email verification process.
+
+This is useful if you are migrating users that have already been verified, or
+if you want to create a superuser that can log in immediately.
+
+This command requires either a mandatory ``username`` argument, which should be a
+valid username for a user of your site, or the :option:`--all` flag if you wish to
+verify all users of your site.
+
+.. code-block:: bash
+
+    $ pootle verify_user username
+
+
+:option:`--all`
+  Verify all users of the site
+
+
 .. _commands#running:
 
 Running WSGI servers
--- a/pootle/apps/accounts/management/commands/__init__.py
+++ b/pootle/apps/accounts/management/commands/__init__.py
@@ -37,6 +37,9 @@ class UserCommand(BaseCommand):
         return self.usage(self.subcommand).replace('%prog', self.prog_name)
 
     def check_args(self, *args):
-        if len(args) != len(self.args.split(" ")):
+        min_args = len([a for a in self.args.split(" ")
+                        if not a.startswith("[")])
+        max_args = len(self.args.split(" "))
+        if not (min_args <= len(args) <= max_args):
             raise CommandError("Wrong number of arguments\n\n%s"
                                % self.usage_string())
--- a/pootle/apps/accounts/management/commands/verify_user.py
+++ b/pootle/apps/accounts/management/commands/verify_user.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from optparse import make_option
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import CommandError
+
+from . import UserCommand
+import accounts
+
+
+class Command(UserCommand):
+    args = "[user]"
+    help = "Verify a user of the system without requiring email verification"
+    shared_option_list = (
+        make_option('--all', dest='all', action="store_true",
+                    help='Verify all users'),
+    )
+    option_list = UserCommand.option_list + shared_option_list
+
+    def handle(self, *args, **kwargs):
+        self.check_args(*args)
+        verify_all = kwargs.get("all")
+
+        # Either [user] OR --all should be set
+        both_or_neither = ((args and verify_all)
+                           or (not args and not verify_all))
+        if both_or_neither:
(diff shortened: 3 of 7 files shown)


unho:
Is the first commit the same as in PR #4002?

docs/server/commands.rst
```diff
@@ -636,6 +636,71 @@ and the process will start a watchdog to track any client-side scripts for
...
+^^^^^^^^^^^
+
+.. versionadded:: 2.7.1
+
+Verify a user without the user having to go through email verification process.
+
+This is useful if you are migrating users that have already been verified, or
+if you want to create a superuser that can log in immediately.
+
+This command requires either a mandatory username argument, which should be a
```
unho: Please highlight `username`.

docs/server/commands.rst
```diff
@@ -636,6 +636,71 @@ and the process will start a watchdog to track any client-side scripts for
...
+valid username for a user of your site, or the `--all` flag if you wish to
```
unho: Please highlight `--all` using `:option:`.

pootle/apps/accounts/management/commands/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+            return User.objects.get(username=username)
+        except User.DoesNotExist:
+            raise CommandError("User %s does not exist" % username)
+
+    def get_real_users(self):
+        User = get_user_model()
+        system = User.objects.get_system_user()
+        nobody = User.objects.get_nobody_user()
+        default = User.objects.get_default_user()
+        return User.objects.exclude(pk__in=[system.pk, nobody.pk, default.pk])
```
unho: This actually looks like something that belongs to `User` manager.

tests/fixtures/models/user.py
```diff
@@ -59,6 +59,21 @@ def admin(transactional_db):
...
-def member(db):
+def member(transactional_db):
     """Require a member user."""
     return _require_user('member', 'Member')
+
+
+@pytest.fixture
+def member_with_email(transactional_db):
+    """Require a member user."""
+    user = _require_user('member', 'Member')
```
unho: Please use a different username for this test user.

tests/models/user.py
```diff
@@ -0,0 +1,195 @@
...
+    # Verify user
+    accounts.utils.verify_user(member_with_email)
+
+    # Get the verified email object
+    EmailAddress.objects.get(user=member_with_email, email="member@this.test",
+                             primary=True, verified=True)
+
+
+@pytest.mark.django_db
+def test_verify_user_without_existing_email(member):
```
unho: I think it is preferable to create a new user fixture instead of reusing `member`.

pootle/apps/accounts/management/commands/verify_user.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # Either [user] OR --all should be set
+        both_or_neither = ((args and kwargs.get("all"))
+                           or (not args and not kwargs.get("all")))
+        if both_or_neither:
+            raise CommandError("You must either provide a [user] to verify or "
+                               "use '--all' to verify all users\n\n%s"
+                               % self.usage_string())
+
+        if kwargs.get("all"):
```
unho: You are getting it again so why not cache it.

tests/models/user.py
phlax: `member` is used extensively in the tests - and is kinda why it was chosen

unho: Yes, but in previous tests it wasn't changed, just used.

docs/server/commands.rst
```diff
@@ -636,6 +636,75 @@ and the process will start a watchdog to track any client-side scripts for
...
+
+.. versionadded:: 2.7.1
+
+Verify a user without the user having to go through email verification process.
+
+This is useful if you are migrating users that have already been verified, or
+if you want to create a superuser that can log in immediately.
+
+This command requires either a mandatory `username` argument, which should be a
+valid username for a user of your site, or the `--all` flag if you wish to
```
unho: Please highlight `--all` using `:option:`.

tests/models/user.py
phlax: ok - ill use a `different_member` for transactional tests

docs/server/commands.rst
phlax: its there on l704

pootle/apps/accounts/managers.py
```diff
@@ -65,6 +65,12 @@ def get_default_user(self):
     def get_nobody_user(self):
         return super(UserManager, self).get_queryset().get(username='nobody')
 
+    def get_real_users(self):
+        return (super(UserManager, self).get_queryset()
+                .exclude(pk__in=[self.get_system_user().pk,
+                                 self.get_nobody_user().pk,
+                                 self.get_default_user().pk]))
```
unho: Please use instead this syntax to align with the rest of the file:

```python
        return super(UserManager, self).get_queryset().exclude(
            pk__in=[
                self.get_system_user().pk,
                self.get_nobody_user().pk,
                self.get_default_user().pk,
            ]
        )
```

docs/server/commands.rst
unho: Yes, but it is not being highlighted in this line.

phlax: k - understood

julen:
I honestly don't think this functionality and command belongs to Pootle, but allauth — I'd consider moving this PR over there.

phlax:
@julen perhaps it should - but for 2.7.1 release we are trying to ease the pain of migration - so it would be good to land

julen:
Fair enough. Consider doing so afterwards though.

julen:
I can read that *This is useful if you are migrating users that have already been verified, or if you want to create a superuser that can log in immediately.*. I couldn't recommend this command for the first use-case, but rather a data migration.

dwaynebailey:
lgtm from me.  Though can we please add a caveat in the command description so that people know the consequence i.e. that the users that they've migrated are actually now not verified.

PR #4025, opened 2015-09-03
Change Submission ordering to [creation_time, pk]

phlax:
Ordering submissions by creation_time alone is unreliable on mysql as it only has precision to the second and there is a risk of multiple submissions having the same time

### DIFF
--- a/pootle/apps/pootle_statistics/migrations/0002_update_submission_ordering.py
+++ b/pootle/apps/pootle_statistics/migrations/0002_update_submission_ordering.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_statistics', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='submission',
+            options={'ordering': ['creation_time', 'pk'], 'get_latest_by': 'creation_time'},
+        ),
+    ]
--- a/pootle/apps/pootle_statistics/models.py
+++ b/pootle/apps/pootle_statistics/models.py
@@ -84,7 +84,41 @@ class TranslationActionTypes(object):
     NEEDS_WORK = 5
 
 
-class SubmissionManager(models.Manager):
+class SubmissionQuerySet(models.QuerySet):
+
+    def _earliest_or_latest(self, field_name=None, direction="-"):
+        """
+        Overrides QuerySet._earliest_or_latest to add pk for secondary ordering
+        """
+        order_by = field_name or getattr(self.model._meta, 'get_latest_by')
+        assert bool(order_by), "earliest() and latest() require either a "\
+            "field_name parameter or 'get_latest_by' in the model"
+        assert self.query.can_filter(), \
+            "Cannot change a query once a slice has been taken."
+        obj = self._clone()
+        obj.query.set_limits(high=1)
+        obj.query.clear_ordering(force_empty=True)
+        # add pk as secondary ordering for Submissions
+        obj.query.add_ordering('%s%s' % (direction, order_by),
+                               '%s%s' % (direction, "pk"))
+        return obj.get()
+
+    def earliest(self, field_name=None):
+        return self._earliest_or_latest(field_name=field_name, direction="")
+
+    def latest(self, field_name=None):
+        return self._earliest_or_latest(field_name=field_name, direction="-")
+
+
+class BaseSubmissionManager(models.Manager):
+
+    def get_queryset(self):
+        return SubmissionQuerySet(self.model, using=self._db)
+
+
+class SubmissionManager(BaseSubmissionManager):
+
+    use_for_related_fields = True
 
     def get_queryset(self):
         """Mimics `select_related(depth=1)` behavior. Pending review."""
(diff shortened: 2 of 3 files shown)


pootle/apps/pootle_statistics/migrations/0002_auto_20150903_1841.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
dwaynebailey: Please name migration files using a human readable name to describe what its doing, rather then the default timestamp

julen:
Can you justify why you think patching Django's internals is a good idea? Likewise, how did you end up doing so?

FWIW I'm -1 on the approach — it's hackish and error-prone, as it deals with private undocumented APIs that are subject to change without further notice.

phlax:
@julen 
the reasoning for doing this is that mysql timestamps use unix timestamps without microseconds.

This means that when you do submissions.earliest()/latest() on a set that has 2 submissions with the same timestamp - you will not get the correct submission.

Afaict overriding the latest/earliest methods in the queryset is the only way to add a secondary ordering field - using the _earliest_or_latest method was far less code than overriding both the earliest/latest methods.

Im surprised that you are -1 as you are using mysql and have a large busy site which is where the problem is mostly likely to occur.

If we dont patch this then the only reliable way i can see to get the earliest/latest submissions is to not use creation_time at all - ie submissions.latest("pk")

iafan:
@phlax, can you shed some more light on where we need this sorting to work?

If this is for some internal sync/export algorithms, then using dates is never a good idea, despite the date resolution. Sequential change ids for submissions is what should be used instead to get the fault-tolerant sorting.

If this is only to sort units in the UI, then sorting by both date and private key is a way to go, but I'm not sure if it's worth the trouble of messing with internals, as Julen says. Does Django allow one to get results sorted by two properties without having to do black magic?

phlax:
@iafan at the moment anywhere that order submissions is prone to getting them out of order if there are 2 with the same timestamp - so that could be eg comments in the ui - or trying to find the last state change/edit in the backend.

The problem is likely to be a very rare one - but as i discovered writing tests - it gives unpredictable results and is hard to diagnose.

I wouldnt really call using custom queryset black magic - its a pretty well documented path:

https://docs.djangoproject.com/en/1.8/topics/db/managers/#calling-custom-queryset-methods-from-the-manager

i do take the point about using an internal method - ie _earliest_or_latest - but even if i had copied out the public earliest/latest methods this is what they would look like:

https://github.com/django/django/blob/master/django/db/models/query.py#L538:L542

so i figured it was ok to just override the one method

iafan:
> at the moment anywhere that order submissions is prone to getting them out of order if there are 2 with the same timestamp - so that could be eg comments in the ui - or trying to find the last state change/edit in the backend.

My main point here is that if this sorting is used in a backend, we'd better move away from using timestamps for sorting altogether.

For pure UI-related sorting, getting "unpredictable" results for events done within the datetime field resoultion is fine.

dwaynebailey:
@iafan this was uncovered when testing the deletion of user submissions.  In this case we are wanting to restore the comments, state and translation that existed before the evil user made contributions.  So we don't have revision numbers in a number of those cases (comments and states).  But we do have timestamps.

The timestamps work fine except on MySQL.  Where as @phlax said it has second resolution.  Only MySQL 5.6.N (where I can't remember N at the moment) bring in millisecond timestamps.  So any rapid submission would make us unable to sort.  And I don't see us wanting or even able to raise our required MySQL version to that.  PostgreSQL doens't have this issue as timestamps have much higher resolution.

We could put revisions onto these other change type, but that seems like overkill for this case.  It might be the right long terms approach though. But to land merging users and deleting users we need this to allow the feature to work properly.

iafan:
Thanks for the detailed explanation, @dwaynebailey!

Using timestamps is extremely error-prone. For example, the server clock can go off by several seconds and then adjusted by ntpd at random times. And it doesn't matter how good the timestamp resolution on your SQL server is — the algorithm will fail.

So I believe using revisions is not an overkill — it's a necessity. I understand you wanted to implement some quick fix that lowers the probability of the error, but I guess our ultimate goal is to make fault-tolerant algorithms, not probablilistic ones.

iafan:
Having said that, please don't treat this as -1 for the PR from me. I'm personally ok if you treat this as a mid-term solution, and just want to emphasize how brittle this approach is.

unho:
I must say that I neither like that hacking into private API methods.

unho:
> i do take the point about using an internal method - ie _earliest_or_latest - but even if i had copied out the public earliest/latest methods this is what they would look like:
> 
> https://github.com/django/django/blob/master/django/db/models/query.py#L538:L542

I don't think overriding those two public methods is such a big problem. As I previously stated fiddling with a private interface doesn't feel the best way to extend the behavior.

phlax:
cool @unho ill add them in then

pootle/apps/pootle_statistics/migrations/0002_auto_20150903_1841.py
unho: This is still not addressed.

unho:
lgtm from my side, but given all the controversy in this PR please wait for one or more gtms before merging.

dwaynebailey:
I'm of the same view as @iafan I see this as an imperfect solution needed to address our immediate need without having to go and refactor large parts of Pootle.  As long as we are aware that this needs to fixed long term or could catch us when we upgrade Django then I'm LGTM on this.

phlax:
@dwaynebailey ive opened a ticket for the broader issue of how best to order submissions

julen:
A last note to mention I'm in favor of addressing this the proper way sooner than later — dealing with private APIs is utterly uncomfortable.

phlax:
@julen to address the issue of using the private method i included the public earliest/latest methods.

I agree that we should address this sooner rather than later though - as it is basically a hack.

PR #4026, opened 2015-09-04
string complete for Irish

kscanne:


dwaynebailey:
Landed in 0686805017008a89f50d992ee56ecd8675bdf18e - thanks @kscanne 

I'll also push this up to pootle.locamotion.org

PR #4034, opened 2015-09-06
Make Redis version check CRITICAL

### LINKED ISSUE #4033: Make redis version check CRITICAL
If the redis version is not sufficiently recent the site will fail.

Currently there is a CRITICAL check for the presence of redis, but the version check is only a WARNING - i think it should be CRITICAL

The hint should also say what the minimum redis version is

phlax:
Currently a site will fail if the Redis version is too old, but the
check is only a WARNING.

This commit also adds the minimum version requirement to the hint.

Fixes #4033

### DIFF
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -94,6 +94,17 @@ def check_redis(app_configs=None, **kwargs):
             id="pootle.C001",
         ))
     else:
+        redis_version = tuple(int(x) for x
+                              in (queue.connection
+                                       .info()["redis_version"].split(".")))
+        if redis_version < REDIS_MINIMUM_REQUIRED_VERSION:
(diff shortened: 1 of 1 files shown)


dwaynebailey:
This doesn't change it from Warning to Critical.  Sorry commented on some old change. Otherwise lgtm

phlax:
Ive updated the PR to properly set the check to critical.

The other question was about which version of Redis is the actual minimum required.

If the version is too low Redis fails with an error about "wrong number of argument to set command". Looking here:

http://redis.io/commands/SET

it seems like the minimum version is 2.6.12 - but currently the check is for >= 2.8.4 - so i think that should be fine.

unho:
lgtm

Issue #3727, opened 2015-05-12
Add support of `wait` mode in sync commands


ta2-1:
All sync tasks should be moved to RQ according to #3577. But some of our users can have problems if we apply it.

What if `update_stores` tasks shouldn't run for a `Project` before all `sync_stores` tasks are finished for this `Project`? How to handle it with sync commands and multiple workers?

We can create a command that checks if any sync job is running for a project (or a translation project). And then we can ask our users to adjust the way how they work with translation data.

Or we can add checking possibility to `update_stores` and `sync_stores` commands themselves.
For instance `./manage.py sync_stores --project=foo --wait` finishes after all sync tasks created for the `foo` project are finished.

iafan:
I think that what we have discuss was:

1) to add a `--nowait` parameter for async operation. By default, the commands would run in a synchronous mode. This is to make the command-line API backward-compatiple.
2) in addition to `--nowait`, add the support for a separate command (let's call it `can_update_stores`) that would check if the project has no pending `sync_stores` jobs and is open for updates.

dwaynebailey:
:+1: @iafan 

Though I would suggest maybe these alterations:

1. If you don't supply ``--nowait`` that we report that other jobs are busy and exit with a non-zero exit status.
2. That might make the need for ``can_update_stores`` if we can simply test for the exit status to determine if we exited because of running jobs.

iafan:
Sometimes we would like to know whether to deal with the project within current sync cycle before we start updating stores, so I think having a standalone `can_update_stores` command makes sense anyway.

ta2-1:
Since sync tasks are executed in a separate process created by rqworker it's impossible to wait when they finish via standard multi process API. Redis DB is a shared resource here. So we can check if a task is finished getting data from Redis keys directly.
Let’s call `TranslationProject` busy if there exists `STARTED`, `QUEUED` or `DEFERRED` sync job for it.
We can check if `TranslationProject` is busy easily. It’s necessary to check a status of a last job for this translation project.
In this issue we won’t support when `sync_stores` or `update_stores` commands are run from different places at the same moment. If we get result for `TranslationProject` that it’s not busy we don't take into account that somebody can run `sync_stores` or `update_stores` command for this `TranslationProject` and make it busy at the next moment after we get the result.

Another thing: if we run sync tasks for many `TranslationProject`s in one command (with `--project` option only or without `--language`, `--project` options), it can be not efficient to check all items in Redis DB very often.
[...]

PR #4038, opened 2015-09-07
Adding --no-rq flag to pootle command

phlax:
This commit adds a --no-rq flag to the pootle command

Before running, it checks if there is a rqworker running and if so:

   - if --noinput is set it issues a warning
   - otherwise, it issues warning and asks user if they wish to proceed

This is useful for the following use cases:
  - (bash) scripting installation/migration
  - debugging - eg with the python debugger (pdb)

I havent added docs but i will if we are happy to add this.

### DIFF
--- a/pootle/runner.py
+++ b/pootle/runner.py
@@ -11,9 +11,12 @@ import os
 import sys
 from argparse import ArgumentParser, SUPPRESS
 
+from django.conf import settings
 from django.core import management
 
 import syspath_override
+
+from core.utils.redis_rq import rq_workers_are_running
 
 
 #: Length for the generated :setting:`SECRET_KEY`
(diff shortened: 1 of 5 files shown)


ta2-1:
I think that `--sync` name is confusing. We are going to move `update_stores`/`sync_stores` to RQ and run commands in synchronous (wait) mode for backward compatibility (see #3727). 
This PR allows to ignore using of RQ and to run everything in single process (even stats refreshing). We can apply this after testing for debug purposes and rename `--sync` to `--debug-single-process` as @iafan suggested.

phlax:
@ta2-1 @iafan how about we rename it to `--no-async`

The reason being that its mainly for installation/upgrade/migration purposes - the debugging aspect is kind of secondary

iafan:
`--no-async` doesn't give a clue that you won't use RQ at all. One can run the job with RQ and wait for it to complete (with `--wait` parameter), which can also be treated as a non-asynchronous mode. This is why I thought that `--debug-single-process` would convey the meaning better. You may want to consider `--single-process` or `--no-rq`.

I also wanted to also emphasize that this is not a regular mode people should be using in a production environment, hence the `debug` prefix. You may come up with some better variant.

phlax:
> You may want to consider --single-process or --no-rq.

Yep - either of these sound good to me.

iafan:
Pick your favorite. )

pootle/runner.py
```diff
@@ -156,6 +160,26 @@ def init_command(parser, settings_template, args):
...
+    """
+    try:
+        queue = get_queue()
+        Worker.all(queue.connection)
+        if len(queue.connection.smembers(Worker.redis_workers_keys)):
+            print("Warning: you currently have redis workers running. "
+                  "Running in synchronous mode may conflict with jobs "
+                  "that are dispatched to rqworker. It is safer to stop "
+                  "worker when using synchronous commands")
+    except:
```
unho: `except Exception:` please.

dwaynebailey:
I'm gtm on this after you've reviewed my comments and made changes as/if needed. No need for further review from me.

Issue #3050, opened 2013-07-10
Plural field goes missing when uploading file


transl8bzimport:
*Version: 2.5.0*

*Originally posted by Henrik:*

*Created [attachment 940](http://bugs.locamotion.org/attachment.cgi?id=940)*

Empty .ts file

Problem seem with Qt ts files. I could not reproduce with .po files

Howto reproduce:
1) untranslated .ts file with plural field
2) translate .ts file offline and upload to project. Merge with current file

Expected outcome:
Both translated fields are visible from pootle

Actual:
Only one translation field is visible. However translation is not lost. Update against templates will return the field and translations are still present.

Attaching two simple .ts files to help with reproducing the bug

phlax:
This is a very old bug, so rather than test it on 2.5 I decided to find out if it was still relevant.

Attempting to upload a .ts file gets the error:

"'tsfile' object has no attribute 'parseheader'"

Looking through the code it seems that its no longer possible to upload any filetype other than .po, as only po files have the parseheader method.

As this needs to be fixed first, I have raised a ticket for this problem here #3966

ta2-1:
To test this on current master I had to import .ts file manually via `./manage.py shell` similar like it's done in test from this commit https://github.com/translate/pootle/commit/f93a5805e65386c3ed264c9df70fe21984fb852e. 
Now I reckon that unit with plurals imported from .ts file is rendered correctly and all plural fields are visible and close the issue.

PR #4042, opened 2015-09-09
Fix plural of multistring getting lost

egr-kittner:
The to_db function of MultiStringField does not add
the plural place holder. This means that plural sources
that have only one string (as is the case in Qt TS files)
loose there plural attribute and will then be mistreated as
singular strings.

**NOTE**: I only tested this with pootle-2.5 since I don't have a working 2.7 installation right now, but the affected code is still the same

### DIFF
--- a/pootle/apps/pootle_store/fields.py
+++ b/pootle/apps/pootle_store/fields.py
@@ -42,7 +42,12 @@ def to_db(value):
         if list_empty(value.strings):
             return ''
         else:
-            return SEPARATOR.join(value.strings)
+            if getattr(value, "plural", False):
+                strings = value.strings[:]
+                strings.append(PLURAL_PLACEHOLDER)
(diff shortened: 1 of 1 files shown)


egr-kittner:
I forgot to mention, this might also fix #3050

ta2-1:
@egr-kittner thank you for this PR. 
We've added small test in f93a5805e65386c3ed264c9df70fe21984fb852e for checking plurals after importing of .ts file (in http://doc.qt.io/qt-4.8/linguist-ts-file-format.html format). From which I can say that it has been already fixed.
So I'm closing the PR. But if you think that it still can be the issue, please reopen.

ta2-1:
@egr-kittner you were right, test added in f93a580 didn't test all things. 
Thank you again your change was applied in #4707 PR.

Issue #3548, opened 2015-01-26
refresh_stats --calculate-wordcount is dangerous and expensive


dwaynebailey:
If you run ``refresh_stats --calculate-wordcount`` it will create a job per unit.  Thus a relatively small project could spawn a massive amount of jobs and chock legitimate jobs.  The name looks harmless and looks in fact like something you want to choose.  It was used to help transition ``WORDCOUNT_FUNC`` changes.

So the use may still be useful.  So I would suggestion doing some or all of the following:

1. Rename the option and mark it as expensive.
2. Add a ``low`` priority queue where this and other tasks could reside.

julen:
So `--calculate-wordcount` should be [moved away from `refresh_stats`](https://github.com/translate/pootle/commit/fffcce4e6dbea7eb357cacf56ed8f235107bab01#commitcomment-11282376) into its own standalone command (note it's not even available in `refresh_stats_rq`).

julen:
Just renamed the issue so it reflects the end goal.

Once this is done, we have to drop the old `refresh_stats` command altogether (and rename `refresh_stats_rq` to `refresh_stats`).

phlax:
We have dropped calculate_wordcount for the 2.7.1 release - we can add it back at a later date if necessary. I have changed the title to reflect that.

PR #4049, opened 2015-09-09
Removing old refresh_stats and renaming refresh_stats_rq

phlax:
After some discussion we thought it would be ok to drop calculate_wordcount for now (touch #3548)

This commit simply renames `refresh_stats_rq` to `refresh_stats`

I will update docs...


unho:
lgtm

phlax:
@unho docs updated

unho:
Please include this change in the release notes and squash both commits.

docs/server/commands.rst
```diff
@@ -88,49 +88,6 @@ refresh_stats
...
-.. django-admin:: refresh_stats_rq
-
-refresh_stats_rq
-^^^^^^^^^^^^^^^^
-
-.. versionadded:: 2.7
-
-Refreshes all calculated statistics ensuring that they are up-to-date.
-
 A background process will create a task for every file to make sure calculated
```
unho: A few lines below there is a note https://github.com/translate/pootle/pull/4049/files#diff-32d9c83230475a8ff37b1695b1606581R97 referring to the old `refresh_stats`, a note that I don't think it is necessary anymore.

phlax:
@unho updated

docs/server/commands.rst
unho: Done.

unho:
lgtm

PR #4051, opened 2015-09-10
Fix to handle empty/duplicate emails in verify_user

phlax:
The system should not allow empty or duplicate emails to be verified.

This PR adds validation to verify_user and tests to ensure it cant happen


unho:
lgtm

Issue #3796, opened 2015-06-04
Duplicate emails simply hang the interface


dwaynebailey:
If your Pootle setup has duplicate emails.  Then the signin simply hangs and doesn't report any errors.  Some of these could be eliminated by cleanup but I suspect other types of errors also wouldn't be reported.  My ideal is that this type of error is communicated to the user who can ask admins to intervene.

```pytb
Traceback (most recent call last):
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 111, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/views/generic/base.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 29, in _wrapper
    return bound_func(*args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 25, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/allauth/account/views.py", line 94, in dispatch
    return super(LoginView, self).dispatch(request, *args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/allauth/account/views.py", line 62, in dispatch
    **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/newrelic-2.50.0.39/newrelic/hooks/framework_django.py", line 827, in wrapper
    return wrapped(*args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/django/views/generic/base.py", line 87, in dispatch
    return handler(request, *args, **kwargs)
  File "/var/www/sites/translations.documentfoundation.org/env/local/lib/python2.7/site-packages/allauth/account/views.py", line 78, in post
[...]

PR #4052, opened 2015-09-10
Logging login errors and re-raising as RuntimeError

phlax:
If an error occurs during login - eg if there are duplicate emails - the login
UI exposes the error (or hangs if settings.DEBUG = True).

This commit catches the error so that the user gets an appropriate message and
sysadmin can see what error occurred.

Touch #3796

### DIFF
--- a/pootle/apps/accounts/views.py
+++ b/pootle/apps/accounts/views.py
@@ -7,14 +7,34 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import logging
+
 from django.core.urlresolvers import reverse
 from django.shortcuts import redirect
 
 from allauth.account.views import LoginView
+from allauth.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.helpers import _add_social_account
 from allauth.socialaccount.models import SocialLogin
 
 from .forms import SocialVerificationForm
+
+
+logger = logging.getLogger(__name__)
+
+
+class PootleLoginView(LoginView):
+
+    def form_valid(self, form):
+        success_url = self.get_success_url()
+        try:
+            return form.login(self.request, redirect_url=success_url)
+        except ImmediateHttpResponse as e:
+            return e.response
+        except Exception as e:
+            logger.exception("%s %s" % (e.__class__.__name__, e))
+            raise RuntimeError("An error occurred logging you in. Please "
+                               "contact your system administrator")
 
 
 class SocialVerificationView(LoginView):
--- a/pootle/apps/accounts/urls.py
+++ b/pootle/apps/accounts/urls.py
@@ -10,10 +10,14 @@
 from django.conf.urls import patterns, url
 
 
-from .views import SocialVerificationView
+from .views import PootleLoginView, SocialVerificationView
 
 
-urlpatterns = patterns('',
+urlpatterns = patterns(
+    '',
+    url(r"^login/$",
+        PootleLoginView.as_view(),
+        name="account_login"),
     url(r'^social/verify/$',
         SocialVerificationView.as_view(),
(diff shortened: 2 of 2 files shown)


pootle/apps/accounts/views.py
```diff
@@ -7,16 +7,36 @@
...
+
+    def form_valid(self, form):
+        success_url = self.get_success_url()
+        try:
+            return form.login(self.request, redirect_url=success_url)
+        except ImmediateHttpResponse as e:
+            return e.response
+        except Exception as e:
+            logger.error("%s %s" % (e.__class__.__name__, e))
+            raise RuntimeError("An error occurred logging you in. Please "
```
unho: Is this valid english?

pootle/apps/accounts/views.py
```diff
@@ -7,16 +7,36 @@
...
+class PootleLoginView(LoginView):
+
+    def form_valid(self, form):
+        success_url = self.get_success_url()
+        try:
+            return form.login(self.request, redirect_url=success_url)
+        except ImmediateHttpResponse as e:
+            return e.response
+        except Exception as e:
+            logger.error("%s %s" % (e.__class__.__name__, e))
```
unho: IMHO it is preferable to use `logger.exception` to log the full traceback.

pootle/apps/accounts/views.py
phlax: I believe so - but i can change it to "an error occurred while logging you in" if thats clearer

pootle/apps/accounts/views.py
phlax: well as you rightly pointed out some of the traceback will get printed with the RuntimeError

Im not really comfortable with this solution of not handling the errors properly - but for now its better than exposing them to the user

pootle/apps/accounts/views.py
```diff
@@ -7,16 +7,36 @@
...
+class PootleLoginView(LoginView):
+
+    def form_valid(self, form):
+        success_url = self.get_success_url()
+        try:
+            return form.login(self.request, redirect_url=success_url)
+        except ImmediateHttpResponse as e:
+            return e.response
+        except Exception as e:
+            logger.exception("%s %s" % (e.__class__.__name__, e))
```
unho: Is it actually necessary to pass that string? I mean, `logger.exception` already prints the whole traceback whose last line is the exception name and its message, isn't it?

phlax: I think its clearer if you *log* the actual exception/message, because even tho logger.exception prints the tb its lost amongst the tb from the RuntimeException

unho: I am sorry, but I don't fully understand that comment. Can you please rephrase?

phlax: @unho after further investigation i was mistaken about how logging.exception worked.

I do however feel that this is a better error message for a a sysadmin because on the line which has ERROR it has what kind of error as well as the error message - although i accept this exception is there half way through the traceback.

unho: It is exactly at the end of the traceback.

phlax: No the runtime error is at the end of the traceback

unho:
lgtm

julen:
Sorry I'm -1 on this approach as already commented on #4011. This should perfectly be doable by overriding the adapter's `send_mail()` method. Can we revert please?

julen:
Besides that, this is not the right way to [override a form in allauth (check `ACCOUNT_FORMS`)](http://django-allauth.readthedocs.org/en/latest/configuration.html).

julen:
Sorry it's a view — disregard my last comment.

phlax:
@julen please check the bug that this was addressing - #3796 - send_mail doesnt come into it

PR #4054, opened 2015-09-10
Adding CACHE check in runner 

phlax:
This prevents the pootle command from a raising `ImproperlyConfigured` error if caches is not setup correctly


unho:
This only works for `pootle command` but not for `manage.py command`. I tried to figure out a way to do this in a way that works for both with no success. So I think it is ok to make `pootle` to display a nicer message and keep `manage.py` as is, because even if `ImproperlyConfigured` traceback is ugly it should provide enough information for the most likely most advanced users running `manage.py`.

So lgtm.

PR #4060, opened 2015-09-12
Merge delete

### LINKED ISSUE #4059: merge_user should also remove the old user
Currently ``merge_user`` does not actually remove the old user.  You need to run ``purge_user`` following ``merge_user`` to completely remove the user.  This is dangerous and error prone, especially on older instances that have a large number of users.

phlax:
This commit makes merge_user delete user after merge.

To retain user after merge you can use --no-delete option

Docs updated.

Fixes #4059


unho:
lgtm after squashing both commits.

PR #4072, opened 2015-09-16
Exclude system users from duplicate email checks

### LINKED ISSUE #4071: False positive in `check_duplicate_emails`
The newly-introduced `check_duplicate_emails` check goes through all users without excluding users that are fine to have *empty* or invalid email addresses (*nobody* and *default*), hence resulting in a false positive test.

phlax:
Fixes #4071


unho:
lgtm

PR #4073, opened 2015-09-16
Releasing 2.7.1

unho:


docs/releases/2.7.1.rst
```diff
@@ -105,8 +124,62 @@ Command changes and additions
...
+  suggestions.
+
+- Suggestion related events are now displayed on the timeline.
+
+- Critical and not critical failing checks are now displayed separately in the
+  editor.
+
+- Potential errors when managing the suggestions are now displayed to users.
+
+- Now users can reject their own suggestions even if they don't have enough
```
julen: I'd say this is a fixed regression where users couldn't reject their own suggestions without reviewing rights.

julen:
Thanks for taking care of the changelog btw :+1: — we should everyone take better habit of keeping it up-to-date as we go.

PR #4087, opened 2015-09-22
Elasticsearch fix

### LINKED ISSUE #4085: Elasticsearch 1.7.0 breaks Pootle if ES not set up
With `elasticsearch==1.7.0` doing `es.ping()` on a not-setup server raises `ConnectionError`.

This breaks the default install here:

https://github.com/translate/pootle/blob/master/pootle/core/search/backends/elasticsearch.py#L29

phlax:
This PR:
- pins elasticsearch version to 1.6
- handles connectionerrors gracefully when checking for ES server

fixes #4085

### DIFF
--- a/pootle/core/search/backends/elasticsearch.py
+++ b/pootle/core/search/backends/elasticsearch.py
@@ -13,6 +13,7 @@ __all__ = ('ElasticSearchBackend',)
 
 try:
     from elasticsearch import Elasticsearch
+    from elasticsearch.exceptions import ConnectionError
 except:
     Elasticsearch = None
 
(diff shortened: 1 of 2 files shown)


pootle/core/search/backends/elasticsearch.py
```diff
@@ -26,7 +27,12 @@ def __init__(self, config_name):
         self._create_index_if_missing()
 
     def _server_setup_and_alive(self):
-        return self._es is not None and self._es.ping()
+        if self._es is not None:
```
julen: In general, I'd [bail out early](https://philikon.wordpress.com/2010/01/26/flat-is-better-than-nested-bail-out-early/):

```python
if self._es is None:
    return False

try:
    return self._es.ping()
except ConnectionError:
    return False
```

unho: I also find this easier to read.

requirements/base.txt
```diff
@@ -12,7 +12,7 @@ django-transaction-hooks
 # Libraries
 cssmin
 diff-match-patch>=20121119
-elasticsearch
+elasticsearch==1.6.0
```
unho: As @dwaynebailey commented in the commit `<1.7.0` feels better.

pootle/core/search/backends/elasticsearch.py
phlax: ill update

PR #4095, opened 2015-09-22
l10n: fix various interpolations

dwaynebailey:
Seems we have a few strings that are doing the interpolation in a non-robust way and also in a way that ensures that they can't be localised.

* ``_('A %s b' % var)`` - is broken as it can't be localised
* ``_('A %s b') % var`` - is more correct
* ``_('A %s b', var)`` - is, I believe, more robust

### DIFF
--- a/pootle/apps/import_export/utils.py
+++ b/pootle/apps/import_export/utils.py
@@ -27,26 +27,26 @@ def import_file(file, user=None):
     f = getclass(file)(file.read())
     if not hasattr(f, "parseheader"):
         raise UnsupportedFiletypeError(_("Unsupported filetype '%s', only PO "
-                                         "files are supported at this time\n")
-                                       % file.name)
+                                         "files are supported at this time\n",
+                                         file.name))
     header = f.parseheader()
     pootle_path = header.get("X-Pootle-Path")
     if not pootle_path:
         raise MissingPootlePathError(_("File '%s' missing X-Pootle-Path "
-                                       "header\n") % file.name)
+                                       "header\n", file.name))
 
     rev = header.get("X-Pootle-Revision")
     if not rev or not rev.isdigit():
         raise MissingPootleRevError(_("File '%s' missing or invalid "
-                                      "X-Pootle-Revision header\n")
-                                    % file.name)
+                                      "X-Pootle-Revision header\n",
+                                      file.name))
     rev = int(rev)
 
     try:
         store, created = Store.objects.get_or_create(pootle_path=pootle_path)
     except Exception as e:
         raise FileImportError(_("Could not create '%s'. Missing "
-                                "Project/Language? (%s)") % (file.name, e))
+                                "Project/Language? (%s)", (file.name, e)))
 
     try:
         store.update(overwrite=True, store=f, user=user,
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -388,7 +388,7 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
     def clean(self):
         if self.code in RESERVED_PROJECT_CODES:
             raise ValidationError(
-                _('"%s" cannot be used as a project code' % (self.code,))
+                _('"%s" cannot be used as a project code', self.code)
             )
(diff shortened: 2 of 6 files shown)


pootle/apps/pootle_project/models.py
```diff
@@ -388,7 +388,7 @@ def delete(self, *args, **kwargs):
     def clean(self):
         if self.code in RESERVED_PROJECT_CODES:
             raise ValidationError(
-                _('"%s" cannot be used as a project code' % (self.code,))
+                _('"%s" cannot be used as a project code') % (self.code,)
```
unho: Why not `_('"%s" cannot be used as a project code', self.code)`?

pootle/apps/import_export/utils.py
```diff
@@ -27,26 +27,26 @@ def import_file(file, user=None):
...
+                                      "X-Pootle-Revision header\n",
+                                      file.name))
     rev = int(rev)
 
     try:
         store, created = Store.objects.get_or_create(pootle_path=pootle_path)
     except Exception as e:
         raise FileImportError(_("Could not create '%s'. Missing "
-                                "Project/Language? (%s)") % (file.name, e))
+                                "Project/Language? (%s)", file.name, e))
```
unho: Perhaps the tuple should be kept.

pootle/apps/pootle_project/models.py
dwaynebailey: To be honest really wasn't sure about that one

pootle/apps/import_export/utils.py
dwaynebailey: I don't now TBH

dwaynebailey: Seems you're right, broke the build

pootle/apps/pootle_project/models.py
unho: Well, you changed it in other places to be like this.

dwaynebailey: Except its a (x,) instance.

unho: No, it doesn't. Using `_('"%s" cannot be used as a project code', self.code)` should just work.

dwaynebailey:
@translate/core looking for a GTM here please

jleclanche:
lgtm.

BTW, I noticed you turned `% (conn_settings)` into `% conn_settings`. As a rule of thumb, I always parenthesize string arguments to string formats. It's a pass-through syntax wise and has two benefits:
* When adding a new argument, the syntax is immediately clear (no need to add parentheses)
* It distinguishes string arguments to tuple or dynamic arguments. Eg the case where you do

```py
args = "foo", "bar", "baz"
s = "hi %s, hello %s, greetings %s" % args
```

pootle/checks.py
```diff
@@ -114,8 +114,8 @@ def check_redis(app_configs=None, **kwargs):
         workers = Worker.all(queue.connection)
     except Exception as e:
         conn_settings = queue.connection.connection_pool.connection_kwargs
-        errors.append(checks.Critical(_("Could not connect to Redis (%s)") % (e),
-            hint=_("Make sure Redis is running on %(host)s:%(port)s") % (conn_settings),
+        errors.append(checks.Critical(_("Could not connect to Redis (%s)", e),
+            hint=_("Make sure Redis is running on %(host)s:%(port)s") % conn_settings,
```
unho: Just curious why the `%` is used since in other places you switched to a comma.

jleclanche: It's correct. ugettext doesn't support dict formatting. cf
https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#standard-translation

dwaynebailey:
@jleclanche thanks for the review.  My logic for ``conn_settings`` was that it is a dictionary, so the brackets where redundant and actually a bit confusing.   I might have misunderstood your comment though.  So I'm landing without it unless you can help with my broken understanding.

jleclanche:
As I mentioned on IRC, I was actually agreeing with your change - feel free to land it. Sorry it didn't quite come through ;)

PR #4108, opened 2015-09-27
Docs: Add CONTRIBUTING guide

dwaynebailey:
This lifts heavily from https://github.com/translate/pootle/wiki/Styleguide-%28proposal%29

I've written this more to provide some guideance to users who will see it in the Github link when they create a new Github issue.  Based in some part on a few unhelpful issues being reported without basic info thus creating the need to ask questions that could have been addressed by the reporter.

Please review this for readability.  The idea is to keep is clean and clear.

I realise it duplicates other stuff, but I'd prefer to have something working now then mash all those fixes together.

If there is something desperate you feel should be here, please flag it, but likley I'll ask that you write that so that it doesn't block this landing.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,0 +1,47 @@
+How to contribute
+=================
+
+Reporting bugs
+--------------
+
+Write your bug report so that you ensure developers are fixing the bug rather
+than spending time trying to understand your bug report.  If we can't see it or
+replicate it, likely we can't fix it.
+
+Checklist:
+
+#. Bug description is clear yet concise.
+#. Clear instructions to reliably replicate the issue.
+#. Screenshots rather than descriptions of screens.
+#. Full traceback if one occured.
+#. Links to the issue.
+
+Commits and pull requests
+-------------------------
+
+Keep the commit log as healthy as the code. It is one of the first places new
+contributors will look at the project.
+
+#. No more than one change per commit. There should be no changes in a commit
+   which are unrelated to its message.
+#. Every commit should pass all tests on its own.
+#. Follow `these conventions <http://chris.beams.io/posts/git-commit/>`_ when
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Translations
+------------
+
+Please translate `online <http://pootle.locamotion.org/projects/pootle/>`_
+
+Need help?
+----------
+
+You can always ask for help in the `#pootle-dev
```
julen: Sounds like routing people to a support channel, which `#pootle-dev` is not.

julen:
The main problem I see here is that it's talking about issues, commits and PRs even before telling people to do their homework: double-check the docs before reporting issues; unless trivial fixes, do not write any code without consulting about it first —even more, and this applies to anyone: think if any code is needed at all ([The Best Code is No Code At All](http://blog.codinghorror.com/the-best-code-is-no-code-at-all/)). This might sound common sense, but practice shows that people not always apply it.

unho:
Just wondering if that doc should just point to http://docs.translatehouse.org/projects/pootle/en/latest/developers/hacking.html

CONTRIBUTING.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Commits and pull requests
+-------------------------
+
+Keep the commit log as healthy as the code. It is one of the first places new
+contributors will look at the project.
+
+#. No more than one change per commit. There should be no changes in a commit
+   which are unrelated to its message.
+#. Every commit should pass all tests on its own.
```
unho: Ideally yes, but I am ok with just asking the PR to pass all tests.

unho:
Replaced by PR #4228. Requested changes on this PR have been applied there. So closing this one.

Issue #4096, opened 2015-09-22
Update templates docs to reflect current behavior


unho:
As pointed out by @julen http://docs.translatehouse.org/projects/pootle/en/latest/features/templates.html needs to be updated to reflect the current behavior and workflow.

PR #4109, opened 2015-09-28
Revise 'template' docs

dwaynebailey:
This addresses #4096 

I'm happy to land, just wanted a few pedantic editors to review what I've crafted.


julen:
:+1: looks good to me.

Just two pedantic comments: 
* The language name is *Templates*, not *Template*.
* For clarity, it'd be worth to have the "language code" *templates* italicized, just so the reader knows when we talk about templates generally, and when about the special language specifically.

PR #4112, opened 2015-09-28
Static pages: integrated CodeMirror editor

### LINKED ISSUE #3346: Add syntax highlighting editor to our static pages
Our static pages are HTML, and it would be nice to have HTML syntax highlight in there.

My personal strong preference is http://codemirror.net/ since it is very solid, has a great API, and allows to easily write custom highlighters, which we can make the use of in other places in the app.

julen:
This PR integrates the CodeMirror editor for static pages, and compiles support for modes into separate chunks.

The highlighting changes according to the configured markup module configured in the settings. It'd be great if you can test with `DEBUG` set as `True` and `False`.

Fixes #3346.

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -31,6 +31,8 @@ complete information.
 Details of changes
 ==================
 
+- The editor for static pages now highlights the content's markup
+  (:issue:`3346`).
 - Pulled latest translations.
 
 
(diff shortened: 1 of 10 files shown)


julen:
I also didn't push the first commit directly to master because it's tightly related to this feature — it's dead code otherwise.

unho:
I tried to test it with no success. Can you please provide some test instructions?

julen:
There are no special instructions: just make sure the deps are up-to-date (`npm install`) and build the JS code.

unho:
Tested with DEBUG on and off for HTML, Markdown and reStructuredText and it seems to work (despite highlight support for rst and Markdown seems to be a bit flawed).

gtm on my side.

Issue #2563, opened 2012-11-21
Factor out merging logic


julen:
Current merging logic implemented for uploads is not reusable so we can't use it in other tasks such as in management commands. So first we need to decouple the merging logic from uploads.

jleclanche:
@dwaynebailey i guess this is what we were talking about re conflict management

PR #4114, opened 2015-09-29
[WIP] Making stores importable and exportable without a file on disk

phlax:
This commit:
- makes get_file_mtime not fail if there is no file on disk
- only updates file_mtime if get_file_mtime returns a value
- moves the check file mtime logic into update_stores
- removes the `only_newer` param from Store.update and associated tests

Touch #2563


pootle/apps/pootle_store/models.py
```diff
@@ -1861,13 +1867,6 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
             self.parse(store=store)
             return
 
-        disk_mtime = self.get_file_mtime()
-        if only_newer and disk_mtime == self.file_mtime:
```
unho: Is `only_newer` actually necessary now?

pootle/apps/pootle_store/models.py
```diff
@@ -1972,7 +1971,7 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
                                                        update_unitids, user,
                                                        revision=update_revision,
                                                        submission_type=submission_type)
-            self.file_mtime = disk_mtime
+            self.file_mtime = self.get_file_mtime() or self.file_mtime
```
unho: Is this not something that is better done inside `get_file_mtime()` instead of being repeated three times here?

phlax: if im honest im not sure what the best thing to set it to is.

If the file has gone away/is not there then perhaps we should set it to datetime_min - atm it just leaves it the same as it is.

unho: Has gone away meaning that it was removed from disk?

phlax: yep

unho: I don't see the problem with the file being away. Code copes with it.

pootle/apps/pootle_store/models.py
phlax: im not sure - i would think so - but its still used in Store.sync so i havent removed.

This clause defo does have to be removed as the "disk_mtime == self.file_mtime" is unreliable if the file is not on disk

pootle/apps/pootle_store/models.py
phlax: @unho i did this but had to revert.

I realised that in the management command it calls get_file_mtime() and tests if the result is None

unho: ok, then lets keep it as is.

pootle/apps/pootle_store/models.py
unho: Yes, but `update()` doesn't call `sync()` and `sync()` doesn't call `update()`. So unless there is any place where `update()` is being called with `only_newer`, `only_newer` should be safe to remove.

phlax: you're right - ill remove

phlax: @unho - i removed and had to update a load of tests to accomodate

phlax:
@julen are you happy with this change?

pootle/apps/pootle_store/models.py
```diff
@@ -1832,16 +1838,14 @@ def get_common_dbids(self, opcodes, old_unitid_list, old_unitids):
 
         return update_dbids
 
-    def update(self, overwrite=False, store=None, only_newer=False, user=None,
-               submission_type=None):
+    def update(self, overwrite=False, store=None,
+               user=None, submission_type=None):
```
unho: I don't there is any need to move `user` to this line.

phlax: theres no need but i do think it looks a bit tidier and reads better

julen:
What's the expected behavior for `update_stores --force` when the store is missing from disk?

pootle/apps/pootle_store/models.py
```diff
@@ -1589,8 +1589,14 @@ def parse(self, store=None):
             return
 
     def get_file_mtime(self):
+        try:
+            disk_mtime = self.file.getpomtime()[0]
+        except ValueError:
+            # dont fail if missing from disk - leave file_mtime unchanged
```
julen: *Don't*

pootle/apps/pootle_store/models.py
```diff
@@ -1589,8 +1589,14 @@ def parse(self, store=None):
...
+            return
```
julen: Explicit > implicit: `return None`. Ditto elsewhere.

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -45,9 +45,18 @@ def handle_translation_project(self, translation_project, **options):
 
     def handle_store(self, store, **options):
         overwrite = options.get('overwrite', False)
-        force = options.get('force', False)
-
-        store.update(overwrite=overwrite, only_newer=not force)
+        if not options.get('force', False):
+            disk_mtime = store.get_file_mtime()
+            if not disk_mtime:
```
julen: Considering `store.get_file_mtime()` returns `None` when `mtime` is unavailable:
```python
if disk_mtime is None:
```

pootle/apps/pootle_store/models.py
julen: I'm not either sure I understand the purpose of the `or self.file_mtime` part. Independently of its meaning, I'm concerned it's a detail being taken care of in a ad-hoc fashion — easy to miss out.

pootle/apps/pootle_store/models.py
unho: My point is that this just adds another diff line when there is absolutely no need.

phlax: the extra line is necessary to be pep8 compliant

phlax:
> What's the expected behavior for update_stores --force when the store is missing from disk?

Well it should be do nothing - ive updated the management command to reflect that

pootle/apps/pootle_store/models.py
phlax: > I'm not either sure I understand the purpose of the or self.file_mtime part. Independently of its meaning, I'm concerned it's a detail being taken care of in a ad-hoc fashion — easy to miss out.

@julen im not sure i follow - do you have any specific suggestions?

the bigger question for me is should a store that previously had a file but now doesn't, set its file_mtime back to datetime_min or should it keep the file_mtime that it had.

 Im not sure it matters that much as the comparison in update_stores at least is only whether file_mtime has changed not whether its newer.

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -45,9 +45,18 @@ def handle_translation_project(self, translation_project, **options):
...
-        force = options.get('force', False)
-
-        store.update(overwrite=overwrite, only_newer=not force)
+        disk_mtime = store.get_file_mtime()
+        if disk_mtime is None:
+            logging.debug(u"File missing from disk, skipping "
+                          u"%s" % store.pootle_path)
+            return
+        if not options.get('force', False):
+            if disk_mtime == store.file_mtime:
```
unho: Why not put all the conditions in the same line?

```python
        if not options.get('force', False) and disk_mtime == store.file_mtime:
```

dwaynebailey:
@phlax wouldn't update_store --force with a missing FS store force the Store to be obsoleted?

phlax:
> @phlax wouldn't update_store --force with a missing FS store force the Store to be obsoleted?

sorry - youre right - ill investigate further

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -45,9 +45,18 @@ def handle_translation_project(self, translation_project, **options):
...
         overwrite = options.get('overwrite', False)
-        force = options.get('force', False)
-
-        store.update(overwrite=overwrite, only_newer=not force)
+        disk_mtime = store.get_file_mtime()
+        if disk_mtime is None:
+            logging.debug(u"File missing from disk, skipping "
+                          u"%s" % store.pootle_path)
+            return
+        if not options.get('force', False):
```
dwaynebailey: Note as long as ``--force`` has a default value this is always defined.  So no need to use ``.get`` with [default values](https://docs.python.org/2/library/optparse.html#default-values) outside of the actual option definition.  And it should be ``options.force`` when you actually need to use the setting.

phlax: >  And it should be options.force when you actually need to use the setting.
"force" is a dictionary key not an attribute so its not possible to do `options.force` only `options['force']`

phlax:
@julen @unho @dwaynebailey 

ive updated this PR to fix the issue with `update_stores` with `--force`

i also had to make a further change to prevent `Store.update` from setting the `last_sync_revision` if no file is present

pootle/apps/pootle_app/management/commands/update_stores.py
dwaynebailey: @phlax, not according to the optparse docs.

> As it parses the command line, optparse sets attributes of the options object returned by parse_args() based on user-supplied command-line values. When parse_args() returns from parsing this command line, options.filename will be "outfile" and options.verbose will be False.

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -44,10 +44,18 @@ def handle_translation_project(self, translation_project, **options):
...
+            logging.debug(u"File missing from disk, skipping "
+                          u"%s" % store.pootle_path)
+            return
+        if not options['force'] and disk_mtime == store.file_mtime:
+            # The file on disk wasn't changed since the last sync
+            logging.debug(u"File didn't change since last sync, skipping "
+                          u"%s" % store.pootle_path)
+            return
+        store.update(overwrite=options['overwrite'],
+                     only_newer=not options['force'])
```
julen: This line is unintended I guess.

pootle/apps/pootle_store/models.py
julen: It looked to me `self.file_mtime` was some value retrieved from elsewhere, but now that I re-read along with your comment, I realize it's exactly the same value the store already has.

I believe this, although more verbose, expresses the intention better:
```python
disk_mtime = self.get_file_mtime():
if disk_mtime is not None:
    self.file_mtime = disk_mtime
```

pootle/apps/pootle_store/models.py
```diff
@@ -1972,10 +1969,12 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
                                                        update_unitids, user,
                                                        revision=update_revision,
                                                        submission_type=submission_type)
-            self.file_mtime = disk_mtime
+            self.file_mtime = self.get_file_mtime() or self.file_mtime
 
-            if (filter(lambda x: changes[x] > 0, changes) and
-                store == self.file.store):
+            update_sync_revision = (filter(lambda x: changes[x] > 0, changes)
+                                    and self.file
```
julen: `and` should go in the previous line. Same for the `and` below this line.

phlax: surely not - its far more readable to see the operator up front

julen: I'm not arguing what's more readable; just asking for the change [to be consistent](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#if-statements).

pootle/apps/pootle_store/models.py
phlax: well maybe - but imho its 2 extra lines of unnecessary code

unho:
lgtm

julen:
I'm happy with the PR :+1: — [I raised some comments on IRC](https://botbot.me/freenode/pootle-dev/2015-10-14/?msg=51859060&page=1) which we already discussed with @phlax.

If this can wait for a few days I'd prefer to hear from @ta2-1 before it gets merged though.

phlax:
i reckon all of the changes from this PR are now handled in #4158

PR #4122, opened 2015-10-02
Removed dynamic model.choice on Project.localfiletype

### LINKED ISSUE #4110: Using dynamic choices on model causes migration issues
There are some dynamic choice on models - if a deployment env is set up differently to the env in which the last migration was created - it causes makemigrations to want to create a new migration.

See here:

https://github.com/translate/pootle/blob/master/pootle/apps/pootle_project/models.py#L159:L160

Generally choices shouldnt be dynamic in this way:

https://docs.djangoproject.com/en/dev/ref/models/fields/#choices

See also #3687

phlax:
Removed the model choices as they were causing migration issues.

Added check in Manager.create and Model.save to ensure localfiletype
is valid

Fixes #4110

### DIFF
--- a/pootle/apps/pootle_app/forms.py
+++ b/pootle/apps/pootle_app/forms.py
@@ -16,6 +16,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from pootle_language.models import Language
 from pootle_project.models import Project, RESERVED_PROJECT_CODES
+from pootle_store.filetypes import filetype_choices
 from pootle_store.models import Store
 
 
@@ -76,6 +77,11 @@ class ProjectForm(forms.ModelForm):
             value = self.cleaned_data.get('localfiletype', None)
             if not value:
                 value = self.instance.localfiletype
+            filetypes = [x[0] for x in filetype_choices]
+            if value not in filetypes:
+                raise forms.ValidationError(
+                    "Unrecognised project filetype: %s" % value
+                )
             return value
 
         def clean_treestyle(self):
--- a/pootle/apps/pootle_project/migrations/0002_remove_dynamic_model_choices_localfiletype.py
+++ b/pootle/apps/pootle_project/migrations/0002_remove_dynamic_model_choices_localfiletype.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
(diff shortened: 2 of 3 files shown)


unho:
Reviewed and tested locally. gtm from my side. Please wait for another gtm.

julen:
Not defining a `choices` means automatic validation is left behind. Keeping that in mind, this looks like a perfect match for [Django 1.8's `forms.ChoiceField` accepting a callable for `choices`](https://github.com/django/django/commit/74e1980cf96eb45079bef464fabdcbe0a6db2423). While one could potentially insert invalid data by using the ORM, `localfiletype` is only exposed via forms, where validation happens.

phlax:
> Not defining a choices means automatic validation is left behind

yep, thats why i added the create/save validation. The only problem i see with this approach is that it doesnt validate on update.

It defo should have validation on any forms too.

julen:
My point is you don't need to introduce any manual validation code, but ensure the form knows about the dynamic choices.

unho:
IMHO the best is to validate both in the model and the form. Django does this automatically for you if you specify the choices in the model, so I think it is better to keep that behavior. Also doing the validation just in the form might be problematic in case the model data does not come from the form.

phlax:
> ... ensure the form knows about the dynamic choices.

@julen ive added form validation also - are you happy for this to land?

pootle/apps/pootle_app/forms.py
```diff
@@ -16,9 +16,11 @@
 
 from pootle_language.models import Language
 from pootle_project.models import Project, RESERVED_PROJECT_CODES
+from pootle_store.filetypes import filetype_choices
 from pootle_store.models import Store
 
 
+
```
unho: Unnecessary blank line. Please remove.

julen:
It's enough to validate the form as I already mentioned in the previous comment.

phlax:
>  It's enough to validate the form as I already mentioned in the previous comment.

What happens if the Project is created programatically?

julen:
> What happens if the Project is created programatically?

[While one could potentially insert invalid data by using the ORM, localfiletype is only exposed via forms, where validation happens.](https://github.com/translate/pootle/pull/4122#issuecomment-145495244)

julen:
Looks good to me.

PR #4128, opened 2015-10-05
Use additional local (external) TMs

### LINKED ISSUE #4102: Add ability to create additional local TMs populated from specified files
Local TM is fast and easy to load.  But it only loads into one TM backend and only loads units that are in Pootle. The goal is to allow us to load any file that TTK understands into a separate TM store that uses elasticsearch.

Steps:

- Add additional TM backend for Locally loaded TM sources i.e. you have two, local TM and external resource.
  - We don’t address any non-blocking UI and assume the speed is the same for local and external TM resources from Elasticsearch.
- Needs the concept of weighting but only at the level of between TM sources.
  - We weight on match quality times a weighting for the TM source. I.e. Local has weighting of 1 and external weighting of 0.9. Then a result of 100% match will have a final quality of 100 for local and 90 for external.  I.e. Actual match * Weighting = final match quality.
- Extend the TM loading tool to be able to point to files and folders to be able to load translation memory resources into the TM server.
  - Should be able to point to a file or a directory
  - Must give the TM a name i.e. “LibreOffice 5.0” that will be used in the TM results (Local TM does this already)
  - Should be able to traverse over a checkout with multiple languages in GNU/non-GNU style.
This should be the same loading tool, i.e. it is loading local or external depending on how we configure (Need to check if this is insane)
- Fix TM results to correctly display source when we don’t know the author which will mostly be the case of External TM.

unho:
This adds the ability to create additional translation memories, and use
them on the editor. Results are now pondered using the result score
and a weighting factor specific to each TM server (between 0 and 1).

Fixes #4102.

### DIFF
--- a/docs/features/translation_memory.rst
+++ b/docs/features/translation_memory.rst
@@ -31,10 +31,14 @@ Configuring Translation Memory
 ------------------------------
 
 Translation Memory will work out of the box with a default Pootle installation.
-There are two methods of getting Translation Memory.
+There are three methods of getting Translation Memory.
 
 1. Amagama - for remote Translation Memory
 2. Elasticsearch - for local Translation Memory
+3. Elasticsearch - for external Translation Memory
+
+
+.. _translation_memory#amagama:
 
 Amagama based remote TM
 ~~~~~~~~~~~~~~~~~~~~~~~
@@ -47,17 +51,49 @@ If you want to setup and connect to your own TM server then the
 :setting:`AMAGAMA_URL` will allow you to point to a private TM server.
 
 
-.. _translation_memory#local_translation_memory:
+.. _translation_memory#elasticsearch_based_tms:
 
-Elasticsearch based local TM
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Elasticsearch-based TMs
+~~~~~~~~~~~~~~~~~~~~~~~
 
 .. versionadded:: 2.7
 
-By default this is configured to operate if local TM is available.  To use it
-you will need to install `Elasticsearch 
-<https://www.elastic.co/products/elasticsearch>`_ and populate the initial
-translation memory using the :djadmin:`update_tmserver` command:
+Pootle can also retrieve TM matches stored on Elasticsearch-based TM servers.
+These TM servers require
+`Elasticsearch <https://www.elastic.co/products/elasticsearch>`_ to be
+installed and running.
+
+.. note:: Elasticsearch depends on Java. Note that some systems may ship with
+  OpenJDK, however `elasticsearch recommends using Oracle JDK
+  <https://www.elastic.co/guide/en/elasticsearch/reference/1.6/setup-service.html#_installing_the_oracle_jdk>`_.
+
+
+Pootle supports two types of Elasticsearch-based TMs:
+
+- **Local TM**: (just one, named ``local``) is populated using translations
+  stored in Pootle database and every new translation gets automatically
+  imported to it.
+
+- **External TMs**: (several) are populated from translation files specifically
+  provided by the server admins, and are not automatically updated.
+
+
+Both local and external TM settings can be adjusted in
+:setting:`POOTLE_TM_SERVER`.
+
+You may want to disable Amagama by setting :setting:`AMAGAMA_URL` to ``''`` if
+you are using Elasticsearch-based TMs, though both can operate together.
+
+
+.. _translation_memory#local_translation_memory:
+
+Elasticsearch-based local TM
+++++++++++++++++++++++++++++
+
+.. versionadded:: 2.7
+
+To use it, the ``local`` TM must be enabled on :setting:`POOTLE_TM_SERVER` and
+will need to be populated using the :djadmin:`update_tmserver` command:
 
 .. code-block:: bash
 
@@ -66,12 +102,68 @@ translation memory using the :djadmin:`update_tmserver` command:
 
 Once populated Pootle will keep Local TM up-to-date.
 
-.. note:: Elasticsearch depends on Java. Note that some systems may ship with
-  OpenJDK, however `elasticsearch recommends using Oracle JDK
-  <https://www.elastic.co/guide/en/elasticsearch/reference/1.6/setup-service.html#_installing_the_oracle_jdk>`_.
+
+.. _translation_memory#external_translation_memories:
+
+Elasticsearch-based external TMs
+++++++++++++++++++++++++++++++++
+
+.. versionadded:: 2.7.3
+
+In order to use them they must be enabled on :setting:`POOTLE_TM_SERVER` and
+you will need to populate them using the :djadmin:`update_tmserver` command
+specifying the TM to use with :option:`--tm` and the project with
+:option:`--project`:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --tm=external --project=Pidgin af.po gl.tmx
 
 
-Local TM settings can be adjusted in :setting:`POOTLE_TM_SERVER`.
+A given TM can host translations for several projects. Just specify them with
+:option:`--project`:
 
-You may want to disable Amagama by setting :setting:`AMAGAMA_URL` to ``''`` if
-you are using Elasticsearch local TM, though the two will operate together.
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --tm=external --project=GNOME pt.tmx eu.po xh.po
+
+
+It is possible to have several Elasticsearch-based external TM servers working
+at once, along with the Elasticsearch-based local TM server. In order to do so
+just add new entries to :setting:`POOTLE_TM_SERVER`:
+
+.. code-block:: python
+
+    POOTLE_TM_SERVER = {
+
+        ...
+
+        'libreoffice': {
+            'ENGINE': 'pootle.core.search.backends.ElasticSearchBackend',
+            'HOST': 'localhost',
+            'PORT': 9200,
+            'INDEX_NAME': 'whatever',
+            'WEIGHT': 0.9,
+            'MIN_SCORE': 'AUTO',
+        },
+    }
+
+Make sure :setting:`INDEX_NAME <POOTLE_TM_SERVER-INDEX_NAME>` is unique. You
+might also want to tweak :setting:`WEIGHT <POOTLE_TM_SERVER-WEIGHT>` to change
+the score of the TM results in relation to other TM servers (valid values are
+between ``0.0`` and ``1.0``).
+
+To use these additional external TMs you will need to populate them using the
+:djadmin:`update_tmserver` command specifying the TM server with
+:option:`--tm`:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --tm=libreoffice --project=LibreOffice af.po gl.tmx
+
+
+Check :djadmin:`update_tmserver` for more options.
+
+Note that Pootle will not push new translations to these TM servers unless you
+explicitly use the :djadmin:`update_tmserver` command, giving you full control
+of which translations make into them.
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -21,6 +21,7 @@ Major Changes
 =============
 
 - Pulled latest translations.
+- Added support for Elasticsearch-based external Translation Memory servers.
 
 
 Below we provide much more detail. These are by no means exhaustive, view the
@@ -34,8 +35,21 @@ Details of changes
 - The editor for static pages now highlights the content's markup
   (:issue:`3346`).
 - Pulled latest translations.
-- :djadmin:`update_tmserver`: renamed :option:`--overwrite` to
-  :option:`--refresh`.
+- :djadmin:`update_tmserver`:
+
+  - Renamed :option:`--overwrite` to :option:`--refresh`.
+  - Added support for Elasticsearch-based external Translation Memory servers,
+    which can be populated from translation files. This effectively brings the
+    ability to display TM results from different TM servers, sorting them by
+    their score.
+
+- :setting:`POOTLE_TM_SERVER`:
+
+  - The ``default`` TM server on  has been renamed to ``local``. Make sure to
+    adjust your settings.
+  - Added a new :setting:`WEIGHT <POOTLE_TM_SERVER-WEIGHT>` option to raise or
+    lower the TM results score for each specific TM server.
+
 - The Apertium MT backend has been dropped.
 
 
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -403,8 +403,10 @@ update_tmserver
 .. versionadded:: 2.7
 
 .. versionchanged:: 2.7.3 Renamed :option:`--overwrite` to :option:`--refresh`.
+   Now it is also possible to import translations from files.
 
-Updates the ``default`` server in :setting:`POOTLE_TM_SERVER`.  The command
+
+Updates the ``local`` server in :setting:`POOTLE_TM_SERVER`.  The command
 reads translations from the current Pootle install and builds the TM resources
 in the TM server.
 
@@ -412,6 +414,10 @@ If no options are provided, the command will only add new translations to the
 server. Use :option:`--refresh` to also update existing translations that have
 been changed, besides adding any new translation. To completely remove the TM
 and rebuild it adding all existing translations use :option:`--rebuild`.
+
+If no specific TM server is specified using :option:`--tm`, then the default
+``local`` TM will be used. If the specified TM server doesn't exist it will
+be automatically created for you.
 
 To see how many units will be loaded into the server use :option:`--dry-run`,
 no actual data will be loaded or deleted (the TM will be left unchanged):
@@ -421,6 +427,55 @@ no actual data will be loaded or deleted (the TM will be left unchanged):
     $ pootle update_tmserver --dry-run
     $ pootle update_tmserver --refresh --dry-run
     $ pootle update_tmserver --rebuild --dry-run
+
+
+This command also allows to read translations from files and build the TM
+resources in the external TM server. In order to do so it is mandatory to
+provide the :option:`--tm` and :option:`--project` options, along with some
+filename argument to import translations from. The provided project name will
+be displayed on TM matches in the translation editor:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --tm=libreoffice --project="LibreOffice 4.3 UI" TM_LibreOffice_4.3.gl.tmx
+
+
+By default the command will only add new translations to the server. To rebuild
+the server from scratch use :option:`--rebuild` to completely remove the TM and
+rebuild it before importing the translations:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --rebuild --tm=mozilla --project="Foo 1.7" foo.po
+
+
+Option :option:`--refresh` doesn't apply when adding translations from files
+on disk.
+
+To see how many units will be loaded into the server use :option:`--dry-run`,
+no actual data will be loaded:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --dry-run --tm=mozilla --project="Foo 1.7" foo.po
+   175045 translations to index
+
+
+This command is capable of importing translations from several files in
+different formats at once:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --tm=mozilla --project="Foo 1.7" foo.po bar.tmx eggs.xliff bacon.ts
+
+
+Use :option:`--target-language` to specify the target language ISO code for the
+imported translations in case it is not possible to guess it from the
+translation files or if the code is incorrect:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --target-language=af --tm=mozilla --project="Foo 1.7" foo.po bar.tmx
 
 
 .. _commands#vfolders:
--- a/docs/server/settings.rst
+++ b/docs/server/settings.rst
@@ -281,32 +281,62 @@ Translation environment configuration settings.
 
 .. setting:: POOTLE_TM_SERVER
 
-.. versionadded:: 2.7
-
 ``POOTLE_TM_SERVER``
+  .. versionadded:: 2.7
+
+  .. versionchanged:: 2.7.3
+
+     Added the :setting:`WEIGHT <POOTLE_TM_SERVER-WEIGHT>` option. Also added
+     another default TM used to import external translations from files.
+
+
   Default:
 
   .. code-block:: python
 
     {
-        'default': {
+        'local': {
             'ENGINE': 'pootle.core.search.backends.ElasticSearchBackend',
             'HOST': 'localhost',
             'PORT': 9200,
             'INDEX_NAME': 'translations',
+            'WEIGHT': 1,
+            'MIN_SCORE': 'AUTO',
+        },
+        'external': {
+            'ENGINE': 'pootle.core.search.backends.ElasticSearchBackend',
+            'HOST': 'localhost',
+            'PORT': 9200,
+            'INDEX_NAME': 'external-translations',
+            'WEIGHT': 0.9,
             'MIN_SCORE': 'AUTO',
         },
     }
+
 
   This is configured to access a standard Elasticsearch setup.  Change the
   settings for any non-standard setup.  Change ``HOST`` and ``PORT`` settings
   as required.
 
   Use ``MIN_SCORE`` to set the Levenshtein Distance score.  Set it to ``AUTO``
-  so that Eslasticsearch will adjust the required score depending on the length
+  so that Elasticsearch will adjust the required score depending on the length
   of the string being translated. Elasticsearch documentation provides further
   details on `Fuzzy matching
   <https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness>`_.
+
+  The default ``local`` TM is automatically updated every time a new
+  translation is submitted. The other TMs are not automatically updated so they
+  can be trusted to provide selected high quality translations.
+
+  .. setting:: POOTLE_TM_SERVER-INDEX_NAME
+
+  Every TM server must have its own unique ``INDEX_NAME``.
+
+  .. setting:: POOTLE_TM_SERVER-WEIGHT
+
+  ``WEIGHT`` provides a weighting factor to alter the final score for TM
+  results from this TM server. Valid values are between ``0.0`` and ``1.0``,
+  both included. Defaults to ``1.0`` if not provided.
 
 
 .. setting:: POOTLE_MT_BACKENDS
--- a/pootle/apps/pootle_app/management/commands/update_tmserver.py
+++ b/pootle/apps/pootle_app/management/commands/update_tmserver.py
@@ -217,6 +217,10 @@ class Command(BaseCommand):
 
         # If files to import have been provided.
         if len(args):
+            if self.is_local_tm:
+                raise CommandError('You cannot add translations from files to '
+                                   'a local TM.')
+
             self.target_language = options.pop('target_language')
             self.project = options.pop('project')
 
@@ -226,6 +230,9 @@ class Command(BaseCommand):
             self.parser = FileParser(index=self.INDEX_NAME,
                                      language=self.target_language,
                                      project=self.project)
+        elif not self.is_local_tm:
+            raise CommandError('You cannot add translations from database to '
+                               'an external TM.')
         else:
             self.parser = DBParser(index=self.INDEX_NAME)
 
--- a/pootle/apps/pootle_app/management/commands/update_tmserver.py
+++ b/pootle/apps/pootle_app/management/commands/update_tmserver.py
@@ -16,6 +16,7 @@ import sys
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
 
 from elasticsearch import helpers, Elasticsearch
+from translate.storage import factory
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
@@ -26,31 +27,14 @@ from pootle_store.models import Unit
 BULK_CHUNK_SIZE = 5000
 
 
-class Command(BaseCommand):
-    help = "Load Local Translation Memory"
-    option_list = BaseCommand.option_list + (
-        make_option('--refresh',
-                    action="store_true",
-                    dest='refresh',
-                    default=False,
-                    help='Process all items, not just the new ones, so '
-                         'existing translations are refreshed'),
-        make_option('--rebuild',
-                    action="store_true",
-                    dest='rebuild',
-                    default=False,
-                    help='Drop the entire index on start and update '
-                         'everything from scratch'),
-        make_option('--dry-run',
-                    action="store_true",
-                    dest='dry_run',
-                    default=False,
-                    help='Report only the number of translations to index '
-                         'and quit'),
-    )
+class DBParser(object):
 
-    def _parse_translations(self, **options):
+    def __init__(self, *args, **kwargs):
+        super(DBParser, self).__init__(*args, **kwargs)
+        self.INDEX_NAME = kwargs.pop('index', None)
 
+    def get_units(self, *filenames):
+        """Gets the units to import and its total count."""
         units_qs = Unit.simple_objects \
             .exclude(target_f__isnull=True) \
             .exclude(target_f__exact='') \
@@ -73,7 +57,121 @@ class Command(BaseCommand):
                 'store__translation_project__language__code'
             ).order_by()
 
-        total = units_qs.count()
+        return units_qs.iterator(), units_qs.count()
+
+    def get_unit_data(self, unit):
+        """Return dict with data to import for a single unit."""
+        fullname = (unit['submitted_by__full_name'] or
+                    unit['submitted_by__username'])
+
+        email_md5 = None
+        if unit['submitted_by__email']:
+            email_md5 = md5(unit['submitted_by__email']).hexdigest()
+
+        return {
+            '_index': self.INDEX_NAME,
+            '_type': unit['store__translation_project__language__code'],
+            '_id': unit['id'],
+            'revision': int(unit['revision']),
+            'project': unit['store__translation_project__project__fullname'],
+            'path': unit['store__pootle_path'],
+            'username': unit['submitted_by__username'],
+            'fullname': fullname,
+            'email_md5': email_md5,
+            'source': unit['source_f'],
+            'target': unit['target_f'],
+        }
+
+
+class FileParser(object):
+
+    def __init__(self, *args, **kwargs):
+        super(FileParser, self).__init__(*args, **kwargs)
+        self.INDEX_NAME = kwargs.pop('index', None)
+        self.target_language = kwargs.pop('language', None)
+        self.project = kwargs.pop('project', None)
+
+    def get_units(self, *filenames):
+        """Gets the units to import and its total count."""
+        units = []
+
+        for filename in filenames:
+            store = factory.getobject(filename)
+            if not store.gettargetlanguage() and not self.target_language:
+                raise CommandError("Unable to determine target language for "
+                                   "'%s'. Try again specifying a fallback "
+                                   "target language with --target-language" %
+                                   filename)
+
+            self.filename = filename
+            units.extend([unit for unit in store.units if unit.istranslated()])
+
+        return units, len(units)
+
+    def get_unit_data(self, unit):
+        """Return dict with data to import for a single unit."""
+        target_language = unit.gettargetlanguage()
+        if target_language is None:
+            target_language = self.target_language
+
+        return {
+            '_index': self.INDEX_NAME,
+            '_type': target_language,
+            '_id': unit.getid(),
+            'revision': 0,
+            'project': self.project,
+            'path': self.filename,
+            'username': None,
+            'fullname': None,
+            'email_md5': None,
+            'source': unit.source,
+            'target': unit.target,
+        }
+
+
+class Command(BaseCommand):
+    help = "Load Translation Memory with translations"
+    option_list = BaseCommand.option_list + (
+        make_option('--refresh',
+                    action='store_true',
+                    dest='refresh',
+                    default=False,
+                    help='Process all items, not just the new ones, so '
+                         'existing translations are refreshed'),
+        make_option('--rebuild',
+                    action='store_true',
+                    dest='rebuild',
+                    default=False,
+                    help='Drop the entire TM on start and update everything '
+                         'from scratch'),
+        make_option('--dry-run',
+                    action='store_true',
+                    dest='dry_run',
+                    default=False,
+                    help='Report the number of translations to index and quit'),
+        # External TM specific options.
+        make_option('--tm',
+                    action='store',
+                    dest='tm',
+                    default='local',
+                    help="TM to use. TM must exist on settings. TM will be "
+                         "created on the server if it doesn't exist"),
+        make_option('--target-language',
+                    action='store',
+                    dest='target_language',
+                    default='',
+                    help="Target language to fallback to use in case it can't "
+                         "be guessed for any of the input files."),
+        make_option('--project',
+                    action='store',
+                    dest='project',
+                    default='',
+                    help='Project to use when displaying TM matches for this '
+                         'translations.'),
+    )
+
+    def _parse_translations(self, *args, **options):
+        units, total = self.parser.get_units(*args)
 
         if total == 0:
             self.stdout.write("No translations to index")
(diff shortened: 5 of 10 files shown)


unho:
Added some new code, docs, and did some extensive changes on the command.

Please review.

docs/features/translation_memory.rst
```diff
@@ -74,4 +78,77 @@ Once populated Pootle will keep Local TM up-to-date.
...
+<https://www.elastic.co/products/elasticsearch>`_ and populate the initial
+translation memory using the :djadmin:`update_external_tmserver` command:
+
+.. code-block:: bash
+
+   (env) $ pootle update_external_tmserver af.po gl.tmx
+
+
+It is possible to have several Elasticsearch based external TM servers working
+at once, along with Elasticsearch based local TM servers. In order to do so
```
dwaynebailey: an ES based local TM server.. surely, singular

docs/features/translation_memory.rst
```diff
@@ -74,4 +78,77 @@ Once populated Pootle will keep Local TM up-to-date.
...
+            'INDEX_NAME': 'whatever',
+            'TM_NAME': 'LibreOffice 4.2',
+            'WEIGHT': 0.9,
+            'IS_LOCAL': False,
+            'MIN_SCORE': 'AUTO',
+        },
+    }
+
+Make sure ``IS_LOCAL`` is set to ``False`` and that ``INDEX_NAME`` is not used
+by any other TM server.
```
dwaynebailey: is unique.

dwaynebailey: ..iq unique

docs/features/translation_memory.rst
```diff
@@ -74,4 +78,77 @@ Once populated Pootle will keep Local TM up-to-date.
...
+
+Other values that you might want to tweak are ``TM_NAME`` (used to identify the
+TM in the translation editor) and ``WEIGHT`` which is used to change the
+relevance of the TM results in relation to other TM servers (valid values are
+between 0 and 1).
+
+To use them you will need to install
+`Elasticsearch <https://www.elastic.co/products/elasticsearch>`_ and populate
+them using the :djadmin:`update_external_tmserver` command:
```
dwaynebailey: Repeated paragraph

docs/features/translation_memory.rst
```diff
@@ -74,4 +78,77 @@ Once populated Pootle will keep Local TM up-to-date.
...
+explicitly use the :djadmin:`update_external_tmserver` command, giving you full
+control of which translations make into them.
+
+.. note:: Elasticsearch depends on Java. Note that some systems may ship with
+  OpenJDK, however `elasticsearch recommends using Oracle JDK
+  <https://www.elastic.co/guide/en/elasticsearch/reference/1.6/setup-service.html#_installing_the_oracle_jdk>`_.
+
+
+You may want to disable Amagama by setting :setting:`AMAGAMA_URL` to ``''`` if
+you are using Elasticsearch TM, though the two will operate together.
```
dwaynebailey: This is repeating what is already said in local TM.  And the pragraph above.  Rather change structure then repeating the same stuff that someone will forget to maintain.

docs/server/settings.rst
```diff
@@ -294,20 +298,50 @@ Translation environment configuration settings.
             'HOST': 'localhost',
```
dwaynebailey: Why not change this entry to ``local`` instead of ``default``, since default doesn't mean anything anymore

docs/server/settings.rst
```diff
@@ -294,20 +298,50 @@ Translation environment configuration settings.
...
   of the string being translated. Elasticsearch documentation provides further
   details on `Fuzzy matching
   <https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness>`_.
 
+  ``IS_LOCAL`` tells whether the TM is a local TM or an external TM. Local TMs
+  are populated using translations done within Pootle, and they are
+  automatically updated every time a new translation is sent. External TMs are
+  populated using translations coming from files holding translations, such as
+  regular translation files or translation memories (TMX files). External TMs
+  are thus not automatically updated.
```
dwaynebailey: Which is why it should be called ``WRITABLE``.  It is quite possible that a remote TM will be writable.  ``IS_LOCAL`` doesn't describe what you just wrote.

docs/server/settings.rst
```diff
@@ -281,9 +281,13 @@ Translation environment configuration settings.
 
 .. setting:: POOTLE_TM_SERVER
 
-.. versionadded:: 2.7
```
dwaynebailey: I'm not sure you want to lose that.  Without it yu don't know when this feature actually arrived.

docs/server/settings.rst
```diff
@@ -294,20 +298,50 @@ Translation environment configuration settings.
...
+  regular translation files or translation memories (TMX files). External TMs
+  are thus not automatically updated.
+
+  ``INDEX_NAME`` must be unique, so every TM server must have its own different
+  index.
+
+  ``TM_NAME`` is used to identify the source TM for every TM match in the
+  translation editor, but only for external TMs.
+  
+  ``WEIGHT`` provides a way to change the relevance for all the TM results in
```
dwaynebailey: s/relevance/score

docs/server/settings.rst
```diff
@@ -294,20 +298,50 @@ Translation environment configuration settings.
...
+  relation to other TM servers. Valid values are between ``0`` and ``1``.
```
dwaynebailey: 0.0 and 1.0 I think communicates that its a float.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+    def _parse_translations(self, *args, **options):
+        units = []
+
+        for filename in args:
+            store = factory.getobject(filename)
+            if (not store.gettargetlanguage() and
+                not self.target_language and
+                not options['dry_run']):
+
+                raise CommandError("Unable to get target language for '%s'. "
```
dwaynebailey: s/get/determine

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+    def _parse_translations(self, *args, **options):
+        units = []
+
+        for filename in args:
+            store = factory.getobject(filename)
+            if (not store.gettargetlanguage() and
+                not self.target_language and
+                not options['dry_run']):
```
dwaynebailey: ``--dry-run`` should fail just like any other run.  All it means is that it doesn't import things.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python
```
dwaynebailey: This whole command smacks of code repeats.  Refactor please.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+                'fullname': None,
+                'email_md5': None,
+                'source': unit.source,
+                'target': unit.target,
+            }
+
+    def handle(self, *args, **options):
+
+        if not getattr(settings, 'POOTLE_TM_SERVER', False):
+            raise CommandError('POOTLE_TM_SERVER setting is missing.')
```
dwaynebailey: Surely the checks catch these?

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+
+        if not len(args):
+            raise CommandError('Some file to import must be specified.')
+
+        self.target_language = options.pop('target_language', '')
+        tm_server = options.pop('tm_server', '')
```
dwaynebailey: Set proper defaults i.e. None in ``make_options`` so that the code is more reable.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+
+        if not tm_server:
```
dwaynebailey: Start with the default in ``make_option`` then you can drp all this setup code.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+            # Use default external TM.
+            tm_server = 'external'
+
+        try:
+            self.tm_settings = settings.POOTLE_TM_SERVER[tm_server]
+        except KeyError:
+            raise CommandError('Specified TM is not on POOTLE_TM_SERVER.')
```
dwaynebailey: s/TM/Translation Memory

is not defined in...

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+
+        if 'IS_LOCAL' not in self.tm_settings:
+            raise CommandError('IS_LOCAL missing for specified TM server.')
```
dwaynebailey: Why do we need that setting? Just have a valid default value and please lets call this ``WRITABLE``

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+        elif self.tm_settings['IS_LOCAL']:
+            raise CommandError('Translations can only be imported into an '
+                               'external TM')
```
dwaynebailey: I honestly don't think this is true.  They're just resources.

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
                     id="pootle.W015",
                 ))
 
+    if settings.POOTLE_TM_SERVER:
+        tm_indexes = []
+
+        for server in settings.POOTLE_TM_SERVER:
```
dwaynebailey: I can't help but feel that there has to be a better way to check these configurations.

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                    _("Duplicate '%s' INDEX_NAME in POOTLE_TM_SERVER.",
+                      settings.POOTLE_TM_SERVER[server]['INDEX_NAME']),
+                    hint=_("Set different INDEX_NAME for all servers in "
+                           "POOTLE_TM_SERVER."),
+                    id="pootle.C009",
+                ))
+            else:
+                tm_indexes.append(settings.POOTLE_TM_SERVER[server]['INDEX_NAME'])
+
+            if 'WEIGHT' not in settings.POOTLE_TM_SERVER[server]:
```
dwaynebailey: ``WEIGHT`` should have a default value if not set which should be 1.0.

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                errors.append(checks.Warning(
+                    _("POOTLE_TM_SERVER['%s'] has no WEIGHT.", server),
+                    hint=_("Set a WEIGHT for POOTLE_TM_SERVER['%s'].", server),
+                    id="pootle.W018",
+                ))
+            elif (settings.POOTLE_TM_SERVER[server]['WEIGHT'] < 0 or
+                  settings.POOTLE_TM_SERVER[server]['WEIGHT'] > 1):
+                errors.append(checks.Critical(
+                    _("POOTLE_TM_SERVER['%s'] doesn't have a WEIGHT less than "
```
dwaynebailey: WEIGHT must be between 0 and 1

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                      "0 or greater than 1.", server),
+                    hint=_("Set a WEIGHT between 0 and 1 (both included) for "
+                           "POOTLE_TM_SERVER['%s'].", server),
+                    id="pootle.C010",
+                ))
+
+            if 'IS_LOCAL' not in settings.POOTLE_TM_SERVER[server]:
+                errors.append(checks.Critical(
+                    _("POOTLE_TM_SERVER['%s'] has no IS_LOCAL.", server),
```
dwaynebailey: This reallly shouldn't be a requirement or be Critical.  Set sensible defaults.

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                    hint=_("Set IS_LOCAL for POOTLE_TM_SERVER['%s'].", server),
+                    id="pootle.C011",
+                ))
+            elif not isinstance(settings.POOTLE_TM_SERVER[server]['IS_LOCAL'], bool):
+                errors.append(checks.Warning(
+                    _("POOTLE_TM_SERVER['%s']['IS_LOCAL'] doesn't have a "
+                      "boolean value.", server),
```
dwaynebailey: ...should be True or False.

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                    hint=_("Set a boolean value for "
+                           "POOTLE_TM_SERVER['%s']['IS_LOCAL'].", server),
+                    id="pootle.W019",
+                ))
+
+            if 'TM_NAME' not in settings.POOTLE_TM_SERVER[server]:
```
dwaynebailey: You shouldn't be required to give TMs a human readable name

pootle/checks.py
```diff
@@ -262,6 +262,73 @@ def check_settings(app_configs=None, **kwargs):
...
+                errors.append(checks.Critical(
+                    _("POOTLE_TM_SERVER['%s'] has no TM_NAME.", server),
+                    hint=_("Set TM_NAME for POOTLE_TM_SERVER['%s'].", server),
+                    id="pootle.C012",
+                ))
+            elif not settings.POOTLE_TM_SERVER[server]['TM_NAME']:
```
dwaynebailey: These two are the same checks.

pootle/core/search/base.py
```diff
@@ -16,6 +16,8 @@
 class SearchBackend(object):
     def __init__(self, config_name=None, **kwargs):
         self._setup_settings(config_name)
+        self.weight = 1
```
dwaynebailey: You've set this as , so why are you insisting that it be explicitly set elsewhere?

pootle/settings/60-translation.conf
```diff
@@ -66,6 +66,35 @@ POOTLE_TM_SERVER = {
...
+        # For valid values for MIN_SCORE see
+        # http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/common-options.html#fuzziness
+        'MIN_SCORE': 'AUTO',
+    },
+    'external': {
+        'ENGINE': 'pootle.core.search.backends.ElasticSearchBackend',
+        'HOST': 'localhost',
+        'PORT': 9200,
+        'INDEX_NAME': 'external-translations',
+        # Name of the Translation Memory.
```
dwaynebailey: Don't repeat comments it only makes for more noise, maintenance burdern and confusion to someone setting this up.  Comment the first instance fully instead.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+            target_language = unit.gettargetlanguage()
+            if target_language is None:
+                target_language = self.target_language
+
+            yield {
+                "_index": self.INDEX_NAME,
+                "_type": target_language,
+                "_id": unit.getid(),
+                'revision': 0,
+                'project': self.tm_settings['TM_NAME'],
```
dwaynebailey: Alarm.  This wasn't the design goal.  A single external tm index must be able to hold multiple projects, just as local tm does.  Using ``TM_NAME`` to achieve this is problematic.  If you import 20 external TM resources you will have to setup 20 TM servers.  That was never the idea.  You should be ablt to have 1 external TM server and load named projects into those.

pootle/templates/editor/units/xhr_tm.html
```diff
@@ -9,7 +9,9 @@
           dir="<%= store.source_dir %>"><%= PTL.editor.doDiff(unit.source[0], sugg.source) %></div>
         <div class="suggestion-translation" lang="<%= store.target_lang %>"
           dir="<%= store.target_dir %>"><%= PTL.utils.fancyHl(sugg.target) %></div>
-        <% if (sugg.project) { %>
+        <% if (!sugg.is_local) { %>
```
dwaynebailey: TM is TM.  Overloading the user with information about external or internal is not helpful.  They likely already know this by the project name.  Lets have type of string that converys the information.  The only difference would be that external in our case would not know the contributors name (though I can imagine some formats could).

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
```diff
@@ -0,0 +1,151 @@
...
+                    action='store',
+                    dest='tm_server',
+                    help="TM to use. TM must exist on settings. TM will be "
+                         "created on the server if it doesn't exist"),
+        make_option('--target-language',
+                    action='store',
+                    dest='target_language',
+                    help="Target language to fallback to use in case it can't "
+                         "be guessed for any of the input files."),
+    )
```
dwaynebailey: YOu should be able to supply a project name as an option, see later as I diagree that one tm == one project.

julen:
As I see this is a clone of the existing local TM functionality with two/three differences:
* The index is not auto-populated
* It allows to import translations from files
* Results from multiple configured indexes are pondered

Is my understanding correct?

dwaynebailey:
@julen that sounds about correct.

docs/server/settings.rst
julen: If this is needed, for me `AUTO_UPDATE` describes better the intended result.

In any case, all indexes are writable by definition, and the distinction between local/external seems completely unnecessary: I cannot think of a use-case where deployments would want to auto-update translations to different indexes simultaneously. The fact that an index is auto-updated could potentially be signaled by index's naming convention.

julen:
Ok, my initial reaction to this PR has been that there's quite a bit of duplication that can be avoided. As a start I see no need for a specialized `update_external_tmserver` command when this functionality can be easily integrated into the existing one. I see quite a bit of duplicated logic, and this affects docs as well. Think of maintenance burden.

julen:
Regarding `TM_NAME`, it results misleading to me: it's writing such information into the existing `project` field, hence assuming a mapping between a configured index and a project. I'd rather name this `DISPLAY_NAME` and index it as a separate field, making it also useful for the situation where DB-populated data and externally-loaded data co-exist. So no need for special-casing. The UI could display messages such as:

`DISPLAY_NAME` unset, auto-updated, populated from DB (current scenario)
> Translated by \<fullname\> in \<project\> project

`DISPLAY_NAME = 'Pootle'`, auto-updated, populated from DB
> Translated by \<fullname\> in Pootle (\<project\> project)

`DISPLAY_NAME = 'GNOME'`, not updated automatically, populated from external files
> Translated in GNOME

I believe this extra field could be added independently of this PR.

docs/server/settings.rst
unho: Will use `AUTO_UPDATE`.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
unho: Then this will require to bring back `--tm-name` option (better if renamed to `--project`) that I removed in a previous iteration. As a consequence we might safely drop `TM_NAME` from the settings.

pootle/checks.py
unho: They are not. The first checks if the server settings have a 'TM_NAME' key while the second checks that its value is actually something.

But I am dropping `TM_NAME` from settings, so nevermind.

pootle/core/search/base.py
unho: Despite we fallback to 1 we are not telling that anywhere since I feel that it is preferable to have complete settings that explicitly state it. Thus the check.

docs/server/settings.rst
unho: Recovered back a few lines below.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
unho: IMHO when providing `--dry-run` we should not fail because it is is not possible to determine the target language, since we just want to report the number of translations to index.

pootle/core/search/base.py
dwaynebailey: Put it in the config decription. Boring people with extra info is not helpful when they don't need to config it if they don't need or understand it.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
unho: It is necessary to tell if the server is either local or external, so when running this command we don't inadvertently push translations to a local TM.

pootle/checks.py
unho: `SearchBackend` sets it.

unho: But as told somewhere else checks are to ensure complete and explicit settings are used.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
dwaynebailey: Sounds good.

julen: Oh, I missed this comment. You can probably disregard [my general comment on `TM_NAME`](https://github.com/translate/pootle/pull/4128#issuecomment-145881223) if the goals differ from the code.

docs/server/settings.rst
julen: It'd be great to hear from others (@translate/core); this seems like a potentially redundant setting.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
unho: Will do since I brought back `--project` so people can push different files to a TM but specifying different projects. Thus a `libreoffice` TM can be pushed translations for project "LibreOffice 4.3 UI" and "LibreOffice 4.4 Docs". Or even mix completely different projects, for example Inkscape and XChat.

docs/server/settings.rst
unho: I noticed that renaming to `AUTO_UPDATE` might allow in the future to have local TMs that are not automatically updated and external TMs that can be automatically updated. This feels like a mess, but I am not sure if this is what @dwaynebailey had in mind.

pootle/core/search/base.py
unho: Done.

pootle/templates/editor/units/xhr_tm.html
unho: Changes in this template are now gone since we ask to provide a project name when importing the translations.

pootle/apps/pootle_app/management/commands/update_external_tmserver.py
dwaynebailey: No I think ``--dry-run`` means run the command in a non-destructive way.  Running it to see if things work then running it and they fail is broken in my camp.

unho: Ok.

docs/server/settings.rst
dwaynebailey: I think it says 2.7.3 versionchanged, which isn't the same thing

unho: It says first versionadded 2.7 and then versionchanged 2.7.3 listing the changes.

docs/server/settings.rst
unho: Not a bad idea.

docs/server/settings.rst
dwaynebailey: I'm OK with ``AUTO_UPDATE`` but I'm also OK to use the name ``local`` instead of ``default`` to signal.  Though I think I prefer the first as there isn't any inferred behaviour.

unho: @dwaynebailey Will need clarification on if we should allow to push translations from files to local TMs, or if we should allow external TMs to be automatically updated.

unho:
Pushed update to address most of the concerns. Still have to refactor the management command to extract common behavior with `update_tmserver`.

pootle/checks.py
unho: I don't like it either, but can't think of a better way right now.

dwaynebailey:
@unho I'm with @julen in that I don't think we need a second command and that the commands do almost exactly the same thing

docs/server/settings.rst
dwaynebailey: My bad, sorry for the noise

docs/server/settings.rst
unho: On a second thought decided that it is best to keep it as `default`.

unho:
This includes a commit that is in PR #4132.

Added refactoring of management commands. I feel like if #4133 should be fixed in order to merge them into a single `update_tmserver` management command, since right now there is still a bit of not shared code.

docs/server/settings.rst
unho: Reason to keep is to keep backwards compatibility and renaming to `local` actually doesn't contribute anything.

docs/server/settings.rst
unho: It would be great if GitHub allowed to hide this discussion threads...

dwaynebailey: It would, just delete them in this case.

docs/server/settings.rst
dwaynebailey: I agree on the backward, but actually it does contribute as it makes it clear what the TM does, ``default`` indicates nothing.

unho: Lets try to see this from a different angle. If we merge both management commands that means that you you will be able to specify the TM using `--tm` and since the settings do not enforce in any way that it can only be a single local TM, then having `local` instead of `default` is actually not contributing anything either. At least `default` is now treated as the default TM.

unho:
Command have been merged together.

unho:
@translate/core Merged both commands together. Docs have been updated. Reviews appreciated.

dwaynebailey:
LGTM with some comments, sorry did them against the commits.  I did a light review of the docs and only picked up on items where the wording was confusing or a bit difficult to understand.

julen:
I already [showed my doubts about having an `AUTO_UPDATE` setting](https://github.com/translate/pootle/pull/4128#discussion_r41271640), and reviewing this again has only increased such doubts.

On the one hand, auto-updating from DB **only** applies to the configured local index. So correct me if I'm wrong, but I don't see an scenario where deployments would want to have multiple local indexes auto-updated automatically — it'd just contribute to data redundancy.

Having said that, I'd suggest to require the local TM to be under the `POOTLE_TM_SERVER['local']` entry by convention (meaningful naming has already been [proposed by @dwaynebailey too](https://github.com/translate/pootle/pull/4128#discussion_r41207817)), and documented it as such.

Therefore there's no need for the `AUTO_UPDATE` setting. If for some reason auto-updating wants to be turned off (I cannot see under which circumstances this applies for a TM already fed from the DB), it's just a matter of renaming away from `local` to anything else.

I'm holding my review to hear further opinions on this.

unho:
IMHO it makes sense dropping `AUTO_UPDATE` and instead rely on `POOTLE_TM_SERVER['local']`. I would also like to see other opinions before proceeding with the removal.

unho:
Actually I think it is better to keep the naming, so instead of relying on `local` lets use already existing `default`.

julen:
Default can mean anything, `local` is explicit and meaningful. This was already [pointed out before](https://github.com/translate/pootle/pull/4128#discussion_r41605176).

dwaynebailey:
@unho just posting here to make things clear.  I agree with @julen ``local`` is explicit and clear and that we don't need ``AUTO_UPDATE``.

unho:
Removed `AUTO_UPDATE`.

docs/features/translation_memory.rst
```diff
@@ -47,17 +51,50 @@ If you want to setup and connect to your own TM server then the
 :setting:`AMAGAMA_URL` will allow you to point to a private TM server.
 
 
+.. _translation_memory#elasticsearch_based_tms:
+
+Elasticsearch based TMs
```
julen: *Elasticsearch-based*. Ditto elsewhere.

docs/features/translation_memory.rst
```diff
@@ -47,17 +51,50 @@ If you want to setup and connect to your own TM server then the
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 2.7
+
+Pootle can also retrieve TM matches stored on Elasticsearch based TM servers.
+These Elasticsearch based TMs require
```
julen: *These servers require*

docs/features/translation_memory.rst
```diff
@@ -47,17 +51,50 @@ If you want to setup and connect to your own TM server then the
...
+
+- **External TMs**: (several) are populated from translation files specifically
+  provided by the server admins, and are not automatically updated.
+
+
+Both local and external TM settings can be adjusted in
+:setting:`POOTLE_TM_SERVER`.
+
+You may want to disable Amagama by setting :setting:`AMAGAMA_URL` to ``''`` if
+you are using Elasticsearch based TMs, though the two will operate together.
```
julen: *though both can operate together.*

docs/features/translation_memory.rst
```diff
@@ -47,17 +51,50 @@ If you want to setup and connect to your own TM server then the
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+++++++++++++++++++++++++++++
 
 .. versionadded:: 2.7
 
-By default this is configured to operate if local TM is available.  To use it
-you will need to install `Elasticsearch 
-<https://www.elastic.co/products/elasticsearch>`_ and populate the initial
-translation memory using the :djadmin:`update_tmserver` command:
+By default this is configured to operate if local TM is available. To use it
```
julen: I'm not sure this sentence makes sense. Is *Local TM is configured by default.* what you mean?

docs/features/translation_memory.rst
```diff
@@ -66,12 +103,67 @@ translation memory using the :djadmin:`update_tmserver` command:
...
 
-Local TM settings can be adjusted in :setting:`POOTLE_TM_SERVER`.
+Elasticsearch based external TMs
+++++++++++++++++++++++++++++++++
 
-You may want to disable Amagama by setting :setting:`AMAGAMA_URL` to ``''`` if
-you are using Elasticsearch local TM, though the two will operate together.
+.. versionadded:: 2.7.3
+
+By default this is configured to operate if any external TM is available. To
```
julen: Same as in local TM, not sure what this first sentence is supposed to mean.

docs/releases/dev.rst
```diff
@@ -34,8 +35,15 @@ Details of changes
 - The editor for static pages now highlights the content's markup
   (:issue:`3346`).
 - Pulled latest translations.
-- :djadmin:`update_tmserver`: renamed :option:`--overwrite` to
-  :option:`--refresh`.
+- :djadmin:`update_tmserver`:
+
+  - Renamed :option:`--overwrite` to :option:`--refresh`.
+  - Added support for Elasticsearch-based external Translation Memory servers,
+    which can be populated using translations from on disk selected translation
```
julen: *from translation files* suffices.

docs/server/commands.rst
```diff
@@ -403,6 +403,9 @@ update_tmserver
 .. versionadded:: 2.7
 
 .. versionchanged:: 2.7.3 Renamed :option:`--overwrite` to :option:`--refresh`.
+   Also now it is possible to populate TMs using translations from specified
```
julen: Isn't *Now it is also possible to import translations from files.* much simpler?

docs/server/commands.rst
```diff
@@ -413,6 +416,10 @@ server. Use :option:`--refresh` to also update existing translations that have
 been changed, besides adding any new translation. To completely remove the TM
 and rebuild it adding all existing translations use :option:`--rebuild`.
 
+If no specific TM server is specified using :option:`--tm`, then the default
+``default`` TM will be used. If the specified TM server doesn't exist it will
```
julen: Doesn't TM default to local TM, hence `local`? If so, you might wanna check other occurrences of `default` in the docs.

docs/server/commands.rst
```diff
@@ -423,6 +430,55 @@ no actual data will be loaded or deleted (the TM will be left unchanged):
     $ pootle update_tmserver --rebuild --dry-run
 
 
+This command also allows to read translations from translation files on disk
```
julen: *to read translations from files and build the TM resources*...

docs/server/commands.rst
```diff
@@ -423,6 +430,55 @@ no actual data will be loaded or deleted (the TM will be left unchanged):
...
+To see how many units will be loaded into the server use :option:`--dry-run`,
+no actual data will be loaded:
+
+.. code-block:: bash
+
+   (env) $ pootle update_tmserver --dry-run --tm=mozilla --project="Foo 1.7" foo.po
+   175045 translations to index
+
+
+This command is capable of importing translations from several files with
```
julen: *in different formats*

docs/server/settings.rst
```diff
@@ -281,33 +281,59 @@ Translation environment configuration settings.
...
 
   Use ``MIN_SCORE`` to set the Levenshtein Distance score.  Set it to ``AUTO``
-  so that Eslasticsearch will adjust the required score depending on the length
+  so that Elasticsearch will adjust the required score depending on the length
   of the string being translated. Elasticsearch documentation provides further
   details on `Fuzzy matching
   <https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness>`_.
 
+  The default ``local`` TM is automatically updated every time a new
+  translation is sent. The other TMs are not automatically updated so they can
```
julen: *is submitted.*

docs/server/settings.rst
```diff
@@ -281,33 +281,59 @@ Translation environment configuration settings.
...
+  be trusted to provide selected high quality translations.
+
+  ``INDEX_NAME`` every TM server must have its own unique index.
```
julen: It seems like some punctuation is missing or the sentence is incomplete.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -73,7 +57,121 @@ def _parse_translations(self, **options):
...
+                    unit['submitted_by__username'])
+
+        email_md5 = None
+        if unit['submitted_by__email']:
+            email_md5 = md5(unit['submitted_by__email']).hexdigest()
+
+        return {
+            "_index": self.INDEX_NAME,
+            "_type": unit['store__translation_project__language__code'],
+            "_id": unit['id'],
```
julen: I know this is a copy&paste, however I'd advise against mixing quoting in the same dict.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -73,7 +57,121 @@ def _parse_translations(self, **options):
...
+            target_language = self.target_language
+
+        return {
+            "_index": self.INDEX_NAME,
+            "_type": target_language,
+            "_id": unit.getid(),
+            'revision': 0,
+            'project': self.project,
+            'path': self.filename,
+            'username': 'anonymous',
```
julen: Shouldn't this be `nobody`? Also not sure if `fullname` should be set

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -86,63 +184,66 @@ def _parse_translations(self, **options):
...
 
         if i != total:
             self.stdout.write("Expected %d, loaded %d." % (total, i))
 
-
-    def handle(self, *args, **options):
+    def _get_settings(self, *args, **options):
         if not getattr(settings, 'POOTLE_TM_SERVER', False):
-            raise CommandError("POOTLE_TM_SERVER is missing from your settings.")
+            raise CommandError('POOTLE_TM_SERVER setting is missing.')
```
julen: `ImproperlyConfiguredError`?

julen: Same few lines below.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -86,63 +184,66 @@ def _parse_translations(self, **options):
...
-                'target': unit['target_f'],
-            }
+            yield self.parser.get_unit_data(unit)
 
         if i != total:
             self.stdout.write("Expected %d, loaded %d." % (total, i))
 
-
-    def handle(self, *args, **options):
+    def _get_settings(self, *args, **options):
```
julen: This method doesn't return anything (an implicit `None` aside); it doesn't make sense to call it as such (`_get_...`)

pootle/checks.py
```diff
@@ -262,6 +262,50 @@ def check_settings(app_configs=None, **kwargs):
                     id="pootle.W015",
                 ))
 
+    if settings.POOTLE_TM_SERVER:
+        tm_indexes = []
+
+        if not 'local' in settings.POOTLE_TM_SERVER:
```
julen: Why force a `local` TM on everyone?

pootle/checks.py
julen: If there's a default why bother/flood server admins?

docs/features/translation_memory.rst
unho: This was copied from previously existing paragraph. Just reworded it.

docs/features/translation_memory.rst
unho: Reworded as well.

docs/server/commands.rst
unho: Yes. Changed.

docs/server/commands.rst
unho: Changed, and checked for other appearances.

julen:
Worth stressing in the release notes that `default` has been renamed to `local`, and the change in settings is necessary for those who want to keep the auto-updating behavior.

docs/server/settings.rst
unho: Fixed.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Fixed.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Renamed to `_set_...`.

unho: Sorry, renamed to `_initialize`.

pootle/checks.py
unho: Removed.

docs/server/settings.rst
```diff
@@ -281,33 +281,59 @@ Translation environment configuration settings.
...
+  
+  ``WEIGHT`` provides a weighting factor to alter the final score for TM
```
dwaynebailey: Use ``.. setting:: POOTLE_TM_SERVER-WEIGHT`` and reference it with ``:setting:`WEIGHT <POOTLE_TM_SERVER-WEIGHT>```.  Same with the other options and references that have been introduced.

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: @julen I guess depends on what the UI shows.  But since this isn't Pootle's `nobody` I'm not sure it makes sense.  But I don't think ``anonymous`` is good either as I'm not sure what will display.  My guess is that it probably should be `None` since we don't know and the UI should do the right thing.

pootle/checks.py
```diff
@@ -262,6 +262,50 @@ def check_settings(app_configs=None, **kwargs):
...
+                tm_indexes.append(settings.POOTLE_TM_SERVER[server]['INDEX_NAME'])
+
+            if 'WEIGHT' not in settings.POOTLE_TM_SERVER[server]:
+                errors.append(checks.Warning(
+                    _("POOTLE_TM_SERVER['%s'] has no WEIGHT.", server),
+                    hint=_("Set a WEIGHT for POOTLE_TM_SERVER['%s'].", server),
+                    id="pootle.W018",
+                ))
+            elif not (0.0 <= settings.POOTLE_TM_SERVER[server]['WEIGHT'] <= 1.0):
+                errors.append(checks.Critical(
```
dwaynebailey: This isn't critical.  Rather set bounds where this is used in real life so that its always between 0.0-1.0.  Also this isn't critical its at best a Warning.  Critical means things stop working.  Stopping Pootle because the TM is slightly misconfigure isn't a good reason to not run Pootle.

pootle/checks.py
```diff
@@ -262,6 +262,50 @@ def check_settings(app_configs=None, **kwargs):
                     id="pootle.W015",
                 ))
 
+    if settings.POOTLE_TM_SERVER:
+        tm_indexes = []
+
+        if 'local' not in settings.POOTLE_TM_SERVER:
+            errors.append(checks.Critical(
+                _("POOTLE_TM_SERVER['local'] is missing.", server),
```
dwaynebailey: Seriously I don't think we even need to warn.  There might be a reason you don't have it on.  We should simply have a good basic config, we do.  Work correctly without any present, I think we do.

pootle/checks.py
```diff
@@ -262,6 +262,50 @@ def check_settings(app_configs=None, **kwargs):
...
+                hint=_("Set POOTLE_TM_SERVER['local']."),
+                id="pootle.C008",
+            ))
+
+        for server in settings.POOTLE_TM_SERVER:
+            if 'INDEX_NAME' not in settings.POOTLE_TM_SERVER[server]:
+                errors.append(checks.Critical(
```
dwaynebailey: Not Critical.  We just then don't have TM and we Warn.

pootle/checks.py
```diff
@@ -262,6 +262,50 @@ def check_settings(app_configs=None, **kwargs):
...
+                    _("POOTLE_TM_SERVER['%s'] has no INDEX_NAME.", server),
+                    hint=_("Set an INDEX_NAME for POOTLE_TM_SERVER['%s'].",
+                           server),
+                    id="pootle.C009",
+                ))
+            elif settings.POOTLE_TM_SERVER[server]['INDEX_NAME'] in tm_indexes:
+                errors.append(checks.Critical(
```
dwaynebailey: Not Critical, I'd guess disable TM in that case.  But not sure how hard this is in the code.  So maybe leave this for now.

pootle/core/search/backends/elasticsearch.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, config_name):
         super(ElasticSearchBackend, self).__init__(config_name)
         self._es = self._get_es_server()
         self._create_index_if_missing()
+        if self._es is not None:
+            self.weight = self._settings.get('WEIGHT', self.weight)
```
dwaynebailey: Set the bounds here.

dwaynebailey:
Not sure if it is worth having a check to see if the first TM is still called 'default' and not 'local'.  Because if it isn't renamed it won't autoupdate.

docs/server/settings.rst
unho: Done.

pootle/checks.py
unho: This is already removed, but didn't push the changes yet.

pootle/checks.py
unho: Removing this check would require to adjust the code in other places in order to ensure we have a `INDEX_NAME`. I prefer to keep that check just here. And this is a critical error since otherwise the TM is unusable.

pootle/checks.py
unho: We are telling to have unique indexes to avoid having more than one entry in `POOTLE_TM_SERVERS` to point to the same index and inadvertently push stuff to the wrong TM. For example think of:

```
    {
        'local': {
            'INDEX_NAME': 'translations',
        },
        'whatever': {
            'INDEX_NAME': 'translations',
        },
    }
```

which would allow the `translations` index to be automatically updated, but also allows translations to be pushed from files.

pootle/checks.py
unho: This is a warning, and they will be warned only if they don't have set the right settings.

pootle/checks.py
unho: Will then downgrade this to warning and set this to `0.0` if a less than zero number is provided, or set to `1.0` if a greater than one is provided.

pootle/apps/pootle_app/management/commands/update_tmserver.py
julen: :+1: on UI doing the right thing and using `None`.

pootle/core/search/backends/elasticsearch.py
unho: Done.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: If `None` is set we get the following which works for me:

ab

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Actually I find preferable that the command just displays an error message than throwing a traceback. So I won't use `ImproperlyConfigured` exception here.

unho:
> Not sure if it is worth having a check to see if the first TM is still called 'default' and not 'local'. Because if it isn't renamed it won't autoupdate.

Not sure how to achieve that since you can have have a `default` TM that not necessarily is the old `default` and you don't necessarily want to have the `local` TM enabled.

dwaynebailey:
@unho not sure but it echos what @julen commented on that it needs to be stressed that this changed otherwise local tm won't be updating.

pootle/checks.py
dwaynebailey: Sure, but that is not a Critical issue.  Critical wil not allow Pootle to start, I don't see any of these as potential data loss so I wouldn't regard it as critical.

dwaynebailey:
@unho LGTM from me

pootle/checks.py
unho: Ok, downgrading to warning.

unho:
Double checked that all the concerns have been addressed or rejected with justification.

I already have on gtm, but waiting for another one given the problematic this PR was.

julen:
Ship it @unho.

Issue #4135, opened 2015-10-09
update_tmserver: Use index hot-swapping to keep TM available while updating it


unho:
See https://groups.google.com/forum/#!topic/elasticsearch/tlmU_7kSrfg and https://www.elastic.co/blog/changing-mapping-with-zero-downtime

From https://github.com/translate/pootle/issues/4120#issuecomment-144801389

PR #4139, opened 2015-10-12
update_tmserver: Clarify command behavior

### LINKED ISSUE #4136: update_tmserver: Refactor options
Right now `update_tmserver` accepts several options (`--rebuild`, `--overwrite`) but it is not clear what they are for. After a discussion with @dwaynebailey we outlined three scenarios:

1) Add translations added since last indexed revision. This creates the index if missing, and adds translations from start if there is no last indexed revision. This is just calling `update_tmserver`.
2) Destroy index if it exists, create the index, and add all translations. This is calling `update_tmserver --rebuild`.
3) Add all translations from start, regardless of if there is last indexed revision, so this updates previously indexed translations in case they were changed afterwards and keeps previously indexed translations that are now gone (e.g. translations from removed projects). This creates the index if missing (this is scenario 1) with no previous index). What makes this scenario different from scenario 2) is that the index is not destroyed. This is calling `update_tmserver --refresh`.

Note that `--dry-run` can also be used for any of those three scenarios. In this case no changes to the index are done (no destroy, no create, no update).

unho:
--overwrite has been renamed to --refresh as it is more explanatory.

Rewrote docs to make it clear what is the command behavior for each
scenario. The note on "TM server remains available when you rebuild"
has been removed as it is not true, see #4135.

Fixes #4136.

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -394,6 +394,7 @@ Translation Memory
 These commands allow you to setup and manage :doc:`Translation Memory
 </features/translation_memory>`.
 
+
 .. django-admin:: update_tmserver
 
 update_tmserver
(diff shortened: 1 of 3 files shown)


docs/server/commands.rst
```diff
@@ -401,17 +402,46 @@ update_tmserver
...
-By default the command will only add new translations to the server.  To
-rebuild the server from scratch use :option:`--rebuild`, this will completely
-remove the TM and rebuild it.  To ensure that the TM server remains available
-when you rebuild you can add :option:`--overwrite`.
+If no options are provided, the command will only add new translations to the
+server:
+
+.. code-block:: bash
+
+    $ pootle update_tmserver
```
julen: This command snippet adds no value. The same applies to the rest of the snippets.

docs/server/commands.rst
```diff
@@ -401,17 +402,46 @@ update_tmserver
 
 .. versionadded:: 2.7
 
+.. versionchanged:: 2.7.3 Renamed :option:`--overwrite` to :option:`--refresh`
+   and clarified the behavior of the command.
```
julen: *"clarified the behavior of the command"* sounds like avoidable noise. Same for the release notes.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -29,13 +29,12 @@
...
+        make_option('--refresh',
                     action="store_true",
-                    dest='overwrite',
+                    dest='refresh',
                     default=False,
-                    help='Process all items, not just the new ones (useful to '
-                         'overwrite properties while keeping the index in a '
-                         'working condition)'),
+                    help='Process all items, not just the new ones, so '
+                         'translations are refreshed'),
```
julen: *existing translations*?

julen:
Looks good to merge, provided comments have been addressed.

docs/server/commands.rst
unho: Removed all of them except the one for `--dry-run`. I think it is a good idea to at least keep one of the usage examples.

unho:
Addressed most of the comments. gtm?

dwaynebailey:
lgtm, not sure the ``--dry-run`` examples add any more value either, but I'm not blocking on that.

PR #4141, opened 2015-10-14
Static pages: implemented live preview

### LINKED ISSUE #3766: Add live preview when editing static documents
When editing static document, let's split the screen vertically if it's width is more than 1920px, and horizontally otherwise, and add a live rendering when contents of the HTML editor area change. In the preview mode, use the same wrapper divs to make sure we pick up all styles and render the page as close to the final output as possible.

Quick illustration: http://www.htmlinstant.com/ or http://htmledit.squarefree.com/

Mockup screenshot:
static_preview

julen:
This PR implements the live preview feature for static pages.

I'm putting this up for testing and further feedback. While at its current state this is fully functional, I'd still like to make a couple of changes before considering to merge it:
* Fix the styling of the content editing/previewing boxes and adjust their heights
* Make the `<ContentPreview />` component _dumb_ (i.e. do not hold any local state, just accept props)

Fixes #3766.

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -32,8 +32,8 @@ complete information.
 Details of changes
 ==================
 
-- The editor for static pages now highlights the content's markup
-  (:issue:`3346`).
+- The editor for static pages now highlights the content's markup and displays a
+  live preview of the rendered contents (:issue:`3346`, :issue:`3766`).
 - Pulled latest translations.
 - :djadmin:`update_tmserver`:
 
--- a/pootle/apps/staticpages/urls.py
+++ b/pootle/apps/staticpages/urls.py
@@ -39,7 +39,16 @@ admin_patterns = patterns('',
 )
 
 
+xhr_patterns = patterns('',
+    url(r'^preview/?$',
+        'staticpages.views.preview_content',
+        name='pootle-xhr-preview'),
+)
+
+
 urlpatterns = patterns('',
     url(r'^pages/',
         include(page_patterns)),
+    url(r'^xhr/',
+        include(xhr_patterns)),
 )
--- a/pootle/apps/staticpages/views.py
+++ b/pootle/apps/staticpages/views.py
@@ -17,6 +17,7 @@ from django.views.generic import (CreateView, DeleteView, TemplateView,
                                   UpdateView)
 
 from pootle.core.http import JsonResponse, JsonResponseBadRequest
+from pootle.core.markup.filters import apply_markup_filter
 from pootle.core.views import SuperuserRequiredMixin
 from pootle_misc.util import ajax_required
 
@@ -223,3 +224,16 @@ def legal_agreement(request):
 
     rendered_form = _get_rendered_agreement(request, form_class())
     return JsonResponse({'form': rendered_form})
+
+
+@ajax_required
+def preview_content(request):
+    """Returns content rendered based on the configured markup settings."""
+    if 'text' not in request.POST:
+        return JsonResponseBadRequest({
+            'msg': _('Text is missing'),
+        })
+
+    return JsonResponse({
+        'rendered': apply_markup_filter(request.POST['text']),
+    })
--- a/pootle/static/js/admin/general/components/ContentEditor.js
+++ b/pootle/static/js/admin/general/components/ContentEditor.js
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) Pootle contributors.
+ *
+ * This file is a part of the Pootle project. It is distributed under the GPL3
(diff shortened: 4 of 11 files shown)


dwaynebailey:
Works.  The live rendering is quite slow.  Might explain why github does it on another tab.

The other issue is that if you have content that is in any way long then the preview is not easily available as you need to scroll down.  IIRC the live preview was meant to be on the right.  But either on the right or below I can imagine space will always be an issue.

LGTM.

dwaynebailey:
Oh I tested with HTML and Markdown, both are fine.  Its nice to have code highlighting in the content, helps prevent some errors.

julen:
> The live rendering is quite slow

The server call to retrieve the rendered content is debounced at 500ms, I just adjusted it to 300ms.

> The other issue is that if you have content that is in any way long then the preview is not easily available as you need to scroll down.

This is one of the changes I want to make before merging as written above.

>  IIRC the live preview was meant to be on the right.

This depends on the viewport size. Currently the vertical split happens at widths larger than 1920px.

pootle/static/js/admin/general/components/ContentPreview.css
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) Pootle contributors.
+ *
+ * This file is a part of the Pootle project. It is distributed under the GPL3
+ * or later license. See the LICENSE file for a copy of the license and the
+ * AUTHORS file for copyright and authorship information.
+ */
```
unho: Do we put now the CSS files under the `js` directory?

unho:
Did some quick testing and seems to work. I have found no issue that hasn't been already raised.

pootle/static/js/admin/general/components/ContentPreview.css
julen: These are self-contained components, CSS and JS being a separation of technologies, not concerns. Concerns are being kept together.

julen:
All the issues that I was mentioning in the first message have been ironed out and I think this is in good shape to be merged. The live preview component now also avoids the server round-trip when the markup is HTML, and in such scenario the preview is immediate.

iafan:
@julen, is this live on stage?

julen:
Yep.

iafan:
Is it possible to display the "Allowed markup: HTML" and "Preview will be displayed here" placeholders as semi-transparent (gray) labels, aligned the same way (e.g. in the top left corner)?

iafan:
Also: when calculating the height of the both panes in horizontal split mode, can we make them take into account the height of the page parts above and below? The idea is to avoid vertical scrolling so that the entire page conent fits the window size.

These lines can probably be merged into one and displayed to the right of the Save  buttons:
 ```
Use <a href="#/page-slug">Link text</a> code to provide links between static pages. Such links will be rewritten with full paths to static pages.

In Announcement pages use /xx/ within URLs for a dynamic language code.
```

julen:
All the comments have now been addressed.

iafan:
> The idea is to avoid vertical scrolling so that the entire page content fits the window size.

I believe this one is still not fixed (page height is still bigger than viewport height).

julen:
It was requested for the horizontal split; are you checking the vertical split perhaps?

iafan:
Yes, this split thing is confusing, sorry. When in *vertical* split (one pane on the left, and one on the right) we want the entire page to fit.

Having said that, in horizontal mode page also doesn't fit vertically (tested in Chrome).

julen:
Can you share a screenshot and the screen resolution you have?

iafan:
screenclip

screenclip 1

iafan:
The first screenshot is from a full-screen mode. Display resolution is 2560x1440.

julen:
There's probably some misunderstanding as from our PM you confirmed we want to have the entire *form* visible in the viewport, not the entire *page*.

iafan:
I apologize. Entire page is what I had in mind. I want to avoid any page scrolling (unless the page is too small), because we have scrollable areas inside.

julen:
Cool, all clear now.

julen:
Implemented the requested changes (live on stage), and the resulting experience is much more pleasant and friendly — thanks for the great input, @iafan!

iafan:
I tested this on stage, and though the edit/preview area height is smaller than before, the entire page doesn't fully fit:

screenclip

Tested in Chrome and in Firefox. In Firefox, the 'Save' button is visible, but the page footer (around 50 px) is still below the viewport and causes vertical scrolling.

julen:
Fixed this and other tangential issues we discussed via IM.

iafan:
GTM!

PR #4144, opened 2015-10-16
update_tmserver: Add translations from directories

unho:


docs/server/commands.rst
```diff
@@ -462,11 +462,12 @@ no actual data will be loaded:
...
 
 This command is capable of importing translations from several files in
-different formats at once:
+different formats at once, either specified directly or by passing its
+container directory:
 
 .. code-block:: bash
 
-   (env) $ pootle update_tmserver --tm=mozilla --project="Foo 1.7" foo.po bar.tmx eggs.xliff bacon.ts
+   (env) $ pootle update_tmserver --tm=mozilla --project="Foo 1.7" foo.po bar.tmx fr/ eggs.xliff bacon.ts
```
dwaynebailey: Now we have dirs I think this line is confusing, rather limit it to two normal files and one dir at the end

docs/server/commands.rst
```diff
@@ -462,11 +462,12 @@ no actual data will be loaded:
 
 
 This command is capable of importing translations from several files in
-different formats at once:
+different formats at once, either specified directly or by passing its
+container directory:
```
dwaynebailey: This command is capable of importing translations in multiple formats from several files and directories at once.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -94,8 +94,25 @@ def __init__(self, *args, **kwargs):
     def get_units(self, *filenames):
         """Gets the units to import and its total count."""
         units = []
+        all_filenames = set()
 
         for filename in filenames:
+            if not os.path.exists(filename):
+                continue
```
dwaynebailey: We'd want to tell someone that their file doesn't exist

docs/server/commands.rst
unho: Do you mean reducing the example length or limiting the number of accepted arguments in `update_tmserver` command?

unho: Oh, the line.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Since this is outside the command `self.stdout.write()` can't be used, and raising a `CommandError` will stop the command immediately, so will try using just `print()`.

unho:
Addressed raised concerns.

pootle/apps/pootle_app/management/commands/update_tmserver.py
julen: You can use `self.stdout.write()` and friends if you init the parser objects with a reference to the stdout handler from the command.

unho:
Updated to use `self.stdout.write()`.

dwaynebailey:
lgtm

unho:
Will wait for another gtm before merging this.

julen:
Looks good to me; didn't test.

PR #4146, opened 2015-10-16
update_tmserver: Skip disabled projects by default

### LINKED ISSUE #3919: update_tmserver shouldn't use disabled projects
``update_tmserver`` command shouldn't use disabled projects unless they are specifically requested.

By default we shouldn't use them.  If the user supplies a command line option or uses a specific project name they we should load that project into the TM.

unho:
Fixes #3919.

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -38,6 +38,8 @@ Details of changes
 - :djadmin:`update_tmserver`:
 
   - Renamed :option:`--overwrite` to :option:`--refresh`.
+  - Translations from disabled projects must be explicitly included with
+    :option:`--include-disabled-projects`.
   - Added support for Elasticsearch-based external Translation Memory servers,
     which can be populated from translation files. This effectively brings the
     ability to display TM results from different TM servers, sorting them by
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -403,7 +403,8 @@ update_tmserver
 .. versionadded:: 2.7
 
 .. versionchanged:: 2.7.3 Renamed :option:`--overwrite` to :option:`--refresh`.
-   Now it is also possible to import translations from files.
+   Disabled projects' translations are no longer added by default. It is also
+   possible to import translations from files.
 
 
 Updates the ``local`` server in :setting:`POOTLE_TM_SERVER`.  The command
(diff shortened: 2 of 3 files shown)


dwaynebailey:
lgtm

unho:
Waiting for a second gtm.

docs/releases/dev.rst
```diff
@@ -38,6 +38,8 @@ Details of changes
 - :djadmin:`update_tmserver`:
 
   - Renamed :option:`--overwrite` to :option:`--refresh`.
+  - Translations from disabled project must be explicitly included with
```
julen: *projects*

docs/server/commands.rst
```diff
@@ -403,7 +403,8 @@ update_tmserver
 .. versionadded:: 2.7
 
 .. versionchanged:: 2.7.3 Renamed :option:`--overwrite` to :option:`--refresh`.
-   Now it is also possible to import translations from files.
+   Disabled projects translations are no longer added by default. It is also
```
julen: *projects'*

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -31,6 +31,7 @@ class DBParser(object):
 
     def __init__(self, *args, **kwargs):
         self.INDEX_NAME = kwargs.pop('index', None)
+        self.include_disabled_projects = kwargs.pop('disabled_projs', False)
```
julen: What about naming this `disabled_projects`? (FWIW we never use `projs`)

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -43,7 +44,14 @@ def get_units(self, *filenames):
                 'store',
                 'store__translation_project__project',
                 'store__translation_project__language'
-            ).values(
+            )
+
+        if not self.include_disabled_projects:
```
julen: I'd suggest to name the attribute `.exclude_disabled_projects`, that way the condition becomes `if self.exclude_disabled_projects:` and the cognitive load for the reader is lower.

julen:
GTM provided the suggestions have been addressed.

PR #4147, opened 2015-10-16
Fix several issues with update_tmserver

unho:


pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -200,8 +201,8 @@ def _initialize(self, *args, **options):
         try:
             self.tm_settings = settings.POOTLE_TM_SERVER[options.get('tm')]
         except KeyError:
-            raise CommandError('Specified Translation Memory is not defined '
-                               'in POOTLE_TM_SERVER.')
+            raise CommandError("Translation Memory '%s' is not defined in "
+                               "POOTLE_TM_SERVER setting." % options.get('tm'))
```
dwaynebailey: This is a ``ImproperlyConfigured`` not a ``CommandError``

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -160,12 +161,12 @@ class Command(BaseCommand):
...
                          "be guessed for any of the input files."),
-        make_option('--project',
+        make_option('--display-name',
                     action='store',
                     dest='project',
                     default='',
-                    help='Project to use when displaying TM matches for this '
-                         'translations.'),
+                    help='Name to use as provenance when displaying TM '
+                         'matches for these translations.'),
```
dwaynebailey: You like the word provenance :)

Name used when displaying TM matches for these translations.

docs/server/commands.rst
```diff
@@ -431,13 +431,18 @@ no actual data will be loaded or deleted (the TM will be left unchanged):
 
 This command also allows to read translations from files and build the TM
 resources in the external TM server. In order to do so it is mandatory to
-provide the :option:`--tm` and :option:`--project` options, along with some
-filename argument to import translations from. The provided project name will
-be displayed on TM matches in the translation editor:
+provide the :option:`--tm` and :option:`--display-name` options, along with
+some filename argument to import translations from.
```
dwaynebailey: ...along with some files to import.

docs/server/commands.rst
```diff
@@ -431,13 +431,18 @@ no actual data will be loaded or deleted (the TM will be left unchanged):
...
+
+The display name is a label used to group translations within a TM. A given TM
+can host translations for several display names. A use for the display name
+might be to specify the project from which the translation comes from. The
+display name will be shown on TM matches in the translation editor. Just
+specify it with :option:`--display-name`:
```
dwaynebailey: The display name can be used to specify the name of the project from which the translations originate.

To specify a name use --display-name

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -129,6 +129,7 @@ def get_unit_data(self, unit):
 
 class Command(BaseCommand):
     help = "Load Translation Memory with translations"
+    args = "[files]"
```
dwaynebailey: Don't we also allow no files?  Is it possible to have more than one usage?

dwaynebailey:
Some issues otherwise LGTM

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: IMHO it is better to display the user a specific message telling what is wrong than throwing a traceback that means nothing to the user.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: `files` is enclosed in `[]` to tell that this is optional.

Not sure what you mean with more than one usage.

dwaynebailey: The non-file version.  I just wanted to flag this rather then discuss it.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Changing it.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: So I am keeping this as it is for the reason in my previous comment.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: I think that `[files]` clearly states that files are optional, so this stands for both file and non-file versions.

Displaying two different usage snippets (both at the same time) would require to override the `usage()` method, but I don't think it is actually necessary.

pootle/apps/pootle_app/management/commands/update_tmserver.py
dwaynebailey: Cool I'm fine with that logic.

dwaynebailey:
LGTM

pootle/apps/pootle_app/management/commands/update_tmserver.py
julen: So what about inviting users to review their settings as well? Something like *"Translation Memory '%s' is not defined in the POOTLE_TM_SERVER setting. Please ensure it exists and double-check you typed it correctly."*

julen:
Thank you for taking this @unho, GTM.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Yes, it is better. Changed.

PR #4148, opened 2015-10-16
Render report form subject and body from templates

### LINKED ISSUE #4084: Translation of the Report Email
I would like to translate the words of the report email, if you could integrate this kind of template on the po file, it would be amazing... naturally title of the email included, which it would be ```[(name-site)] Unit #(num) ((lang))```

```
Username: (username)
Current URL: (url)
IP address: (ip_address)
User-Agent: (user_agent)

Unit: (url_string)

Source: (source_string)

Current translation: 

Your question or comment:
```

Thx in advance ;)

unho:
This makes it easier to customize them.

Fixes #4084.

### DIFF
--- a/docs/releases/dev.rst
+++ b/docs/releases/dev.rst
@@ -53,6 +53,7 @@ Details of changes
     lower the TM results score for each specific TM server.
 
 - The Apertium MT backend has been dropped.
+- Report string errors form subject and body can be overriden.
 
 
 ...and lots of refactoring, new tests, cleanups, improved documentation and of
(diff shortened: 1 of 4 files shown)


dwaynebailey:
I'd feel more comfortable if the email was one thing.  I.e. not seperate body and subject items but written in one file.  Otherwise lgtm

julen:
Having separate subject and bodies is fine (allauth does that for instance), besides it's easier and cleaner to implement.

:+1: good to go from my side.

PR #4149, opened 2015-10-16
l10n: Fix mute/unmute checks events in timeline

unho:
Hardcoding types on template doesn't seem like the best way to do this, but putting HTML code in views is neither pretty.


julen:
Thank you @unho, feel free to merge it.

PR #4158, opened 2015-10-23
Store.update cleanup

phlax:
Use cases:
=========
update_stores
------------------
  - without overwrite:
      - update db Store if on disk file_mtime has changed since last sync
      - make db store exact match
  - with `overwrite`: 
      - ignore file_mtimes:
      - make the db store an exact match to on-file store

import_file:
--------------
  - do not update units that have changed in pootle since file was exported (although make them suggestions)

pootle_fs:
------------
  - make db exact copy of file store
  - merge changed units in both - file wins - pootle unit turned to suggestion
  - merge changed units in both - pootle wins - fs unit turned to suggestion 

other use cases:
---------------------
 - import a non-PO file

Proposed changes
===============

- add store_revision to Store.update to override parsed store_revision

store_revision set to Store.max_unit_revision()
-----------------------------------------------------------
- make db exact copy of file store

store_revision set and less than Store.max_unit_revision()
--------------------------------------------------------------------------
- units that have been removed in file and have a revision of less than store_revision in db are obsoleted
- units that have been removed in file and have a revision of greater than store_revision are kept
- units that have been added in the file and dont exist as obsolete units in db are added
- units that have changed in file but have not changed in db since store_revision are updated
- units that have changed in file and have been changed in db since store_revision are treated as follows:
  - if pootle_wins pootle version of the unit is kept and the file version is set to a suggestion
  - if file_wins file version is kept and the pootle version is set to suggestion


dwaynebailey:
@phlax the outline might need clarity between remove and obsolete.  Beacuse IIRC we actually never delete a unit, simply obsolete it.

phlax:
@dwaynebailey - thanks - updated

phlax:
@ta2-1 @dwaynebailey 

A few questions emerging from the changes to Store.update:

1. should Store.update parse_revision or be passed revision?
2. should there be a default where no revision can be parsed - ie Store.get_max_unit_revision() or 0?
3. should units that have not been updated have revision increased when Store is updated?

Questions 1 and 2 are pretty critical to how we can bring back non-PO file importing

phlax:
@ta2-1 @dwaynebailey 

Following a conversation with @dwaynebailey and IMHO, im gonna answer above questions

> 1. should Store.update parse_revision or be passed revision?

store_revision should be passed to Store.update and never parsed in it.

> 2. should there be a default where no revision can be parsed - ie Store.get_max_unit_revision() or 0?

In the case of update_stores it should:
 - set store_revision as Store.last_sync_revision
 - unless --overwrite is set in which case it should set store_revision to Store.get_max_unit_revision() - ie make db exact copy of file.

Otherwise store_revision should be set to 0 which would:
   - where there is conflicting units it would set suggestion for file version (depending on pootle_wins)
   - leave any units in db alone if they are missing in file
   - add any units from file unless they are obsolete in db

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -51,10 +54,38 @@ def handle_translation_project(self, translation_project, **options):
...
+                    store.increment_unsynced_unit_revision(
+                        store.get_max_unit_revision()))
+            store.last_sync_revision = store.get_max_unit_revision()
+            if update_unsynced:
+                logger.info(
+                    u"[update] unsynced %d units in %s [revision: %d]" % (
+                        update_unsynced, store.pootle_path,
+                        store.get_max_unit_revision())
+                    )
+        store.save()
```
ta2-1: @phlax, we call `store.save()` twice: 1st in `store.update()` 2nd – here.

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -51,10 +54,38 @@ def handle_translation_project(self, translation_project, **options):
...
+            store.file.store,
+            store_revision=store_revision,
         )
+        store.file_mtime = disk_mtime
+        if changes and any(x > 0 for x in changes.values()):
+            update_unsynced = None
+            if store.last_sync_revision is not None:
+                updated_unsynced = (
+                    store.increment_unsynced_unit_revision(
+                        store.get_max_unit_revision()))
```
julen: When calling functions/methods by breaking lines, let's keep the final `)` in its own line. The same applies everywhere in this PR.

pootle/apps/pootle_store/models.py
```diff
@@ -1854,146 +1724,84 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
...
-                    self.get_max_unit_revision())
-                )
+            if old_state < PARSED:
+                self.state = PARSED
+            else:
+                self.state = old_state
+            has_changed = any(x > 0 for x in changes.values())
+            self.save(update_cache=has_changed)
+            if has_changed:
+                logger.info(
```
julen: Note the `log()` function serves its own purpose and logs to the `action` named logger. This change will break existing logging configurations. Since introducing logging-related changes in this PR is not relevant, I'd appreciate taking such changes out and keeping logging as it was.

pootle/apps/pootle_store/models.py
```diff
@@ -2392,8 +2200,10 @@ def all_pootle_paths(self):
 
     def getitem(self, item):
         """Returns a single unit based on the item number."""
-        return self.units[item]
-
+        # store.units[item] is unreliable so we need to enumerate
```
julen: For any future reader, it'd be good to elaborate on what and why is unreliable.

tests/fixtures/mock.py
```diff
@@ -17,7 +17,12 @@
 # the time it's run the decorators have already been applied to the
 # functions, therefore the patching has no effect
 mp = monkeypatch()
-mp.setattr('django.utils.functional.cached_property', property)
+
+# not sure why this was monkeypatched
```
julen: It was a footgun (799a5861). Please take this change out unless you really know you're fixing it (in such case no point in keeping commented code around)

pootle/apps/pootle_store/models.py
phlax: added ref to #4160  which deals with the issue

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -18,6 +18,9 @@
 from pootle_translationproject.models import scan_translation_projects
 
 
+logger = logging.getLogger(__name__)
```
julen: As already mentioned, please omit any logging-related changes.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,226 @@
...
+        return [unitid for unitid, unit in self.db_units.items()
+                if unit['state'] != OBSOLETE]
+    active_units = cached_property(_active_units)
+
+    @cached_property
+    def db_units(self):
+        """All of the db units regardless of state or revision"""
+        db_units = OrderedDict()
+        _db_units = self.db_store.unit_set.values('unitid', 'state', 'id',
+                                                  'index', "revision",
```
julen: Please do not mix quoting.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,226 @@
...
+            """
+            file_unit = self.file_units[uid]
+            db_unit = self.db_units[uid]
+
+            db_source = multistring_to_python(db_unit["source_f"])
+            db_target = multistring_to_python(db_unit["target_f"])
+
+            source_same = (
+                (file_unit["source"] == ""
+                 and db_source == "")
```
julen: As already outlined in the past, [wrap conditionals expressions around the operators](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#if-statements).

pootle/apps/pootle_store/models.py
```diff
@@ -1857,143 +1727,80 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
...
+                self.state = PARSED
+            else:
+                self.state = old_state
+            has_changed = any(x > 0 for x in changes.values())
+            self.save(update_cache=has_changed)
+            if has_changed:
+                log(u"[update] %s units in %s [revision: %d]"
+                    % (get_change_str(changes),
+                       self.pootle_path,
+                       self.get_max_unit_revision()))
```
julen: Can you please keep exactly the previous code for the log call? There's no actual change, and the diff adds no more than noise in here and in the commit log.

tests/fixtures/models/store.py
```diff
@@ -15,14 +16,164 @@
 
 import pytest
 
+from ...utils import update_store, create_store
+
+
+DEFAULT_STORE_UNITS_1 = [
+    ("Unit 1", "Unit 1"),
+    ("Unit 2", "Unit 2")]
```
julen: Ending `]` in its own line please. The same applies everywhere. Also for dicts and function and method calls.

tests/models/store.py
```diff
@@ -145,8 +153,10 @@ def test_update_save_changed_units(ru_update_save_changed_units_po):
             assert unit.mtime == updated_unit.mtime
 
 
+# TEST DISABLED: last_sync_revision is no longer set in Store.update
```
julen: For disabling tests a better flag is to `@pytest.mark.xfail('reason')`.

tests/models/store.py
```diff
@@ -277,5 +292,158 @@ def test_update_upload_old_revision_new_unit(en_tutorial_po):
     file_name = "tests/data/po/tutorial/en/tutorial_update_old_unit.po"
     _update_from_upload_file(en_tutorial_po, file_name)
 
-    # the new unit has not been added
-    assert len(en_tutorial_po.units) == 1
+    # OLD BEHAVIOUR: the new unit has not been added
```
julen: How relevant is it to mention a behavior change here? This would be for docs/release notes, not sure how it fits here though.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,226 @@
...
+    def __init__(self, db_store, file_store, file_revision):
+        self.db_store = db_store
+        self.file_store = file_store
+        self.db_revision = db_store.get_max_unit_revision()
+        self.file_revision = file_revision
+
+    def _active_units(self):
+        return [unitid for unitid, unit in self.db_units.items()
+                if unit['state'] != OBSOLETE]
+    active_units = cached_property(_active_units)
```
julen: Any reason why the decorator is not being used here?

pootle/apps/pootle_store/models.py
phlax: @julen im going to say no to this request - as the title of the PR suggests - the aim was to clean up Store.update - and this change does that.

pootle/apps/pootle_store/diff.py
phlax: because i had been trying to work out why cached_property wasnt working - ill revert

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -51,10 +51,36 @@ def handle_translation_project(self, translation_project, **options):
...
     def handle_store(self, store, **options):
-        store.update(
-            overwrite=options['overwrite'],
-            only_newer=not options['force']
+        if not store.file:
+            return
+        disk_mtime = store.get_file_mtime()
+        if not options["force"] and disk_mtime == store.file_mtime:
+            # The file on disk wasn't changed since the last sync
+            logger.debug(u"File didn't change since last sync, skipping "
```
ta2-1: NameError: global name 'logger' is not defined

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -51,10 +51,36 @@ def handle_translation_project(self, translation_project, **options):
...
+        store.file_mtime = disk_mtime
+        if changes and any(x > 0 for x in changes.values()):
+            update_unsynced = None
+            if store.last_sync_revision is not None:
+                updated_unsynced = store.increment_unsynced_unit_revision(
+                    update_revision
+                )
+            store.last_sync_revision = update_revision
+            if update_unsynced:
+                logger.info(u"[update] unsynced %d units in %s [revision: %d]"
```
ta2-1: NameError: global name 'logger' is not defined

pootle/apps/pootle_store/models.py
```diff
@@ -1867,143 +1737,81 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
...
+        return update_revision, changes
```
ta2-1: @phlax `update_revision` should be set before we return it.

pootle/apps/pootle_store/models.py
```diff
@@ -1867,143 +1737,81 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
...
         # Lock store
         logging.debug(u"Updating %s", self.pootle_path)
         old_state = self.state
         self.state = LOCKED
-        self.save()
+        self.save(update_cache=False)
 
-        store_revision = parse_pootle_revision(store)
-        max_unit_revision = self.get_max_unit_revision()
-        update_revision = Revision.incr()
```
ta2-1: @phlax it was set here. And you put it to function parameter.

pootle/apps/pootle_store/models.py
phlax: k @ta2-1 i reckon this is fixed

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @cached_property
+    def updated_db_units(self):
+        return [unitid for unitid, unit in self.db_units.items()
+                if unit['revision'] > self.file_revision
+                and not unit["state"] == OBSOLETE]
+
+    def get_updated_dbids(self):
+        """Returns a set of unit DB ids to be updated.
+        :return: a set of unit DB ids to be updated.
```
julen: No added value in repeating the line above.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        update_dbids = set()
+
+        def _has_changed(uid):
+            """ check if the source and targets match
```
julen: Nit: check grammar.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,222 @@
...
+        return update_dbids
+
+    def diff(self):
+        """Return a dictionary of change actions"""
+        _diff = {}
+        _diff["index"] = self.get_indexes_to_update()
+        _diff["obsolete"] = self.get_units_to_obsolete()
+        _diff["add"] = self.get_units_to_add()
+        _diff["update"] = self.get_units_to_update()
+        return _diff
```
julen: Since there's no use in the extra variable, I'd directly return the dictionary (aligns with the rest of the codebase):
```python
return {
  'index': self.get_indexes_to_update(),
  'obsolete': self.get_units_to_obsolete(),
  'add': self.get_units_to_add(),
  'update': self.get_units_to_update(),
}
```

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def get_units_to_obsolete(self):
+        return [unit['id'] for unitid, unit in self.db_units.items()
+                if unitid not in self.file_unit_ids
+                and unitid in self.active_units
```
julen: Nit: [wrap between `and`/`or`](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#if-statements). Same elsewhere.

pootle/apps/pootle_store/diff.py
unho: +1

pootle/apps/pootle_store/models.py
```diff
@@ -49,13 +48,15 @@
...
+from .diff import StoreDiff
 from .fields import (TranslationStoreField, MultiStringField,
                      PLURAL_PLACEHOLDER, SEPARATOR)
 from .filetypes import factory_classes
-from .util import (OBSOLETE, UNTRANSLATED, FUZZY, TRANSLATED, get_change_str,
-                   parse_pootle_revision)
+from .util import OBSOLETE, UNTRANSLATED, FUZZY, TRANSLATED, get_change_str
 
 
+logger = logging.getLogger(__name__)
```
julen: This can be removed.

pootle/apps/pootle_store/models.py
```diff
@@ -1867,143 +1737,82 @@ def update(self, overwrite=False, store=None, only_newer=False, user=None,
...
-                'unitid', 'state', 'id', 'index',
-            )
-            for (unitid, state, dbid, index) in unit_values:
-                if state == OBSOLETE:
-                    old_obsolete_unitid_set.add(unitid)
+        update_revision = None
+        changes = {}
+        try:
+            to_change = StoreDiff(self, store, store_revision).diff()
+            change = False
```
julen: What about having the logic to determine if a store has changes in a static method of `StoreDiff`? The method can be tested on its own, and this becomes easier to read (I'm pretty happy with what you have already achieved in `Store.update()`, this would make it a breeze to look at though):
```python
to_change = StoreDiff(self, store, store_revision).diff()
if StoreDiff.has_changes(to_change):
     ...
```

pootle/apps/pootle_store/models.py
julen: Arguably we'd break on `%`. However if you check the rest of the codebase we'd go on with this as follows:
```python
args = (foo, bar, baz)
log('my %s foo %s bar %s message' % args)
```

pootle/apps/pootle_store/diff.py
phlax: this is a policy that **really needs to be changed**

tests/fixtures/models/store.py
```diff
@@ -15,14 +16,174 @@
...
+
 
 TEST_UPDATE_PO = "tests/data/po/tutorial/en/tutorial_update.po"
 TEST_EVIL_UPDATE_PO = "tests/data/po/tutorial/en/tutorial_update_evil.po"
 
+UPDATE_STORE_TESTS = OrderedDict()
+UPDATE_STORE_TESTS['min_empty'] = {"update_store": (0, [])}
+UPDATE_STORE_TESTS['min_new_units'] = {
+    "update_store": (0, DEFAULT_STORE_UNITS_3)
+    }
```
julen: Nit: mind indent of ending brace

tests/fixtures/models/store.py
```diff
@@ -15,14 +16,174 @@
...
+                         in DEFAULT_STORE_UNITS_3]
+
 
 TEST_UPDATE_PO = "tests/data/po/tutorial/en/tutorial_update.po"
 TEST_EVIL_UPDATE_PO = "tests/data/po/tutorial/en/tutorial_update_evil.po"
 
+UPDATE_STORE_TESTS = OrderedDict()
+UPDATE_STORE_TESTS['min_empty'] = {"update_store": (0, [])}
+UPDATE_STORE_TESTS['min_new_units'] = {
+    "update_store": (0, DEFAULT_STORE_UNITS_3)
```
julen: Nit: always use a trailing comma — it helps reduce future diff noise. Same elsewhere.

tests/fixtures/models/store.py
```diff
@@ -15,14 +16,174 @@
 
 import pytest
 
+from ...utils import update_store, create_store
+
+
+DEFAULT_STORE_UNITS_1 = [("Unit 1", "Unit 1"),
+                         ("Unit 2", "Unit 2")]
+
+DEFAULT_STORE_UNITS_2 = [("Unit 3", "Unit 3"),
```
julen: For lists/dicts with many items, always [break immediately after the opening brace](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#continuing-long-statements).

pootle/apps/pootle_store/diff.py
julen: It's just a matter of style, there's no right answer or good/bad, so please stick to it.

phlax: Well readability is a measurable metric - and im pretty sure this makes code less readable

At the moment the policies are a mess and contradictory - and most of the code base doesnt even meet pep8 - we need to review these policies

pootle/apps/pootle_store/models.py
phlax: there are examples of quite a few styles - i have been using this one because it keeps to the various policies and is more readable imho

pootle/apps/pootle_store/diff.py
dwaynebailey: Adding my 2c.  If there is a contradiction in the styleguide then lets clean it up please.  Most style is a choice and the readability comes from consistent application.  If it's not working then lets discuss one aspect and fix it.

Note the Pootle code-base is old.  So don't measure inconstancy as not complying with the style guide.  Just that that code hasn't got cleaned up yet.

In terms of PEP8.  It really isn't the be all and end all.  I like my code to comply.  But even in PEP8 there are mutually exclusive options as well as options that make code less readable, so we have to make choices.  As far as PEP8 goes now we do comply fully to the choices made in ``pootle/tools/pep8.sh``.  We fail on Travis if any of these PEP8 requirements fail ``E10,E11,E222,E251,E261,E262,E27,E401,E70,E711,E712,E713,E721,W191,W291,W292,W293,W391,W60``.

Also not that we haven't applied pep8 to tests, they where far too messy, but we could expand our coverage there if needed.

We can always expand the list of checks that we require to pass, but that would mean:

1. Cleaning up cases in the code to get 100% pass for that PEP8 test
2. Making a choice where there is a conflict of options
3. Adding it to ``pep8.sh`` so that we catch these in Travis

@phlax if you have ones that continue to flag in your editor (as I assume its base choice is not conflicting) feel free to report an issue as I'm a happy PEP8 cleaner.

unho: Yes, readability counts. Personally instead of adding brackets I would just indent it for this particular case. But in the rest of the cases I would rely on the styleguide.

Please note that coding guidelines were written after debating them to deal with specific problems we hit. I am open to change them or align more with PEP8, but only if that doesn't bring back those problems that we got rid of by using the current guidelines.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,235 @@
...
+        return [unitid for unitid, unit in self.db_units.items()
+                if unit['revision'] > self.file_revision
+                and not unit["state"] == OBSOLETE]
+
+    def get_updated_dbids(self):
+        """Returns a set of unit DB ids to be updated.
+        """
+        update_dbids = set()
+
+        def _has_changed(uid):
```
ta2-1: @phlax, I found the reason why we have a speedup. Number of updated units in [this PR (pootle_store/models.py#L1614)] (https://github.com/phlax/pootle/blob/aecd9338daa6e3f987bccfdfe4230aaa3742a216/pootle/apps/pootle_store/models.py#L1614) differs from [current HEAD (pootle_store/models.py#L1689)](https://github.com/translate/pootle/blob/f7f17237fa8f10ebfaf8366a76843c1ee2826883/pootle/apps/pootle_store/models.py#L1689)
You filter unit list here and return `changed` units only. And you don't take into account `state`, `developer_comment`, `translator_comment`, `locations`, `context` unit fields (see [unit.update](https://github.com/phlax/pootle/blob/aecd9338daa6e3f987bccfdfe4230aaa3742a216/pootle/apps/pootle_store/models.py#L695))

And yeah, it's very strange that don't have tests for this.

phlax: @ta2-1 - ive added a UnitDiff class and i reckon ive addressed your concern.

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,306 @@
...
+    def file_locations(self):
+        return self.db_unit["locations"].split("\n")
+
+    @property
+    def locations_changed(self):
+        return self.db_unit["locations"] != self.file_locations
+
+    @property
+    def source_changed(self):
+        return not ((self.file_unit["source"] == "" and self.db_source == "")
```
phlax: @ta2-1 the one remaining concern i have is whether this test is exactly correct for detecting db/file unit differences - similar for target_changed

pootle/apps/pootle_store/diff.py
```diff
@@ -0,0 +1,302 @@
...
+                dev_comment = unit.getnotes(origin="developer")
+                translator_comment = unit.getnotes(origin="translator")
+                funit = {"unitid": unit.getid(),
+                         "context": unit.getcontext(),
+                         "locations": unit.getlocations(),
+                         "source": unit.source,
+                         "target": unit.target,
+                         "state": unit.get_state_n(),
+                         "developer_comment": dev_comment,
+                         "translator_comment": translator_comment}
```
unho: Can you please put it like:

```python
                file_units[unit.getid()] = {
                    'unitid': unit.getid(),
                    'context': unit.getcontext(),
                    'locations': unit.getlocations(),
                    'source': unit.source,
                    'target': unit.target,
                    'state': unit.get_state_n(),
                    'developer_comment': dev_comment,
                    'translator_comment': translator_comment,
                }
```

It is easier to read and doesn't use extra variables.

phlax:
@ta2-1 ive done a last minute re-implementation of the updateable unit comparison

@dwaynebailey - good news we get to keep the 40% speedup!

dwaynebailey:
@phlax that's like winning the lottery! I've had a ook at UnitDiff, Greek to me :( Ironically there is a but just reported about previous msgid, so likely your changes make detecting those changes easier also.

ta2-1:
@dwaynebailey  This is not the lottery) We selected all units to be updated via `SELECT IN` query and built `Unit` Django model objects and then tested if anything was changed via `Unit.update()`.
In this PR some comparison code was duplicated (moved) from `Unit.update()` to light and efficient classes (`DBUnit`, `FileUnit`) this allowed us to decrease number of unit ids for `SELECT IN` query and select units that were changed only.

dwaynebailey:
@ta2-1 don't worry the lottery reference was about the changes still keeping the 40% gain which I was worried we wouldn't keep when we found that we weren't flagging updates for other meta data changes.  But thanks for the explanation anyway ;)

tests/models/store.py
```diff
@@ -145,8 +153,9 @@ def test_update_save_changed_units(ru_update_save_changed_units_po):
             assert unit.mtime == updated_unit.mtime
 
 
-@pytest.mark.django_db
-def test_update_set_last_sync_revision(ru_update_set_last_sync_revision_po):
+@pytest.mark.xfail("last_sync_revision is no longer set in Store.update "
```
ta2-1: @phlax, did we tested that command works?

julen:
Again, thank you for taking up this massive work!

unho:
Very nice work. And thanks for taking the bullet for us. Congratulations!

phlax:
big thanks to @ta2-1 who has contributed hugely to landing this

PR #4167, opened 2015-11-03
Add timestamp to TM entries

### LINKED ISSUE #4117: Add timestamp to our local TM entries
When updating our local TM in Elasticsearch, let's add a timestamp of the submission (in UTC) as a separate field in the entry structure. In the translation editor, when rendering 'Similar translations' block, let's display the metadata like this:

`Translated by <username> in “<some_project>” project 2 days ago`

(where `2 days ago` is an extra part that is rendered based on the stored date). For entries that don't have a timestamp, we'll fallback to the current message format:

`Translated by <username> in “<some_project>”`

Rationale: knowing when the particular translation was added allows translators to put more weight to the translations that were done recently.

unho:
Fixes #4117.

### DIFF
--- a/pootle/apps/pootle_app/management/commands/update_tmserver.py
+++ b/pootle/apps/pootle_app/management/commands/update_tmserver.py
@@ -20,7 +20,9 @@ from translate.storage import factory
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+from django.utils import dateparse
 
+from pootle.core.utils import dateformat
 from pootle_store.models import Unit
 
 
@@ -57,6 +59,7 @@ class DBParser(object):
                 'revision',
                 'source_f',
                 'target_f',
+                'submitted_on',
                 'submitted_by__username',
                 'submitted_by__full_name',
                 'submitted_by__email',
@@ -76,6 +79,14 @@ class DBParser(object):
         if unit['submitted_by__email']:
             email_md5 = md5(unit['submitted_by__email']).hexdigest()
 
+        iso_submitted_on = unit.get('submitted_on', None)
+
+        display_submitted_on = None
+        if iso_submitted_on:
+            display_submitted_on = dateformat.format(
+                dateparse.parse_datetime(str(iso_submitted_on))
+            )
+
         return {
             '_index': self.INDEX_NAME,
             '_type': unit['store__translation_project__language__code'],
@@ -88,6 +99,8 @@ class DBParser(object):
             'email_md5': email_md5,
             'source': unit['source_f'],
             'target': unit['target_f'],
+            'iso_submitted_on': iso_submitted_on,
+            'display_submitted_on': display_submitted_on,
         }
 
 
@@ -153,6 +166,8 @@ class FileParser(object):
             'email_md5': None,
             'source': unit.source,
             'target': unit.target,
+            'iso_submitted_on': None,
+            'display_submitted_on': None,
         }
 
 
--- a/pootle/core/search/backends/elasticsearch.py
+++ b/pootle/core/search/backends/elasticsearch.py
@@ -89,6 +89,9 @@ class ElasticSearchBackend(SearchBackend):
                         'username': hit['_source']['username'],
                         'fullname': hit['_source']['fullname'],
                         'email_md5': hit['_source']['email_md5'],
+                        'iso_submitted_on': hit['_source'].get('iso_submitted_on', None),
+                        'display_submitted_on': hit['_source'].get('display_submitted_on',
+                                                                None),
                         'score': hit['_score'] * self.weight,
                     })
                 else:
--- a/pootle/templates/editor/units/xhr_tm.html
+++ b/pootle/templates/editor/units/xhr_tm.html
@@ -10,7 +10,13 @@
         <div class="suggestion-translation" lang="<%= store.target_lang %>"
           dir="<%= store.target_dir %>"><%= PTL.utils.fancyHl(sugg.target) %></div>
         <% if (sugg.project) { %>
+        <% if (sugg.iso_submitted_on) { %>
+        <div class="tm-context">
+          <%= interpolate(gettext('Translated by %(fullname)s in &ldquo;<span title="%(path)s">%(project)s</span>&rdquo; project %(time_ago)s'), {fullname: sugg.fullname, path: sugg.path, project: sugg.project, time_ago: `<time class="extra-item-meta js-relative-date" title="${sugg.display_submitted_on}" datetime="${sugg.iso_submitted_on}">&nbsp;</time>`}, true) %>
+        </div>
+        <% } else { %>
         <div class="tm-context"><%= interpolate(gettext('Translated by %(fullname)s in &ldquo;<span title="%(path)s">%(project)s</span>&rdquo; project'), {fullname: sugg.fullname, path: sugg.path, project: sugg.project}, true) %></div>
+        <% } %>
         <% } %>
       </div>
     </div>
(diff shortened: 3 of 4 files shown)


pootle/core/search/backends/elasticsearch.py
```diff
@@ -80,6 +80,9 @@ def search(self, unit):
                 translation_pair = hit['_source']['source'] + hit['_source']['target']
                 if translation_pair not in counter:
                     counter[translation_pair] = 1
+                    submitted_on = None
+                    if 'submitted_on' in hit['_source']:
+                        submitted_on = hit['_source']['submitted_on']
```
julen: Did you try `submitted_on = hit['_source'].get('submitted_on', None)`? This would require a single-line change.

pootle/templates/editor/units/xhr_tm.html
```diff
@@ -10,8 +10,14 @@
         <div class="suggestion-translation" lang="<%= store.target_lang %>"
           dir="<%= store.target_dir %>"><%= PTL.utils.fancyHl(sugg.target) %></div>
         <% if (sugg.project) { %>
+        <% if (sugg.submitted_on) { %>
+        <div class="tm-context">
+          <%= interpolate(gettext('Translated by %(fullname)s in &ldquo;<span title="%(path)s">%(project)s</span>&rdquo; project <time class="extra-item-meta js-relative-date" title="%(submitted_format)s" datetime="%(submitted_isoformat)s">&nbsp;</time>'), {fullname: sugg.fullname, path: sugg.path, project: sugg.project, submitted_format: sugg.submitted_on, submitted_isoformat: sugg.submitted_on|isoformat}, true) %>
```
julen: Can we abstract the HTML tags away so we reduce the risk of localizers breaking actual functionality? Note the whole `<time ....>&nbsp;</time>` is avoidable noise we can reduce to a placeholder.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -153,6 +157,7 @@ def get_unit_data(self, unit):
             'email_md5': None,
             'source': unit.source,
             'target': unit.target,
+            'submitted_on': make_aware(datetime.now()),
```
julen: This would be a valid *import time*, however as I understand it, we are looking for *translation time*. When importing from files, we effectively don't know when a unit was translated, so I think this line is wrong and would result misleading. Thoughts @translate/core?

dwaynebailey: Agree with @julen, we want times only when we know they are valid.  We can't tell when a unit was actually translated when we import a file, thus the UI when we don't know the time should apply to almost any file that can't give us a unit translation time (for now I think we can just assume that none can give that).

unho: Ok, setting it to `None`.

unho:
Please review again.

For testing you need to either run `update_tmserver` (to update from DB) or translate some string in order to get some TM results with the submitted datetime.

julen:
The results always display *A few seconds ago*. I presume you need to call `helpers.updateRelativeDates()` right after injecting the content in the UI (check elsewhere in the editor). Note this is yet another reason why we are moving away to the `<TimeSince />` component.

unho:
@julen Perhaps it is better to just do it using React. If you can craft a working patch I will append it to this PR.

julen:
I'm afraid that's beyond my priorities right now. Calling the method as I suggested is totally fine — we'll eventually replace it with the afore-mentioned component anyway.

friedelwolff:
I'm not at all familiar with this new code, but if this is coming from PO files, we can check the PO-Revision-Date as a better estimation - I think most tools update that these days. If this tool does incremental import (only new units are added), only the new units will have the newer date. Does that help?

iafan:
I think it's perfectly fine to use import time as translation time. In the end, it doesn't really matter if the user has been working on this translation locally for some time, it only matters when the new translation has been applied as a submission (became public). This time will be the same as the translation time as displayed on the unit.

julen:
@iafan if I understood your comment correctly, you are referring to uploads, note however the discussion is about populating the TM from any files by using the `update_tmserver` command — these translations are independent from submissions, as opposed to uploads. Uploads will already create submissions, and (as per this PR) their submission time will be used when indexing them in the TM. So the behavior is already as you describe it.

julen:
To reiterate what I already mentioned before, as a translator I find it irrelevant when translations were *added* to the TM and it adds no value TBH. It can be useful for admins or when performing bulk-remove actions, but that's a separate concern. Added to that, it's misleading for translators, since it reports incorrect information.

As an example of where this could cause trouble, let's say a language team phased out using a specific term in their translations, however some admin imported an old-fashioned file for this language (*errare humanum est*). 
Translators will be puzzled to receive **recent results** where some anonymous user translated units using the old terminology. Also, those translators who are not familiar with the terminology will either a) rely on the wrong TM matches, or b) be in doubt as they'll see different terms being displayed in Terminology and TM.

dwaynebailey:
Can I bring the timestamp on import to a close.  @friedelwolff's suggestion would work though we don't do incremental updates with this command, so no go for now.  @julen hits the nail on the head, this information is wrong and therefore misleading.

Conclusion: there should be no timestamp for these.

Now I'm not sure if the code reflects this, @unho?

iafan:
> you are referring to uploads, note however the discussion is about populating the TM from any files by using the update_tmserver command

Sorry, I missed that. For `update_tmserver`, it's definitely better not to set the date, so I agree with you and @dwaynebailey.

unho:
@dwaynebailey Current status of this PR is that `update_tmserver` only sets the timestamp for translations pulled from Pootle DB (translations uploaded or done directly on Pootle, so not translations coming from files on disk).

unho:
Now the right `<time>` is set on the template and thus the date is correctly updated.

@translate/core Please review and test again.

unho:
Adjusted to store both the ISO and the display submission datetimes on the TM server, and renamed keys used to store such values.

Please review and test again.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -76,6 +79,14 @@ def get_unit_data(self, unit):
         if unit['submitted_by__email']:
             email_md5 = md5(unit['submitted_by__email']).hexdigest()
 
+        iso_submitted = None
```
julen: I recall this was `iso_submitted = unit.get('submitted_on', None)` before, which is the preferred idiomatic Python way to set default values. Also I realize the `_on` from the original variable is missing here and in the `display_` variant.

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -76,6 +79,14 @@ def get_unit_data(self, unit):
...
+        display_submitted = None
+        if unit['submitted_on']:
+            iso_submitted = unit['submitted_on']
+            display_submitted = dateformat.format(
+                    dateparse.parse_datetime(str(iso_submitted))
```
julen: Nit: 4-space indent

julen:
GTM minor comments aside.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Changed.

pootle/apps/pootle_app/management/commands/update_tmserver.py
unho: Changed.

Issue #4173, opened 2015-11-11
Elasticseach TM query gets 'string indices must be integers' exception


dwaynebailey:
This error occurs sporadically on ES queries.  Its pretty hard to replicate but in debugging with @phlax it seems like it could be a timeout issue.

`exceptions:TypeError: string indices must be integers`

```pytb
Traceback (most recent call last):
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.54.0.41/newrelic/api/web_transaction.py", line 711, in __iter__
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.54.0.41/newrelic/api/web_transaction.py", line 1087, in __call__
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 187, in __call__
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 111, in get_response
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/db/transaction.py", line 394, in inner
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.54.0.41/newrelic/hooks/framework_django.py", line 499, in wrapper
  File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 52, in _wrapped_view_func
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_misc/util.py", line 59, in wrapper
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_store/decorators.py", line 66, in decorated_f
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_store/views.py", line 833, in get_edit_unit
  File "/var/www/sites/mozilla/src/pootle/apps/pootle_store/models.py", line 912, in get_tm_suggestions
  File "/var/www/sites/mozilla/src/pootle/core/search/broker.py", line 51, in search
  File "/var/www/sites/mozilla/src/pootle/core/search/backends/elasticsearch.py", line 78, in search
```

dwaynebailey:
Replicating the actual conditions seems very hard.  So I decided to replicate the error based on my own theory that the results where not as expected.

Using this code

```py
json_string_nores = """{
      "took" : 63,
      "timed_out" : false,
      "_shards" : {
              "total" : 5,
              "successful" : 5,
[...]

PR #4174, opened 2015-11-11
Elastisearch: catch non dict results

dwaynebailey:
Part of addressing the cause of #4173 

I want to deploy this to mlo to see if we can catch the traceback from #4173, just want some gtm's that this isn't broken or going down the wrong path.


pootle/core/search/backends/elasticsearch.py
```diff
@@ -75,6 +77,11 @@ def search(self, unit):
             }
         )
 
+        if not isinstance(es_res['hits'], dict):
```
phlax: hi @dwaynebailey

you might need something more like:

```
if not isinstance(es_res, dict) or not isinstance(es_res.get("dict"), dict)
```

pootle/core/search/backends/elasticsearch.py
```diff
@@ -75,6 +77,21 @@ def search(self, unit):
...
+            return []
+
+        if not isinstance(es_res['hits'], dict):
+            logging.error("Elasticsearch 'es_res['hits']' is '%s' not 'dict', '%s'" %
+                          (type(es_res['hits']), es_res['hits']))
+            return []
+
+        if not isinstance(es_res['hits']['hits'], list):
+            logging.error("Elasticsearch 'es_res['hits']['hits']' is '%s' not 'list', '%s'" %
+                          (type(es_res['hits']i['hits']), es_res['hits']['hits']))
```
phlax: theres a typo here

dwaynebailey:
@phlax I'm not sure it was needed really.  But made this a bit more robust for each level of the es_res structure.  It catches your concern about other data being missing.  But tbh the only way I could get the TypeError failure was when `es_res['hits']` was a string instead of a dict.  In all other cases another error would occur.

pootle/core/search/backends/elasticsearch.py
phlax: theres a typo here

julen:
If the end goal is catching the actual error I don't think this is necessary ([check my comment on #4173](https://github.com/translate/pootle/issues/4173#issuecomment-155783027)).

PR #4189, opened 2015-11-16
Display language and project announcements in their overview pages

### LINKED ISSUE #4178: Display language and project announcements in the context of corresponding pages
It seems that now we have an option to specify announcements for languages and projects, but these are not displayed when one goes to the respective language or project page. Let's display announcements there. This relates to implementing in-context announcement editing (see #3984)

unho:
Fixes #4178.

@translate/core Please review.

### DIFF
--- a/pootle/apps/pootle_language/models.py
+++ b/pootle/apps/pootle_language/models.py
@@ -22,6 +22,7 @@ from pootle.core.cache import make_method_key
 from pootle.core.mixins import TreeItem
 from pootle.core.url_helpers import get_editor_filter
 from pootle.i18n.gettext import tr_lang, language_dir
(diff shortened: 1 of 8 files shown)


dwaynebailey:
gtm.  I tested on an instance and the sidebar appears as expected.  There where some place where it appeared or disappeared that where unexpected but in chatting to @unho we seemed to confirm that this is expected behaviour or matches current behaviour.

1. Notice changed.
  1. It appears
  2. Reload
  3. It disappears - this matches current behaviour
2. Language notice changed but in TP
   1. Sidebar appears
   2. Since Language notice is on this sidebar we'd actually expect it to appear.

PR #4190, opened 2015-11-16
Overriding django.core.mail.send_mail to set outgoing email headers

### LINKED ISSUE #4176: Modify outbound email headers to prevent mail from being bounced
We currently use `From: "{{user.display_name}}" <{{user.email}}>` header in the emails we send from Pootle. This doesn't work with all servers. For example, it doesn't work with Yahoo emails as their server bounces our emails with the following error message: `Unauthenticated email from yahoo.com is not accepted due to domain's 550-5.7.1 DMARC policy.`

So let's make sure all outbound emails have the following `From:` and `Reply-to:` headers:

```
From: "{{user.display_name}}" <{{settings.DEFAULT_FROM_EMAIL}}>
Reply-to: "{{user.display_name}}" <{{user.email}}>
```

phlax:
Use `settings.DEFAULT_FROM_EMAIL` as from_email and set the reply-to
header to user's email

Fixes #4176

### DIFF
--- a/pootle/apps/contact/forms.py
+++ b/pootle/apps/contact/forms.py
@@ -16,6 +16,7 @@ from django.utils.translation import ugettext_lazy as _
 from contact_form.forms import ContactForm as OriginalContactForm
 
 from pootle.core.forms import MathCaptchaForm
+from pootle.core.mail import send_mail
 
 
 class ContactForm(MathCaptchaForm, OriginalContactForm):
@@ -55,6 +56,15 @@ class ContactForm(MathCaptchaForm, OriginalContactForm):
 
     def recipient_list(self):
         return [settings.POOTLE_CONTACT_EMAIL]
+
+    def save(self, fail_silently=False):
+        """Build and send the email message."""
+
+        kwargs = self.get_message_dict()
+        kwargs["headers"] = {"Reply-To": kwargs["from_email"]}
+        kwargs["from_email"] = settings.DEFAULT_FROM_EMAIL
+        send_mail(fail_silently=fail_silently, **kwargs)
+
 
 # Alters form's field order. Use `self.field_order` when in Django 1.9+
 ContactForm.base_fields = OrderedDict(
--- a/pootle/core/mail.py
+++ b/pootle/core/mail.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.conf import settings
+from django.core.mail import get_connection, EmailMultiAlternatives
+
+
+def send_mail(subject, message, from_email, recipient_list,
+              fail_silently=False, auth_user=None, auth_password=None,
+              connection=None, html_message=None, headers=None):
+    """Override django send_mail function to allow use of custom email headers.
+    """
+
+    connection = connection or get_connection(username=auth_user,
+                                              password=auth_password,
+                                              fail_silently=fail_silently)
+
+    mail = EmailMultiAlternatives(subject, message,
+                                  from_email, recipient_list,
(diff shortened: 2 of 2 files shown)


unho:
Tried to test this using `python -m smtpd -n -c DebuggingServer localhost:1025` but the emails don't display the `Reply-to:` header. Perhaps I am doing something wrong, @phlax can you provide test instructions?

phlax:
@unho - youre right - i need to override the render_mail in accounts.adapter.PootleAdapter to make it work

julen:
Maybe the original report wasn't crystal clear on the actual issue, but problems with Yahoo! email addresses arise when using the **contact form**, not during registration/pw reset, as the former sets the user's full name + email address in the `From:` header. The way to solve this is by setting the [`Reply-To:` header](https://help.yahoo.com/kb/SLN24016.html) in the contact form, [as already suggested](https://github.com/translate/pootle/issues/4176#issuecomment-156035796). Please consider discarding this PR.

phlax:
@julen that is exactly what this PR does - ie overrides django's send_mail to allow the inclusion of additional headers.

If i understand correctly - what is now required is to override the contact form to ensure it uses the overriden send_mail

pootle/core/mail.py
```diff
@@ -0,0 +1,24 @@
+from django.conf import settings
```
phlax: added

pootle/apps/contact/forms.py
```diff
@@ -56,6 +57,17 @@ def from_email(self):
     def recipient_list(self):
         return [settings.POOTLE_CONTACT_EMAIL]
 
+    def save(self, fail_silently=False):
+        """
+        Build and send the email message.
```
julen: Nit: mind the [docstring conventions](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#docstrings). Same elsewhere.

julen:
Mind squashing the commits please? Also, let's avoid present continuous in commit messages as already discussed. GTM after that.

pootle/core/mail.py
```diff
@@ -0,0 +1,31 @@
...
+from django.core.mail import get_connection, EmailMultiAlternatives
+
+
+def send_mail(subject, message, from_email, recipient_list,
+              fail_silently=False, auth_user=None, auth_password=None,
+              connection=None, html_message=None, headers=None):
+    """
+    Override django send_mail function to use from email as `reply-to` and
+    site email as `from`
```
unho: Add final period and please use `Reply-To` and `From` to refer to the headers.

pootle/apps/contact/forms.py
```diff
@@ -56,6 +57,17 @@ def from_email(self):
...
+
+        """
+        kwargs = self.get_message_dict()
+        kwargs["headers"] = {"reply-to": kwargs["from_email"]}
```
unho: Please use `Reply-To`.

pootle/apps/contact/forms.py
```diff
@@ -56,6 +57,17 @@ def from_email(self):
...
+        kwargs["from_email"] = settings.DEFAULT_FROM_EMAIL
```
unho: Just wondering if this would be easier to read if it is an update call:

```python
    kwargs.update({
        'headers': {
            'Reply-To': kwargs['from_email'],
        },
        'from_email': settings.DEFAULT_FROM_EMAIL,
    })
```

phlax: i dont think so

julen: Just a matter of choice — personally I prefer `.update()`.

unho:
Tested and it works. GTM after nits.

pootle/apps/contact/forms.py
phlax: well - for me it kinda depends on context - but here where we are capturing an existing value and setting to a different key - and then updating the original key - i think its clearer to this step-by-step

unho: Works for me. But please address the other nits, specifically renaming to `Reply-To` to use exactly the same case that is expected.

PR #4191, opened 2015-11-16
Re-enable download for terminology project files

### LINKED ISSUE #4129: A user can't download the terminology files from a Terminology project
I know we switched off download of ``pootle-terminology.po`` files as they cause a number of problem mostly because there is no associated file.  But this seems to have extended to also include the 'Terminology' project.

I can't recall if at the time we decided it was an OK compromise.  But now with @phlax's work which means we can have non-existent files we should likely be able to bring this back. At the very least we shouldn't be disabling download of files within Terminology ad only restrict ``pootle-terminology.po``

phlax:
Fixes #4129


unho:
lgtm

phlax:
thanks - i think this is covered by tests added in #4158 - im also doing some UI tests to confirm that the problem is fixed

PR #4192, opened 2015-11-16
Consolidating force and overwrite options in sync_stores

### LINKED ISSUE #3843: Consolidate --force and --overwrite options in sync_stores
There are two options --force and --overwrite that seem to do exactly the same things.  I'm not sure if codewise they do but I think from a user perspective they should.  Reading the docs it is completely unclear how they are different.

--force - downloads regardless of the last_sync_revision optimisation
--overwrite - downloads regardsless of last sync and also overwrite files.

In which case how does --force not overwrite files?

phlax:
...and consolidating update_structure, conservative, and only_newer
in Store.sync into a single force option.

Fixes #3843


phlax:
@ta2-1 @dwaynebailey im not sure if i have understood the ticket 100% (#3843) - can you take a look and see if this PR meets the expected behaviour

unho:
No docs? Also include it in release notes as this must be reported.

phlax:
@unho re docs - will do - im gonna add WIP on this as it requires some docs updates - and im not entirely clear yet what they are

pootle/apps/pootle_store/models.py
```diff
@@ -2044,22 +2044,20 @@ def serialize(self):
...
 
-        #TODO only_newer -> not force
-        if (only_newer and self.file.exists() and
+        if (not force and self.file.exists() and
             self.last_sync_revision >= last_revision):
             logging.info(u"[sync] No updates for %s after [revision: %d]" %
                 (self.pootle_path, self.last_sync_revision))
             return
 
-        if not self.file.exists() and not skip_missing:
```
unho: Why is this removed?

tests/models/directory.py
```diff
@@ -65,7 +65,7 @@ def test_delete_mark_obsolete_resurrect_sync(fr_tutorial_subdir_to_remove_po):
         unit.save()
 
     # Recover store and directory by syncing
-    updated_store.sync(only_newer=False)
+    updated_store.sync()
```
unho: Shouldn't this be replaced by `force=True`?

pootle/apps/pootle_store/models.py
phlax: @unho im glad you asked 8)

because in L2049 it tests for `skip_missing and not self.file.exists()` and returns if it exists - its superfluous

tests/models/directory.py
phlax: you are probably right - ill add

pootle/apps/pootle_store/models.py
unho: Thanks! Please put this specific cleanup on a separate commit and merge it directly to master. Don't forget to include this explanation on the commit message!

PR #4193, opened 2015-11-16
Ignore non project filetypes when uploading zip files

### LINKED ISSUE #4124: Uploads perform poorly when some non localisation error is picked up
On a Mac you could have stray ``.DS_Store`` files.  Also when using POedit it will create a ``.mo`` files whenever you save the ``.po`` file.

Errors such as ``Unknown filetype (en_ZA/firefox/browser/chrome/overrides/.DS_Store)`` are reported in this case.

Whenever such an error occurs then the upload reports the error and fails to complete.  We should I think ignore errors unrelated to the translations files we are uploading.  And at least execute correctly for those where there are no errors.

phlax:
Fixes #4124

### DIFF
--- a/pootle/apps/import_export/views.py
+++ b/pootle/apps/import_export/views.py
@@ -64,10 +64,12 @@ def export(request):
         return download(f.getvalue(), "%s.zip" % (prefix), "application/zip")
 
 
-def handle_upload_form(request):
+def handle_upload_form(request, project):
     """Process the upload form."""
     if request.method == "POST" and "file" in request.FILES:
         upload_form = UploadForm(request.POST, request.FILES)
+        project_filetypes = [project.localfiletype,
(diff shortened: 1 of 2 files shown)


unho:
lgtm

pootle/apps/import_export/views.py
```diff
@@ -64,10 +64,12 @@ def export(request):
         return download(f.getvalue(), "%s.zip" % (prefix), "application/zip")
 
 
-def handle_upload_form(request):
+def handle_upload_form(project, request):
```
julen: I think it's preferable (although not necessary) to add any new parameters at the end. I'd still consider doing that to keep `request` as the first parameter, which is a kinda convention for Django code dealing with views.

julen:
:+1: looks good to me.

Issue #4142, opened 2015-10-14
Allow import attribution overriding


dwaynebailey:
When importing it is not possible to override the attribution given.
1. If using the UI the uploader will be attributed.  This is problematic if an admin is uploading a file on behalf of someone else.
2. When using `import` command the attribution is to the system user.  I had a case where I was resolving conflicts and wanted to attribute the author but couldn't.

This has two implementation needs:
1. Allow `--author` to be specified for the `import` command and use that for attribution.  If the user does not exist then fail.
2. If you have admin rights in the UI add a dropdown that allows you to select the user to attribute.  I'm not sure we need to track somewhat like Git with Author and Commiter.  Could be helpful for the timeline.

dwaynebailey:
Lets aim to do the cli for 2.7.3, the UI can wait

phlax:
i believe this is complete - @dwaynebailey ?

dwaynebailey:
Upload from the UI will attribute correctly, and also allows admins to override attribution.

PR #4194, opened 2015-11-16
Allow user attribution on CLI file import

phlax:
Fixes first half of #4142

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -563,6 +563,15 @@ counter is older than on Pootle, that is someone has translated while the file
 was offline, then it will be rejected.  Otherwise the translations in the file
 are accepted.
 
+Available options:
+
+:option:`--user`
+  .. versionadded:: 2.7.3
+
+  Import file(s) as given user. The user with the provided username must exist.
+
+  Default: ``system``.
+
 
 .. _commands#manually_installing_pootle:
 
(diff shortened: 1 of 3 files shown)


unho:
Would be nice if you can add this change to release notes as well. Please squash both commits as well since docs changes are not so extensive.

docs/server/commands.rst
```diff
@@ -563,6 +563,16 @@ counter is older than on Pootle, that is someone has translated while the file
 was offline, then it will be rejected.  Otherwise the translations in the file
 are accepted.
 
+Available options:
+
+:option:`--user`
+  .. versionadded:: 2.7.3
+
+  Import file(s) as given user. If set this option should the username of a
```
julen: The sentence is incomplete, there seems to be some word missing. I'd consider rewording it as simple as *The user with the provided username must exist.*

pootle/apps/import_export/management/commands/import.py
```diff
@@ -9,19 +9,37 @@
 
 import os
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+from optparse import make_option
 from zipfile import is_zipfile, ZipFile
 
 from django.core.management.base import BaseCommand, CommandError
+from django.contrib.auth import get_user_model
```
julen: Nit: `django.contrib` comes before `django.core`

julen:
Docs are broken on Travis. GTM after fixing that and addressing the comments.

PR #4195, opened 2015-11-16
Only show/process upload/download form if user has translation power

### LINKED ISSUE #4125: Users with only suggest rights are able to upload translations
It seems that if you have only suggestion rights to a project, that when you upload a file and there are no conflicts and such, your translations will become real translations and not be limited to your rights.

Someone with only suggestion rights should not be able to upload translations, but they should be able to submit suggestions via an upload.

phlax:
Ref #4125

### DIFF
--- a/pootle/apps/pootle_translationproject/views.py
+++ b/pootle/apps/pootle_translationproject/views.py
@@ -10,6 +10,7 @@
 from django.conf import settings
 from django.shortcuts import render
 
+from import_export.views import handle_upload_form
 from pootle.core.browser import (get_children, get_table_headings, get_parent,
                                  get_vfolders)
 from pootle.core.decorators import (get_path_obj, get_resource,
@@ -61,18 +62,13 @@ def browse(request, translation_project, dir_path, filename=None):
     ctx.update(get_browser_context(request))
 
     # TODO improve plugin logic
-    if "import_export" in settings.INSTALLED_APPS and request.user.is_authenticated():
-        from import_export.views import handle_upload_form
-
-        ctx.update(handle_upload_form(request, project))
-
-        has_download = (not translation_project.is_terminology_project and
-                        (check_permission('translate', request) or
-                         check_permission('suggest', request)))
-        ctx.update({
-            'display_download': has_download,
-            'has_sidebar': True,
-        })
+    if "import_export" in settings.INSTALLED_APPS:
+        if not translation_project.is_terminology_project:
+            if request.user.is_authenticated():
+                if check_permission('translate', request):
+                    ctx.update(handle_upload_form(request, project))
+                ctx.update({'display_download': True,
+                            'has_sidebar': True})
 
     stats = request.resource_obj.get_stats()
 


unho:
This is actually not fixing the issue by allowing users with suggestion rights to upload translations as suggestions rather than as translations. It just enforces to upload as translations.

phlax:
@unho you're quite right ill change the commit message to reflect that

unho:
@phlax If you are going forward with this fix, please also leave some comment on the original issue about it being fixed in a rather different way.

dwaynebailey:
@phlax this changes the way this behave in a rather fundamental way.  Before anyone who was authenticated could download any file. Now only if you have translate rights can you download.  I'm not sure if this has unintended consequences e.g. someone download es for cross-reference, etc.  I can't think of any concrete use case in operation now.  But be aware its changing more than just lets expose Upload to those with translate rights.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -128,16 +128,13 @@ def browse(request, translation_project, dir_path, filename=None):
...
         from import_export.views import handle_upload_form
-
         ctx.update(handle_upload_form(request))
-
-        has_download = (not translation_project.is_terminology_project and
-                        (check_permission('translate', request) or
-                         check_permission('suggest', request)))
         ctx.update({
-            'display_download': has_download,
+            'display_download': True,
```
unho: I think we might want to allow all authenticated users to download the translation files.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -128,18 +128,14 @@ def browse(request, translation_project, dir_path, filename=None):
...
-            'has_sidebar': True,
-        })
+    if "import_export" in settings.INSTALLED_APPS:
+        if not translation_project.is_terminology_project:
+            if request.user.is_authenticated():
+                if check_permission('translate', request):
+                    from import_export.views import handle_upload_form
+                    ctx.update(handle_upload_form(request, project))
+                ctx.update({'display_download': True,
+                            'has_sidebar': True})
```
unho: Nit: Keep previous layout for updating the dict, including trailing comma after each key-value pair.

phlax: im gonna pass on this - it complies with the policies

pootle/apps/pootle_translationproject/views.py
```diff
@@ -128,18 +128,14 @@ def browse(request, translation_project, dir_path, filename=None):
...
-                         check_permission('suggest', request)))
-        ctx.update({
-            'display_download': has_download,
-            'has_sidebar': True,
-        })
+    if "import_export" in settings.INSTALLED_APPS:
+        if not translation_project.is_terminology_project:
+            if request.user.is_authenticated():
+                if check_permission('translate', request):
+                    from import_export.views import handle_upload_form
```
unho: Nit: I usually put a blank line after imports to make code more readable.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -128,18 +128,14 @@ def browse(request, translation_project, dir_path, filename=None):
...
-
-        has_download = (not translation_project.is_terminology_project and
-                        (check_permission('translate', request) or
-                         check_permission('suggest', request)))
-        ctx.update({
-            'display_download': has_download,
-            'has_sidebar': True,
-        })
+    if "import_export" in settings.INSTALLED_APPS:
+        if not translation_project.is_terminology_project:
```
unho: IMHO we must allow people to download terminology. I am not 100% sure we should forbid people to upload it.

phlax: @unho this is fixed in a different PR

pootle/apps/pootle_translationproject/views.py
phlax: agreed im not sure why it is even there - ill try and move it - maybe circular import

unho: I doubt it is a circular import. Might be there because most users don't necessarily see the upload form, thus we only import this in case they will see it.

pootle/apps/pootle_translationproject/views.py
unho: If you mean #4191 I don't see how it is enabled there. Probably this is only because of it not being rebased on top of this PR. So if it is being handled elsewhere it is ok for me.

unho:
lgtm

pootle/apps/pootle_translationproject/views.py
phlax: yes exactly

Issue #4205, opened 2015-11-20
`pootle test_checks` dies with AttributeError


imankulov:
Hi! 

I found out that `test_checks` doesn't work properly in current master. Let me please report two (hopefully, minor) bugs in the same report.

1. If no parameters provided, command doesn't fail as it should. [This test](https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/management/commands/test_checks.py#L70) doesn't work, because `('' and '') == bool(None)` is actually `False`

2. When started with no `--checks` option, it dies immediately with AttributeError

```
$ pootle test_checks --source='Foo' --target='Foo %s'
2015-11-20 18:47:35,468 INFO Loading custom settings from '/home/pootle/.pootle/pootle.conf'...
2015-11-20 18:47:35,898 INFO Using Python PO
2015-11-20 18:47:36,810 ERROR Problem with check filter 'gconf': 'GnomeChecker' object has no attribute 'locations'
2015-11-20 18:47:36,812 ERROR Problem with check filter 'validxml': 'LibreOfficeChecker' object has no attribute 'locations'
2015-11-20 18:47:36,813 ERROR Problem with check filter 'credits': 'MozillaChecker' object has no attribute 'locations'
Traceback (most recent call last):
  File "/home/pootle/env/bin/pootle", line 9, in <module>
    load_entry_point('Pootle==2.7.3.dev20151120180616', 'console_scripts', 'pootle')()
  File "/home/pootle/pootle/pootle/runner.py", line 319, in main
    django_settings_module='pootle.settings')
  File "/home/pootle/pootle/pootle/runner.py", line 299, in run_app
    management.execute_from_command_line(command)
  File "/home/pootle/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/home/pootle/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 377, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/pootle/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/home/pootle/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/home/pootle/env/local/lib/python2.7/site-packages/django/core/management/base.py", line 533, in handle
    return self.handle_noargs(**options)
[...]

PR #4208, opened 2015-11-21
Make test_checks work with no options and against TTK checks

dwaynebailey:
This addresses #4205 

1. Ensuring we bail correctly when no options are given.  Also making the logic of what is required more explicit.
2. Working with TTK checks


unho:
lgtm.

Issue #4215, opened 2015-11-24
update_stores logic is no longer called when TP is initialised


phlax:
This is a regression from #3877 

If this is the desired behaviour ill make the necessary changes to restore it.

The other option is explicitly calling ``update_stores`` after adding TP - and changing docs to reflect that.

In order to fix i probably need to move the last_sync_revision logic back out of the management command so it can be called from other points in the code more easily

i also think i need to check anywhere that store.update is called and make sure it doesnt expect old last_sync_revision behaviour

dwaynebailey:
I'm trying to think of cases or instance that would have triggered:

1. When a TP was added and files are on disk IIRC that would load units into the TP
2. When a new language is created and no files are on disk it will create the TP from templates.
3. If update_store is called for a language that doesn't have a TP then the TP is created and the stores are loaded.

I think that's the previous behaviour in terms of interaction related to TPs.

PR #4218, opened 2015-11-24
TranslationProject initialisation cleanup

phlax:
Touch #4215

### DIFF
--- a/docs/server/installation.rst
+++ b/docs/server/installation.rst
@@ -46,7 +46,7 @@ Setting up the virtual environment
 
 In order to install Pootle first create a virtual environment. The virtual
 environment allows you to install dependencies independent of your system
-packages. 
+packages.
 
 Please install ``virtualenv`` from your system packages, e.g. on Debian:
 
@@ -167,29 +167,17 @@ Before you run Pootle for the first time, you need to create the schema for
 the database and populate it with initial data. This is done by executing the
 :djadmin:`migrate` and :djadmin:`initdb` management commands:
 
+.. note:: You will need to have an :ref:`RQ worker running
+   <installation#running-rqworker>` to complete this. Alternately, you can
+   use the :option:`--no-rq`.
+
 .. code-block:: bash
 
   (env) $ pootle migrate
   (env) $ pootle initdb
 
-
-.. _installation#refreshing-stats:
-
-Refreshing stats
-----------------
-
-On first installation you will need to generate the statistics from your
-database. You will need to have an :ref:`RQ worker running 
-<installation#running-rqworker>` to complete this.
-
-.. code-block:: bash
-
-   (env) $ pootle refresh_stats
-
-This command will dispatch jobs to the RQ worker and may take some time.
-
-If you wish to run :djadmin:`refresh_stats` in the foreground without using the RQ
-worker you can use the :option:`--no-rq` option.
+Running :djadmin:`initdb` will take some time as it will create the default
+Projects and Stores.
 
 
 .. _installation#admin-user:
--- a/tests/fixtures/models/language.py
+++ b/tests/fixtures/models/language.py
@@ -33,7 +33,7 @@ def _require_language(code, fullname, plurals=2, plural_equation='(n != 1)'):
 
 
 @pytest.fixture
-def english(root):
+def english(root, system):
     """Require the English language."""
     return _require_language('en', 'English')
 
@@ -84,3 +84,18 @@ def russian(english):
 def fish(english):
     """Require the Fish language ><(((º>"""
     return _require_language(code='fish', fullname='Fish')
+
+
+# due to issues with tests not having a clean slate ie:
+# (https://github.com/translate/pootle/issues/3898)
+# please do not use the klingon fixtures 8)
+@pytest.fixture
+def klingon(english):
+    """Require the Klingon language."""
+    return _require_language('kl', 'Klingon')
+
+
+@pytest.fixture
+def klingon_vpw(english):
+    """Require the Klingon language (VPW dialect)."""
+    return _require_language('kl_VPW', 'Klingon vegan peace warriors')
--- a/tests/models/revision.py
+++ b/tests/models/revision.py
@@ -25,7 +25,6 @@ def test_max_revision(af_tutorial_po):
     initial_max_revision = Unit.max_revision()
     initial_revision = Revision.get()
     assert initial_max_revision == initial_revision
-    assert initial_max_revision == 1
 
     # Let's make 10 translation updates, this must also update their
     # revision numbers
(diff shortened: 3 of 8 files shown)


phlax:
@ta2-1 if you get a chance can you take a look at this  - tia

phlax:
@julen i reckon this is pretty much cleaned up - if you or ta2-1 could take a look that would be really appreciated

docs/server/installation.rst
```diff
@@ -172,23 +172,12 @@ the database and populate it with initial data. This is done by executing the
...
-
-.. _installation#refreshing-stats:
-
-Refreshing stats
-----------------
-
-On first installation you will need to generate the statistics from your
-database. You will need to have an :ref:`RQ worker running 
+You will need to have an :ref:`RQ worker running 
 <installation#running-rqworker>` to complete this.
```
unho: This warning should be before the snippet with the commands to run.

docs/server/installation.rst
```diff
@@ -172,23 +172,12 @@ the database and populate it with initial data. This is done by executing the
...
 
-.. code-block:: bash
-
-   (env) $ pootle refresh_stats
-
 This command will dispatch jobs to the RQ worker and may take some time.
```
unho: Which command? `migrate`? `initdb`?

tests/fixtures/models/language.py
```diff
@@ -27,13 +27,14 @@ def _require_language(code, fullname, plurals=2, plural_equation='(n != 1)'):
         'nplurals': plurals,
         'pluralequation': plural_equation,
     }
+
```
unho: Can we just avoid adding these blank lines? Or at least put them on a separate commit that can be directly landed as this change is not directly related with any change in the surrounding code. Note that I am ok to keep this to allow this PR to be merged asap, so consider this as a reminder for the future.

tests/fixtures/models/language.py
```diff
@@ -84,3 +85,18 @@ def russian(english):
...
+@pytest.fixture
+def klingon(english):
+    """Require the Klingon language."""
+    return _require_language('kl', 'Klingon')
+
+
+@pytest.fixture
+def klingon_vpw(english):
+    """Require the Klingon language (VPW dialect)."""
+    return _require_language('kl_VPW', 'Klingon vegan peace warriors')
```
unho: You definitely need to do a Star Trek marathon. Perhaps twice :)

tests/models/revision.py
```diff
@@ -25,7 +25,6 @@ def test_max_revision(af_tutorial_po):
     initial_max_revision = Unit.max_revision()
     initial_revision = Revision.get()
     assert initial_max_revision == initial_revision
-    assert initial_max_revision == 1
```
unho: Is this change related to the rest of the changes on the PR?

phlax: yes - the test fails without removing this line

tests/models/translationproject.py
```diff
@@ -0,0 +1,81 @@
...
+from pootle_language.models import Language
+from pootle_project.models import Project
+from pootle_translationproject.models import TranslationProject
+
+
+@pytest.mark.django_db
+def test_tp_create_fail(tutorial, english):
+
+    # trying to create a TP with no Project raises a RelatedObjectDoesNotExist
+    # which can be caught with Project.DoesNotExist
```
unho: Nit: I don't particularly like texts starting with lowercase and without ending period. Every time I see one of these I have doubts about if part of the comment has been removed at the beginning or at the end, or in other words if comment is incomplete.

docs/server/installation.rst
unho: This is because some readers that are in a hurry might just copy and paste the commands, run them and then finish reading the instructions and notice that they ran something they actually shouldn't.

tests/fixtures/models/language.py
```diff
@@ -27,13 +27,14 @@ def _require_language(code, fullname, plurals=2, plural_equation='(n != 1)'):
...
     }
+
     language, created = Language.objects.get_or_create(**criteria)
 
     return language
 
 
 @pytest.fixture
-def english(root):
+def english(root, system):
```
unho: This looks like some bugfix that should go on its own commit, isn't it?

tests/fixtures/models/language.py
```diff
@@ -84,3 +85,18 @@ def russian(english):
 def fish(english):
     """Require the Fish language ><(((º>"""
     return _require_language(code='fish', fullname='Fish')
+
+
+# due to issues with tests not having a clean slate ie:
+# (https://github.com/translate/pootle/issues/3898)
+# please do not use the klingon fixtures 8)
```
unho: Which is the point on adding languages that cannot be used?

phlax: they are used

tests/fixtures/models/language.py
phlax: no - its not - its never been necessary before

tests/models/translationproject.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def _test_tp_stores_match(tp1, tp2):
+    # for testing tp creation from templates
+    assert tp1.stores.count() == tp2.stores.count()
+    tp1_stores = tp1.stores.all()
+    tp2_stores = tp2.stores.all()
+
+    for i, store1 in enumerate(tp1_stores):
+        store2 = tp2_stores[i]
```
unho: Perhaps simpler:

```python
    for store1, store2 in zip(tp1_stores, tp2_stores):
        assert store1.units == store2.units
```

tests/models/translationproject.py
```diff
@@ -0,0 +1,81 @@
...
+        assert store1.units == store2.units
```
unho: Perhaps simpler:

```python
    for store1, store2 in zip(tp1_stores, tp2_stores):
        assert store1.units == store2.units
```

phlax: nice - but not important right now - this function is not currently used

tests/models/translationproject.py
```diff
@@ -0,0 +1,81 @@
...
+    assert tp.stores.count() == template_tp.stores.count()
+    assert (
+        [(s, t)
+         for s, t
+         in template_tp.stores.first().units.values_list("source_f",
+                                                         "target_f")]
+        == [(s, t)
+            for s, t
+            in tp.stores.first().units.values_list("source_f",
+                                                   "target_f")])
```
unho: I bet we can make this a bit more unreadable by using Perl. What about using a helper function for getting the list for each of the TPs here?

unho:
lgtm

tests/models/translationproject.py
phlax: post 2.7.3 these tests will be tidied

PR #4226, opened 2015-11-26
Allow initdb to be run without creating projects

phlax:
This is potentially useful if eg you dont want the tutorial loaded

It also allows travis to run without importing the entire terminology
project if Stores are to be loaded when TPs are created.

...just updating docs...


julen:
:+1:

PR #4227, opened 2015-11-26
Enable PEP257 testing and fix hanging fruit

### LINKED ISSUE #4212: Automate pep257 - docstring - checks
PEP257 defines docstring conventions and we can test for them using the [pep257 tool](http://pep257.readthedocs.org/en/latest/error_codes.html)

dwaynebailey:
Fixes #4212 but should land post 2.7.3

* Enables PEP257 checking in Travis (cost 7s).  
* Makes current clean checks hard requirements
* Fixes some of the low hanging fruit

Big remaining issues are:
* Missing docstrings
* Single single first line (jury out on if we actually want this
* D4 checks which relate to word choice Return over Returns, Test over Tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ script:
   # FIXME reenable migration testing
   #- tests/version-migration/migrate.sh
   - make pep8
+  - pep257
   - py.test --cov=. -vv
 notifications:
   email:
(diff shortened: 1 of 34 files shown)


julen:
I didn't check the actual contents yet, nevertheless in terms of the Travis cost, I think we'd be better off placing this as a separate task, so that it's not run 3 times but just once. Basically the same applies to any other non DB-dependant checks.

dwaynebailey:
@julen lets deal with separation of meta from real tests in another issue.  7s is a small cost and I'd rather not confuse issues.

julen:
Agreed — since you stressed the penalty thought you'd be worried about it right now.

dwaynebailey:
@julen nah the other way around, wanted to stress that it wasn't costing minutes

unho:
lgtm

phlax:
lgtm

PR #4228, opened 2015-11-26
Add CONTRIBUTING guide

### LINKED ISSUE #3999: Add a CONTRIBUTING page to Pootle Github project
We have the CONTRIBUTING page @jleclanche suggested and also examples such as [this](https://github.com/skidding/cosmos/blob/master/CONTRIBUTING.md)

This bug is to track the review and massage of those to build our own Pootle CONTRIBUTING page.

unho:
This replaces PR #4108.

Fixes #3999.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,0 +1,68 @@
+How to contribute
+=================
+
+Reporting bugs
+--------------
+
+Before reporting please:
+
+- Double-check `the documentation
+  <http://docs.translatehouse.org/projects/pootle/>`_.
+- Make sure `the issue hasn't been already reported
+  <https://github.com/translate/pootle/issues/>`_.
+
+
+Write your bug report so that you ensure developers are fixing the bug rather
+than spending time trying to understand it.  If we can't see it or replicate it,
+likely we can't fix it.
+
+Checklist:
+
+#. Bug description is clear yet concise.
+#. Clear instructions to reliably replicate the issue.
+#. Screenshots rather than descriptions of screens.
+#. Full traceback if one occurred.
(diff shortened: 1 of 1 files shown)


julen:
I'd like to propose the following fixes: https://github.com/unho/pootle/compare/docs/contributing...julen:pr/4228-fixes

Ideally I'd prefer this to be markdown rather than rST too, not sure what your impressions are on the matter.

phlax:
i tend to avoid markdown these days as ive hit compatibility issue -  eg on pypi with it

unho:
I am ok with your suggestions except https://github.com/unho/pootle/commit/f63a5b26507ab6592e22c61c6fd940153d51f14f since IMHO if we tell people to ask for the better way to contribute we should point them to the Pootle development IRC channel (#pootle-dev) instead to the support one (#pootle). Anyway I will pick the "seek for help" rewording from that commit, and the rest of the changes.

Regarding using Markdown, we are using reStructuredText for the README, so I think it is better to stick with it.

phlax:
we have to use rst for the README or it wont render on pypi

julen:
I won't bike-shed, however note this doc is to be consumed by GH, not PyPI.

OTOH, the motivation behind the IRC change is the following: currently, many of the issues being opened are actual support questions. Another way to avoid that is to point people to a support channel, which #pootle-dev is not.

unho:
Ok, will use #pootle.

unho:
Included all @julen suggestions.

unho:
@translate/core Any gtm?

phlax:
gtm

unho:
Ok, I am merging this one too.

Issue #4239, opened 2015-11-27
Pyflakes cleanup


phlax:
There are a lot of pyflakes issues atm

Not sure if there any exceptions that anyone would like to see:

- ?

julen:
Can you please give some context and share which ones are these issue? TIA.

phlax:
@julen i put up the ticket for people to add any exceptions to pyflakes that they would like to see

personally i dont want to see any - but the obvious one that comes to mind is import modules for re-import. In this case we seem to have decided to use an ``__all__`` list to express why the imports are there - so that seems to be a non-issue

phlax:
once we have the pyflakes count to 0 ill close this ticket - which we are very close too - i just cleaned up any that were non-functional changes. There are a couple left - list comprehensions resetting local vars i think

julen:
So are you referring to these?
```bash
$ pyflakes . | grep -v migrations
./manage.py:15: 'syspath_override' imported but unused
./setup.py:164: 'syspath_override' imported but unused
./pootle/checks.py:250: 'textile' imported but unused
./pootle/checks.py:252: 'markdown' imported but unused
./pootle/checks.py:254: 'docutils' imported but unused
./pootle/runner.py:17: 'syspath_override' imported but unused
./pootle/apps/pootle_app/apps.py:18: 'pootle_checks' imported but unused
./pootle/apps/virtualfolder/helpers.py:44: list comprehension redefines 'dir_path' from line 35
./pootle/core/helpers.py:140: list comprehension redefines 'units' from line 134
./pootle/core/markup/filters.py:76: 'textile' imported but unused
./pootle/core/markup/filters.py:78: 'markdown' imported but unused
./pootle/core/markup/filters.py:80: 'docutils' imported but unused
```

phlax:
yep - thats them

julen:
* `syspath_override` issues can be silenced/ignored — it's used but it only has side-effects at import time (see #3608 too).
* The import for `pootle_checks` is needed AFAIK to register the checks themselves on app init.
* The markup checks can be silenced/ignored too: they only try to determine if such packages are available. This is code that might be going away depending on the verdict of https://github.com/translate/pootle/issues/4184#issuecomment-157179358

So from my understanding the actual issues left alive are:
```
./pootle/apps/virtualfolder/helpers.py:44: list comprehension redefines 'dir_path' from line 35
./pootle/core/helpers.py:140: list comprehension redefines 'units' from line 134
```

[...]

PR #4240, opened 2015-11-27
Pyflakes cleanup

phlax:
ref #4239

### DIFF
--- a/pootle/apps/pootle_app/views/admin/__init__.py
+++ b/pootle/apps/pootle_app/views/admin/__init__.py
@@ -7,6 +7,11 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-from .languages import *
-from .projects import *
-from .users import *
+from .languages import LanguageAdminView, LanguageAPIView
+from .projects import ProjectAdminView, ProjectAPIView
+from .users import UserAdminView, UserAPIView
+
+
+__all__ = (
+    LanguageAdminView, LanguageAPIView, ProjectAdminView, ProjectAPIView,
+    UserAdminView, UserAPIView)
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -18,7 +18,7 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.urlresolvers import reverse
-from django.db import models, transaction
+from django.db import models
 from django.db.models import F
 from django.template.defaultfilters import truncatechars
 from django.utils import timezone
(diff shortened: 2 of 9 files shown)


pootle/apps/pootle_app/views/admin/__init__.py
```diff
@@ -7,6 +7,9 @@
...
 
-from .languages import *
-from .projects import *
-from .users import *
+from .languages import LanguageAdminView, LanguageAPIView
+from .projects import ProjectAdminView, ProjectAPIView
+from .users import UserAdminView, UserAPIView
+
+(LanguageAdminView, LanguageAPIView, ProjectAdminView, ProjectAPIView,
+ UserAdminView, UserAPIView)
```
unho: Not sure I like this kind of change.

phlax: import * is not explicit and is regarded as bad

https://www.python.org/dev/peps/pep-0008/#id17

unho: I am ok with explicit import, but is it necessary to have this tuple?

phlax: without it triggers a pyflakes unused local var error

dwaynebailey: Does feel like a cludge to get tests to pass.  I'm OK with that but can we at least be consistent in placement/use of tuple?

unho: At least put two blank lines after the imports.

pootle/apps/pootle_store/models.py
```diff
@@ -18,7 +18,7 @@
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.urlresolvers import reverse
-from django.db import models, transaction
+from django.db import models
```
unho: Can you please directly land the adding missing imports and remove unused imports changes?

pootle/apps/pootle_app/views/admin/__init__.py
phlax: > Does feel like a cludge to get tests to pass. I'm OK with that but can we at least be consistent in placement/use of tuple?

Yep it is a kludge but its kinda *the* kludge. We could use the __all__ var ie:

```
from foo import ClassForReexport, OtherClassForReexport


__all__ = [ClassForReexport, OtherClassForReexport]
```
but its really only a cosmetic difference

phlax: i think using `__all__` would serve the purpose of shutting codacy up at least 8/

unho: +1 on using `__all__` as long it doesn't introduce any other issue.

dwaynebailey:
@phlax seriously I'd just have landed the missing and uneeded imports, they really don't need a PR.  Can we please have proper commit messages.  This introduced ``__all__ `` but does nothing to explain why.  Leaving it to some sucker at some unknown future date left wondering why.

The ``__all__`` still feels like a real kludge but I'm OK with it landing.

GTM with an actual human facing commit message.

PR #4241, opened 2015-11-27
Moving initdb functions into InitDB class

### LINKED ISSUE #4225: Create an InitDB class
i think it would be cleaner to put the functions in pootle.core.initdb into a class

phlax:
Fixes #4225

### DIFF
--- a/pootle/apps/pootle_app/management/commands/initdb.py
+++ b/pootle/apps/pootle_app/management/commands/initdb.py
@@ -15,7 +15,7 @@ os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
 
 from django.core.management.base import NoArgsCommand
 
-from pootle.core.initdb import initdb
+from pootle.core.initdb import InitDB
 
 
 class Command(NoArgsCommand):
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_app/management/commands/initdb.py
```diff
@@ -33,7 +33,7 @@ class Command(NoArgsCommand):
 
     def handle_noargs(self, **options):
         self.stdout.write('Populating the database.')
-        initdb(options["create_projects"])
+        InitDB().init(options["create_projects"])
```
julen: Maybe the naming is misleading and the intention is different, so correct me if I'm wrong, but I find this construct totally counter-intuitive — there's `__init__` for a purpose.

phlax: fair point - ill change

unho:
I have no additional concerns. lgtm once the raised concern is fixed.

pootle/apps/pootle_app/management/commands/initdb.py
unho: I thought that this would be changed to pass the options to `__init__` instead of `initdb`.

phlax: no - i just changed the name of the method - i didnt want to trigger the db install on object creation for several reasons

unho: I was talking about using `InitDB(options["create_projects"]).init_db()` instead.

phlax: seem more like an option to the init_db method than the constructor to me

julen:
Man, mind your commit messages :pray:

PR #4244, opened 2015-11-28
[WIP] Pep8: cleanups

phlax:


pootle/core/mixins/treeitem.py
```diff
@@ -434,7 +449,7 @@ def clear_all_cache(self, children=True, parents=True):
         self.mark_all_dirty()
         self.clear_dirty_cache(children=children, parents=parents)
 
-    ################ Update stats in Redis Queue Worker process ###############
+    # ############### Update stats in Redis Queue Worker process #############
```
unho: Why this blank space in the middle of the hashes?

pootle/core/mixins/treeitem.py
```diff
@@ -679,8 +701,10 @@ def update_cache_job(instance):
     # Django should take care of setting this up, but it doesn't yet
     # (fixed in Django 1.10):
     # https://code.djangoproject.com/ticket/16734
-    script_name = (u'/' if settings.FORCE_SCRIPT_NAME is None
-                        else force_unicode(settings.FORCE_SCRIPT_NAME))
+    script_name = (
+        u'/'
+        if settings.FORCE_SCRIPT_NAME is None
+        else force_unicode(settings.FORCE_SCRIPT_NAME))
```
unho: IMHO the string should be on the first line and the rest should be vertically aligned with it.

pootle/core/mixins/treeitem.py
```diff
@@ -642,8 +662,10 @@ def create_job(self, status=None, depends_on=None):
         Creates Job object with given job ID
         """
         args = (self.instance,)
-        return Job.create(self.func, args=args, id=self.id, connection=self.connection,
-                          depends_on=depends_on, status=status)
+        return Job.create(
+            self.func, args=args,
+            id=self.id, connection=self.connection,
+            depends_on=depends_on, status=status)
```
unho: I remember this code as something that I had to read several times to understand how the stats caching worked, so I would prefer to keep this even more readable than it was. My suggestion is to put an argument per line if possible.

phlax:
this has been addressed by @dwaynebailey 8)

PR #4245, opened 2015-11-28
Improving PEP8 performance

dwaynebailey:
This increases our PEP8 coverage as follows:

- One issue is taken down to zero (ensuring we don't regress but not worrying too much about creating new errors that we don't block on)
- Add checks that we've cleaned into our travis pep8 check

We got to this coverage:
```
92      E265 block comment should start with '# '
5       E501 line too long (94 > 84 characters)
```


pootle/apps/pootle_app/models/permissions.py
```diff
@@ -160,10 +162,10 @@ class Meta:
...
     directory = models.ForeignKey('pootle_app.Directory', db_index=True,
                                   related_name='permission_sets')
-    positive_permissions = models.ManyToManyField(Permission, db_index=True,
-            related_name='permission_sets_positive')
-    negative_permissions = models.ManyToManyField(Permission, db_index=True,
-            related_name='permission_sets_negative')
+    positive_permissions = models.ManyToManyField(
+        Permission, db_index=True, related_name='permission_sets_positive')
+    negative_permissions = models.ManyToManyField(
+        Permission, db_index=True, related_name='permission_sets_negative')
```
unho:
```python
    positive_permissions = models.ManyToManyField(
        Permission,
        db_index=True,
        related_name='permission_sets_positive',
    )
    negative_permissions = models.ManyToManyField(
        Permission,
        db_index=True,
        related_name='permission_sets_negative',
    )
```

See http://docs.translatehouse.org/projects/pootle/en/latest/developers/styleguide.html#fields-in-models-and-forms

phlax: i think its much cleaner how @dwaynebailey has done it

pootle/apps/pootle_store/views.py
```diff
@@ -503,11 +504,11 @@ def get_units(request):
...
 
         if sort_by_field is None or sort_on == 'units':
-            # Since `extra()` has been used before, it's necessary to explicitly
-            # request the `store__pootle_path` field. This is a subtetly in
-            # Django's ORM.
-            uid_list = [u['id'] for u in step_queryset.values('id',
-                                                              'store__pootle_path')]
+            # Since `extra()` has been used before, it's necessary to
+            # explicitly request the `store__pootle_path` field. This is a
+            # subtetly in Django's ORM.
```
unho: Typo in "subtetly"?

pootle/apps/pootle_translationproject/models.py
```diff
@@ -464,13 +461,11 @@ def gettermmatcher(self):
...
             self.non_db_state.termmatcher = Matcher(
-                    terminology_stores.iterator(),
-            )
+                terminology_stores.iterator())
             self.non_db_state.termmatchermtime = mtime
 
         return self.non_db_state.termmatcher
 
-
     ###########################################################################
```
unho: Instead of removing previous blank line, unindent this `##########` line.

pootle/apps/pootle_store/views.py
MozillaPootleL10nBot: This is a pep8 fix, not a spelling or grammar fix.  I didn't read any of the comments except very superficially.  So I won't be fixing any of these.  Feel free to land them after.

pootle/apps/pootle_store/models.py
```diff
@@ -1294,14 +1288,10 @@ def toggle_qualitycheck(self, check_id, false_positive, user):
...
-            field=SubmissionFields.NONE,
-            unit=self,
-            store=self.store,
-            type=sub_type,
-            quality_check=check
-        )
+                         translation_project=self.store.translation_project,
+                         submitter=user, field=SubmissionFields.NONE,
+                         unit=self, store=self.store, type=sub_type,
+                         quality_check=check)
```
unho: Previous code allowed to easily spot each argument on its own line. Is it not possible to keep it that way by just appending a comma after the last argument?

The same goes for several other similar changes, but would like to first hear other opinions about this particular instance.

dwaynebailey: @unho either are fine and valid.  I wans't primarily trying to increase readability around the long list of parameters dimension.  Primarily my aim is PEP8 which generally then does increase readability, something I'm actually finding with long lines that nobody wanted to fold.  Once I fold them I find I can read the statement without tracking my eyes.

The original has its own problems in that creation_time should also be wrapped in which case it's a proper list.  I'd rather leave the ideas of is this OK in the fix vs should it be a list to the next person who needs to work on the code.  As in some cases vertical space is harder to read then horizontal space.

julen:
Thank you for taking care of this @dwaynebailey.

I don't know if my understanding is correct, but I see this PR is looking towards being fully PEP8 compliant. It's worth noting that there are situations where such tests are meant to be invalid, since we make some exceptions on top of PEP8 completely on purpose: soft limit on line length, alignment of line continuations, and a few more. I'd strongly like to keep applying *some* of these exceptions.

If my understanding hasn't been correct, you can entirely disregard my comment.

unho:
I have the same feeling as @julen. Personally I prefer to keep some of the styleguide recommendations, but if we are going to stop applying them, like for example in commit "PEP8: E129 visually indented line with same indent as next logical line", then lets include the appropriate changes in the styleguide as well.

I am ok with some of the changes proposed to be directly merged. This should lighten the PR a bit. The commits I agree with are:

- PEP8: E711 comparison to None should be 'if cond is not None:'
- PEP8: E502 the backslash is redundant between brackets
- PEP8: ensure E228 and E231
- PEP8: E122 continuation line missing indentation or outdented
- PEP8: E124 closing bracket does not match visual indentation
- PEP8: E502 the backslash is redundant between brackets
- PEP8: E131 continuation line unaligned for hanging indent
- PEP8: E227 missing whitespace around bitwise or shift operator
- Cleanup: reflow # comments
- PEP8: E302 expected 2 blank lines
- PEP8: E241 multiple spaces after ':'
- PEP8: E123 closing bracket does not match indentation


There are other commits where I am ok with most of the changes, but not all of them.

dwaynebailey:
@julen 

> I don't know if my understanding is correct, but I see this PR is looking towards being fully PEP8 compliant.

The initial aim of the PR wasn't to be fully PEP8 compliant (technically that is impossible as some PEP8 suggestions would conflict, although I haven't found that tbh).  The initial aim was to clean up a category so that it could be 100% compliant and then make it a hard requirement to prevent regression.

But it became clear that:

1. Where we had soft issues, noise masked real correctable issues.  So you need zero or things will sneak in.
2. It was pretty easy to comply with some subtle changes.

Where it was impossible I left things as is for someone to track later.
 
> It's worth noting that there are situations where such tests are meant to be invalid, since we make some exceptions on top of PEP8 completely on purpose: soft limit on line length, alignment of line continuations, and a few more. I'd strongly like to keep applying some of these exceptions.

So from my review I have found the following:

1. The soft limit of 84 essentially hides a few that could be better done in 79, and I've managed to get almost everything below 79 without any objectionable gymnastics.  That said I actually am testing with 84 but did some cleanup with 79.  My finding is that long lines are almost always more readable when you don't need to track right.
2. Even with 84 we had a lot of incursions, down to 5 now though and we can actually get to zero but someone else would need to fix that as it would likely involve code change, or a style change that I wasn't happy with.
3. A number of the systems I found being used to split lines e.g. \ could be done better.  So although they worked with our styleguide they worked better when done differently.
4. Line continuation alignment is actually problematic in `if` blocks as it is confusing as to what is part of the condition and where the statement block begins.  Using a PEP8 compliant indent makes it much clearer what is the condition and where the code block starts.  This happens because we use () to surround complex conditions.  So this is one in our existing styleguide that I think we need to change as its in fact impairing readability.
5. Continuation of elaborate fn.fn().fn() I found worked just as well and many times cleaner by doing PEP8 compliant splits.  Without doing that we have to drop that check and then we regress with I think limited benefits as the PEP8 one works as is.

I think moving this forward can we please:

1. Separate concerns.  I'd like to land this, so can we please not major on minors.  It was a massive task to get things clean.  I'd really not like to see this railroaded for quite honestly less than 1% of cases. Lets focus on the major here which is getting an increase is automation.
2. Anyone is free to adjust anything back if its unreadable when they actually go to read that code, as long as they stay compliant.
3. For anything that we don't actually test we regress.  And regressing for the 1 place, that you're unlikely to need to look at ever, seems to me to defeat the object.

Lets land this unless there is a major issue on the specific check.  I created them in batches so that I could (relatively easily) drop ones that the team doesn't agree with.  But lets not block the whole thing because of minor changes that I've had to do to get us compliant.

Sorry if this sounds like a rant but I don't want to create too much dialogue on something that's a cleanup where the aim is the end benefit and where the pain that people are raising is so much smaller then the current pain we experience.

phlax:
great work @dwaynebailey and a massive improvement on what is there already.

with the caveat that i havent run the code at all - it lgtm

dwaynebailey:
So I'm landing this without E129 @unho raised some concerns.  And E501 (line length) @phlax had concerns on some of the approaches I took to get to under 84.

So I'm landing without those and will PR them on their own.

PR #4246, opened 2015-11-30
Cleanups: drop use of % in logging

dwaynebailey:
logging can accept variables for interpolation as parameters.  This
means they're only interpolated if something is logged.  Using % means
that the interpolation is always happening (as far as I understand).

### DIFF
--- a/pootle/apps/accounts/utils.py
+++ b/pootle/apps/accounts/utils.py
@@ -173,11 +173,11 @@ class UserPurger(object):
         self.revert_units_state_changed()
 
         # Delete remaining submissions.
-        logger.debug("Deleting remaining submissions for: %s" % self.user)
+        logger.debug("Deleting remaining submissions for: %s", self.user)
         self.user.submission_set.all().delete()
 
         # Delete remaining suggestions.
-        logger.debug("Deleting remaining suggestions for: %s" % self.user)
+        logger.debug("Deleting remaining suggestions for: %s", self.user)
         self.user.suggestions.all().delete()
 
     @write_stdout(" * Removing units created by: %(user)s... ")
@@ -194,7 +194,7 @@ class UserPurger(object):
 
             if not other_subs.exists():
                 unit.delete()
-                logger.debug("Unit deleted: %s" % repr(unit))
+                logger.debug("Unit deleted: %s", repr(unit))
 
     @write_stdout(" * Reverting unit comments by: %(user)s... ")
     def revert_units_commented(self):
@@ -215,12 +215,12 @@ class UserPurger(object):
                 unit.translator_comment = last_comment.new_value
                 unit.commented_by = last_comment.submitter
                 unit.commented_on = last_comment.creation_time
-                logger.debug("Unit comment reverted: %s" % repr(unit))
+                logger.debug("Unit comment reverted: %s", repr(unit))
             else:
                 unit.translator_comment = ""
                 unit.commented_by = None
                 unit.commented_on = None
-                logger.debug("Unit comment removed: %s" % repr(unit))
+                logger.debug("Unit comment removed: %s", repr(unit))
 
             # Increment revision
             unit._comment_updated = True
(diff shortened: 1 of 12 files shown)


phlax:
hmm - my feeling is stick to interpolation its more pythonic and it works everywhere

dwaynebailey:
Perhaps some other motivators:

- Pythonic for %, maybe but consider that:
  - There's also ``.format()``
  - It's also the [proposed model of how to do this](https://docs.python.org/2/library/logging.html#logging.Logger.debug)
- We're inconsistent and use both for no other reason then I think we forgot that logging doesn't need %.
- And perhaps these [better motivations](https://www.simonmweber.com/2014/11/24/python-logging-traps.html) which make things more reliable it would seem.

pootle/apps/accounts/utils.py
```diff
@@ -287,16 +287,14 @@ def revert_units_reviewed(self):
                 previous_review = reviews.latest('pk')
                 unit.reviewed_by = previous_review.submitter
                 unit.reviewed_on = previous_review.creation_time
-                logger.debug("Unit reviewed_by reverted: %s"
-                             % (repr(unit)))
+                logger.debug("Unit reviewed_by reverted: %s", (repr(unit)))
```
unho: Unnecessary parentheses around `repr`.

pootle/apps/accounts/utils.py
```diff
@@ -287,16 +287,14 @@ def revert_units_reviewed(self):
...
             else:
                 unit.reviewed_by = None
                 unit.reviewed_on = None
 
                 # Increment revision
                 unit._target_updated = True
-                logger.debug("Unit reviewed_by removed: %s"
-                             % (repr(unit)))
+                logger.debug("Unit reviewed_by removed: %s", (repr(unit)))
```
unho: Unnecessary parentheses around `repr`.

pootle/apps/accounts/utils.py
```diff
@@ -334,8 +332,7 @@ def revert_units_state_changed(self):
                 # Increment revision
                 unit._state_updated = True
                 unit.save()
-                logger.debug("Unit state reverted: %s"
-                             % (repr(unit)))
+                logger.debug("Unit state reverted: %s", (repr(unit)))
```
unho: Unnecessary parentheses around `repr`.

pootle/apps/pootle_app/management/commands/update_stores.py
```diff
@@ -56,8 +56,8 @@ def handle_store(self, store, **options):
         disk_mtime = store.get_file_mtime()
         if not options["force"] and disk_mtime == store.file_mtime:
             # The file on disk wasn't changed since the last sync
-            logging.debug(u"File didn't change since last sync, skipping "
-                          u"%s" % store.pootle_path)
+            logging.debug(u"File didn't change since last sync, "
+                          "skipping %s", store.pootle_path)
```
unho: String should start with `u"`.

pootle/apps/pootle_store/models.py
```diff
@@ -1849,9 +1849,9 @@ def update_from_disk(self, overwrite=False):
                 )
             self.last_sync_revision = update_revision
             if update_unsynced:
-                logging.info(u"[update] unsynced %d units in %s [revision: %d]"
-                             % (update_unsynced, self.pootle_path,
-                                update_revision))
+                logging.info(u"[update] unsynced %d units in %s "
+                             "[revision: %d]", update_unsynced,
```
unho: String should start with `u"`.

pootle/apps/pootle_app/management/commands/update_stores.py
dwaynebailey: Not in a line continuation I believe.

unho:
Nits aside, lgtm.

pootle/apps/pootle_app/management/commands/update_stores.py
unho: Perhaps it is not a requirement (I thought it was but couldn't find anything supporting this idea) but IMHO it make easier to spot this is the continuation of the previous line unicode string.

PR #4247, opened 2015-11-30
[RFC] Cleanup/comprehension

dwaynebailey:
I'm not sure if I got this 100% correct.  But 2.7 introduces set/dict comprehension which I'm trying to use here.


unho:
lgtm

PR #4248, opened 2015-11-30
[RFC] Cleanup: Use @staticmethod

dwaynebailey:
Some of these might not be valid, but a few are.  We should mark things as `@staticmethod` if they are indeed that.  I'm not sure how much of a perf boost is involved but it certainly makes it clear.  Some of what I've selected may in fact not be valid thus the RFC.

### DIFF
--- a/pootle/apps/contact/forms.py
+++ b/pootle/apps/contact/forms.py
@@ -54,7 +54,8 @@ class ContactForm(MathCaptchaForm, OriginalContactForm):
             self.cleaned_data['email']
         )
 
-    def recipient_list(self):
+    @staticmethod
+    def recipient_list():
         return [settings.POOTLE_CONTACT_EMAIL]
 
     def save(self, fail_silently=False):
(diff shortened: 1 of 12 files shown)


pootle/apps/contact/forms.py
```diff
@@ -54,7 +54,8 @@ def from_email(self):
             self.cleaned_data['email']
         )
 
-    def recipient_list(self):
+    @staticmethod
```
unho: This is extending a class that had this as a regular method. So undo this change.

pootle/apps/contact/views.py
```diff
@@ -47,7 +47,8 @@ def get_initial(self):
 
         return initial
 
-    def get_success_url(self):
+    @staticmethod
```
unho: Please undo this since it is extending a class that didn't had this as a staticmethod.

unho:
I have doubts about most of the changes here. I think that there are several methods that should be kept as regular ones so they correctly override methods from parent classes. The only one I don't it is a problem is the one in `pootle/core/search/backends/elasticsearch.py`.

dwaynebailey:
Dropping, minimal value

PR #4249, opened 2015-11-30
[RFC] Cleanup/quantifiedcode fixes

dwaynebailey:
Various cleanups picked up but https://www.quantifiedcode.com/

Some I'm not sure about, some are pretty straight forward and could land as is.  Would like some feedback as to which can land easily and which might be contentious.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -111,16 +111,15 @@ else:
             build_icons_inc = True
 
 if build_icons_inc:
-    icons_txt_file = open(icons_inc_file_name, "w")
-    for icon_image in os.listdir(icons_dir):
-        icon_name = icon_image[:icon_image.rfind(".")]
-        print >>icons_txt_file, ".. |icon:" + icon_name + "| " + \
-                                "image:: /" + icons_dir + "/" + icon_image
-        print >>icons_txt_file, "                      :alt: " + \
-                                icon_name.replace("-", " ").replace("_", " ") + \
-                                " icon"
-        print >>icons_txt_file
-    icons_txt_file.close()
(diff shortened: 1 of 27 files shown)


unho:
I like all the changes except "Avoid `None` as a redundant second argument to `dict.get()`". Can you please explain that one a bit more?

dwaynebailey:
@unho I had similar reservations to ``dict.get('somthing', None)``.  By default ``get()`` will return ``None`` if the key is not found, thus ``None`` in this case is redundant and thus not needed.  I was in two minds as to whether we should all just know that or if it should be explicit.  Thus its here for comments.

Maybe more of a concern is that this does expose code like:

```py
a = dict.get('some', None)
if a:
    do_something()
```
When in fact this should be:

```py
a = dict.get('some', None)
if a in not None:
    do_something()
```

unho:
I withdraw my previous block on "Avoid `None` as a redundant second argument to `dict.get()`" after reading https://docs.python.org/2/library/stdtypes.html#dict.get

So lgtm.

unho:
Anyway I would include the explanation on the commit message, with the https://docs.python.org/2/library/stdtypes.html#dict.get link.

julen:
I'm +1 on keeping `dict.get('foo', None)`, it's way more explicit than the default behavior. With this it's immediately clear subsequent checks need to be against `None`: `x is None`, `x is not None`.

The rest of the commits look perfect.

dwaynebailey:
Landed this without the ``dict.get(,None)`` changes, decided to take @julen's take on this specifically because its explicit.

PR #4252, opened 2015-12-01
Cleanup: if conditions following failed .get()

dwaynebailey:
.get() by default will return None if the key is not found in the
dictionary.

Thus if the following does not find the key then x == None.

x = dict.get('something', None)

This requires any conditional to test against None as follows:
if x is not None:

But in place in the code we're doing:
if x:

This correct that and simplifies some conditionals if needed.

### DIFF
--- a/pootle/apps/pootle_app/forms.py
+++ b/pootle/apps/pootle_app/forms.py
@@ -75,7 +75,7 @@ class ProjectForm(forms.ModelForm):
 
         def clean_localfiletype(self):
             value = self.cleaned_data.get('localfiletype', None)
-            if not value:
+            if value is None:
                 value = self.instance.localfiletype
             filetypes = [x[0] for x in filetype_choices]
             if value not in filetypes:
@@ -86,7 +86,7 @@ class ProjectForm(forms.ModelForm):
 
         def clean_treestyle(self):
             value = self.cleaned_data.get('treestyle', None)
-            if not value:
+            if value is None:
                 value = self.instance.treestyle
             return value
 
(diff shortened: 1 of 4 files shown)


julen:
:+1: thank you @dwaynebailey.

dwaynebailey:
@julen @phlax raised a concern that there may be cases where the old version is valid.  I'll need to look at these again.  But this is the bulk of them.

pootle/apps/pootle_app/forms.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, *args, **kwargs):
 
         def clean_localfiletype(self):
             value = self.cleaned_data.get('localfiletype', None)
-            if not value:
+            if value is None:
```
unho: What if `localfiletype` is `''` (empty string)?

unho: I have the same concern for several other similar changes.

julen: This is a valid concern in this context (form field values of character types; their `to_python()` will return `''` for empty values), and probably the best bet is to replace `None` with `''` as the default in the `.get()` call.

unho: +1

dwaynebailey:
@translate/core I'm not going to be able to look at this.  It did expose some potential issues and i'm not the best person to check those out.  So if anyone wants to pick this up, please do.

phlax:
i would rather leave logic changes to when the logic is being changed

dwaynebailey:
Going to close this.  Current pushback is to clean then when we clean it.

PR #4253, opened 2015-12-01
Reenable pep257, __all__ should contain strings

dwaynebailey:
http://docs.pylint.org/features.html#id5

invalid-all-object (E0604): Invalid object %r in `__all__`, must contain only strings Used when an invalid (non-string) object occurs in `__all__`.

Changing this allows us to reenable pep257 checks on Travis.

### DIFF
--- a/pootle/apps/pootle_app/views/admin/__init__.py
+++ b/pootle/apps/pootle_app/views/admin/__init__.py
@@ -13,5 +13,5 @@ from .users import UserAdminView, UserAPIView
 
 
 __all__ = (
-    LanguageAdminView, LanguageAPIView, ProjectAdminView, ProjectAPIView,
-    UserAdminView, UserAPIView)
+    'LanguageAdminView', 'LanguageAPIView', 'ProjectAdminView',
+    'ProjectAPIView', 'UserAdminView', 'UserAPIView')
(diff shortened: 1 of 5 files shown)


unho:
I remember seeing some chat re this, but as far as I can tell this lgtm.

phlax:
@unho the chat was that as the pep257 tool is not entirely stable - i would rather it was not included in travis - although i support us keeping to the pep257 standard.

the other chat was that we are introducing another standard (ie pylint) when we havent finished cleaning up pep8 - which i dont think is wise

dwaynebailey:
Yip agree with @phlax, this can rest until pep8 is in and stable.

julen:
Can someone share a link to the mentioned chat?

unho:
@julen This was a chat on Flowdock. I am afraid I can't share any link here.

julen:
Thanks for clarifying @unho. It'd be good to have such conversations in the open though.

julen:
The PR looks good to me.

unho:
lgtm.

phlax:
im still -1 on automating tests with a tool that i have already wasted time tracking down (unresolved) bugs on

dwaynebailey:
@phlax this landed without automation enabled.

Personally I think we can enable it.  As apart from the fixes to ``__all__`` included here and previously, the tool works like charm.  But I'm not going to fight about it.

But please don't complain when reviews include docstring fix requests that this would have caught, <cough>.

PR #4255, opened 2015-12-01
PEP8 all the way to zero

dwaynebailey:
These are the last PEP8 cleanups which raised some contention.  They result in a zero count for pep8 v1.5.7 and have only 7 E731 errors for pep v 1.6.2


unho:
"PEP8: E129 visually indented line with same indent as next logical line" commit would require altering the styleguide accordingly.

phlax:
> "PEP8: E129 visually indented line with same indent as next logical line" commit

Im not so happy with this commit either - i would really rather pull these conditions out of the if statement

pootle/apps/pootle_app/forms.py
```diff
@@ -64,13 +64,14 @@ def __init__(self, *args, **kwargs):
             ).count
 
             if has_stores:
-                #self.fields['localfiletype'].widget.attrs['disabled'] = True
+                # self.fields['localfiletype'].widget.attrs['disabled'] = True
```
unho: What about instead removing that commented code?

pootle/apps/pootle_app/models/directory.py
```diff
@@ -57,7 +57,7 @@ class Meta(object):
         ordering = ['name']
         app_label = "pootle_app"
 
-    ############################ Properties ###################################
+    # # # # # # # # # # # # # #  Properties # # # # # # # # # # # # # # # # # #
```
unho: I suppose that using other character doesn't fix the problem entirely because other rule would enforce a blank space to be after the initial hash, isn't it?

unho:
I don't particularly like most of the changes on "PEP8: E501 line too long". I think we need to get to some agreement on which is the best way to get long lines wrapped.

pootle/apps/pootle_app/forms.py
dwaynebailey: Likely that's a better approach.  But we're left with cases where temporary code is wanting to be commented out.  We could decide to ignore this check.

phlax:
> that said interestingly @phlax prefers this style of if layout

```
if (a
    and b
    or c):
    do_something()
```
eh?

im not sure how i gave you that idea. to repeat - i *never* put if conditions on multiple lines - and this example is also not valid pep8-wise

dwaynebailey:
@phlax sorry of I put words in your mouth.  But the number of times I've heard, "Please no W503".  And that example is W503! Even if the specific multiline `if` isn't what you would do, it is W503 with the operator not at the end of the line.

phlax:
for the avoidance of doubt - my preferred solution
```
worth_doing = (
    some_long_condition is True
    and some_other_condition is False)

if worth_doing:
    do_it_well()

```

dwaynebailey:
@phlax man sometimes github issues and PRs aren't the place to talk.  That example was __not__ meant as a proposal on how to format if statements.  But an illustration of the working of _operator at the start_!  Thus I will not pick apart the use of ``is True`` in your own example ;)

dwaynebailey:
Taking a step back.  We seem to be missing the high level picture here.  Which was __increase automated linting__.  The decisions we really need to make is do we land or do we reject XNNN test?

I don't think its very productive to debate the styleguide here as the context is wrong and my aims are __zero__ _not_ __perfection__. 

I have a number of motivations and justifications but I won't bore you with them here.

So going forward:
* Are there tests we can sign off on or should we turn it into a default ignore check?  These are E129, E265, E266, E501?
* Are there tests we want to keep but want to see them reworked in their own PR?

unho:
> So going forward:
> * Are there tests we can sign off on or should we turn it into a default ignore check?  These are E129, E265, E266, E501?
> * Are there tests we want to keep but want to see them reworked in their own PR?

- E129: IMHO this all comes from wrapping long lines. So this requires discussing styleguide. I would ignore it for now until we reach some consensus. I am ok with keeping this on a separate PR.
- E265: It exposes the commented code issue. I think it is better to keep the check enabled and remove commented code to get this to zero.
- E266: I would prefer to ignore it.
- E501: I am not sure we should enforce this, but I prefer "PEP8: E501 line length - early exit" and "PEP8: E501 hard one" commits to be landed. Note that there is other commit for this check that I prefer to be held until we discuss long lines wrap (this is related to E129 above).

pootle/apps/pootle_app/forms.py
julen: I don't see compelling reasons to keep commented out code — we use VCSs after all, and if there's something important to warn other developers, commented out code without any extra details is not the best approach IMHO.

julen:
The E266 commit reads (@dwaynebailey):
> This does feel like its not the right approach.  I wondered if many of
those comments add any value.

Some of these comments look valid to me, as a way to warn devs specific methods are being overriden from parent classes. Not everyone is familiar with parent classes (which might even come from 3rd party libraries), and unfortunately this implicitness is something OOP carries on by design — I'd love if there was a better way to signal this though.

julen:
I'm happy with the changes, provided other stylistic concerns are addressed separately.

julen:
Ah, an extra (general) thought on line length: these are frequently caused due to excessive nesting in the code, which is a code smell. Splitting methods/functions into smaller manageable pieces, exiting early (as in 6134021) and [keeping the code flat](https://philikon.wordpress.com/2010/01/26/flat-is-better-than-nested-bail-out-early/) always helps.

dwaynebailey:
E265 has landed on master, code was tracked down to its arrival and removed.

dwaynebailey:
E501 c10183ce041e357ec6d4b80ee59247a4d451c32b and 34fec4a981e782ac9e570f58ede71786b7e71895 landed on master as per https://github.com/translate/pootle/pull/4255#issuecomment-161264894

dwaynebailey:
E266 8cddba0f13460e7b86ab67dc32aba6575aa77814 (I should have revised the comment).  The comments must stay as per https://github.com/translate/pootle/pull/4255#issuecomment-161320591 so I adopted the reformatted header to work around the check.

dwaynebailey:
OK, last push.  I assume I interpreted everyone's feedback as required.

E501 landed, I'm trusting with E129 gone there is some general happiness on the styles I've used, if not please reflow when you next need to hack on it.
E129 disabled - there was just too much disagreement and pushback so I've marked that as off by default.

julen:
Which one was the E129 commit?

dwaynebailey:
@julen [E129](http://pep8.readthedocs.org/en/latest/intro.html#error-codes) visually indented line with same indent as next logical line

Essentially it is raised by our ``if`` blocks of:

```py
if (one or
    two or
    three):
    do_something()
```
The error is because ``three`` and ``do_something()`` are at the same indent, thus making it difficult to distinguish between the conditions and the statements.

julen:
Ok thanks for clarifying. Knowing that, I'm completely fine with it — in fact, it serves a good purpose.

PR #4260, opened 2015-12-02
Add sorting by contributions to contributors and rewrite docs

unho:


unho:
Updated to address raised concerns.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -44,6 +51,10 @@ def handle_all(self, **options):
         if options["revision"]:
             units = units.filter(revision__gte=options["revision"])
 
+        if options["sort_by"] not in ["name", "contributions"]:
+            # Specified sorting is not valid. Fall back to default.
+            options["sort_by"] = "name"
```
dwaynebailey: [``choiced=``](https://docs.python.org/2/library/optparse.html#optparse.Option.choices) in ``make_options`` will do this for free.

dwaynebailey:
Otherwise lgtm

pootle/apps/pootle_app/management/commands/contributors.py
unho: Investigating that.

phlax:
nice one @unho !

Issue #4213, opened 2015-11-24
Changing language code breaks Pootle


unho:
The problem is that basically changing a language code doesn't trigger the necessary changes on `pootle_path` for the related TPs and its Directories, Stores and VFolderTreeItems. Thus stats don't work, and other Pootle features don't work as expected (like syncing files to disk). There are two proposed tasks to fix this issue:

1. Forbid to change the language code for existing languages. This can be landed on its own but 2) is required to avoid losing the ability to alter the language code.
2. Add a management command to alter a given language code that would perform all the required changes:

- Disable all the language TPs so they don't get updates either from disk or from the UI,
- Sync all language stores to disk to ensure no translations are lost,
- Flush stats and caches related to this language,
- Change the language code,
- Copy files on disk from old locations to new locations (research how this might affect `update_stores` since we would like to avoid triggering any update since we just synced translations),
- Change `pootle_path` for the affected objects so the new language code is used,
- Recreate stats for the whole language (preferably TP by TP),
- Any other relevant change?
- Enable all the language TPs

The expected final outcome is that no attributions or permission assignments are lost, the language code is renamed and Pootle is still working.

phlax:
ref #4152

dwaynebailey:
I'd be concernd that that the command does too much, like syncing to disk.  I'd say that is the admin's task not the commands task.  I know why we want to do it, but it has unintended consequences if the admin doesn't know if will do something that actually is unrelated to the intention.

unho:
If sync to disk is not included in the command then we need to ask the user running it to manually disable the TPs before syncing, and then enabling them back. I think it is better to include this on the command.

dwaynebailey:
Disabling is unrelated to syncing.

dwaynebailey:
Syncing will change files on the filesystem or make commits all of which is likely unexpected.  Disabling is not destructive, sync_stores is.

julen:
Unless I'm missing some detail, this is no different from renaming a project — I'd consider it a dupe of #3764.

unho:
It is indeed related. I wouldn't consider this necessarily a dupe.

I suppose that 1) can be achieved with something similar to https://github.com/translate/pootle/commit/7348f8a767efeba7e5b18552adbac803f75ef9f7 ?

julen:
> I wouldn't consider this necessarily a dupe.

Mind elaborating why?

unho:
[...]

PR #4262, opened 2015-12-02
Languages admin: prevent from modifying existing language codes

unho:
This is a workaround until #4213 is properly addressed.


julen:
Basically it's the same as `ProjectForm`, it'll work fine so ship it.

unho:
Thanks. That is the lgtm I was waiting for. Merging it.

PR #4263, opened 2015-12-02
Travis: run META build tests apart from platform

dwaynebailey:
Run a META build which does docs, pep8, etc apart from the real code check.

It does make the Travis file a bit unwieldy though.  But we push about a minute of tests into META.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ script:
   - if [[ $META ]]; then make docs; fi
   - if [[ $META ]]; then make mo-all; fi
   - if [[ $META ]]; then make pep8; fi
+  - if [[ $META ]]; then ./setup.py sdist; fi
   # - if [[ $META ]]; then pep257; fi
   - if [[ ! $META ]]; then make assets; fi
   - if [[ ! $META ]]; then DJANGO_SETTINGS_MODULE=pootle.settings ./setup.py build_checks_templates; fi


.travis.yml
```diff
@@ -25,25 +26,23 @@ before_script:
...
-  - TESTPID=$!
-  - sleep 10
-  - kill -2 $TESTPID
+  - if [[ $META ]]; then python -m compileall -q -f .; fi
+  - if [[ $META ]]; then make docs; fi
+  - if [[ $META ]]; then make mo; fi
+  - if [[ $META ]]; then make pep8; fi
+  # - if [[ $META ]]; then pep257; fi
+  - if [[ ! $META ]]; then make assets; fi
+  - if [[ ! $META ]]; then DJANGO_SETTINGS_MODULE=pootle.settings ./setup.py build_checks_templates; fi
```
unho: I think this is something we want to run just once.

.travis.yml
```diff
@@ -25,25 +26,23 @@ before_script:
...
+  - if [[ ! $META ]]; then ./manage.py migrate --noinput --traceback; fi
+  - if [[ ! $META ]]; then ./manage.py migrate | (egrep "Your models have changes" && exit 1 || exit 0); fi
```
unho: This also looks like something we want to run once.

phlax:
@dwaynebailey you could maybe cut down the unwieldiness by switching the METAVAR and doing 

```
  - if [[ ! $NOTMETA ]]; then PYTEST CMD; fi 

```

.travis.yml
```diff
@@ -25,25 +26,23 @@ before_script:
...
+  - if [[ ! $META ]]; then ./manage.py runserver --traceback; fi &
+  - if [[ ! $META ]]; then TESTPID=$!; fi
+  - if [[ ! $META ]]; then sleep 10; fi
+  - if [[ ! $META ]]; then kill -2 $TESTPID; fi
   # FIXME reenable migration testing
   #- tests/version-migration/migrate.sh
-  - make pep8
-  # - pep257
-  - py.test --cov=. -vv
+  - if [[ ! $META ]]; then py.test; fi
```
unho: If the tests don't use the databases, then this is something we want to run just once.

phlax: the tests **do** use the database! 

also please dont get rid of cov or the -vv

unho: Then forget about my comment.

.travis.yml
dwaynebailey: I see $META as not part of the core code.  So no don't want to pull that out just yet.

.travis.yml
dwaynebailey: Ditto as above

julen:
Thank you for taking care of this @dwaynebailey; however, as you say, I find this completely unwieldy.

What about getting some help from [Tox](https://testrun.org/tox/latest/)? It helps keep things [extremely simple for Travis](https://github.com/jleclanche/django-push-notifications/blob/master/.travis.yml), yet [configurable](https://www.dominicrodger.com/tox-and-travis.html).

dwaynebailey:
@julen I looked at tox previously and found it confusing.  But that link its pretty neat.

If you're OK with it, I'd prefer to land this before trying the tox route.

julen:
If it helps having this in quickly go for it, but please ensure Tox is added sooner than later.

unho:
lgtm.

Tox looks great as it seems that allows to offload the docs build and other tasks to a completely separated job quite easily.

dwaynebailey:
This landed in c841e11bf227b92027c2ea43f1942f5f7225a9e3...31d5d23d9628629b7f583fbf91a6db057159854a

PR #4264, opened 2015-12-02
Add Unit priority field and some fixes/tests for VirtualFolders

phlax:


pootle/apps/pootle_store/__init__.py
```diff
@@ -0,0 +1,2 @@
+
+default_app_config = 'pootle_store.apps.PootleStoreConfig'
```
unho: Missing licensing.

pootle/apps/virtualfolder/__init__.py
```diff
@@ -0,0 +1,2 @@
+
+default_app_config = 'virtualfolder.apps.PootleVirtualFolderConfig'
```
unho: Missing licensing.

pootle/apps/pootle_store/util.py
```diff
@@ -138,3 +138,9 @@ def parse_pootle_revision(store):
 
 def get_state_name(code, default="untranslated"):
     return STATES_NAMES.get(code, default)
+
+
+def vfolders_installed():
+    if 'virtualfolder' in settings.INSTALLED_APPS:
+        return True
+    return False
```
unho: The following is much simpler:

```python
    return 'virtualfolder' in settings.INSTALLED_APPS
```

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,13 @@
+
+
```
unho: Missing licensing.

pootle/apps/pootle_store/models.py
```diff
@@ -652,6 +659,21 @@ def getorig(self):
...
+        vfolder_priorities = (
+            self.vfolders.values_list("priority", flat=True))
+
+        if vfolder_priorities:
+            priority = max(vfolder_priorities)
+        else:
+            priority = 1
+
+        if priority != self.priority:
+            Unit.objects.filter(pk=self.pk).update(priority=priority)
```
unho: Is there any reason why `self.update(priority=priority)` doesn't work?

phlax: its the objects.update method that needs to be used here to avoid hitting save

pootle/apps/pootle_store/receivers.py
```diff
@@ -0,0 +1,10 @@
+
+from django.db.models.signals import post_save
```
unho: Missing licensing.

pootle/apps/pootle_store/models.py
```diff
@@ -356,6 +359,11 @@ class Unit(models.Model, base.TranslationUnit):
                                     db_index=True, related_name='reviewed')
     reviewed_on = models.DateTimeField(db_index=True, null=True)
 
+    # this is calculated from virtualfolders if installed and linked
+    priority = models.FloatField(
+        db_index=True, default=1,
+        validators=[MinValueValidator(0.9)])
```
unho: The minimum value is `0.0`.

unho: Since this is calculated from vfolders, perhaps it is pointless to enforce any value checking here.

unho: Actually `0.0` is not even valid: https://github.com/phlax/pootle/blob/unit_priority/pootle/apps/virtualfolder/models.py#L202-L203

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,13 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from virtualfolder.models import VirtualFolder
+
+
+@receiver(post_save, sender=VirtualFolder)
+def vfolder_unit_priority_handler(sender, **kwargs):
+
+    for unit in kwargs["instance"].units.all():
```
unho: Perhaps this should be moved to `VirtualFolder.save()` because that way we can check if the `priority` actually changed and then avoid updating any unit.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,13 @@
+
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from virtualfolder.models import VirtualFolder
+
+
+@receiver(post_save, sender=VirtualFolder)
```
unho: Please note that priorities must be recalculated when a VirtualFolder is deleted.

pootle/apps/pootle_store/receivers.py
```diff
@@ -0,0 +1,10 @@
...
+from django.dispatch import receiver
+
+from pootle_store.models import Unit
+
+
+@receiver(post_save, sender=Unit)
+def unit_priority_handler(sender, **kwargs):
```
unho: Since this only makes sense when `virtualfolder` app is enabled, what about moving this handler to that app?

pootle/apps/pootle_store/models.py
phlax: good spot on the typo

validation is always good

i know that mlo uses sub 1 values so ill set to 0.1

pootle/apps/pootle_store/receivers.py
phlax: yep - good idea

pootle/apps/virtualfolder/receivers.py
phlax: hmm - i like the idea of not checking if it hasnt changed but i would rather not move it into save - ill check now

pootle/apps/virtualfolder/receivers.py
phlax: good spot!

pootle/apps/virtualfolder/receivers.py
unho: We are not having in account the scenario where a VirtualFolder changes its location, and then suddenly a bunch of units are not included on it.

pootle/apps/pootle_store/models.py
unho: That will require altering docs and changing https://github.com/translate/pootle/blob/master/pootle/apps/virtualfolder/models.py#L202-L203

pootle/apps/virtualfolder/receivers.py
phlax: surely its units list would change in that case - no?

pootle/apps/pootle_store/receivers.py
unho: If you do so, perhaps you can also move https://github.com/translate/pootle/blob/master/pootle/apps/virtualfolder/models.py#L459-L493 there to have all the receivers in that app in the same place.

pootle/apps/pootle_store/models.py
phlax: we wont need to change anything if we set it to 0.1 - validation will pass and the default hasnt changed afaict

unho: What if someone specifies a `0.09` priority for a virtual folder?

pootle/apps/virtualfolder/receivers.py
unho: For the "virtualfolder location changed" scenario this should be moved to `VirtualFolder.save()` since in https://github.com/translate/pootle/blob/master/pootle/apps/virtualfolder/models.py#L144-L171 all relationships with units are unlinked and then relinked. There is the only place where you can recalculate priority for previously linked units that are no longer linked.

pootle/apps/pootle_store/models.py
phlax: ok - im setting the min to 0

everyone is happy - just that there is not actually a way to set it to 0

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,60 @@
...
+def relate_unit(sender, instance, created=False, **kwargs):
+    """Add newly created units to the virtual folders they belong, if any.
+
+    When a new store or translation project, or even a full project is added,
+    some of their units might be matched by the filters of any of the
+    previously existing virtual folders, so this signal handler relates those
+    new units to the virtual folders they belong to, if any.
+    """
+    if not created:
+        instance.set_priority()
```
unho: Unit priority must be set this in this handler only when it is created. Already existing units already have the right priority because post_save handler for VirtualFolder already set it. But new units might be added to a store included in a virtual folder, and thus its priority must be calculated here.

phlax: no - i merged the 2 functions

unho: You are right, this is being calculated below for created units (after actually linking it with vfolders).

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,60 @@
...
+
+    for vf in VirtualFolder.objects.iterator():
+        for location in vf.all_locations:
+            if not pootle_path.startswith(location):
+                continue
+
+            for filename in vf.filter_rules.split(","):
+                if pootle_path == "".join([location, filename]):
+                    vf.units.add(instance)
+                    instance.set_priority()
```
unho: This might be triggered for each and any existing vfolder. Better put it after the loops to trigger it just once.

pootle/apps/pootle_store/models.py
```diff
@@ -356,6 +359,11 @@ class Unit(models.Model, base.TranslationUnit):
                                     db_index=True, related_name='reviewed')
     reviewed_on = models.DateTimeField(db_index=True, null=True)
 
+    # this is calculated from virtualfolders if installed and linked
+    priority = models.FloatField(
+        db_index=True, default=1,
+        validators=[MinValueValidator(0)])
```
unho: Nit: Use the same style for defining fields that is used in the rest of the models in this file.

pootle/apps/pootle_store/models.py
```diff
@@ -652,6 +660,24 @@ def getorig(self):
...
+    def calculate_priority(self):
+        if not vfolders_installed():
+            return 1.0
+
+        priority = (
+            self.vfolders.order_by("-priority")
+                         .values_list("priority", flat=True)
+                         .first())
+        if priority is None:
+            priority = 1.0
```
unho: Nit: Just return the value.

pootle/apps/pootle_store/migrations/0006_set_unit_priorities.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def forwards(apps, schema_editor):
+
+    from pootle_store.util import vfolders_installed
```
unho: Unnecessary import?

pootle/apps/pootle_store/models.py
phlax: it needs to return a default - ie 1.0 - and 0 is valid  - so you cant do

```
return self.vfolders(...etc...) or 1.0
```

unho: I was talking about:

```python
        if priority is None:
            return 1.0
        return priority
```

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+@receiver(post_save, sender=Unit)
+def vfolder_unit_postsave_handler(sender, instance, created=False, **kwargs):
+    """Add newly created units to the virtual folders they belong, if any.
+
+    When a new store or translation project, or even a full project is added,
+    some of their units might be matched by the filters of any of the
+    previously existing virtual folders, so this signal handler relates those
+    new units to the virtual folders they belong to, if any.
+    """
+
```
unho: Nit: Remove blank line after docstring. Don't we have now a check for this?

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+from pootle_store.models import Unit
+from pootle_store.util import OBSOLETE
+
+from .models import VirtualFolder, VirtualFolderTreeItem
+from .signals import vfolder_post_save
+
+
+def update_vfolder_tree(vf, store):
+    # Create missing VirtualFolderTreeItem tree structure after
+    # adding this new unit.
```
unho: Which new unit? You are passing here a store.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+            if unit_added:
+                break
+
+        if unit_added:
+            update_vfolder_tree(vf, unit.store)
+
+
+@receiver(pre_save, sender=VirtualFolder)
+def vfolder_unit_priority_presave_handler(sender, instance, **kwargs):
+
```
unho: Adding a detailed docstring won't harm.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+            unit.set_priority()
+            stores_affected.add(unit.store)
+
+    for store in stores_affected:
+        update_vfolder_tree(original, store)
+
+
+@receiver(vfolder_post_save, sender=VirtualFolder)
+def vfolder_unit_priority_handler(sender, instance, **kwargs):
+
```
unho: Adding a brief docstring won't harm.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+            update_vfolder_tree(vf, unit.store)
+
+
+@receiver(pre_save, sender=VirtualFolder)
+def vfolder_unit_priority_presave_handler(sender, instance, **kwargs):
+
+    if instance.id is None:
+        return
+
+    original = VirtualFolder.objects.get(pk=instance.pk)
```
unho: Since this is `pre_save` wouldn't `instance` lack the changes we want to avoid by doing this?

unho: Forget about this.

pootle/apps/virtualfolder/receivers.py
phlax: fair enough ill change

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+from django.dispatch import receiver
+
+from pootle_store.models import Unit
+from pootle_store.util import OBSOLETE
+
+from .models import VirtualFolder, VirtualFolderTreeItem
+from .signals import vfolder_post_save
+
+
+def update_vfolder_tree(vf, store):
```
unho: Perhaps you might want to use the parent directory instead of the store.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+@receiver(vfolder_post_save, sender=VirtualFolder)
+def vfolder_unit_priority_handler(sender, instance, **kwargs):
+
+    for unit in instance.units.all():
+        unit.set_priority()
+
+
+@receiver(pre_save, sender=Unit)
+def vfolder_unit_unobsoleted(sender, instance, created=False, **kwargs):
+    if instance.state == OBSOLETE:
```
unho: Adding a docstring would help.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        Unit.objects.get(pk=instance.pk, state=OBSOLETE)
+    except Unit.DoesNotExist:
+        return
+    add_unit_to_vfolders(instance)
+
+
+@receiver(pre_save, sender=Unit)
+def vfolder_unit_obsoleted(sender, instance, created=False, **kwargs):
+    if instance.state != OBSOLETE:
```
unho: Adding a docstring would help.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -0,0 +1,133 @@
...
+        return
+    try:
+        Unit.objects.get(pk=instance.pk, state__gt=OBSOLETE)
+    except Unit.DoesNotExist:
+        return
+    for vfolder in instance.vfolders.all():
+        update_vfolder_tree(vfolder, instance.store)
```
unho: Since we are removing this unit from the virtual folders it is in, I think it is safer to use the following instead:

```python
    for vfti in instance.store.parent_vf_treeitems.iterator():
        vfti.update_all_cache()
```

This way we avoid calling `get_or_create()` in `update_vfolder_tree()`.

pootle/apps/virtualfolder/receivers.py
ta2-1: Can we run signal for obsolete unit? Or is it just security check?

phlax: not sure i follow

tests/models/virtualfolder.py
```diff
@@ -152,3 +152,118 @@ def test_vfolder_with_no_filter_rules():
...
+    # obsolete unit is not in the vfolder
+    assert obsolete_unit not in vfolder.units.all()
+
+    obsolete_unit.state = TRANSLATED
+    obsolete_unit.save()
+
+    # unobsoleted unit is in the vfolder
+    assert obsolete_unit in vfolder.units.all()
+
+    to_obsolete = vfolder.units.all()[0]
```
unho: Nit: Use `first()` instead of `all()[0]`. Same for other places where `[0]` is being used.

pootle/apps/virtualfolder/receivers.py
ta2-1: This is pre_save for unobsolete. So why don't we do anything if unit is obsolete?

tests/models/virtualfolder.py
```diff
@@ -152,3 +152,118 @@ def test_vfolder_with_no_filter_rules():
...
+
+    # obsoleted unit is not in the vfolder
+    assert to_obsolete not in vfolder.units.all()
+
+
+@pytest.mark.django_db
+def test_vfolder_unit_priorities(site_matrix):
+    from pootle_pytest.factories import VirtualFolderFactory
+
+    from pootle_store.models import Unit
```
unho: Any reason why this imports are not at module level? I have the feeling I already asked this in the past.

phlax: yep because if you import from (pootle) models to early it causes problems.

in this case it may well be ok - ill test...

tests/models/virtualfolder.py
phlax: no that is not a good idea - as currently there is an issue with mtime which is how first is defined.

this test really does not the obsolete unit was retrieved

tests/models/virtualfolder.py
```diff
@@ -152,3 +152,118 @@ def test_vfolder_with_no_filter_rules():
...
+
+    assert all(
+        priority == 1
+        for priority
+        in Unit.objects.values_list("priority", flat=True))
```
unho: Is it safe to assume that there is no virtual folder with priority different than 1?

tests/models/virtualfolder.py
```diff
@@ -152,3 +152,118 @@ def test_vfolder_with_no_filter_rules():
...
+
+    assert all(
+        priority == 3
+        for priority
+        in vfolder0.units.values_list("priority", flat=True))
+
+    assert all(
+        priority == 1.0
+        for priority
+        in Unit.objects.filter(vfolders__isnull=True)
```
unho: Again assuming that there is no other virtual folder, thus you can be skipping some other stores besides `store0.po`. Perhaps it is better to check for all units that don't belong to that store rather than units that are not on any virtual folder.

pootle/apps/virtualfolder/receivers.py
unho: > not sure i follow

@phlax We have the option to read the code, or read a more friendly docstring. I prefer the docstring. See `vfolder_unit_postsave_handler` below for a helpful docstring.


> This is pre_save for unobsolete. So why don't we do anything if unit is obsolete?

@ta2-1 Is that no handled by `vfolder_unit_obsoleted` receiver below?

unho:
Finished review. Raised several nits and doubts.

tests/models/virtualfolder.py
phlax: @unho - this is a test - its not an assumption

pootle/apps/virtualfolder/receivers.py
phlax: ill add a comment but here it is psuedo code

```
function to_find_units_that_have_been_obsoleted:
     if unit_is_obsolete_now and unit_was_not_before:
            the_unit_was_obsoleted
```

the reverse is true for the other function

tests/models/virtualfolder.py
phlax: yes

phlax: in fact we are not assuming this we are **testing** this

phlax: or rather testing what units return when  we know that there are no vfolders on the site.

tests/models/virtualfolder.py
phlax: also this is a test for *priority*, there is a different test for *membership*

tests/models/virtualfolder.py
unho: I just was a bit confused because of the possibility that virtual folders created on other test functions could be interfering here.

pootle/apps/virtualfolder/receivers.py
phlax: >>    not sure i follow

> @phlax We have the option to read the code...

i was replying to @ta2-1's comment - apologies for confusion

tests/models/virtualfolder.py
phlax: thats not how unit tests work

ta2-1:
GTM

PR #4265, opened 2015-12-02
Remove unit ordering

phlax:
On sites with a large number of Units - doing any db operation with
Unit.objects is pretty much unusable due to a v expensive sort

In general sorting on anything non-unique is expensive, even more so if there are multiple sort params - so best not
to make it the default.

This commit removes that ordering

### DIFF
--- a/pootle_pytest/fixtures/models/store.py
+++ b/pootle_pytest/fixtures/models/store.py
@@ -132,7 +132,8 @@ def _setup_store_test(store, member, member2, test):
 
     store_revision, units_update = test["update_store"]
     revision_min = store.get_max_unit_revision()
-    units_before = [unit for unit in store.unit_set.all()]
+    units_before = [
+        unit for unit in store.unit_set.all().order_by("index")]
 
     fs_wins = test.get("fs_wins", True)
     if fs_wins:
--- a/pootle/apps/pootle_store/diff.py
+++ b/pootle/apps/pootle_store/diff.py
@@ -84,7 +84,7 @@ class StoreDiff(object):
         unit_fields = ("unitid", "state", "id", "index", "revision",
                        "source_f", "target_f", "developer_comment",
                        "translator_comment", "locations", "context")
-        for unit in self.db_store.unit_set.values(*unit_fields):
+        for unit in self.db_store.unit_set.values(*unit_fields).order_by("index"):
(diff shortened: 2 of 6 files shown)


pootle_pytest/fixtures/models/store.py
```diff
@@ -132,7 +132,8 @@ def _setup_store_test(store, member, member2, test):
 
     store_revision, units_update = test["update_store"]
     revision_min = store.get_max_unit_revision()
-    units_before = [unit for unit in store.unit_set.all()]
+    units_before = [
+        unit for unit in store.unit_set.all().order_by("index")]
```
unho: I think this fits in a single line.

julen:
Quick test with migrations applied, in any `/translate/` page without any sorting set I get:
```python
Internal Server Error: /xhr/units/
  File "django/core/handlers/base.py", line 111, in get_response
  File "pootle/apps/pootle_misc/util.py", line 59, in wrapper
    return f(request, *args, **kwargs)
  File "pootle/apps/pootle_store/views.py", line 528, in get_units
    'store__pootle_path')]
  File "django/db/models/query.py", line 626, in values
    return self._clone(klass=ValuesQuerySet, setup=True, _fields=fields)
  File "django/db/models/query.py", line 961, in _clone
    c._setup_query()
  File "django/db/models/query.py", line 1129, in _setup_query
    self.query.add_fields(self.field_names, True)
  File "django/db/models/sql/query.py", line 1670, in add_fields
    "Choices are: %s" % (name, ", ".join(names)))
FieldError: Cannot resolve keyword 'sort_by_field' into field. Choices are: commented_by, commented_by_id, commented_on, context, creation_time, developer_comment, id, index, locations, mtime, qualitycheck, reviewed_by, reviewed_by_id, reviewed_on, revision, source_f, source_hash, source_length, source_wordcount, state, store, store_id, submission, submitted_by, submitted_by_id, submitted_on, suggestion, target_f, target_length, target_wordcount, translator_comment, unitid, unitid_hash, vfolders
```

Changing sort order to any non-default value makes it usable again.

phlax:
thanks @julen ill track this down now

phlax:
cool @julen @ta2-1 this issue is fixed by the change that @ta2-1 and i had been discussing.

PR #4266, opened 2015-12-03
Add site_matrix test fixture with lots of units

phlax:
This commit adds a fixture for factory generating:
- 2 langs
- 2 projects
- a tp for each proj/lang
- 3 stores for each tp
- 8 units for each store (2 each of translated, untranslated, fuzzy, obsolete)

The units mtimes are spread out to allow sort/filter testing of unit lists based on mtime

### DIFF
--- a/pootle_pytest/factories.py
+++ b/pootle_pytest/factories.py
@@ -7,9 +7,13 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from hashlib import md5
+
 import factory
 
 from django.utils import timezone
+
+import pootle_store
 
 
 class SubmissionFactory(factory.django.DjangoModelFactory):
@@ -41,3 +45,107 @@ class AgreementFactory(factory.django.DjangoModelFactory):
 
     class Meta(object):
         model = 'staticpages.Agreement'
+
+
+class DirectoryFactory(factory.django.DjangoModelFactory):
+
+    @factory.lazy_attribute
+    def pootle_path(self):
+        if self.parent is None:
+            return "/"
+        return (
+            "%s/%s"
+            % (self.parent.pootle_path.rstrip("/"),
+               self.name))
+
+    class Meta(object):
+        model = 'pootle_app.Directory'
+        django_get_or_create = ("pootle_path", )
+
+
+class LanguageFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_language.Language'
+        django_get_or_create = ("code", )
+
+    code = factory.Sequence(lambda n: 'language%s' % n)
+    fullname = factory.Sequence(lambda n: 'Language %s' % n)
+
+
+class ProjectFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_project.Project'
+        django_get_or_create = ("code", )
+
+    code = factory.Sequence(lambda n: 'project%s' % n)
+    fullname = factory.Sequence(lambda n: 'Project %s' % n)
+    pootle_path = factory.LazyAttribute(lambda p: "/projects/%s" % p.code)
+
+
+class StoreFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_store.Store'
+        django_get_or_create = ("pootle_path", )
+
+    @factory.lazy_attribute
+    def pootle_path(self):
+        return (
+            "%s/%s"
+            % (self.translation_project.pootle_path.rstrip("/"),
+               self.name))
+
+    name = factory.Sequence(lambda n: 'store%s.po' % n)
+    parent = factory.LazyAttribute(
+        lambda s: s.translation_project.directory)
+
+
+class TranslationProjectFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_translationproject.TranslationProject'
+
+
+class UnitFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_store.Unit'
+
+    state = pootle_store.util.UNTRANSLATED
+
+    @factory.lazy_attribute
+    def index(self):
+        # returns an incrementing index relative to the store
+        return self.store.unit_set.count()
+
+    @factory.lazy_attribute
+    def unitid(self):
+        return (
+            "%s_unit%s"
+            % (pootle_store.util.get_state_name(self),
+               self.index))
+
+    @factory.lazy_attribute
+    def unitid_hash(self):
+        return md5(self.unitid.encode("utf-8")).hexdigest()
+
+    @factory.lazy_attribute
+    def source_f(self):
+        return (
+            "%s Source %s %s"
+            % (pootle_store.util.get_state_name(self).capitalize(),
+               self.store.pootle_path,
+               self.index))
+
+    @factory.lazy_attribute
+    def target_f(self):
+        state_name = pootle_store.util.get_state_name(self)
+        if state_name in ["translated", "fuzzy", "obsolete"]:
+            return (
+                "%s Target %s %s"
+                % (state_name.capitalize(),
+                   self.store.pootle_path,
+                   self.index))
+        return ""
--- a/pootle_pytest/fixtures/site.py
+++ b/pootle_pytest/fixtures/site.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+import shutil
+
+import pytest
+
+
+@pytest.fixture
+def site_matrix(request, system, settings):
+
+    from pootle_pytest.factories import (
+        ProjectFactory, DirectoryFactory, LanguageFactory,
+        TranslationProjectFactory, StoreFactory, UnitFactory)
+
+    from pootle_store.models import UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE
+
+    # create root and projects directories
+    DirectoryFactory(name="projects", parent=DirectoryFactory(parent=None))
+
+    # add 2 languages
+    languages = [LanguageFactory() for i in range(0, 2)]
+
+    for i in range(0, 2):
+        # add 2 projects
+        project = ProjectFactory(source_language=languages[0])
+
+        for language in languages:
+            # add a TP to the project for each language
+            tp = TranslationProjectFactory(project=project, language=language)
+
+            for i in range(0, 3):
+                # add 3 stores
+                store = StoreFactory(translation_project=tp)
+
+                # add 8 units to each store
+                for state in [UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE]:
+                    for i in range(0, 2):
+                        UnitFactory(store=store, state=state)
+
+    def _teardown():
+        # required to get clean slate 8/
+        for trans_dir in os.listdir(settings.POOTLE_TRANSLATION_DIRECTORY):
+            if trans_dir.startswith("project"):
+                shutil.rmtree(
+                    os.path.join(
+                        settings.POOTLE_TRANSLATION_DIRECTORY, trans_dir))
+
+    request.addfinalizer(_teardown)
--- a/pootle_pytest/plugin.py
+++ b/pootle_pytest/plugin.py
@@ -37,6 +37,7 @@ from fixtures.import_export_fixtures import (
     FILE_IMPORT_FAIL_TESTS,
     file_import_failure, ts_directory, en_tutorial_ts)
 from fixtures.revision import revision
+from fixtures.site import site_matrix
 from fixtures.views import admin_client
 
 
@@ -70,4 +71,4 @@ __all__ = (
     'english_tutorial', 'french_tutorial', 'italian_tutorial',
     'russian_tutorial', 'spanish_tutorial', 'templates_tutorial',
     'delete_pattern', 'en_tutorial_ts', 'file_import_failure', 'ts_directory',
-    'revision', 'admin_client')
+    'revision', 'admin_client', 'site_matrix')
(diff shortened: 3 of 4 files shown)


pootle_pytest/plugin.py
```diff
@@ -51,23 +52,23 @@ def pytest_generate_tests(metafunc):
...
+    "es_tutorial_subdir_remove_po", "fr_tutorial_remove_sync_po",
+    "fr_tutorial_subdir_to_remove_po", "issue_2401_po", "it_tutorial_po",
+    "param_update_store_test", "po_directory", "ru_tutorial_po",
+    "ru_update_save_changed_units_po", "ru_update_set_last_sync_revision_po",
+    "store_diff_tests", "templates_tutorial_pot", "test_get_units_po",
+    "afrikaans_tutorial", "afrikaans_vfolder_test", "arabic_tutorial_obsolete",
+    "english_tutorial", "french_tutorial", "italian_tutorial",
+    "russian_tutorial", "spanish_tutorial", "templates_tutorial",
+    "delete_pattern", "en_tutorial_ts", "file_import_failure", "ts_directory",
+    "revision", "admin_client", "site_matrix")
```
ta2-1: @phlax, is it possible to have atomic changes when one test added only?

ta2-1:
@phlax, can you describe what site_matrix test tests?

pootle_pytest/plugin.py
phlax: well there arent any tests actually added - but i can separate out adding the factories from adding the fixture if you think that is helpful

julen:
Yes I'm also missing some background and description here.

phlax:
@ta2-1 its a fixture not a test - and in this commit is not used - its used by the code in otherwise unrelated PRs that i have in the pipeline

pootle_pytest/factories.py
```diff
@@ -41,3 +45,112 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+    @factory.lazy_attribute
+    def pootle_path(self):
+        if self.parent is None:
+            return "/"
+        return (
+            "%s/%s"
+            % (self.parent.pootle_path.rstrip("/"),
+               self.name))
+
+    class Meta:
```
julen: As I recall from some PR this will fail validation, as you need to inherit from object.

phlax: @julen can you provide a link - i have never seen Meta inheriting from object - but may well be wrong

pootle_pytest/factories.py
```diff
@@ -41,3 +45,112 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+
+
+def get_state_name(unit):
+    if unit.state == pootle_store.util.TRANSLATED:
+        return "translated"
+    elif unit.state == pootle_store.util.FUZZY:
+        return "fuzzy"
+    elif unit.state == pootle_store.util.OBSOLETE:
+        return "obsolete"
+    else:
```
julen: `else` is redundant here (let's avoid unnecessary nesting). Arguably the same applies to `elif`. But since this is just a mapping, you can also opt to use a simple dict instead:
```python
return {
  pootle_store.util.TRANSLATED: 'translated',
  pootle_store.util.FUZZY: 'fuzzy',
  pootle_store.util.OBSOLETE: 'obsolete',
}.get(unit.state, 'untranslated')
```

pootle_pytest/plugin.py
```diff
@@ -51,23 +52,23 @@ def pytest_generate_tests(metafunc):
...
-    'fr_tutorial_subdir_to_remove_po', 'issue_2401_po', 'it_tutorial_po',
-    'param_update_store_test', 'po_directory', 'ru_tutorial_po',
-    'ru_update_save_changed_units_po', 'ru_update_set_last_sync_revision_po',
-    'store_diff_tests', 'templates_tutorial_pot', 'test_get_units_po',
-    'afrikaans_tutorial', 'afrikaans_vfolder_test', 'arabic_tutorial_obsolete',
-    'english_tutorial', 'french_tutorial', 'italian_tutorial',
-    'russian_tutorial', 'spanish_tutorial', 'templates_tutorial',
-    'delete_pattern', 'en_tutorial_ts', 'file_import_failure', 'ts_directory',
-    'revision', 'admin_client')
+    "admin", "default", "evil_member", "member", "member2",
```
julen: There's no reason to change quotes here.

phlax: good point - ill change

pootle_pytest/factories.py
julen: No, me neither, only in this PR of @dwaynebailey. Unfortunately I cannot find it as there's so many noise around such changes that makes it impossible to locate it.

phlax: @dwaynebailey know anything about Meta(object?) and validation?

julen:
Is the plan to start replacing the use of in-DB fixtures with such factories that can be created in-memory? I'd love to hear so :grin:

phlax:
> Is the plan to start replacing the use of in-DB fixtures with such factories that can be created in-memory? I'd love to hear so 

definitely possible - but not for the tests im currently writing which tests code that reads from the db

julen:
...but we test on MySQL and PostgreSQL. Avoiding hitting the DB is best practice if it can be done so — tests are usually ensured to run blazingly fast.

phlax:
its actually something of a bonus that we test on all 3 dbs because all have their quirks that mean sometimes tests will only fail on one of the dbs - but in reality the test should have failed on all.

a good example of this is row ordering - sometimes you will get back the results you expected because the rows were added that way - not because they were properly sorted - usually this gets picked up by postgres. Mysql and sqlite have similar quirks.

but yep - best practice is defo to avoid the db if at all possible 8)

phlax:
> Is the plan to start replacing the use of in-DB fixtures

also worth mentioning is that these factories could be used with factory.build() to create in-mem objects - just not thats not how im using them

pootle_pytest/factories.py
dwaynebailey: This commit 5d0f7b77da62342ef2c2949da8de6a7090c12fd0 is about it, picked up here https://www.quantifiedcode.com/app/issue_class/86d4d89288cc4da9a6c1ad122f00771c

pootle_pytest/factories.py
```diff
@@ -41,3 +45,110 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+
+class DirectoryFactory(factory.django.DjangoModelFactory):
+
+    @factory.lazy_attribute
+    def pootle_path(self):
+        if self.parent is None:
+            return "/"
+        return (
+            "%s/%s"
+            % (self.parent.pootle_path.rstrip("/"),
```
unho: It is not necessary to use `.rstrip("/")` if the string is `"%s%s"`.

unho: Same a few lines below for `StoreFactory`.

pootle_pytest/factories.py
```diff
@@ -41,3 +45,110 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+
+    @factory.lazy_attribute
+    def pootle_path(self):
+        return ("%s/%s"
+                % (self.translation_project.pootle_path.rstrip("/"),
+                   self.name))
+
+    name = factory.Sequence(lambda n: 'store%s.po' % n)
+    parent = factory.LazyAttribute(
+        lambda s: s.translation_project.directory)
```
unho: Wouldn't this prevent a TP from having a tree directory structure?

pootle_pytest/factories.py
phlax: hmm - it might not be - but there was a situation that came up where it was - so i would rather err on the side of caution

pootle_pytest/factories.py
```diff
@@ -41,3 +45,110 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+    class Meta:
+        model = 'pootle_store.Unit'
+
+    unitid = factory.Sequence(lambda n: 'unit%s' % n)
+    index = factory.Sequence(lambda n: n)
+    state = pootle_store.util.UNTRANSLATED
+
+    @factory.lazy_attribute
+    def index(self):
+        return self.store.unit_set.count()
```
unho: Is it ok to set the same index for all units in the same store?

pootle_pytest/factories.py
phlax: not sure i follow - but i think the TP directory structure is set up automagically

pootle_pytest/factories.py
phlax: its not the same index - it increments as the units are added

unho: Thanks for the explanation. That was something I couldn't figure out based on the code. Can you please add a comment explaining this?

pootle_pytest/factories.py
```diff
@@ -41,3 +45,110 @@ class AgreementFactory(factory.django.DjangoModelFactory):
...
+    return "untranslated"
+
+
+class UnitFactory(factory.django.DjangoModelFactory):
+
+    class Meta:
+        model = 'pootle_store.Unit'
+
+    unitid = factory.Sequence(lambda n: 'unit%s' % n)
+    index = factory.Sequence(lambda n: n)
```
phlax: i need to remove this tho @unho

pootle_pytest/factories.py
unho: Yes, it should work. But this feels overcomplicated.

pootle_pytest/factories.py
unho: What I was trying to say is that this way a Store parent is always the TP directory, independently of how deeply nested it is on the directory tree within the TP. This actually prevents going up through the trail for a store (unless it is directly below the TP). Or perhaps I am missing something?

pootle_pytest/factories.py
julen: In any case the `index` is set twice: as a class attribute and a method which creates the index. Not sure if both are needed.

unho: Doesn't happen the same for `unitid`?

pootle_pytest/fixtures/site.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.fixture
+def site_matrix(request, system, settings):
+
+    from pootle_pytest.factories import (
+        ProjectFactory, DirectoryFactory, LanguageFactory,
+        TranslationProjectFactory, StoreFactory, UnitFactory)
+
+    from pootle_store.models import UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE
```
unho: Is there any reason not to import these at module level?

pootle_pytest/factories.py
phlax: @julen youre right - just spotted this also - https://github.com/translate/pootle/pull/4266#discussion_r46545399

pootle_pytest/fixtures/site.py
```diff
@@ -0,0 +1,55 @@
...
+            tp = TranslationProjectFactory(project=project, language=language)
+
+            for i in range(0, 3):
+                # add 3 stores
+                store = StoreFactory(translation_project=tp)
+
+                # add 8 units to each store
+                for state in [UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE]:
+                    [UnitFactory(store=store, state=state)
+                     for i in range(0, 2)]
```
unho: Is it not clearer by not using list comprehension?

pootle_pytest/fixtures/site.py
phlax: you cant import (pootle) models at module level in fixtures

pootle_pytest/fixtures/site.py
phlax: im not sure - i kinda felt that having the factory up front was more explicit in terms of its vars

pootle_pytest/fixtures/site.py
unho: Are the factories affected as well? I would appreciate including a comment stating that Pootle models can't be imported at module level.

phlax: @unho im not commenting this or else i would need to comment it in a 100 places

if you write tests - you know

pootle_pytest/fixtures/site.py
unho: I would be ok with it if it was meant to store the result in a variable, like `languages` a few lines above. But it is not, and using a regular for looks more readable:

```python
                for state in [UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE]:
                    for i in range(0, 2):
                        UnitFactory(store=store, state=state)
```

vs

```python
                for state in [UNTRANSLATED, TRANSLATED, FUZZY, OBSOLETE]:
                    [UnitFactory(store=store, state=state)
                     for i in range(0, 2)]
```

pootle_pytest/factories.py
phlax: i think what you are saying is that the matrix doesnt currently implement subfolders - and that would be entirely correct

unho: That is not 100% true as in https://github.com/translate/pootle/pull/4266/files#diff-2bbf96c64475cba15bc4e18a425f8818R25 you are initializing a directory and its subfolder.

The problem is that any store's parent is always its TP's directory. Therefore you can have subfolders on a TP, and those subfolders can have stores (and other subfolders with stores, and so on), but none of this subfolders's stores will point to its actual parent (the subfolder they are in) but to its TP's directory.

phlax: @unho this code works - i know this from *a lot* of using it and the dir structure is correct

please feel free to test it out

phlax: @unho 

```
(Pdb) Unit.objects.all()[0].store.parent.pootle_path
u'/af/terminology/'
```
this is taken from livestaging - so the factory code is correct in how it constructs stores

unho: You should be checking a unit in a store that is in a subfolder, not some terminology unit since Terminology project is flat.

unho:
lgtm provided that:

- Name clash for `index` and `unitid` in `UnitFactory` is addressed,
- Subfolder issue is fixed in the future (see discussion on this https://github.com/translate/pootle/pull/4266#discussion_r46544756)

pootle_pytest/fixtures/site.py
julen: @unho has a valid point: the end goal of list comprehensions is creating lists by definition. Here there's no use in creating a list.

PR #4271, opened 2015-12-04
[WIP] Use tox for cleaner Travis builds

dwaynebailey:
This implements tox to build Pootle.

This is mostly done and builds docs, mo, pep8 and pep257 apart from the real tests.  Various databases work as expected.  Coveralls works.  Caching of pip works.

A few things still need to be ported:

- Some minor checks: check page building, unmigrated models
- Forcing Django version to latest bugfix in a main release
- Need to check if caching makes sense in the tox venvs


.travis.yml
```diff
@@ -20,32 +20,9 @@ before_install:
...
   - pip install --upgrade pip # FIXME remove once Travis support pip7 by default
-  - CFLAGS="-O0" pip install -r requirements/travis.txt
-  - pip install "Django>=$DJANGO_VERSION,<"$(echo $(echo $DJANGO_VERSION | cut -d"." -f1).$(( $(echo $DJANGO_VERSION | cut -d"." -f2) + 1 )) )
-  - pip install .
-  - pip freeze  # Print all installed versions for reference.
-before_script:
-  - if [[ $DATABASE_BACKEND == 'postgres' ]]; then psql -c 'create database pootle;' -U postgres; fi
-  - if [[ $DATABASE_BACKEND == 'mysql_innodb' ]]; then mysql -e 'create database pootle CHARACTER SET utf8 COLLATE utf8_general_ci;'; fi
-  - if [[ $DATABASE_BACKEND == 'mysql_innodb' ]]; then mysql -e "SET GLOBAL wait_timeout = 36000;"; fi
+  - pip install tox
```
unho: Is it not better to put this on `requirements/travis.txt`?

dwaynebailey: Likely not.  Not sure if we even need travis.txt.  You want venv installation to happen in tox not in .travis.yml, so you only want that one install there.

phlax: could we keep a tests.text or similar with the other deps - for setting up test env not on travis?

dwaynebailey: There is already one.  Likely travis.txt can go away as its done differently in tox

julen:
s/parrallel/parallel/ in the commit message.

And I don't see why Django 1.8+1.9 need to be tested here.

dwaynebailey:
@julen Django 1.8/1.9 - just me experimenting to see if we can override Django.  We can't because pip doesn't do resolutions very nicely.  This will go away.

dwaynebailey:
Landed in 3a39f1ebfff726fddd28225fafd0a7d79b406dfd

PR #4272, opened 2015-12-04
Remove unit select relateds from Unit qss

phlax:
requires #4265 

In general i think its a bad idea to have select_relateds on querysets.

It forces all code to pull out additional table joins irrespective of whether they are used.

Also, the units share parent Store/TP/Projects so i think it would be better to cache info about the parents than pulling the same info out for every unit.

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -234,14 +234,6 @@ def stringcount(string):
 
 class UnitManager(models.Manager):
 
-    def get_queryset(self):
-        """Mimics `select_related(depth=1)` behavior. Pending review."""
-        return (
-            super(UnitManager, self).get_queryset().select_related(
-                'store', 'submitted_by', 'commented_by', 'reviewed_by',
-            )
-        )
-
     def live(self):
         """Filters non-obsolete units."""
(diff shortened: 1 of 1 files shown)


unho:
As far as I can tell looks fine. I checked in the code for `Store.units` to see if it is safe to drop the `select_related` and I think it is. I am not so sure about dropping it on the model manager. Would like to hear from others.

@phlax Did you run some profiler tool to see if it impacts performance and how much?

julen:
Among other views I specifically checked the edit widget view (`get_edit_unit`) where most of this was affecting, and all looks good to me, and I'm happy to see it merged. The initial query count also goes down as expected (24 down from 30).

@unho it's a good idea to remove this from the default model manager; it was kept there only to avoid potential breakage when migrating away from `select_related(depth=<N>)`. The `depth` parameter was removed in Django 1.7 for good reasons, but it is pending review in many managers because this has been used since the very early DB-based Pootle days.

unho:
I am not against removing stuff that prevents the code to be more performant, just a bit worried about side effects we might not want to suffer.

ta2-1:
I'm ok to merge this now and do some optimisations later if it's necessary.

Issue #4152, opened 2015-10-20
Better language code handling


phlax:
@dwaynebailey and i have been discussing ways to improve language code handling

The main issues we are seeking to address are:
- Encourage users to use BCP-compliant language codes
- Make it easier to add new TranslationProjects without having to add a language first

The approaches that we have discussed are:
- Preloading ISO639 lang codes as Pootle Languages (potentially in initdb)
- Preloading ISO639 lang codes in a KnownLanguages table and making it easier to initiate a lang if it is known
- A BCP validator - with warning to user if creating non-compliant code
- Some combination of above

leonardcj:
Please avoid using the ISO639-1 macrolanguage codes where possible.  I'm trying to convince CLDR and glibc to avoid them totally.  

https://en.wikipedia.org/wiki/ISO_639_macrolanguage

PR #4280, opened 2015-12-07
Improved CLDR support

imankulov:
Hey there! 

Just wanted to raise one question. Not sure how to discuss this, maybe we can do that in comments to this pull request.

By default when you initialize Pootle with `initdb`, it generates a bunch of languages with default settings for plural forms. The data for this is taken from the translate toolkit database. The thing is, I feel like the [CLDR database](http://cldr.unicode.org/) becomes the standard de-facto for such kind of information.

It looks like by this point we accumulate a number inconsistencies between CLDR and "classic" translate toolkit. I happen to spot some of them, like:
- Number of languages in Turkish and Hungarian: 2 (TTK) vs 1 (CLDR).
- Number of languages in Russian and Polish: 3 (TTK) vs 4 (CLDR).

It would be great to have at least an option to use CLDR database as a valid source for plural forms, and this PR while doesn't provide a complete solution, tries to solve our own problem in an opportunistic manner. Namely it:
- Adds support for optional import languages and plural definitions from CLDR database (uses babel)
- Allows Pootle admins to convert the language database from TTK to CLDR (well, there's just a function, which isn't exposed in CLI in any way)
- Fixes the `nplurals` check issue: if number of plurals in a database doesn't match the number of plurals, known to TTK, `nplurals` check always fails.
- Extends the `pluralequation` database field (255 symbols isn't enough for some babel-generated plural expressions)

I hope it makes sense and you guys can consider merging this PR in. Thank you!

### DIFF
--- a/pootle/core/initdb.py
+++ b/pootle/core/initdb.py
@@ -36,7 +36,7 @@ CLDR_FALLBACK_ALIASES = {
 
 class InitDB(object):
 
-    def init_db(self, create_projects=True, cldr=False, aliases=CLDR_FALLBACK_ALIASES):
+    def init_db(self, create_projects=True, cldr=False):
         """Populate the database with default initial data.
 
         This creates the default database to get a working Pootle installation.
@@ -319,25 +319,6 @@ class InitDB(object):
             code = attrs.pop('code')
             self._create_object(Language, code=code, defaults=attrs)
 
-
-    def update_languages_to_cldr(self):
-        """
-        Take all existing languages, and update definitions according to the latest CLDR definition rules database
-
-        It's not used anywhere in the code, but if you need, you can just call it manually.
-        """
-        from pootle_language.models import Language
-        cldr_languages = {l['code']: l for l in self.get_cldr_languages()}
-        for lang_obj in Language.objects.all():
-            try:
-                cldr_lang = cldr_languages[lang_obj.code]
-            except KeyError:
-                continue
-            for k, v in cldr_lang.items():
-                setattr(lang_obj, k, v)
-                lang_obj.save()
-
-
     def get_cldr_languages(self):
         """
         Helper function to extract CLDR information. Heavily relied on babel functionality
@@ -351,7 +332,8 @@ class InitDB(object):
 
             plurals_str = plural.to_gettext(locale.plural_form)
             nplurals, pluralequation = re_plurals.match(plurals_str).groups()
-            lang_aliases = {lang}.union(set(self.aliases.get(lang, [])))
+            lang_aliases = set(CLDR_FALLBACK_ALIASES.get(lang, []))
+            lang_aliases.add(lang)
             for alias in lang_aliases:
                 yield {
                     'code': alias,
--- a/pootle/core/initdb.py
+++ b/pootle/core/initdb.py
@@ -7,6 +7,7 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import re
 import logging
 
 from django.contrib.auth import get_user_model
@@ -23,11 +24,19 @@ from staticpages.models import StaticPage as Announcement
 
 
 logger = logging.getLogger(__name__)
+re_plurals = re.compile(r'nplurals=(\d); plural=(.+)$')
+
+
+CLDR_FALLBACK_ALIASES = {
+    # new name: list of fallback names
+    'zh_Hans': ['zh_CN'],
+    'zh_Hant': ['zh_TW', 'zh_HK'],
+}
 
 
 class InitDB(object):
 
-    def init_db(self, create_projects=True):
+    def init_db(self, create_projects=True, cldr=False, aliases=CLDR_FALLBACK_ALIASES):
         """Populate the database with default initial data.
 
         This creates the default database to get a working Pootle installation.
@@ -42,7 +51,10 @@ class InitDB(object):
         self.create_pootle_permission_sets()
         if create_projects:
             self.create_default_projects()
-        self.create_default_languages()
+        if cldr:
+            self.create_default_languages_cldr()
+        else:
+            self.create_default_languages()
 
     def _create_object(self, model_klass, **criteria):
         instance, created = model_klass.objects.get_or_create(**criteria)
@@ -201,12 +213,11 @@ class InitDB(object):
     def require_english(self):
         """Create the English Language item."""
         criteria = {
-            'code': "en",
             'fullname': u"English",
             'nplurals': 2,
             'pluralequation': "(n != 1)",
         }
-        en, created = self._create_object(Language, **criteria)
+        en, created = self._create_object(Language, code='en', defaults=criteria)
         return en
 
     def create_root_directories(self):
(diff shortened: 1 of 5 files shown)


phlax:
hi @imankulov - thanks for the PR

this is definitely something we have been thinking about - see here for discussion - #4152 

...and is something @dwaynebailey probably wants to have some input on

unho:
There is also some work and requests on the TTK side:

- https://github.com/translate/translate/issues/3271
- https://github.com/translate/translate/pull/126
- https://github.com/translate/translate/issues/2900
- https://github.com/translate/translate/issues/2453

pootle/core/initdb.py
```diff
@@ -295,3 +306,56 @@ def create_default_languages(self):
...
+        You can optionally set up a list of 'fallback aliases' (mappings from language definitions existing in CLDR
+        to locale names which you need in your Pootle installation)
+        """
+        from pootle_language.models import Language
+        for attrs in self.get_cldr_languages():
+            code = attrs.pop('code')
+            self._create_object(Language, code=code, defaults=attrs)
+
+
+    def update_languages_to_cldr(self):
```
unho: Defined but unused.

pootle/core/initdb.py
```diff
@@ -201,12 +213,11 @@ def create_pootle_permission_sets(self):
     def require_english(self):
         """Create the English Language item."""
         criteria = {
-            'code': "en",
             'fullname': u"English",
             'nplurals': 2,
             'pluralequation': "(n != 1)",
         }
-        en, created = self._create_object(Language, **criteria)
+        en, created = self._create_object(Language, code='en', defaults=criteria)
```
unho: The changes on `require_english` seem to be unnecessary.

pootle/core/initdb.py
```diff
@@ -295,3 +306,56 @@ def create_default_languages(self):
...
+        from babel import Locale, localedata, plural
+
+        for lang in localedata.locale_identifiers():
+            locale = Locale(lang)
+            if not locale.english_name:
+                continue
+
+            plurals_str = plural.to_gettext(locale.plural_form)
+            nplurals, pluralequation = re_plurals.match(plurals_str).groups()
+            lang_aliases = {lang}.union(set(self.aliases.get(lang, [])))
```
unho: Where is `self.aliases` being set?

pootle/core/initdb.py
imankulov: @unho Correct. Not sure it we should keep it or remove, actually. I myself used this function to convert my current database to CLDR-compliant (called manually from `pootle shell`), and probably someone else can find it useful as well... 

I can expose it as an option to `initdb` management command, or to ensure `initdb --cldr` always calls this function too.

pootle/core/initdb.py
imankulov: oops, nowhere :) Will provide a fix shortly

pootle/core/initdb.py
unho: No, `initdb` is meant to be used once to initialize the database from scratch, not to tweak it afterwards. If you think that this code might be used somewhere else in the future keep it on a branch on your fork, but we are not going to land unused code.

Re adding this feature of converting the languages to CLDR, we would like to wait for input from @dwaynebailey.

pootle/core/initdb.py
```diff
@@ -221,7 +232,7 @@ def create_template_languages(self):
         untranslated template files.
         """
         self._create_object(
-            Language, **dict(code="templates", fullname="Templates"))
+            Language, code='templates', defaults=dict(fullname="Templates"))
```
unho: This change seems unnecessary as well.

pootle/core/initdb.py
imankulov: `--cldr` option is also never go through the codebase, fixing this as well

pootle/core/initdb.py
```diff
@@ -23,11 +24,19 @@
...
+    # new name: list of fallback names
+    'zh_Hans': ['zh_CN'],
+    'zh_Hant': ['zh_TW', 'zh_HK'],
+}
 
 
 class InitDB(object):
 
-    def init_db(self, create_projects=True):
+    def init_db(self, create_projects=True, cldr=False, aliases=CLDR_FALLBACK_ALIASES):
```
unho: Probably it is not necessary to pass `CLDR_FALLBACK_ALIASES` as it is available in the module.

pootle/core/initdb.py
imankulov: It's unrelated, but necessary I think. 

If something in language settings will be changed (like, the plural expression, as in my case, or a project name, or whatever), `get_or_create` function tries to take the object using all kwargs in criteria, finds nothing, tries to create a new language with code `en` and dies with an exception. My modifications make `get_or_create` search by code only

pootle/core/initdb.py
imankulov: Agree. Will fix that

pootle/core/initdb.py
imankulov: Please see my note reg. English initialization

imankulov:
@unho Thanks a lot for the input. Made fixes for the issues you have mentioned, in a318b1c.

iafan:
FWIW, the main difference between CLDR and TTK is that CLDR also provides plural forms for non-integer numbers, while TTK (and gettext/.PO) supports only integers. This is why you see the difference in formulas.

imankulov:
@iafan Not quite, actually. This is true for Russian and Polish, for example, but when it comes to Turkish and Hungarian, I guess, they simply disagree on what should be considered as a plural form. [I found this read entertaining](http://www.hungarianreference.com/plurals.aspx), by the way

iafan:
In Hungarian, CLDR also lists two forms for cardinals (`one` and `other`), same as in TTK: http://www.unicode.org/cldr/charts/27/supplemental/language_plural_rules.html#hu

imankulov:
@iafan Good point indeed! That's not quite related, but I wonder why babel returns just a single plural form. Either I'm doing something wrong, or I don't understand something, or there is a bug in Babel CLDR parser...

dwaynebailey:
I haven't had time to look at this too closely but I'm not that comfortable with this global view of things.

In terms of language codes, I think the correct approach going forward is to store all language codes using BCP47 and do any code conversion elsewhere.

In terms of plurals forms.  We need to break away from a one plural form model.  I am hosting projects that use different plural form sources (which may or may not align).  Ideally a project should know what form it uses and that should be the one being used.  So Gettext, CLDR or whatever.

imankulov:
@dwaynebailey, I'm all in for per-project instead of global language specifications. Not sure how much efforts it would take though...

Meanwhile, it looks like the Babel CLDR parser we rely upon in this PR, is currently broken and returns invalid results for some languages. Refs python-babel/babel#287

dwaynebailey:
@imankulov not 100% sure the effort myself.  From a Pootle perspective the only thing we really need is the nplurals i.e. the count of plural forms.  In PO we do have the equation because we output it in the PO files.

I'm not sure how much we want to depend on Babel, it gives us much more then we want I think at the moment.

I'm trying to think how best to leverage what you are doing here so that it can be generally helpful for us moving forward.

imankulov:
I think it's time to close it down 😄

PR #4281, opened 2015-12-07
Optimize calculate checks

phlax:
This PR adds an optimized checker for updating Unit checks.

By avoiding iterating full Units from a full Unit set it significantly reduces the amount of memory taken, and speeds up processing

while this duplicates some of the functionality from the Unit model  it is necessary to optimize the calculate_checks command enough to make it usable on a large dataset.

while outside the scope of this PR we could possibly find other parts of the code that might use QualityCheckUpdater optimization afterwards

### DIFF
--- a/pootle/apps/pootle_app/management/commands/calculate_checks.py
+++ b/pootle/apps/pootle_app/management/commands/calculate_checks.py
@@ -33,8 +33,6 @@ class Command(PootleCommand):
         ),
     )
     option_list = PootleCommand.option_list + shared_option_list
-
-    cached_methods = [CachedMethods.CHECKS]
     process_disabled_projects = True
 
     def handle_all_stores(self, translation_project, **options):
--- a/pootle/apps/pootle_app/management/commands/calculate_checks.py
+++ b/pootle/apps/pootle_app/management/commands/calculate_checks.py
@@ -14,11 +14,8 @@ from optparse import make_option
 # This must be run before importing Django.
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
 
-from django.utils import timezone
-
 from pootle.core.mixins.treeitem import CachedMethods
-from pootle_store.models import Store, QualityCheck, Unit
-from pootle_store.util import OBSOLETE
+from pootle.core.checks.checker import QualityCheckUpdater
 
 from . import PootleCommand
 
@@ -41,82 +38,20 @@ class Command(PootleCommand):
     process_disabled_projects = True
 
     def handle_all_stores(self, translation_project, **options):
-        calculate_checks(check_names=options['check_names'],
-                         translation_project=translation_project)
+        logging.info(
+            u"Running %s for %s",
+            self.name, translation_project)
+        QualityCheckUpdater(
+            options['check_names'],
+            translation_project).update()
 
     def handle_all(self, **options):
         if not self.projects and not self.languages:
             logging.info(u"Running %s (noargs)", self.name)
 
             try:
-                calculate_checks(check_names=options['check_names'])
+                QualityCheckUpdater(options['check_names']).update()
             except Exception:
                 logging.exception(u"Failed to run %s", self.name)
         else:
             super(Command, self).handle_all(**options)
-
-
-def calculate_checks(check_names=None, translation_project=None):
-    store_fk_filter = {}
-    unit_fk_filter = {}
-
-    if translation_project is not None:
-        store_fk_filter = {
-            'store__translation_project': translation_project,
-        }
-        unit_fk_filter = {
-            'unit__store__translation_project': translation_project,
-        }
-
-    logging.info('Calculating quality checks for all units...')
-    QualityCheck.delete_unknown_checks()
-
-    checks = QualityCheck.objects.filter(**unit_fk_filter)
-    if check_names is not None:
-        checks = checks.filter(name__in=check_names)
-    checks = checks.values('id', 'name', 'unit_id',
-                           'category', 'false_positive')
-    all_units_checks = {}
-    for check in checks:
-        all_units_checks.setdefault(
-            check['unit_id'], {})[check['name']] = check
-
-    unit_filter = {
-        'state__gt': OBSOLETE
-    }
-    unit_filter.update(store_fk_filter)
-    # unit's query is faster without `select_related('store')`
-    units = Unit.simple_objects.filter(**unit_filter) \
-                               .order_by('store__id')
-    store = None
-    # units are ordered by store, we update dirty cache after we switch to
-    # another store
-    for unit_count, unit in enumerate(units.iterator(), start=1):
-        if store is None or unit.store_id != store.id:
-            if store is not None:
-                store.update_dirty_cache()
-            # we get unit.store only if the store differs from previous
-            store = Store.simple_objects.get(id=unit.store_id)
-
-        # HACKISH: set unit.store to avoid extra querying in
-        # `unit.update_quality_checks()` method
-        unit.store = store
-
-        unit_checks = {}
-        if unit.id in all_units_checks:
-            unit_checks = all_units_checks[unit.id]
-
-        if unit.update_qualitychecks(keep_false_positives=True,
-                                     check_names=check_names,
-                                     existing=unit_checks):
-
-            # update unit.mtime but avoid to use unit.save() because it can
-            # trigger unnecessary things: logging, stats cache updating
-            # TODO: add new action type `quality checks were updated`?
-            Unit.simple_objects.filter(id=unit.id).update(mtime=timezone.now())
-
-        if unit_count % 10000 == 0:
-            logging.info("%d units processed", unit_count)
-
-    if store is not None:
-        store.update_dirty_cache()
--- a/pootle/apps/pootle_app/management/commands/calculate_checks.py
+++ b/pootle/apps/pootle_app/management/commands/calculate_checks.py
@@ -48,10 +48,6 @@ class Command(PootleCommand):
     def handle_all(self, **options):
         if not self.projects and not self.languages:
             logging.info(u"Running %s (noargs)", self.name)
-
-            try:
-                QualityCheckUpdater(options['check_names']).update()
-            except Exception:
-                logging.exception(u"Failed to run %s", self.name)
+            QualityCheckUpdater(options['check_names']).update()
         else:
             super(Command, self).handle_all(**options)
--- a/pootle/core/checks/checker.py
+++ b/pootle/core/checks/checker.py
@@ -44,7 +44,7 @@ class CheckableUnit(UnitProxy):
 
 class UnitQualityCheck(object):
 
-    def __init__(self, unit, checker original_checks,
+    def __init__(self, unit, checker, original_checks,
                  check_names, keep_false_positives=True):
         """Refreshes QualityChecks for a Unit
 
--- a/pootle/core/checks/checker.py
+++ b/pootle/core/checks/checker.py
@@ -1,0 +1,329 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+import time
+
+from django.conf import settings
+from django.utils import timezone
+from django.utils.functional import cached_property
+from django.utils.lru_cache import lru_cache
+
+from pootle_misc.checks import run_given_filters
+from pootle_misc.util import import_func
+from pootle_store.models import QualityCheck, Unit, Store
+from pootle_store.unit import UnitProxy
+from pootle_store.util import OBSOLETE
+from pootle_translationproject.models import TranslationProject
+
+
+logger = logging.getLogger(__name__)
+
+
+class CheckableUnit(UnitProxy):
+    """CheckableUnit wraps a `Unit` values dictionary to provide a `Unit` like
+    instance that can be used by UnitQualityCheck
+
+    At a minimum the dict should contain source_f, target_f, store__id, and
+    store__translation_project__id
+    """
+
+    @property
+    def store(self):
+        return self.store__id
+
+    @property
+    def tp(self):
+        return self.store__translation_project__id
+
+
+class UnitQualityCheck(object):
+
+    def __init__(self, unit, checker original_checks,
+                 check_names, keep_false_positives=True):
+        """Refreshes QualityChecks for a Unit
+
+        As this class can work with either `Unit` or `CheckableUnit` it only
+        uses a minimum of `Unit` attributes from `self.unit`.
+
+        :param unit: an instance of Unit or CheckableUnit
+        :param checker: a Checker for this Unit.
+        :param original_checks: current QualityChecks for this Unit
+        :param check_names: limit checks to given list of quality check names.
+        :param keep_false_positives: when set to `False`, it will unmute any
+            existing false positive checks.
+        """
+        self.checker = checker
+        self.unit = unit
+        self.original_checks = original_checks
+        self.check_names = check_names
+        self.keep_false_positives = keep_false_positives
+        self.unmute_list = []
+
+    @cached_property
+    def check_failures(self):
+        """Current QualityCheck failure for the Unit
+        """
+        if self.check_names is None:
+            return self.checker.run_filters(
+                self.unit, categorised=True)
+        return run_given_filters(
+            self.checker, self.unit, self.check_names)
+
+    @cached_property
+    def checks_qs(self):
+        """QualityCheck queryset for the Unit
+        """
+        return QualityCheck.objects.filter(unit=self.unit.id)
+
+    def delete_checks(self, checks):
+        """Delete checks that are no longer used.
+        """
+        to_delete = self.checks_qs.filter(name__in=checks)
+        if to_delete.exists():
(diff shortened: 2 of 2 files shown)


pootle/apps/pootle_misc/checker.py
```diff
@@ -0,0 +1,288 @@
+# -*- coding: utf-8 -*-
```
dwaynebailey: I think @julen had an idea of not using ``pootle_misc`` going forward but moving such things into ``pootle.core.misc`` IIRC.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+class UnitQualityCheck(object):
+
+    def __init__(self, checker, unit, original_checks,
+                 check_names, keep_false_positives=True):
+        """Refreshes QualityChecks for a Unit
+
+        :param checker: a Checker for this Unit.
+        :param unit: an instance of Unit.
+        :param original_checks: current QCs for this Unit
```
unho: `QCs`? Please use the long version since this is meant to be a human readable text.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        :param check_names: limit checks to given list of quality check names.
+        :param keep_false_positives: when set to `False`, it will activate
+            (unmute) any existing false positive checks.
```
unho: Just saying `unmute` is enough.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        self.unmute_list = []
+
+    @cached_property
+    def check_failures(self):
+        """Current QC failure for the Unit
+        """
+        if self.check_names is None:
+            return self.checker.run_filters(
+                self.unit, categorised=True)
+        else:
```
unho: Unnecessary `else`.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+            return run_given_filters(
+                self.checker, self.unit, self.check_names)
+
+    @cached_property
+    def check_qs(self):
+        """QualityCheck queryset for the Unit
+        """
```
unho: These docstrings fit in a single line.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        return QualityCheck.objects.filter(unit=self.unit.id)
```
unho: Nit: I find easier to use:

```python
        return self.unit.qualitycheck_set.all()
```

phlax: maybe you have not understood the code

unho: Yep, didn't notice it.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def update(self):
+        """Update QCs for a Unit, deleting and unmuting as appropriate.
+        """
+        # update the checks for this unit
+        updated = self.update_checks()
+
+        # delete any remaining checks that were only in the original list
+        deleted = (
+            self.original_checks and self.delete_checks(self.original_checks))
```
unho: Please use a regular `if`.

pootle/core/checks/checker.py
phlax: k - ill add a docstring to explain it better

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # unmute any checks that have been marked for unmuting
+        unmuted = (
+            self.unmute_list and self.unmute_checks(self.unmute_list))
+
+        return (updated or deleted or unmuted)
+
+    def update_checks(self):
```
unho: Nit: Put this method before `update`.

pootle/apps/pootle_app/management/commands/calculate_checks.py
```diff
@@ -36,87 +33,16 @@ class Command(PootleCommand):
...
-
-    cached_methods = [CachedMethods.CHECKS]
     process_disabled_projects = True
 
     def handle_all_stores(self, translation_project, **options):
-        calculate_checks(check_names=options['check_names'],
-                         translation_project=translation_project)
+        QualityCheckUpdater(
+            options['check_names'],
+            translation_project=translation_project)
```
unho: Nit: just pass `translation_project` instead of `translation_project=translation_project` which is fine because it is the second argument.

unho: @phlax Shouldn't some of the methods of `QualityCheckUpdater` be called after spawning it?

phlax: i dont think so - why?

phlax: sorry - i missed your point - yep - ill change

unho: I think you already noticed, but just for the record: Because if you just initialize the object, and the don't call any method, then no quality check is being calculated/updated/whatever.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def __init__(self, check_names=None, translation_project=None,
+                 keep_false_positives=True):
+        """Refreshes QualityChecks for Units
+
+        :param check_names: limit checks to given list of quality check names.
+        :param translation_project: an instance of `TranslationProject` to
+            restrict the update to.
+        :param keep_false_positives: when set to `False`, it will activate
+            (unmute) any existing false positive checks.
```
unho: Nit: As above, just using "unmute" is good enough.

pootle/apps/pootle_app/management/commands/calculate_checks.py
phlax: yep

pootle/core/checks/checker.py
phlax: it was copied from the original code

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+            units = units.filter(
+                store__translation_project=self.translation_project)
+        return units
+
+    def clear_checks(self):
+        QualityCheck.delete_unknown_checks()
+
+    @lru_cache(maxsize=None)
+    def get_checker(self, tp):
+        """Return the site QualityChecker or the QC associated with
```
unho: What does `QC` mean?

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        if self.translation_project is not None:
+            units = units.filter(
+                store__translation_project=self.translation_project)
+        return units
+
+    def clear_checks(self):
+        QualityCheck.delete_unknown_checks()
+
+    @lru_cache(maxsize=None)
+    def get_checker(self, tp):
```
unho: I would rename `tp` to `tp_pk` as it describes it better.

unho: Also this method looks like it belongs to `pootle_translationproject` app rather than here.

pootle/apps/pootle_app/management/commands/calculate_checks.py
```diff
@@ -36,87 +33,16 @@ class Command(PootleCommand):
...
-
-    if translation_project is not None:
-        store_fk_filter = {
-            'store__translation_project': translation_project,
-        }
-        unit_fk_filter = {
-            'unit__store__translation_project': translation_project,
-        }
-
-    logging.info('Calculating quality checks for all units...')
```
unho: This was lost, therefore there might not be any feedback on the command being run in some scenarios.

pootle/core/checks/checker.py
phlax: it wouldnt allow it to be cached without an additional layer of complexity - ie cache expiry

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+            if self.update_translated_unit():
+                updated_count += 1
+
+        return updated_count
+
+    def update_untranslated(self):
+        """Delete QCs for untranslated Units
+        """
+        check_qs = self.check_qs
+        check_qs = check_qs.filter(unit__target_f="")
```
unho: I find it easier to combine both lines:

```python
        check_qs = self.check_qs.filter(unit__target_f="")
```

I think this double assignment was used somewhere else in the PR.

unho: @phlax Please note that there might be translated units with empty target (Mozilla weirdness), so better check by `unit__state=UNTRANSLATED`.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        self.stores.update(
+            set(check_qs.values_list("unit__store__pk", flat=True).distinct()))
```
unho: I don't think it is necessary to convert to a set.

pootle/core/checks/checker.py
phlax: they still wouldnt require checking

pootle/core/checks/checker.py
phlax: distinct can be unreliable - so its safer

pootle/core/checks/checker.py
phlax: any unit that has "" as its target regardless of whether its state is UNTRANSLATED can be safely ignored for checks afaiaa

unho: I find safer to check for `UNTRANSLATED` as I can't predict if in the future a "" target might have some check for it.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def update_translated(self):
+        """Update checks for translated Units
+        """
+        unit_fields = [
+            "id", "source_f", "target_f", "locations", "store__id"]
+
+        tp_key = "store__translation_project__id"
+        if self.translation_project is None:
+            unit_fields.append(tp_key)
```
unho: Just:

```python
        if self.translation_project is None:
            unit_fields.append("store__translation_project__id")
```

so we don't define single use vars.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+
+        translated = self.units.exclude(target_f="").values(*unit_fields)
+
+        updated_count = 0
+        checker = None
+        if self.translation_project is not None:
+            # we only need to get the checker once if TP is set
+            checker = self.get_checker(self.translation_project.id)
```
unho: We get the checker and we don't use it. Why?

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,305 @@
...
+        unit_fields = [
+            "id", "source_f", "target_f", "locations", "store__id"]
+
+        tp_key = "store__translation_project__id"
+        if self.translation_project is None:
+            unit_fields.append(tp_key)
+
+        translated = self.units.exclude(target_f="").values(*unit_fields)
+
+        updated_count = 0
```
unho: Nit: Move `updated_count` to just before the loop where it is changed.

pootle/core/checks/checker.py
unho: Perhaps forgot to pass it to `self.update_translated_unit()` below?

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,314 @@
...
+            checker,
+            unit,
+            self.checks.get(unit.id, {}),
+            self.check_names,
+            self.keep_false_positives)
+        if checker.update():
+            self.stores.add(unit.store)
+            self.units.filter(id=unit.id).update(mtime=timezone.now())
+            return True
+        return False
```
unho: Nit: Some blank lines in this method code will make it easier to read it.

pootle/core/checks/checker.py
phlax: it does get reused but there is an error in the code - which i have fixed - but waiting to push

pootle/apps/pootle_app/management/commands/calculate_checks.py
```diff
@@ -36,87 +33,19 @@ class Command(PootleCommand):
...
-                                     check_names=check_names,
-                                     existing=unit_checks):
-
-            # update unit.mtime but avoid to use unit.save() because it can
-            # trigger unnecessary things: logging, stats cache updating
-            # TODO: add new action type `quality checks were updated`?
-            Unit.simple_objects.filter(id=unit.id).update(mtime=timezone.now())
-
-        if unit_count % 10000 == 0:
-            logging.info("%d units processed", unit_count)
```
unho: We also lose this feedback. When traversing massive queryset do we get enough feedback to know that the command is still running and therefore don't panic?

pootle/core/checks/checker.py
unho: Please replace all `QC` appearance by the long version of it.

pootle/apps/pootle_app/management/commands/calculate_checks.py
phlax: yep - that was there because it took such a ridiculous amount of time/energy to complete - the feedback is now timings between each part of the job - im happy with that

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,314 @@
...
+
+        As this class can work with either `Unit` or `CheckableUnit` it only
+        uses a minimum of `Unit` attributes from `self.unit`.
+
+        :param checker: a Checker for this Unit.
+        :param unit: an instance of Unit or CheckableUnit
+        :param original_checks: current QCs for this Unit
+        :param check_names: limit checks to given list of quality check names.
+        :param keep_false_positives: when set to `False`, it will activate
+            (unmute) any existing false positive checks.
```
unho: Just saying `unmute` is enough.

pootle/core/checks/checker.py
unho: Please rewrite all single line docstrings on the PR.

pootle/core/checks/checker.py
unho: Like:

```python
        deleted = False
        if self.original_checks:
            deleted = self.delete_checks(self.original_checks)
```

The same goes for `unmuted` below.

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,314 @@
...
+        """Update checks for translated Units
+        """
+        unit_fields = [
+            "id", "source_f", "target_f", "locations", "store__id"]
+
+        tp_key = "store__translation_project__id"
+        if self.translation_project is None:
+            unit_fields.append(tp_key)
+
+        translated = self.units.exclude(target_f="").values(*unit_fields)
```
unho: Nit: Move this to just before the loop where it is used.

unho:
lgtm given the raised nits are addressed and this is properly tested to ensure it works.

pootle/core/checks/checker.py
phlax: yep - fixed in latest code - waiting to push

pootle/core/checks/checker.py
unho: Ok, then keep it here.

pootle/core/checks/checker.py
unho: ok

pootle/apps/pootle_app/management/commands/calculate_checks.py
phlax: added a log in command

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,310 @@
...
+        self.check_names = check_names
+        self.translation_project = translation_project
+        self.keep_false_positives = keep_false_positives
+        self.stores = set()
+
+    @cached_property
+    def checks(self):
+        """Existing checks in the database for all units
+        """
+        checks = self.check_qs
```
unho: Please rename variable to `checks_qs`.

pootle/core/checks/checker.py
phlax: actually you are right - i will change

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,324 @@
...
+        self.translation_project = translation_project
+        self.keep_false_positives = keep_false_positives
+        self.stores = set()
+        self._store_to_expire = None
+
+    @cached_property
+    def checks(self):
+        """Existing checks in the database for all units
+        """
+        checks = self.checks_qs
```
unho: Shouldn't this be the following?

```python
        checks = self.checks_qs()
```

pootle/core/checks/checker.py
```diff
@@ -0,0 +1,324 @@
...
+            if checker is None:
+                checker = self.get_checker(unit[tp_key])
+            if checker and self.update_translated_unit(unit, checker=checker):
+                updated_count += 1
+        return updated_count
+
+    def update_untranslated(self):
+        """Delete QualityChecks for untranslated Units
+        """
+        checks_qs = self.checks_qs.exclude(unit__state__gte=OBSOLETE)
```
unho: Shouldn't this be the following?

```python
        checks_qs = self.checks_qs().exclude(unit__state__gte=OBSOLETE)
```

phlax: no its an instance property

pootle/core/checks/checker.py
phlax: no - its a property

unho: Sorry, didn't notice the property.

ta2-1:
GTM

PR #4285, opened 2015-12-13
VFolder optimizations

phlax:
- Don't check vfolders if not in a TP ui screen
   - this is a very expensive query and has no purpose if the pootle_path starts with /projects
- Use a more efficient search for deciding when to show vfolder priority

### DIFF
--- a/pootle/core/helpers.py
+++ b/pootle/core/helpers.py
@@ -23,7 +23,7 @@ from pootle_misc.stats import get_translation_states
 from pootle_store.models import Unit
 from pootle_store.views import get_step_query
 from pootle_translationproject.models import TranslationProject
-from virtualfolder.models import VirtualFolder
+from virtualfolder.models import VirtualFolderTreeItem
 
 from .url_helpers import get_path_parts, get_previous_url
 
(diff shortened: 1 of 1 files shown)


iafan:
I think the idea was to be able to have vfolders on all levels, including anything that starts with `/projects/`.

If the queries behind vfolders are very expensive, we might want to rather think on how to optimize vfolder performance in general.

phlax:
> If the queries behind vfolders are very expensive, we might want to rather think on how to optimize vfolder performance in general.

yep - agreed. Atm though its making this query and its really expensive and pointless

iafan:
Can you please double-check with @unho? I might be wrong, but my understanding is that we already can specify vfolders on project level (and in this case your PR might break some things).

phlax:
yep ill check with him - but atm there are no units with a store that has a pootle_path beginning with /projects - so this is unnecessary

iafan:
Thanks!

unho:
Current status is that you can specify vfolders with any path. But vfolders are only displayed at TP level. The idea is to enable them everywhere in the future.

pootle/core/helpers.py
```diff
@@ -84,10 +84,11 @@ def get_translation_context(request):
     vfolder_pk = getattr(request, 'current_vfolder', '')
     display_priority = False
 
-    if not vfolder_pk:
-        display_priority = VirtualFolder.objects.filter(
-            units__store__pootle_path__startswith=request.pootle_path
-        ).exists()
+    if not vfolder_pk and not request.pootle_path.startswith("/projects"):
```
unho: If we are implementing this optimization as a temporary speedup (until vfolders are displayed outside TP level), then you might want to also leave out requests with `pootle_path` for a specific language (So only calculate if `pootle_path` has at least two items and it doesn't start with `/projects/`, like `/lang/proj/` or `/lang/proj/whatever/`)

phlax: im not seeing how that can be done efficiently without checking if `lang` is a valid lang code

unho: `if not vfolder_pk and len(request.pootle_path.split("/")) < 4 and not request.pootle_path.startswith("/projects")`

unho:
lgtm once concerns addressed.

unho:
Thanks for doing the changes.

lgtm.

PR #4286, opened 2015-12-13
Remove remaining default select_relateds

phlax:
these should be added where required not as the default query

### DIFF
--- a/pootle/apps/pootle_language/models.py
+++ b/pootle/apps/pootle_language/models.py
@@ -25,17 +25,6 @@ from pootle.i18n.gettext import tr_lang, language_dir
 from staticpages.models import StaticPage
 
 
-class LanguageManager(models.Manager):
-
-    def get_queryset(self):
-        """Mimics `select_related(depth=1)` behavior. Pending review."""
-        return (
-            super(LanguageManager, self).get_queryset().select_related(
-                'directory',
-            )
-        )
-
-
 class LiveLanguageManager(models.Manager):
     """Manager that only considers `live` languages.
 
@@ -108,7 +97,7 @@ class Language(models.Model, TreeItem):
     directory = models.OneToOneField('pootle_app.Directory', db_index=True,
                                      editable=False)
 
-    objects = LanguageManager()
+    objects = models.Manager()
     live = LiveLanguageManager()
 
     class Meta(object):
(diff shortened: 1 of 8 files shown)


unho:
The "add where required" part is meant to be part of a separate PR? Have you checked the impact on performance?

phlax:
> Have you checked the impact on performance?

yep - massive improvement

pootle/apps/pootle_language/models.py
```diff
@@ -108,7 +98,7 @@ class Language(models.Model, TreeItem):
     directory = models.OneToOneField('pootle_app.Directory', db_index=True,
                                      editable=False)
 
-    objects = LanguageManager()
+    objects = models.Manager()
```
unho: Is it necessary to keep this line?

pootle/apps/pootle_language/models.py
```diff
@@ -94,6 +83,7 @@ class Language(models.Model, TreeItem):
                           'localization-guide/en/latest/l10n/'
                           'pluralforms.html">'
                           'our page</a> on plural forms.')
+
```
unho: Please remove this blank line.

pootle/apps/pootle_store/models.py
```diff
@@ -1351,11 +1331,6 @@ def get_last_updated_info(self):
 class StoreManager(models.Manager):
```
unho: It seems that now there is not a big difference between this and `simple_objects`, so perhaps `simple_objects` can be dropped as well.

phlax: > It seems that now there is not a big difference between this and simple_objects, so perhaps simple_objects can be dropped as well.

probably - but that would require changing anywhere that currently uses - so i reckon that is a separate PR

pootle/apps/pootle_language/models.py
phlax: it seems that you have to if there is another manager

unho:
lgtm

julen:
As we [commented yesterday on IRC](https://botbot.me/freenode/pootle-dev/2015-12-16/?msg=56261218&page=1), in general this makes things faster except the cross-language resource pages (such as `/projects/<code>/resource/file.foo`). The different is not huge or noticeable for end-users, but it's worth mentioning it here for the record.

phlax:
@julen thanks for this - i am following this up

Issue #4279, opened 2015-12-07
Wordcount code of Pootle gives a wrong output sometimes


JohnnyKing94:
I wanna alert you that sometime the code used by Pootle to count the words gives a wrong output. There are some phrases like this with I made experiments that Pootle calculates **44 words**, when they're ONLY **40**.

```
There is a woman in town by the name of Elsie Dolban. She is eccentric, and that has caused some of the villagers to condemn her. I'm interested in *your* opinion of her.nnSpeak to Inquisitor Roche, he has strong opinions in this matter.
```
I don't know if it's the ```*``` as special character or what... but true is, I used every tools, even the command ```wc -w``` of linux to count the words. now i'm worried about the real word count of all my strings uploaded on the db having special characters...
I alert you to take a look into it.

iafan:
I counted the words manually (just splitting text by space and assuming `nn` here is a line break) and there are 44 words. How did you get 40?

JohnnyKing94:
Sorry i meant 43 words... i used Microsoft Word and the "echo "THE-TEXT | wc -w" command on linux...

iafan:
Ok. So is this `nn` thing in the middle a real line break in the string or do you have `her.nnSpeak` exactly like this and expect it to be counted as one word? Do you probably have a link to the unit in Pootle to verify against?

JohnnyKing94:
Yes i was doing tests with the xliff format and i noted that the backslashes were missing, anyway even with ```\n\n```, still counted as 43... it doesn't change...

unho:
Just checked on the shell:

```python
In [1]: from pootle.core.utils.wordcount import wordcount

In [2]: wordcount("There is a woman in town by the name of Elsie Dolban. She "
  ....:           "is eccentric, and that has caused some of the villagers to "
  ....:           "condemn her. I'm interested in *your* opinion of her.\n\n"
  ....:           "Speak to Inquisitor Roche, he has strong opinions in this "
  ....:           "matter.")
Out[2]: 45
```

JohnnyKing94:
Lol, now it tells 45 :D... i have 44 with the both version "\n\n" and the "nn" on Pootle... with my other tools, Word and the linux command tells me 43... and notepad++ tells me 45... wtf is wrong? ahahahhahahahahah

unho:
Is it ok to say Pootle provides the right count :)

It gets funnier if you try:

```python
In [3]: len(("There is a woman in town by the name of Elsie Dolban. She "
[...]

PR #4287, opened 2015-12-14
Tests: Initial test for both Pootle and TTK wordcounts

unho:
Ref #4279.


unho:
@phlax The PR has been updated to use parameterized tests. If it is ok I will squash and land when I am back.

phlax:
lgtm

PR #4292, opened 2015-12-15
[WIP] Search pootle path using regex

phlax:
At the moment when using Unit.objects.get_by_path it uses extra fields. These fields cause a headache with the queryset as store__pootle_path drops out of the scope of the qs.

Also, if pootle_path is not set it currently adds a search of like "/%" which is uncessary.

This PR:
- adds a PathResolver using django's url resolver
- adds regex filtering of qs rather than using extra fields
- removes unnecessary pootle_path filtering 
- adds some tests for above

**just working on some of the code around split_pootle_path**

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -39,7 +39,8 @@ from pootle.core.mixins import CachedMethods, CachedTreeItem
 from pootle.core.models import Revision
 from pootle.core.search import SearchBroker
 from pootle.core.storage import PootleFileSystemStorage
-from pootle.core.url_helpers import get_editor_filter, split_pootle_path
+from pootle.core.url_helpers import (
+    get_editor_filter, split_pootle_path, PathResolver)
 from pootle.core.utils import dateformat
 from pootle.core.utils.timezone import datetime_min, make_aware
 from pootle_misc.aggregate import max_column
@@ -240,34 +241,14 @@ class UnitManager(models.Manager):
         :param pootle_path: An internal pootle path.
         :param user: The user who is accessing the units.
         """
-        lang, proj, dir_path, filename = split_pootle_path(pootle_path)
 
-        units_qs = self.get_for_user(user)
+        units_qs = self.get_for_user(user).exclude(
+            store__pootle_path__startswith="/templates/")
 
-        # /projects/<project_code>/translate/*
-        if lang is None and proj is not None:
-            if dir_path and filename:
-                units_path = ''.join(['/%/', proj, '/', dir_path, filename])
-            elif dir_path:
-                units_path = ''.join(['/%/', proj, '/', dir_path, '%'])
-            elif filename:
-                units_path = ''.join(['/%/', proj, '/', filename])
-            else:
-                units_path = ''.join(['/%/', proj, '/%'])
-        # /projects/translate/*
-        elif lang is None and proj is None:
-            units_path = '/%'
-        # /<lang_code>/<project_code>/translate/*
-        # /<lang_code>/translate/*
-        else:
-            units_path = ''.join([pootle_path, '%'])
-
-        units_qs = units_qs.extra(
-            where=[
-                'pootle_store_store.pootle_path LIKE %s',
-                'pootle_store_store.pootle_path NOT LIKE %s',
-            ], params=[units_path, '/templates/%']
-        )
+        resolver = PathResolver(pootle_path)
+        if resolver.regex is not None:
+            units_qs = units_qs.filter(
+                store__pootle_path__regex=resolver.regex)
 
         # Non-superusers are limited to the projects they have access to
         if not user.is_superuser:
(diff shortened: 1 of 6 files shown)


pootle/core/url_helpers.py
```diff
@@ -12,7 +12,83 @@
...
+            return "/%s" % self._path.strip("/")
+        return "/%s/" % self._path.strip("/")
+
+    @cached_property
+    def kwa(self):
+        return resolve(self.path).kwargs
+
+    @property
+    def directory(self):
+        if not self.path:
```
unho: Shouldn't this be `self._path`?

phlax: no, i dont thinks so - if self.path is empty dont try to parse

pootle/core/url_helpers.py
```diff
@@ -12,7 +12,83 @@
...
+        return self.kwa.get("language_code", None) or None
+
+    @property
+    def project(self):
+        if not self.path:
+            return None
+        return self.kwa.get("project_code", None) or None
+
+    @property
+    def regex(self):
```
unho: Adding some docstring here would be helpful as it is not obvious what is this meant for.

pootle/core/url_helpers.py
```diff
@@ -12,7 +12,83 @@
...
 import urlparse
 
-from django.core.urlresolvers import reverse
+from django.core.urlresolvers import reverse, resolve
+from django.utils.functional import cached_property
+
+
+class PathResolver(object):
+    """Converts a request.path into parts for matching pootle_path
+    with a regex
```
unho: I would rephrase this:

Normalizes the request.path, extracts its parts and provides a regex useful to match the path.

Or use a even more explanatory docstring with examples.

pootle/apps/pootle_store/models.py
```diff
@@ -40,7 +40,8 @@
 from pootle.core.models import Revision
 from pootle.core.storage import PootleFileSystemStorage
 from pootle.core.search import SearchBroker
-from pootle.core.url_helpers import get_editor_filter, split_pootle_path
+from pootle.core.url_helpers import (
+    get_editor_filter, split_pootle_path, PathResolver)
```
unho: Nit: Use styleguide recommendations on how to wrap long lines: "If you break in a statement with parentheses or braces, align to the braces" in http://docs.translatehouse.org/projects/translate-toolkit/en/latest/developers/styleguide.html#continuing-long-statements

pootle/core/url_helpers.py
unho: Nit: Instead of adding the regex capability on the commit where `PathResolver` is added, I would add it on the "Use PathResolver.regex when filtering by pootle_path" commit. This way it is easier to see why some feature has been initially added.

pootle_pytest/fixtures/models/unit.py
```diff
@@ -0,0 +1,90 @@
+
```
unho: No license notice?

pootle_pytest/plugin.py
```diff
@@ -46,7 +48,8 @@
 PARAMETERS = (
     ("update_store_test_names", UPDATE_STORE_TESTS),
     ("file_import_failure_names", FILE_IMPORT_FAIL_TESTS),
-    ("wordcount_names", WORDCOUNT_TESTS))
+    ("wordcount_names", WORDCOUNT_TESTS),
+    ("unit_path_resolver_names", UNIT_PATH_RESOLVER_TESTS))
```
unho: Nit: We should go back to having the opening and closing brackets on its own line to avoid the diff to include stupid changes like this.

tests/models/unit.py
```diff
@@ -281,3 +281,41 @@ def test_accept_suggestion_update_wordcount(it_tutorial_po, system):
...
+            assert parsed is None
+        else:
+            assert parsed == ""
+
+    units = Unit.objects.order_by("pk")
+    if resolved.regex:
+        expected = units.filter(store__pootle_path__regex=resolved.regex)
+    else:
+        expected = units.all()
+    if result.get("project"):
```
unho: I would add a blank line before this `if` for readability.

pootle/core/url_helpers.py
unho: My fault, I always forget to check if there is any attribute with this name.

pootle/core/decorators.py
```diff
@@ -24,10 +24,10 @@
 from pootle_project.models import Project, ProjectSet, ProjectResource
 from pootle_store.models import Store
 from pootle_translationproject.models import TranslationProject
-from virtualfolder.helpers import extract_vfolder_from_path
+from virtualfolder.helpers import extract_vfolder_from_request
 
 from .exceptions import Http400
-from .url_helpers import split_pootle_path
```
unho: There is another usage of `split_pootle_path` in this file, so either get the import back, or replace it.

phlax:
This is no longer required

PR #4310, opened 2015-12-22
Display user's email address to managers and other cleanups

unho:


pootle/apps/accounts/models.py
```diff
@@ -147,6 +148,15 @@ def has_contact_details(self):
     def twitter_url(self):
         return 'https://twitter.com/{0}'.format(self.twitter)
 
+    @property
+    def is_manager(self):
+        """Tells if the user is a manager for any language, project or TP."""
+        return PermissionSet.objects.filter(
+                    user=self,
```
phlax: you might want to use `user.permission_set.filter` here

pootle/templates/user/profile.html
```diff
@@ -18,8 +18,8 @@ <h1 class="user-name">
         {{ object.display_name }}
       {% endif %}
     </h1>
-    {% if user.is_superuser or request.permissions.administrate %}
-    <p><a href="mailto:{{ object.email }}">{{ object.email }}</a></p>
+    {% if user.is_superuser or user.is_manager %}
```
phlax: my only concern here is that we dont want to make this query more than once in a request - ideally we could cache it - or alternatively we could set it in the view context

unho: Caching it has the downside that we have to get rid of the cache every time the permissions are changed. I am ok with setting this on the context provided by the view.

phlax: cool - i think my concern was basically that is_manager looks like is_superuser and is_staff but is less efficient at retrieving a result

unho:
@phlax Addressed your concerns.

phlax:
@unho how about changing the name to "has_manager_permissions" and making it a method rather than property - to differentiate it from `is_superuser` and `is_staff`

unho:
We have been trying for a long time to get rid of methods that act like properties. So I am -1 on `has_manager_permissions`.

phlax:
i am very up for using properties where its appropriate - but here im feeling that this looks way too much like the users attributes which are very cheap to calculate, whereas this is not necessarily

unho:
@phlax PR updated. Any other concern?

unho:
@translate/core Any gtm?

tests/models/user.py
```diff
@@ -429,3 +429,44 @@ def test_update_user_email_bad_invalid_duplicate(member_with_email, member2):
...
+    # Assign 'administrate' right for 'Afrikaans (Tutorial)' TP and check user
+    # is manager.
+    criteria = {
+        'user': no_perms_user,
+        'directory': afrikaans_tutorial.directory,
+    }
+    ps, created = PermissionSet.objects.get_or_create(**criteria)
+    ps.positive_permissions = [administrate]
+    ps.save()
+    assert no_perms_user.has_manager_permissions() is True
```
julen: This is not Pythonic, and in fact it's recommended against in PEP8: for boolean values one would simply `assert no_perms_user.has_manager_permissions()` and `assert not no_perms_user.has_manager_permissions()`.

pootle/templates/user/profile.html
```diff
@@ -34,6 +34,10 @@ <h3 class="user-score">
       {% endif %}
     </h3>
 
+    <h3 class="user-detailed-stats">
+      <a href="{% url 'pootle-user-stats' object.username %}">{% trans 'User stats' %}</a>
```
julen: It might be worth to align the wording with the user menu and use _User statistics_ instead.

pootle/static/css/user.css
```diff
@@ -64,6 +64,18 @@
     color: #5A8295;
 }
 
+#user-meta a.admin,
```
julen: `a.` is unnecessary in these two selectors, let's remove it.

julen:
This breaks when the visitor is an anonymous user: `self.request.user` is Django's own `AnonymousUser`, which lacks `has_manager_permission`. Note this is not the same as the `nobody` user, which is _sort of_ a regular user (in other contexts we overcome this with `request.profile = User.get(request.user)`).

I couldn't find any mentions or references to it, but I see a new link to user statistics has been added here too. In that context, I'd like to hear from @iafan in terms of the positioning and sizing of such link (see screenshot below).

screen shot 2016-01-19 at 10 44 02

unho:
@julen Fixed all the raised issues.

> I couldn't find any mentions or references to it, but I see a new link to user statistics has been added here too. In that context, I'd like to hear from @iafan in terms of the positioning and sizing of such link (see screenshot below).

I noticed that it is basically impossible to reach some user statistics so I added a link in the place it looked more natural to me: that user's profile. I am completely open to style or place it the way you prefer.

julen:
Thank you, @unho.

I think we can easily make the `has_manager_permission()` method more efficient. How about:
```python
if user.is_anonymous():
    return False
if user.is_superuser:
    return True
(...)
return self.permissionset_set.filter(**criteria).exists()
```

Then `get_context_data()` boils down to `context['user_is_manager'] = user.has_manager_permissions()`.

I'm happy with the changes and they work fine. However as said I'd like to hear from @iafan in terms of the *User statistics* link before merging.

iafan:
@unho I'd call this link "Statistics" to match the menu item name, and would place it below the "Last action: ...." line with the same font size/style.

unho:
@unho I'd call this link "Statistics" to match the menu item name, and would place it below the "Last action: ...." line with the same font size/style.

@iafan You mean below the "Latest activity" line?

Will rename to "Statistics".

unho:
@julen Updated to include your suggestions.

unho:
@julen @iafan Since the inclusion of the Statistics link looks like is going to block this PR, I am moving that particular change to a new PR: #4405

iafan:
> @iafan You mean below the "Latest activity" line?

Yes.

julen:
@unho as mentioned earlier, the changes are fine to be merged.

unho:
Thanks. Merging.

PR #4311, opened 2015-12-22
Enable flake8

dwaynebailey:
This enables flake8 by default.  Seems it is most useful for catching unused imports.

I've used `# noqa` in a few places where we're importing libraries based within `if` statements.

Also note that the build will fail as we have 2 outstanding fixes that I'm hoping someone else can fix:

```
./pootle/apps/virtualfolder/helpers.py:44:35: F812 list comprehension redefines 'dir_path' from line 35
./pootle/core/helpers.py:146:50: F812 list comprehension redefines 'units' from line 140
``

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,3 +57,16 @@ exclude=.svn,CVS,.bzr,.hg,.git,__pycache__,migrations,conf.py,_build,.tox,pootle
 # Default ignore list for ref E121,E123,E126,E226,E24,E704
 ignore=E129,E226,E402,W503
 statistics=True
+
+[flake8]
+# See [pep8] above, [pep8] is the reference, please keep in sync.
+# You would think flake8 would have the decency of reading the pep8 configs,
+# but it doesn't so we have to repeat them.
+max-line-length=84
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: setup.cfg [hunk 1]
```diff
@@ -50,3 +50,16 @@ exclude=migrations,conf.py,.tox,.git,pootle/static/js,pootle/translations,pootle
 # Default ignore list for ref E121,E123,E126,E226,E24,E704
 ignore=E129,E226,E402,W503
 statistics=True
+
+[flake8]
+# See [pep8] above, [pep8] is the reference, please keep in sync.
+# You would think flake8 would have the decency of reading the pep8 configs,
+# but it doesn't so we have to repeat them.
+max-line-length=84
+exclude=migrations,conf.py,.tox,.git,pootle/static/js,pootle/translations,pootle/locale,pootle/assets
```
unho: Use just `pootle/static` to also exclude CSS.

Perhaps `pootle/dbs`, `pootle/exports`, `pootle/log`, `pootle/templates` and `pootle/tools` can also be excluded.

unho:
Landed a fix for those in https://github.com/translate/pootle/commit/bff090ba despite the virtualfolders one will be gone once @phlax lands its superb refactor from last friday.

ta2-1:
lgtm

PR #4314, opened 2015-12-24
Sort imports and automatic import checking

dwaynebailey:
This uses `isort` to ensure the the imports are sorted.

- The imports have been sorted
- Travis is enabled
- setup.cfg contains the config for isort
- A number of files are excluded where they can't be validated automatically by isort, or where we deviate from import sorting e.g. pootlle_pytest, tests, commands and a few others.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,12 @@ env:
   - TOXENV=eslint
   - TOXENV=docs
   - TOXENV=mo
+  - TOXENV=isort
   - TOXENV=pep8_flake8
   - TOXENV=pep257
+matrix:
+  allow_failures:
+    - env: TOXENV=isort
 cache:
   directories:
     - pootle/assets/js/node_modules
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ SPRITE_DIR = ${IMAGES_DIR}/sprite
 FORMATS=--formats=bztar
 TEST_ENV_NAME = pootle_test_env
 
-.PHONY: all build clean sprite test pot mo mo-all help docs assets pep8
+.PHONY: all build clean sprite test pot mo mo-all help docs assets pep8 sort-imports
```
unho: Where is `sort-imports` section on the Makefile?

dwaynebailey: Removed simply running ``isort`` is enough

ta2-1:
lgtm

dwaynebailey:
The Travis part is still present but I've added it to an allowed failures section.  We can add anything that we care about, but not enough to block a build.

PR #4316, opened 2015-12-27
Optimize extract_vfolder_from_path function

phlax:
- doesnt use split_pootle_path
- minimizes db calls

### DIFF
--- a/pootle/apps/virtualfolder/helpers.py
+++ b/pootle/apps/virtualfolder/helpers.py
@@ -7,81 +7,32 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-from pootle.core.url_helpers import get_all_pootle_paths, split_pootle_path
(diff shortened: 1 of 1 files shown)


pootle/apps/virtualfolder/helpers.py
```diff
@@ -7,81 +7,21 @@
...
-
-    The path /gl/firefox/browser/vfolder/chrome/file.po with the vfolder
-    virtual folder on it will be converted to
-    /gl/firefox/browser/chrome/file.po if the virtual folder exists and is
-    public.
-
-    Have in mind that several virtual folders with the same name might apply in
-    the same path (as long as they have different locations this is possible)
-    and in such cases the one with higher priority is returned.
-    """
```
unho: I would prefer to keep some helpful docstring.

unho:
Adding tests for this should be straightforward.

unho:
After checking with @phlax I will be contributing a test for this.

Issue #3659, opened 2015-04-07
`update_stores` fails after renaming directory and store from uppercase to lowercase


ta2-1:
```
Traceback (most recent call last):
  File "/home/l10n/pootle/pootle/apps/pootle_app/management/commands/__init__.py", line 43, in do_translation_project
    process_stores = self.handle_translation_project(tp, **options)
  File "/home/l10n/pootle/pootle/apps/pootle_app/management/commands/update_stores.py", line 41, in handle_translation_project
    translation_project.scan_files()
  File "/home/l10n/pootle/pootle/apps/pootle_translationproject/models.py", line 375, in scan_files
    file_filter,
  File "/home/l10n/pootle/pootle/apps/pootle_app/project_tree.py", line 254, in add_files
    fs_subdir, db_subdir, file_filter)
  File "/home/l10n/pootle/pootle/apps/pootle_app/project_tree.py", line 254, in add_files
    fs_subdir, db_subdir, file_filter)
  File "/home/l10n/pootle/pootle/apps/pootle_app/project_tree.py", line 254, in add_files
    fs_subdir, db_subdir, file_filter)
  File "/home/l10n/pootle/pootle/apps/pootle_app/project_tree.py", line 224, in add_files
    file_filter)
  File "/home/l10n/pootle/pootle/apps/pootle_app/project_tree.py", line 125, in split_files_and_dirs
    for child_path in [child_path for child_path in os.listdir(real_dir)
OSError: [Errno 2] No such file or directory: '/home/l10n/production/serge/po/web_evernote/ar/resources/mail/PremiumGiftSubscriptionConfirmationStencil'
```

It happened because MySQL used case insensitive collation. So probably we shouldn't use it.

julen:
But this is a deployment-specific thing, isn't it? If so, some docs warning about possible issues of using a case-insensitive collation would suffice, as Pootle won't be able to do anything.

iafan:
We don't want to create a hard dependency on a specific collation type in a specific database. Moreover, changing collation type might affect searching functionality.

This is a bug that I think should better be addressed from the client side: if on update_stores we detect that some folder was renamed, we need to act accordingly. The easiest way is to treat it as if the old folder is gone and the new one is created.

ta2-1:
Seems these queries solve the problem
```SQL
ALTER TABLE pootle_store_store MODIFY pootle_path varchar(255) collate utf8_bin;
[...]

PR #4317, opened 2015-12-27
Convert get_for_path to get_translatable

phlax:
- replaces Unit.get_for_path with Unit.get_translatable
- avoids use of path or regex wherever possible

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -83,6 +83,9 @@ CHECKED = 2
 POOTLE_WINS = 1
 FILE_WINS = 2
 
+LANGUAGE_REGEX = r"[^/]{,255}"
+PROJECT_REGEX = r"[^/]{,255}"
+
 
 # # # # # # # # Quality Check # # # # # # #
 
@@ -234,50 +237,62 @@ class UnitManager(models.Manager):
         return self.live() \
                    .filter(store__translation_project__project__disabled=False)
 
-    def get_for_path(self, pootle_path, user):
+    def get_translatable(self, user, project_code=None, language_code=None,
+                         dir_path=None, filename=None):
         """Returns units that fall below the `pootle_path` umbrella.
 
-        :param pootle_path: An internal pootle path.
+        :param project_code: A string for matching the code of a Project.
+        :param language_code: A string for matching the code of a Language.
+        :param dir_path: A string for matching the dir_path and descendants
+           from the TP.
+        :param filename: A string for matching the filename of Stores.
         :param user: The user who is accessing the units.
         """
-        lang, proj, dir_path, filename = split_pootle_path(pootle_path)
+        from pootle_project.models import Project
+
+        if not user.is_superuser:
+            user_projects = Project.accessible_by_user(user)
+            if project_code and project_code not in user_projects:
+                return self.none()
 
         units_qs = self.get_for_user(user)
 
-        # /projects/<project_code>/translate/*
-        if lang is None and proj is not None:
-            if dir_path and filename:
-                units_path = ''.join(['/%/', proj, '/', dir_path, filename])
-            elif dir_path:
-                units_path = ''.join(['/%/', proj, '/', dir_path, '%'])
-            elif filename:
-                units_path = ''.join(['/%/', proj, '/', filename])
-            else:
-                units_path = ''.join(['/%/', proj, '/%'])
-        # /projects/translate/*
-        elif lang is None and proj is None:
-            units_path = '/%'
-        # /<lang_code>/<project_code>/translate/*
-        # /<lang_code>/translate/*
-        else:
-            units_path = ''.join([pootle_path, '%'])
-
-        units_qs = units_qs.extra(
-            where=[
-                'pootle_store_store.pootle_path LIKE %s',
-                'pootle_store_store.pootle_path NOT LIKE %s',
-            ], params=[units_path, '/templates/%']
-        )
-
-        # Non-superusers are limited to the projects they have access to
-        if not user.is_superuser:
-            from pootle_project.models import Project
-            user_projects = Project.accessible_by_user(user)
+        if language_code:
             units_qs = units_qs.filter(
-                store__translation_project__project__code__in=user_projects,
-            )
+                store__translation_project__language__code=language_code)
+        else:
+            units_qs = units_qs.exclude(
+                store__pootle_path__startswith="/templates/")
 
-        return units_qs
+        if project_code:
+            units_qs = units_qs.filter(
+                store__translation_project__project__code=project_code)
+        elif not user.is_superuser:
+            units_qs = units_qs.filter(
+                store__translation_project__project__code__in=user_projects)
+
+        if not (dir_path or filename):
+            return units_qs
+
+        pootle_path = "/%s/%s/%s%s" % (
+            language_code or LANGUAGE_REGEX,
+            project_code or PROJECT_REGEX,
+            dir_path or "",
+            filename or "")
+        if language_code and project_code:
+            if filename:
+                return units_qs.filter(
+                    store__pootle_path=pootle_path)
+            else:
+                return units_qs.filter(
+                    store__pootle_path__startswith=pootle_path)
+        else:
+            # we need to use a regex in this case as lang or proj are not
+            # set
+            if filename:
+                pootle_path = "%s$" % pootle_path
+            return units_qs.filter(
+                store__pootle_path__regex=pootle_path)
 
 
 class Unit(models.Model, base.TranslationUnit):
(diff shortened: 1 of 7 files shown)


unho:
lgtm

unho:
Shouldn't both commits be squashed? Is it really necessary to alter the order of the parameters?

phlax:
> Is it really necessary to alter the order of the parameters?

this method is useful for filtering out the template dirs for a user without any other filter - so it seems to make sense to switch the args and make the path optional.

phlax:
> Shouldn't both commits be squashed?

maybe - but i expected some conversation about the second commit - so kept separate

pootle/apps/pootle_store/models.py
```diff
@@ -235,48 +235,27 @@ def get_for_user(self, user):
...
-                units_path = ''.join(['/%/', proj, '/', filename])
-            else:
-                units_path = ''.join(['/%/', proj, '/%'])
-        # /projects/translate/*
-        elif lang is None and proj is None:
-            units_path = '/%'
-        # /<lang_code>/<project_code>/translate/*
-        # /<lang_code>/translate/*
-        else:
-            units_path = ''.join([pootle_path, '%'])
```
unho: How do we get these by using the regex?

phlax: atm we dont - ill add a WIP header

pootle/core/views.py
```diff
@@ -49,6 +49,102 @@
...
+    @property
+    def filename(self):
+        return self.kwargs.get("filename", "") or ""
+
+    @property
+    def language_code(self):
+        return self.kwargs.get("language_code", None) or None
+
+    @cached_property
+    def path(self):
```
dwaynebailey: Codacy Issue found: [An attribute defined in pootle.core.views line 108 hides this method](https://www.codacy.com/app/dwayne/pootle/file/1742333681/issues/source?bid=1990595&fileBranchId=3039950#l134)

phlax: at last! some useful output from codacy 8/

pootle/apps/pootle_store/models.py
unho: Is this sorted out now?

phlax: yep - i believe so

pootle/core/views.py
```diff
@@ -49,6 +50,115 @@
...
+    project_regex = r"[a-zA-Z0-9\-\_]*"
+
+    @property
+    def regex(self):
+        if not (self.project_code or self.language_code):
+            return None
+        regex = r"^/"
+
+        if self.language_code:
+            regex = r"%s%s/" % (regex, self.language_code)
```
unho: I find it simpler to understand like this:

```python
            regex += r"%s/" % self.language_code
```

phlax: not sure  i agree

pootle/core/views.py
```diff
@@ -49,6 +50,115 @@
...
+        return self.kwargs.get("language_code", None) or None
+
+    @cached_property
+    def path(self):
+        if not self.request_path:
+            return self.request_path
+        path = self._extracted_vfolder_and_path[1]
+        parts = path.split("/")
+        if "." in parts[-1]:
+            return "/%s" % path.strip("/")
```
unho: Simpler:

```python
            return path.rstrip("/")
```

pootle/core/views.py
unho: Perhaps it is less explicit.

pootle/core/views.py
phlax: that doesnt guarantee an initial slash - im not sure if it matters tho

unho: All paths have an initial slash. But yes, it is safer not to assume that.

pootle/apps/pootle_store/models.py
unho: I don't see any test for translation paths (paths with `/translate/` on them).

julen:
Unless I'm doing something wrong, testing this on InnoDB shows a massive performance degradation.

  | *master* | *pr/4317*
---- | ----- | -----
/ab/cd/export-view/?search=notebook&sfields=notes,locations |  12 queries run, total 0.125 seconds | 11 queries run, total 13.01 seconds
/ab/cd/export-view/?filter=incomplete | 16 queries run, total 0.191 seconds | 11 queries run, total 6.777 seconds
/ab/export-view/?filter=incomplete | 13 queries run, total 0.262 seconds | 11 queries run, total 6.861 seconds

Here */ab/cd/* is a TP with ~13000 words, and */ab/* contains ~16000 words. The degradation is noticeable in other views too, so this is not limited to the export view.

It might be the case that queries involving regexps do not use an index at all and they have to scan the entire table.

phlax:
@julen i would expect a signifcant performance hit from moving to regexes.

I am not however seeing anything in the order that you are seeing above.

After repeated testing on our staging server against a lang with ~150000 live units the speed goes from ~.6 second to ~.8 second for the relevant regex query

phlax:
...it could be that you are hitting resource limits in your env and this is forcing sorting etc on the disk, otherwise im not really understanding why there would be such a performance hit.

phlax:
@julen would you be able to slowlog your mysql server to find out which queries are taking a long time

julen:
This is in my local machine with far enough resources to handle this operation.

The slow log follows for accessing */ab/cd/export-view/?filter=incomplete*:
```sql
# Query_time: 2.678598  Lock_time: 0.003926 Rows_sent: 1  Rows_examined: 3200834
SELECT COUNT(*) AS `__count`
FROM `pootle_store_unit`
INNER JOIN `pootle_store_store` ON (`pootle_store_unit`.`store_id` = `pootle_store_store`.`id`)
WHERE (`pootle_store_unit`.`state` > -100
       AND NOT (`pootle_store_store`.`pootle_path` LIKE BINARY '/templates/%')
       AND `pootle_store_store`.`pootle_path` REGEXP BINARY '^/ab/cd/'
       AND (`pootle_store_unit`.`state` = 0
            OR `pootle_store_unit`.`state` = 50));

# Query_time: 4.785047  Lock_time: 0.000177 Rows_sent: 1732  Rows_examined: 3200834
SELECT `pootle_store_unit`.`id`,
       `pootle_store_unit`.`store_id`,
       `pootle_store_unit`.`index`,
       `pootle_store_unit`.`unitid`,
       `pootle_store_unit`.`unitid_hash`,
       `pootle_store_unit`.`source_f`,
       `pootle_store_unit`.`source_hash`,
       `pootle_store_unit`.`source_wordcount`,
       `pootle_store_unit`.`source_length`,
       `pootle_store_unit`.`target_f`,
       `pootle_store_unit`.`target_wordcount`,
       `pootle_store_unit`.`target_length`,
       `pootle_store_unit`.`developer_comment`,
       `pootle_store_unit`.`translator_comment`,
       `pootle_store_unit`.`locations`,
       `pootle_store_unit`.`context`,
       `pootle_store_unit`.`state`,
       `pootle_store_unit`.`revision`,
       `pootle_store_unit`.`creation_time`,
       `pootle_store_unit`.`mtime`,
       `pootle_store_unit`.`submitted_by_id`,
       `pootle_store_unit`.`submitted_on`,
       `pootle_store_unit`.`commented_by_id`,
       `pootle_store_unit`.`commented_on`,
       `pootle_store_unit`.`reviewed_by_id`,
       `pootle_store_unit`.`reviewed_on`,
       `pootle_store_unit`.`priority`,
       `pootle_store_store`.`id`,
       `pootle_store_store`.`file`,
       `pootle_store_store`.`parent_id`,
       `pootle_store_store`.`translation_project_id`,
       `pootle_store_store`.`pootle_path`,
       `pootle_store_store`.`name`,
       `pootle_store_store`.`file_mtime`,
       `pootle_store_store`.`state`,
       `pootle_store_store`.`creation_time`,
       `pootle_store_store`.`last_sync_revision`,
       `pootle_store_store`.`obsolete`
FROM `pootle_store_unit`
INNER JOIN `pootle_store_store` ON (`pootle_store_unit`.`store_id` = `pootle_store_store`.`id`)
WHERE (`pootle_store_unit`.`state` > -100
       AND NOT (`pootle_store_store`.`pootle_path` LIKE BINARY '/templates/%')
       AND `pootle_store_store`.`pootle_path` REGEXP BINARY '^/ab/cd/'
       AND (`pootle_store_unit`.`state` = 0
            OR `pootle_store_unit`.`state` = 50));
```

Note the `BINARY` keyword is being used because we have `utf8_bin` collation (cf. #3659). That doesn't affect performance though — and yeah, I checked temporarily switching to `utf8_general_ci` just in case.

phlax:
@julen so im minded that we can probably get rid of the regexes altogether by not searching by path, we already have the parsed lang/proj/dir_path/filename - so there isnt really that much point to reassembling them in a regex/path.

however, im really keen to find out whats causing such significant differences in our 2 test environments.

unho:
lgtm

ta2-1:
@phlax, tested on translate and export views, it's working not worse. lgtm.

julen:
Wonderful, this now works great. Well done @phlax! :+1:

PR #4322, opened 2016-01-03
Add view tests

phlax:


tests/views/tp.py
```diff
@@ -0,0 +1,218 @@
+# -*- coding: utf-8 -*-
```
ta2-1: We are using `translation_project` instead of `tp` in module names. As I understood a short version was intended here.

ta2-1:
@phlax, I was going to say that tests are great and it's not worth to check everything and the main thing is that tests pass... But they don't now.

phlax:
> @phlax, I was going to say that tests are great and it's not worth to check everything and the main thing is that tests pass... But they don't now.

sure i was trying to fix the calculate_checks on langs issue that only happens on travis

i can remove it if you are gtm re the code - all else passes reliably afaict

ta2-1:
lgtm, since tests will pass again

PR #4325, opened 2016-01-05
Dont initialize children in treeitems if not required

phlax:


pootle/core/mixins/treeitem.py
```diff
@@ -221,7 +221,8 @@ def get_critical_url(self, **kwargs):
 
     def get_stats(self, include_children=True):
         """get stats for self and - optionally - for children"""
-        self.initialize_children()
+        if include_children:
```
ta2-1: @phlax, children are necessary here in any case. This is un-cached TreeItem and any `._calc()` method requires children.

phlax: @ta2-1 - ok ill remove this line

phlax: there was a mixup  - this line is good  - other line removed

ta2-1:
gtm

PR #4339, opened 2016-01-07
Add test for extract_vfolder_from_path function

unho:
Followup for #4316.

### DIFF
--- a/tests/virtualfolder_helpers.py
+++ b/tests/virtualfolder_helpers.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from pootle_translationproject.models import TranslationProject
+from virtualfolder.helpers import extract_vfolder_from_path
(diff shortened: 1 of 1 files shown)


tests/virtualfolder_helpers.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
```
phlax: this line is unnecessary

unho: Perhaps, but I just copied this from other file in tests. I checked 5 of them randomly and they all have this line. Doesn't harm anyway.

tests/virtualfolder_helpers.py
```diff
@@ -0,0 +1,66 @@
...
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from __future__ import absolute_import
```
phlax: i dont think this is necessary either

phlax:
nits aside - lgtm - nice one!

tests/virtualfolder_helpers.py
phlax: yep - but no need to copy out bad code

phlax:
lgtm

PR #4340, opened 2016-01-07
Dango 1.8

### LINKED ISSUE #4233: Upgrade Django dependency to latest stable

dwaynebailey:
Migrating to Django 1.8.  End goal is 1.9 but one step at a time.

Notes:
- There are some commits that need to be squashed
- Management commands aren't extensively tested
- Migrations need review from someone more savvy
- There might be issues in TEMPLATES, especially live vs dev, and expected behaviour currently.
- ContextType changes need someone else's eyes.

Fixes #4233

### DIFF
--- a/pootle_pytest/fixtures/models/permission.py
+++ b/pootle_pytest/fixtures/models/permission.py
@@ -20,7 +20,6 @@ def pootle_content_type(db):
         'model': 'directory',
     }
     content_type, created = ContentType.objects.get_or_create(**args)
-    content_type.name = 'pootle'
     content_type.save()
 
     return content_type
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@
 sudo: false
 language: python
 env:
-  - TOXENV=py27-django17-sqlite
-  - TOXENV=py27-django17-mysql
-  - TOXENV=py27-django17-postgres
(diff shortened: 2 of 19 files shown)


pootle_pytest/fixtures/models/permission.py
```diff
@@ -20,7 +20,6 @@ def pootle_content_type(db):
         'model': 'directory',
     }
     content_type, created = ContentType.objects.get_or_create(**args)
-    content_type.name = 'pootle'
```
unho: Just wondering why this change is not on the "Django 1.8: ContentType name field removed" commit which provides a very detailed explanation.

unho:
> - Management commands aren't extensively tested

Given the changes introduced I don't think that is necessary.

> - Migrations need review from someone more savvy

Migrations look good. Were those automatically created?

> - There might be issues in TEMPLATES, especially live vs dev, and expected behaviour currently.
> - ContextType changes need someone else's eyes.

The ContextType changes are OK. The usage of name comes from some very ancient code that hasn't being touched to avoid breaking things. We can dive in the commit history for the actual reason, but I am not sure that is really necessary.

unho:
Nits aside, lgtm. Some of the commits can be directly merged (for example the `load url from future` one).

PR #4342, opened 2016-01-08
Django 1.8: 'django_language' is no longer read

julen:
At EN we don't use I18N and I didn't explicitly test it, however I believe this is all what's needed.

Refs. https://docs.djangoproject.com/en/1.8/releases/1.8/#features-removed-in-1-8

> The session key `django_language` is no longer read for backwards compatibility.


unho:
lgtm

PR #4344, opened 2016-01-08
Add gitter/travis webhook

phlax:


dwaynebailey:
This will cause noise from private repos won't it?

phlax:
This will cause noise from private repos won't it? <-- the same as currently in IRC

PR #4349, opened 2016-01-08
Django 1.9

dwaynebailey:
Upgrading to Django 1.9.  A few of these commits are really Django 1.8 deprecations that get noisy in 1.9.

Some comments
- SessionAuthenticationMiddleware - will become default in 1.10, but has some implications for users.  The recommendation is to run on 1.7 for a while then and then enable. So really impacts migrating users the most.  Opinions on how to handle this welcome.
- Django transaction hooks might need a check, but then we haven't released a version with this
- contact-form - uses unreleased git version

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -12,7 +12,6 @@ django-redis==4.4.4
 django-rq>=0.8.0,<=0.9.2
 django-sortedm2m==1.3.2
 django-statici18n==1.2.1
-django-transaction-hooks==0.2
 
 # Libraries
 cssmin==0.2.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,6 +1,6 @@
 # Base requirements
 
-Django>=1.8.13,<1.9  # rq.filter: >=1.8,<1.9
+Django>=1.9.9,<1.10  # rq.filter: >=1.9,<1.10
 
 # Django apps
 django-allauth==0.27.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -9,7 +9,7 @@ django-contact-form==1.2
 django-contrib-comments==1.7.3
 django-overextends==0.4.2
 django-redis==4.4.4
-django-rq>=0.8.0,<=0.9.2
+django-rq>=0.9.0,<=0.9.2
 django-sortedm2m==1.3.2
 django-statici18n==1.2.1
 
@@ -23,7 +23,7 @@ lxml>=3.5,<=3.6.4
 python-dateutil==2.5.3
 python-levenshtein==0.12.0
 pytz==2016.6.1
-rq>=0.5.0,<=0.6.0
+rq>=0.5.5,<=0.6.0
 scandir==1.3
 
 # Translate Toolkit
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@
 sudo: false
 language: python
 env:
-  - TOXENV=py27-django18-sqlite PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
-  - TOXENV=py27-django18-mysql PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
-  - TOXENV=py27-django18-postgres PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
+  - TOXENV=py27-django19-sqlite PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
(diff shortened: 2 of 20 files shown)


dwaynebailey:
The following are know deprecation warnings, help to fix appreciates:

* pootle/apps/pootle_store/fields.py:82: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use `Field.from_db_value` instead.
  class MultiStringField(models.Field):
* pootle/apps/staticpages/views.py:110: RemovedInDjango110Warning: `staticpages.views.PageCreateView.get_form` method must define a default value for its `form_class` argument.

requirements/base.txt
```diff
@@ -1,13 +1,13 @@
-Django>=1.8.8,<1.9
+Django>=1.9.1
 
 # Django apps
 django-allauth>=0.24.1
 django-assets>=0.11
-django-contact-form==1.0
+#django-contact-form>=1.1
+-e git+https://github.com/ubernostrum/django-contact-form.git#egg=django-contact-form
```
unho: I think 1.1 is available: https://pypi.python.org/pypi/django-contact-form/1.1

dwaynebailey: It is, but we need 1.2

julen:
@dwaynebailey nice you are taking up this one. Do you have any estimated timing on when you would like to merge it? I'd like to take some time to test, review and understand the upgrade changes before it gets merged, so it'd be good to hear your thoughts on it. Thanks!

julen:
Btw, the transaction hooks change needs to be recorded in the changelog as deployments will need to upgrade their settings when upgrading.

dwaynebailey:
@julen I'd like to merge as soon as possible but with these notes/caveats:

* My first step was to get this working and tests passing (DONE)
* Reduce the deprecation warnings (MOSTLY DONE - the most noisy have already landed, argparse migration is done but needs cleanup)
* Some things should likely land on master before the move to Django 1.9.  All deprecation warnings listed previously.
* I don't want to land this with upstream unreleased versions: contact-form, pytest-django have this issue

Note on the changelog: This needs to be created with a review of the changes that have landed.  Same would go for the SessionAuthenticationMiddleware.

dwaynebailey:
@julen these are the [current outstanding deprecation fixes required](https://github.com/translate/pootle/pull/4349#issuecomment-170118982)

unho:
Just wondering if a check for `transaction hooks` being on settings should be added.

julen:
If I'm not mistaken (didn't test), overextends shouldn't be necessary anymore [as its functionality is built-in](https://code.djangoproject.com/ticket/15053).

dwaynebailey:
@julen would love some further checking.  As [overextends](https://github.com/stephenmcd/django-overextends#project-configuration) mentions running it in Django 1.9+.  So maybe there is stil some stuff that isn't provided by Django.  I'd tackle it apart from this PR if we can.

requirements/base.txt
```diff
@@ -6,7 +6,8 @@ Django>=1.9.7
 django-allauth>=0.25.2
 django-assets>=0.11
 django-contact-form>=1.2
-django-contrib-comments>=1.7.1
+#django-contrib-comments>=1.7.1
```
julen: Note you can already depend on `>=1.7.2` 😉.

dwaynebailey: @julen thanks :)

dwaynebailey:
@julen seems like I've found the last piece of the puzzle to get tests passing, so we have here Django 1.9 support.  Issues remaining:

* 08d3a9b - propagate: True - this needs a better solution of a logging review
* a89afca - we're using git version of overextends, which might be addressed as in https://github.com/translate/pootle/pull/4349#issuecomment-181854246

julen:
Note we'll also need an updated version of django-statici18n; as I understand, a new version should be out soon.

dwaynebailey:
> Note we'll also need an updated version of django-statici18n; as I understand, a new version should be out soon.

I think that's more a Django 1.10 need for https://github.com/zyegfryed/django-statici18n/pull/22 right?

ta2-1:
@dwaynebailey 
re 
https://github.com/translate/pootle/commit/08d3a9b63554d2d8d8041de230dfa097e8695061
will it be worth to change `propagate` attribute in test settings (#4474 )?

dwaynebailey:
@translate/core this is all pretty much done and working.  Of course Django 1.9 is past mainstream support anyway and in extended support (See https://www.djangoproject.com/download/)

I think the plan continues to be to stick with 1.8 LTS which has a year overlap with the next LTS which will be 1.11.

My plan is to continue to keep this PR rebased so that we don't regress against 1.9. And to continue to tackle problems in 1.10 and master (1.11).  I'd suggest that we move to 1.11 around 1.11.1/.2

phlax:
ive looked through the commits and dont see any issues.

i think we should move up to django1.9 asap, else we risk introducing further 1.8 > 1.9 issues.

lgtm, nice one @dwaynebailey

unho:
looks ok

dwaynebailey:
Going to land this because:

* After that last rebase things got broken... again.  So maybe we should just be making sure we're on a newer version instead of having to do massive fixes.
* The last two comments are supportive
* We have a lot of change landing now so lets better get this in while we have such churn
* This allows us to focus on 1.10 issues and move those forward without continual rebasing.

Sorry in advance if somehow this messed with any plans you may have had.

PR #4351, opened 2016-01-09
Django 1.8: NoArgsCommand deprecated

dwaynebailey:
Use BaseCommand and handle()

See
https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/#django.core.management.NoArgsCommand


julen:
Nice, GTM.

PR #4352, opened 2016-01-09
Exception.message deprecated in Python 2.6

dwaynebailey:
Need someone else to look at this to see that I'm not doing something silly here.


julen:
In case you wanna be more explicit, you can also opt to use `str(e)`. The changes look good though :+1:

PR #4353, opened 2016-01-09
Remove cherrypy

dwaynebailey:
Be gone fair maiden.  Seems our only user was 'start' and that was easily converted to simply use runserver.

I might actually just remove start and make people run runserver... Thoughts?


docs/server/commands.rst
```diff
@@ -922,6 +891,16 @@ Version Control support has been removed from Pootle and will reappear in a
 later release.
 
 
+.. django-admin:: run_cherrypy
+
+run_cherrypy
+^^^^^^^^^^^^
+
+.. deprecated:: 2.7
```
julen: Shouldn't this be 2.7.3 instead?

julen:
The PR works fine and looks good to me.

PR #4355, opened 2016-01-11
Simplify vfolder extract algorithm

unho:
Lets suppose a 'whatever' vfolder is created in `/af/firefox/` and it includes
a `/af/firefox/browser/chrome/browser/translation.dtd.po` file, so the
following VirtualFolderTreeItems are created for the vfolder:

`/af/firefox/whatever/`
`/af/firefox/whatever/browser/`
`/af/firefox/whatever/browser/chrome/`
`/af/firefox/whatever/browser/chrome/browser/`

Note that for any of these VirtualFolderTreeItems, the vfolder name is always
placed in the same place in the path.

So if a path like `/af/firefox/whatever/browser/chrome/` is provided for
extraction the matched VirtualFolderTreeItems are:

`/af/firefox/whatever/`
`/af/firefox/whatever/browser/`
`/af/firefox/whatever/browser/chrome/`

which all them belong to the same vfolder, and therefore it is unnecessary
for them to be sorted by vfolder priority.

There can be vfolders with the same name and different location, but
their VirtualFolderTreeItems will have the vfolder name in a different place
in their paths, so only VirtualFolderTreeItems for a single vfolder will be
matched.

If there are vfolders with different name in the same location only
VirtualFolderTreeItems for the same vfolder will be matched as well.

### DIFF
--- a/pootle/apps/virtualfolder/helpers.py
+++ b/pootle/apps/virtualfolder/helpers.py
@@ -7,8 +7,6 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-from pootle.core.url_helpers import get_all_pootle_paths
-
 from .models import VirtualFolderTreeItem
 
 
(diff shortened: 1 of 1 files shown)


unho:
@phlax Further simplified the algorithm.

PR #4356, opened 2016-01-11
Upgrade requirements

unho:


julen:
Looks good to me, not sure about the PG failure on Travis though.

phlax:
> not sure about the PG failure on Travis though.

this has happened a couple of times before seemingly randomly, but i thought it had gone away.

i think there is a problem with ordering not being enforced on units in the export view.

unho:
Updated the PR to include a maximum version for RQ.

Re Travis failure that is some error I already have seen in other PRs that comes and goes.

ta2-1:
I reckon that `rq==0.5.5` doesn't affect us. lgtm.

ta2-1:
@unho with #4360 there is no need to change rq version here.

unho:
@ta2-1 Great! Will drop RQ version change then. And will merge the PR.

PR #4363, opened 2016-01-12
Performance optimizations

phlax:
- ~~optimize directory children vfolder assocations~~
- only set permissions once
- select_relates for directory parents in views

### DIFF
--- a/pootle/apps/pootle_translationproject/views.py
+++ b/pootle/apps/pootle_translationproject/views.py
@@ -25,8 +25,7 @@ from pootle.core.decorators import (
 from pootle.core.helpers import (
     get_sidebar_announcements_context)
 from pootle.core.views import (
-    PootleBrowseView, PootleTranslateView, PootleExportView,
-    requires_permission, set_permissions)
+    PootleBrowseView, PootleTranslateView, PootleExportView)
 from pootle_app.models import Directory
 from pootle_app.models.permissions import (
     check_permission, get_matching_permissions)
@@ -117,10 +116,7 @@ class TPMixin(object):
     """
 
     @redirect_to_tp_on_404
-    @set_permissions
-    @requires_permission("view")
     def dispatch(self, request, *args, **kwargs):
-        # get funky with the request 8/
         return super(TPMixin, self).dispatch(request, *args, **kwargs)
 
     @property
--- a/pootle/apps/pootle_translationproject/views.py
+++ b/pootle/apps/pootle_translationproject/views.py
@@ -156,10 +156,19 @@ class TPDirectoryMixin(TPMixin):
     export_url_path = "pootle-tp-export"
     translate_url_path = "pootle-tp-translate"
 
+    @property
+    def object_related(self):
+        tp_prefix = (
+            "parent__" * self.kwargs.get("dir_path", "").count("/"))
+        return [
+            "%stranslationproject" % tp_prefix,
+            "%stranslationproject__language" % tp_prefix,
+            "%stranslationproject__project" % tp_prefix]
+
     @lru_cache()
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_translationproject/views.py
```diff
@@ -160,10 +156,20 @@ class TPDirectoryMixin(TPMixin):
     export_url_path = "pootle-tp-export"
     translate_url_path = "pootle-tp-translate"
 
+    @property
+    def object_related(self):
+        tp_prefix = (
+            "parent__" * self.kwargs.get("dir_path", "").count("/"))
+        return [
+            "%stranslationproject" % tp_prefix,
+            "%stranslationproject" % tp_prefix,
```
unho: Why is this line included twice?

phlax: to be sure?

il remove

unho:
Would be great if you can include some explanation on the "Only set permissions once on the request" commit message.

phlax:
k - ill amend

pootle/core/browser.py
```diff
@@ -180,10 +176,18 @@ def get_children(directory):
...
-                   for child_dir in directory.child_dirs.live().iterator()]
-
-    stores = [make_store_item(child_store)
-              for child_store in directory.child_stores.live().iterator()]
+    directories = []
+    vfti_pks = vftis_for_children(directory).values_list("pk", flat=True)
+    directories = [
+        make_directory_item(
+            child_dir,
+            sort=("priority" if child_dir.pk in vfti_pks else ""))
```
unho: This won't work as expected. `child_dir` is a `Directory` while `vftis_for_children()` return `VirtualFolderTreeItem` objects, so a `child_dir.pk` will never be in `vfti_pks`.

pootle/core/browser.py
```diff
@@ -180,10 +176,18 @@ def get_children(directory):
...
+        for child_dir
+        in directory.child_dirs.live().iterator()]
+
+    stores = [
+        make_store_item(child_store)
+        for child_store in
+        directory.child_stores.live().iterator()]
```
unho: This is the kind of cleanup I prefer to be on a separate commit when I am reviewing some commit that has actual changes.

unho:
lgtm. Haven't tested for performance.

julen:
After testing the before/after of this PR when accessing overview pages under different directory levels, the query count definitely goes down. The query time on average remains similar; sometimes it's slightly worse though (in the magnitude of a few milliseconds).

I might be looking at the wrong place, so @phlax could you help us clarify under which circumstances we should see the optimization kicking in?

phlax:
hi @julen - yep there are 2 optmizations with this code, both relate to TP or descendant views.

the first optimization prevents the permission code from being called 2x.

the second optimization grabs the parent(s) info in the initial query.

I have had consistent results of a speedup of around 1/3

julen:
The second is the one I was checking. So if I understand correctly this only applies to accessing an overview page for the first time, right? The subsequent reads should be served from the LRU cache. Does it also apply to subdirectories?

On a second look I can see the improvement for the root of TPs, for subdirectories I get mixed results, usually the previous code being slightly faster (again, in the magnitude of a few ms). Does this align with your results?

phlax:
the LRU cache is on the lifetime of the view class so it wont be cached the next time you load the page

> On a second look I can see the improvement for the root of TPs, for subdirectories I get mixed results, usually the previous code being slightly faster (again, in the magnitude of a few ms). Does this align with your results?

no - as i said - i saw a consistent 1/3 improvement - although i didnt test the commits in isolation

PR #4367, opened 2016-01-13
Fix: revert to more liberals in url regexes

phlax:


julen:
I assume some paths might have not been matched previously, is this PR addressing it?

phlax:
hi @julen - yep this is an urgent fix for matching url paths

pootle/apps/pootle_translationproject/urls.py
```diff
@@ -25,30 +25,30 @@
...
 
     # Browser
     url(r'^(?P<language_code>[^/]*)/(?P<project_code>[^/]*)/'
-        r'(?P<dir_path>([a-zA-Z0-9\-\_]*/)*)?$',
+        r'(?P<dir_path>([^/]*/)*)?$',
         TPBrowseView.as_view(),
         name='pootle-tp-browse'),
     url(r'^(?P<language_code>[^/]*)/(?P<project_code>[^/]*)/'
-        r'(?P<dir_path>([a-zA-Z0-9\-\_]*/)*)(?P<filename>.*\.*)?$',
+        r'(?P<dir_path>([^/]*/)*)(?P<filename>.*\.*)?$',
```
unho: A filename is mandatory, so I guess we can drop the final `?`.

julen:
Cool thanks. Do you have any specific pattern to test against? It'd also be great if you could add some explanation in the future.

unho:
lgtm, but haven't tested.

@julen It failed when visiting http://staging.translatehouse.org:8007/af/mozilla_lang/

phlax:
yep, without this PR directories with eg spaces or non a-zA-Z throw an error in their parent directories.

apologies for not giving better explanation on commit, i just wanted to land a fix asap.

we need to add tests to reflect the actual possibilities for paths - ie non a-zA-Z0-9

julen:
Can't we just revert to what we had before the views refactoring? (`r'(?P<dir_path>(.*/)*)(?P<filename>.*\.*)?$'`) This known to work just fine unless there's something new I'm missing.

phlax:
> Can't we just revert to what we had before the views refactoring? 

good spot - the PR doesnt work correctly atm - ill update

julen:
@unho I cannot reproduce with your example. Our staging server runs master and it didn't exhibit any abnormal behavior at e.g. https://translate.stage.evernote.com/ru/android_evernote/ — is the port number relevant?

phlax:
@julen port is not relevant.

what is relevant is whether the directory/tp etc has children with non a-zA-Z0-9-_ chars

phlax:
@julen @unho PR updated - could you review again - ta

phlax:
actually - just seen typo - updating...

julen:
Code looks good to me, feel free to merge it. I also checked we don't have any `[^a-zA-Z_\-]` chars in our file/pathnames. What I still couldn't see is the actual issue in action in the provided URL — got a specific file path?

phlax:
sorry - thats because weve been testing other branches on their

unho:
lgtm

PR #4368, opened 2016-01-13
Cleanup url related Django1.8 deprecations

dwaynebailey:
``patterns()`` and string based ``url()`` views are deprecated in Django 1.8.  This updates and fixes those.


pootle/apps/pootle_language/urls.py
```diff
@@ -7,14 +7,14 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-from django.conf.urls import patterns, url
+from django.conf.urls import url
 
 from .views import (
     LanguageBrowseView, LanguageExportView, LanguageTranslateView)
+from . import views
```
julen: For consistency and to avoid a mixture, it'd be preferable to import from the `views` module when it's already done like that (check one line above). Could you please change this here and in other places where it applies?

julen:
Tested, all fine here :+1: GTM with the nit addressed.

unho:
lgtm despite I would prefer to explicitly import all the view callables from `.views` (or the one that applies).

PR #4372, opened 2016-01-14
Fix: allow admin users to view hidden project resources

phlax:


unho:
To my understanding the hidden resources are the obsolete stores and directories. I don't think those should be displayed even to admins.

phlax:
@unho - hmm - in that case we need to work out why they are being shown in the menu i reckon

ta2-1:
@unho is right. we have only disabled (hidden) projects that should be shown to admin users.

phlax:
PR withdrawn as better fix is to restrict what is shown in menu

PR #4377, opened 2016-01-14
Fix: menu items contained paths from incorrect Stores and obsolete

### LINKED ISSUE #4375: Menus are showing non-live objects in drop-downs

phlax:
- Project.resources was taking over a minute to run in our test env
  - reduced to ~.6 seconds
- Project.resources LIKE query was incorrect and contained paths
  other projects
- Project.resources was not filtering obsolete paths

Fixes #4375

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -19,7 +19,7 @@ from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
-from django.db import connection, models
+from django.db import models
 from django.db.models import Q
 from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
@@ -30,11 +30,12 @@ from django.utils.translation import ugettext_lazy as _
 from pootle.core.cache import make_method_key
 from pootle.core.mixins import CachedTreeItem
 from pootle.core.models import VirtualResource
-from pootle.core.url_helpers import (get_editor_filter, get_path_sortkey,
-                                     split_pootle_path, to_tp_relative_path)
+from pootle.core.url_helpers import (
+    get_editor_filter, get_path_sortkey, split_pootle_path)
 from pootle_app.models.directory import Directory
 from pootle_app.models.permissions import PermissionSet
 from pootle_store.filetypes import factory_classes
+from pootle_store.models import Store
 from pootle_store.util import absolute_real_path
 from staticpages.models import StaticPage
 
@@ -282,21 +283,6 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         """Returns ``True`` if this project is a terminology project."""
         return self.checkstyle == 'terminology'
 
-    @property
-    def vfolders(self):
-        """Return the public virtual folders for this project."""
-        if 'virtualfolder' in settings.INSTALLED_APPS:
-            # This import must be here to avoid circular import issues.
-            from virtualfolder.models import VirtualFolder
-
-            return [vf.tp_relative_path
-                    for vf in VirtualFolder.objects.filter(
-                        units__store__translation_project__project__code=self.code,
-                        is_public=True
-                    ).distinct()]
-
-        return []
-
     @cached_property
     def languages(self):
         """Returns a list of active :cls:`~pootle_languages.models.Language`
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_project/models.py
```diff
@@ -30,11 +30,12 @@
 from pootle.core.cache import make_method_key
 from pootle.core.mixins import CachedTreeItem
 from pootle.core.models import VirtualResource
-from pootle.core.url_helpers import (get_editor_filter, get_path_sortkey,
-                                     split_pootle_path, to_tp_relative_path)
```
unho: `to_tp_relative_path` seems to be unused now, isn't it?

unho: I mean, is it not dead code now?

pootle/apps/pootle_project/models.py
```diff
@@ -314,36 +300,33 @@ def resources(self):
...
+                vfolder__is_public=True,
+                pootle_path__regex="^/[^/].*/%s/" % self.code)
+            if 'virtualfolder' in settings.INSTALLED_APPS
+            else [])
+        resources = sorted(
+            {"/".join(x.split("/")[3:])
+             for x
+             in (set(stores.values_list("pootle_path", flat=True))
+                 | set(dirs.values_list("pootle_path", flat=True))
+                 | set(vftis.values_list("pootle_path", flat=True)
```
unho: Can you call `values_list` on a list?

pootle/apps/pootle_project/models.py
```diff
@@ -314,36 +300,33 @@ def resources(self):
         :cls:`~pootle_store.models.Store` resource paths available for
         this :cls:`~pootle_project.models.Project` across all languages.
         """
-        cache_key = make_method_key(self, 'resources', self.code)
+        from virtualfolder.models import VirtualFolderTreeItem
```
unho: Is it possible to import for an app that is not installed?

pootle/apps/pootle_project/models.py
```diff
@@ -314,36 +300,33 @@ def resources(self):
...
-                           key=get_path_sortkey)
-
+        stores = Store.objects.live().order_by().filter(
+            translation_project__project__pk=self.pk)
+        dirs = Directory.objects.live().order_by().filter(
+            pootle_path__regex=r"^/[^/].*/%s/" % self.code)
+        vftis = (
+            VirtualFolderTreeItem.objects.filter(
+                vfolder__is_public=True,
+                pootle_path__regex="^/[^/].*/%s/" % self.code)
```
unho: Shouldn't be a raw string?

pootle/apps/pootle_project/models.py
```diff
@@ -314,36 +300,33 @@ def resources(self):
...
-
-        # Calculate TP-relative paths and sort them
-        resources_set = {to_tp_relative_path(result[0]) for result in results}
-        resources = sorted(resources_set | set(self.vfolders),
-                           key=get_path_sortkey)
-
+        stores = Store.objects.live().order_by().filter(
+            translation_project__project__pk=self.pk)
+        dirs = Directory.objects.live().order_by().filter(
+            pootle_path__regex=r"^/[^/].*/%s/" % self.code)
```
unho: What is the purpose of the dot in the regex?

pootle/apps/pootle_project/models.py
phlax: yep, its installed in the python sense

pootle/apps/pootle_project/models.py
phlax: ill change, altho im not 100% about the difference

pootle/apps/pootle_project/models.py
phlax: yep, i reckon its superfluous

pootle/apps/pootle_project/models.py
phlax: quite possibly, im going to add a PR very shortly that removes a few unused functions.

i would prefer in this PR just to get this change through

pootle/apps/pootle_project/models.py
unho: The difference is that GitHub highlights the regex special chars.

pootle/apps/pootle_project/models.py
unho: I would prefer the code to be dropped as soon it is dead, but I am ok with leaving this for other PR.

unho:
lgtm. Haven't tested.

pootle/apps/pootle_project/models.py
julen: It's also used in the `Directory` model, and it's more explicit than manually doing splits. I'll bring it back here.

phlax: @julen please dont

it really isnt more explicit - i can read the split code straight away without having to go and check what that "helper" does

julen: I have to disagree here, anyone that reads the method/docstring will have a clue of what it's supposed to do versus trying to guess the meaning of the split. The helper is meant to hide the *how* and show the *what*, i.e show the intention explicitly.

Remember not everyone is familiar with the code :wink: — that's IMHO the reason why this helps and the motivation not to remove it.

phlax: well from debugging this code, right now, im really failing to see the benefit of adding extra code, to do something that python does perfectly well.

julen: FWIW it has to be said there's no extra code being added: for small as it is, it's [reusing it](https://github.com/translate/pootle/blob/e62a1eff750d818f0c2071effe0cd2ed3462b79b/pootle/apps/pootle_app/models/directory.py#L93) instead of duplicating it. Again, I consider important to denote the _intentionality_, especially for anyone who is not familiar with the code, which is not quite clear by reading the `split` IMHO.

julen:
The other day I forgot to mention that this will need a upgrade step for existing deployments either as a data migration or instructions as part of the release notes. In the past at EN we have run this snippet to clear up this cache manually when deploying a new release which included changes affecting the resources drop-down:
```python
from pootle.core.cache import get_cache, make_method_key
cache = get_cache()
cache.delete_pattern(make_method_key('Project', 'resources', '*'))
```

HTH.

PR #4378, opened 2016-01-14
Remove 'start' command, use 'runserver' instead

dwaynebailey:
It doesn't offer any advantage over directly using runserver anymore.


docs/server/commands.rst
```diff
@@ -901,6 +891,16 @@ run_cherrypy
 Run the CherryPy server bundled with the Translate Toolkit.
 
 
+.. django-admin:: start
+
+start
+^^^^^
+
+.. deprecated:: 2.7.3
```
phlax: i think this should be "removed" or similar, if its deprecated its still there for now

unho: There is no "removed". So I am wondering if this should be just removed from docs.

unho:
lgtm once nits addressed.

docs/server/commands.rst
phlax: +1 !

dwaynebailey: 'removed' doesn't exist.  So I added a custom directive to do just that.

dwaynebailey: No they shouldn't just be removed from docs.  We're not doing any deprecation phase and we don't want to carry this around with us following some deprecation strategy.  But we do want to communicate to users that the command is gone and what they should do without it.  That requires it in the docs.  Deprecated works TBH, but you got a new directive

And if any of you don't make use of it may your PRs itch

unho: I still think that if it is gone, it is gone. So don't keep any unnecessary documentation for it.

dwaynebailey: @unho reread my last comment please

Issue #4376, opened 2016-01-14
Do not display unnecessary permissions


unho:
When assigning permissions for users several unnecessary permissions are listed:

- `Can add directory`
- `Can change directory`
- `Can delete directory`

This happens for TP, language and project, and probably in all the permission views.

Expected behavior is not to list those in order to do not overwhelm with long permission choices.

julen:
FTR, these are also displayed in the site-wide permissions view.

I recall I got rid of these in the past with a data migration, but they were being re-created when running migrations because there was no way to omit the default permissions for a model. Now it seems there's a way to prevent that by specifying [`default_permissions`](https://docs.djangoproject.com/en/1.8/ref/models/options/#django.db.models.Options.permissions) in the model's `Meta` class.

The fix proposed in #4379 works fine in my testing @unho. Taken into account the possibility mentioned above though, I believe it'd be cleaner not to manually exclude permissions from being displayed (it becomes harder to maintain) but rather prevent them from being created. It's probably worth giving a shot to this approach (if you haven't already).

unho:
@julen Checked the docs you pointed to and looks like a good way to deal with this, if we include some datamigration to get rid of the already existing Directory permissions. Will tweak the PR in that way for all of you to review (easier to criticize and view the flaws of something specific than some non-concrete idea).

PR #4379, opened 2016-01-14
Do not display directory permissions

unho:


unho:
@julen Tweaked the PR to do it the way you suggested on https://github.com/translate/pootle/issues/4376#issuecomment-171968217. Good to merge?

unho:
Tested that adding a new project with two languages and several subdirectories and files works fine.

julen:
Nice one @unho, works as expected! GTM :+1:

PR #4380, opened 2016-01-14
[WIP] Migrate optparse to argparse

dwaynebailey:
optparse is deprecated in Django and replaces with argparse.

The change is quite invasive.
* We need to move all make_options to add_argument
* ``*args`` is no longer
* opton['opt'] is the correct way to get option settings
* Various argparse related fixes can now happen


dwaynebailey:
Closing this in favour of landing these fixes in a cleaner fashion, and one by one.

PR #4387, opened 2016-01-16
Sync optimizations

phlax:
Optimizations to speed up sync_stores

This provides db performance boost of ~30%- at least with force/overwrite or a fresh sync

Timings for pootle sync_stores --force --overwrite --project=firefox --language=af:

OLD: 2804 queries run, total 3.281 seconds
NEW: 1862 queries run, total 2.377 seconds

**can i please emphasize that this is for a single TP**

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -2276,21 +2276,36 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
             mtime = self.get_cached_value(CachedMethods.MTIME)
             if mtime is None or mtime == datetime_min:
                 mtime = timezone.now()
+            user_displayname = None
+            user_email = None
             if user is None:
+                submissions = self.translation_project.submission_set.exclude(
+                    submitter__username="nobody")
                 try:
-                    submit = self.translation_project.submission_set \
-                                 .filter(creation_time=mtime).latest()
-                    if submit.submitter.username != 'nobody':
-                        user = submit.submitter
-                except ObjectDoesNotExist:
+                    username, fullname, user_email = (
+                        submissions.filter(creation_time=mtime).values_list(
+                            "submitter__username",
+                            "submitter__full_name",
+                            "submitter__email").latest())
+                except Submission.DoesNotExist:
                     try:
-                        lastsubmit = self.translation_project.submission_set \
-                                                             .latest()
-                        if lastsubmit.submitter.username != 'nobody':
-                            user = lastsubmit.submitter
-                        mtime = min(lastsubmit.creation_time, mtime)
+                        _mtime, username, fullname, user_email = (
+                            submissions.values_list(
+                                "creation_time",
+                                "submitter__username",
+                                "submitter__full_name",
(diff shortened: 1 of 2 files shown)


julen:
These are the (by no means scientific) results I get after running `sync_stores --project=<foo>` without any arguments (made several runs of each to confirm):

 | 4 languages | 23 languages
---- | ---- | ---- | ----
master | ~57s | ~5m 44s
PR #4387 | ~56s | ~5m 32s
PR #4387 (without header updates)* | ~55s | ~5m 28s

_* At EN we don't have the need to modify headers at all, so I experimented commenting out the lines which call `update_store_header`_

With bigger data sets this could shrink a few more seconds overall. Changes look :+1: to me for merging.

Apart from that, at EN we'd be interested in having a way to flag not processing headers at all; we can open a ticket if everyone is happy with that (cc @iafan).

julen:
As an aside, while reviewing this I realized the `user=None` parameter for `Store.sync` is somehow _dead_ code: there's no code path passing a user as an argument. Worth checking this separately.

iafan:
@julen I don't think it's worth introducing a new mode for the sake of shaving off just a few seconds.

julen:
Right, was just echoing past thoughts (evernote/pootle#149). (In an ideal scenario were the serialization target is exchangeable, this should be feasible without extra complications though).

phlax:
>  @julen I don't think it's worth introducing a new mode for the sake of shaving off just a few seconds.

just to clarify this PR does not introduce any new "mode", just limits the amount of db requests that are made

iafan:
@phlax I have no objections to the PR itself. I was referring to the Julen's experiment with not updating headers. This is not worth doing just for the sake of the small speed improvements Julen reported.

pootle/apps/pootle_store/models.py
```diff
@@ -2272,21 +2276,35 @@ def update_store_header(self, user=None):
...
+                                "submitter__full_name",
+                                "submitter__email").latest())
                     except ObjectDoesNotExist:
                         pass
+                if user_email:
+                    user_displayname = (
+                        fullname.strip()
+                        if fullname.strip()
+                        else username)
+            elif user.is_authenticated:
```
unho: As commented by @julen https://github.com/translate/pootle/pull/4387#issuecomment-172566849 the `user` passed to `store.sync` is dead code, so the `user` passed to `update_store_header` is dead code too, and this branch is never run.

unho:
lgtm once nit addressed.

phlax:
@julen ive added my timings for this PR - can you check if this reflects what you are seeing - ta.

julen:
@phlax did you make further relevant changes?

phlax:
@julen - none that affect the timings iirc

ta2-1:
@phlax, actually I'd split this PR.
- updating of a header; (Probs I don't follow but why do we change dead code?)
- adding SR; (GTM)
- avoid unnecessary query for modified units; (GTM)

ta2-1:
finally. lgtm.

PR #4388, opened 2016-01-16
Rename pootle_pytest to pytest_pootle

julen:
Reasoning: this is a pytest plugin, not a Pootle plugin.

*(As a side-effect this helps reduce the clash when using CLI autocompletion)*


phlax:
lgtm

PR #4389, opened 2016-01-16
Travis: fix `node_modules` cache location

julen:
While there was a `cache` entry for it, it was surprising that every single `npm install` run was downloading all the dependencies. It turns out the configuration wasn't pointing to the source directory where npm is invoked.


dwaynebailey:
Doh.  Niely spotted, lgtm

PR #4393, opened 2016-01-16
Optimization: dont run system checks in worker processes

phlax:


julen:
From my understanding RQ worker is initialized once, then the process is kept alive listening for incoming jobs, and every time it processes a job it doesn't run any system checks. One can verify this by adding some logging in any system check.

Knowing that something is broken when initializing the worker is not bad *per se* — in fact, it's good to know the worker is in good shape to process jobs.

So unless there's some important detail I'm missing, I'm inclined to discard this micro optimization because it can potentially hide important errors (and make one think everything should work properly) at the expense of a minimum speed-up.

phlax:
afaict it runs in every process, but i can check this

phlax:
yep, here is the last entry from the rq logs of a full TP refresh_stats

```
[2016-01-17T13:14:19]   default: pootle.core.mixins.treeitem.update_cache_job(<Project: Firefox 44 (Aurora)>) (25a77e64-dc01-4439-aa65-5dda377b2810)

{u'time': u'0.001', u'sql': u'SET SQL_AUTO_IS_NULL = 0'}

{u'time': u'0.000', u'sql': u'SET SQL_AUTO_IS_NULL = 0'}

{u'time': u'0.042', u'sql': u"SELECT `accounts_user`.`email` FROM `accounts_user` WHERE NOT (`accounts_user`.`username` IN ('default', 'nobody', 'system')) GROUP BY `accounts_user`.`email` HAVING COUNT(`accounts_user`.`email`) > 1 ORDER BY NULL"}

{u'time': u'0.001', u'sql': u"SELECT `accounts_user`.`id`, `accounts_user`.`password`, `accounts_user`.`last_login`, `accounts_user`.`username`, `accounts_user`.`email`, `accounts_user`.`full_name`, `accounts_user`.`is_active`, `accounts_user`.`is_superuser`, `accounts_user`.`date_joined`, `accounts_user`.`unit_rows`, `accounts_user`.`rate`, `accounts_user`.`review_rate`, `accounts_user`.`hourly_rate`, `accounts_user`.`score`, `accounts_user`.`currency`, `accounts_user`.`is_employee`, `accounts_user`.`twitter`, `accounts_user`.`website`, `accounts_user`.`linkedin`, `accounts_user`.`bio` FROM `accounts_user` WHERE `accounts_user`.`username` = 'admin'"}

{u'time': u'0.000', u'sql': u'SET SQL_AUTO_IS_NULL = 0'}

{u'time': u'0.002', u'sql': u'SELECT `pootle_app_translationproject`.`id`, `pootle_app_translationproject`.`language_id`, `pootle_app_translationproject`.`project_id`, `pootle_app_translationproject`.`real_path`, `pootle_app_translationproject`.`directory_id`, `pootle_app_translationproject`.`pootle_path`, `pootle_app_translationproject`.`creation_time` FROM `pootle_app_translationproject` INNER JOIN `pootle_app_directory` ON ( `pootle_app_translationproject`.`directory_id` = `pootle_app_directory`.`id` ) WHERE (`pootle_app_translationproject`.`project_id` = 18 AND `pootle_app_directory`.`obsolete` = 0)'}
6 queries run, total 0.046 seconds
[2016-01-17T13:14:19]   Job OK
[2016-01-17T13:14:19]   Result is kept for 500 seconds
[2016-01-17T13:14:19]
[2016-01-17T13:14:19]   *** Listening on default...
```
This query here is checking for duplicate emails:

```
{u'time': u'0.042', u'sql': u"SELECT `accounts_user`.`email` FROM `accounts_user` WHERE NOT (`accounts_user`.`username` IN ('default', 'nobody', 'system')) GROUP BY `accounts_user`.`email` HAVING COUNT(`accounts_user`.`email`) > 1 ORDER BY NULL"}
```
and is relatively expensive to run.

The same dump from a full refresh_stats with the provided fix, shows only 3 queries and a similar reduction for every other job.

phlax:
...and a reduction in time of ~5-10x for each job - at least in terms of the sql anyway.

phlax:
not to mention the fact that its repeatedly reconnecting to the db

phlax:
...within each job

julen:
If that's the case this is definitely terrible. To some extent this is described in the [_Performance notes_ section](http://python-rq.org/docs/workers/) of the RQ docs, but I was under the impression that [django-rq's `rqworker` wrapper](https://github.com/ui/django-rq/blob/master/django_rq/management/commands/rqworker.py) took care of it and helped mitigate this.

In fact, prior to commenting I added some logging to verify my understanding of how things worked, and I could only see system checks being run whenever `rqworker` was launched, not during job execution. So I'm curious to see what's different and understand where the checks are triggered. Could you maybe share the snippet you used to verify this, please?

As a side note, I assume [our own `AppConfig`](https://github.com/translate/pootle/blob/e62a1eff750d818f0c2071effe0cd2ed3462b79b/pootle/apps/pootle_app/apps.py#L19) might be producing some side-effects which might be worth revisiting.

phlax:
@julen the one thing im not 100% about is whether it is triggered when DEBUG = False

im just testing this now

julen:
As mentioned before, I used simple logging:
```diff
diff --git a/pootle/checks.py b/pootle/checks.py
index bed4f73..92a2930 100644
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -11,6 +11,10 @@ from django.core import checks
 from django.db import OperationalError, ProgrammingError
 from django.utils.translation import ugettext as _

+import logging
+
+logger = logging.getLogger('rq.worker')
+

 # Minimum Translate Toolkit version required for Pootle to run.
 TTK_MINIMUM_REQUIRED_VERSION = (1, 13, 0)
@@ -47,6 +51,7 @@ def _version_to_string(version, significance=None):

 @checks.register()
 def check_duplicate_emails(app_configs=None, **kwargs):
+    logger.info('CHECKING DUPLICATE EMAILS...')
     from accounts.utils import get_duplicate_emails
     errors = []
     try:
```

Also added some `pdb.set_trace()` to check where the code steps through, and jobs do not run system checks (I'm checking with `DEBUG = True`). This is why I'm curious to see differences.

phlax:
hmmm - k - i also tried to debug this further without much success - i think this might be my bad

Im thinking that the connection debugging i have is wrong - ill investigate this further.

julen:
Cool, no problem. In any case, please let us know if you experience this again in a reproducible manner, as it'd be super ineffective to have a bug like this.

PR #4394, opened 2016-01-16
Stats: add select_related where it is necessary

phlax:
This commit should increase performance of refresh_stats and adds
select_related to:


pootle/apps/pootle_app/models/directory.py
```diff
@@ -154,15 +154,17 @@ def can_be_updated(self):
...
         result = []
         if not self.is_projects_root():
-            # FIXME: can we replace this with a quicker path query?
-            result.extend([item for item in
-                           self.child_stores.live().iterator()])
-            result.extend([item for item in self.child_dirs.live().iterator()])
+            result.extend(
+                self.child_stores.live().select_related(
+                    "parent", "translation_project"))
+            result.extend(self.child_dirs.live())
```
ta2-1: @phlax, I'm not sure about this. You call this "optimization" but I can reckon that parents are not necessary when we calculate stats. Am I wrong here?
Can you tell me when we need parent objects for children?

phlax: dont know atm - but i do know it gets called in the stats code somewhere

personally i think we should rethink the stats code so it calculates the list of objects to expire only once

unho: Is it ok to miss the `iterator()` here?

phlax: i believe so - it will be iterated

Issue #4383, opened 2016-01-15
Enforce correct collation (utf8_bin/utf8_general_ci) collation on case-(in)sensitive columns


phlax:
any columns with pootle_paths are basically case-sensitive as they take their paths from FS paths.

if the db collation schema is set to `utf_general_ci` then there is a risk of multiple objects matching the same path.

this has caused a problem in the past on renaming directories - ref #3659 

UPDATE:
---

After some testing it seems that the django __exact/__iexact operators dont work for mysql.

If you set the collation to ``_bin`` its always case-sensitive - __iexact never works.

Conversely if you set the collation to ``_ci`` its always case-insensitive - __exact never works.

My conclusion for this is that we need to also enforce __ci on those columns that need to be CI - eg lang codes

phlax:
to check the collation of a column:

```
select collation_name from information_schema.columns where table_schema="DBNAME" and table_name="TABLENAME" and column_name="COLUMNAME" ;
```

to update a column
```
alter table DBNAME.TABLENAME modify COLUMNAME varchar(255) character set utf8 collate utf8_bin;
```
``varchar(255)`` should be replaced with the column type.

im not sure if subsequent migration operations on the column would unset the collation - this needs to be checked.

julen:
> if the db collation schema is set to `utf_general_ci` then there is a risk of multiple objects matching the same path.

So according to this, is `utf8_bin` the preferred/suggested collation for deployments using MySQL? If that's the case, can Pootle somehow enforce it or at least initialize fresh DBs using such collation?

I guess there are other considerations when choosing a collation (such as sorting of results), but just wanted to double-check this.

phlax:
> So according to this, is utf8_bin the preferred/suggested collation for deployments using MySQL? If that's the case, can Pootle somehow enforce it or at least initialize fresh DBs using such collation?

yep, we can add a migration setting the column appropriately. My only concern is to ensure that subsequent schema migrations dont overwrite this.

> I guess there are other considerations when choosing a collation (such as sorting of results), but just wanted to double-check this

yes. At the very least alphabetical ordering is different between the collations.

ta2-1:
@phlax, can we point here why language code should be case insensitive? That's not obvious here.
As I understood, language codes need to be CS just because we want to be able to redirect from non-canonical codes to canonical ones. So this will allow us to retrieve the right language object from MySQL db.

phlax:
lang codes are CS, urls are not - the PR redirects accordingly

PR #4400, opened 2016-01-18
Ensure fields have correctly case-in/sensitive collation on mysql

phlax:
Ref: #4383


julen:
I guess this also addresses #3659, right?

phlax:
> I guess this also addresses #3659, right?

i would hope so, but i would like to add a test before closing that bug

ta2-1:
Checked. Migrations work. Tests pass. lgtm.

PR #4404, opened 2016-01-20
Move UserCommand to argparse

dwaynebailey:
Migrate the ``UserCommand`` based management commands to argparse.  Includes tests of the command using ``call_command``.

### DIFF
--- a/pootle/apps/accounts/management/commands/purge_user.py
+++ b/pootle/apps/accounts/management/commands/purge_user.py
@@ -14,6 +14,5 @@ class Command(UserCommand):
     help = "Delete user and all related objects"
 
     def handle(self, **options):
-        super(Command, self).handle(**options)
         for user in options['user']:
             self.get_user(user).delete(purge=True)
--- a/pootle/apps/accounts/management/commands/purge_user.py
+++ b/pootle/apps/accounts/management/commands/purge_user.py
@@ -11,9 +11,9 @@ from . import UserCommand
 
 
 class Command(UserCommand):
-    args = "user"
     help = "Delete user and all related objects"
 
-    def handle(self, *args, **kwargs):
-        super(Command, self).handle(*args, **kwargs)
-        self.get_user(args[0]).delete(purge=True)
+    def handle(self, **options):
+        super(Command, self).handle(**options)
+        for user in options['user']:
+            self.get_user(user).delete(purge=True)
--- a/pootle/apps/accounts/management/commands/update_user_email.py
+++ b/pootle/apps/accounts/management/commands/update_user_email.py
@@ -20,13 +20,24 @@ User = get_user_model()
 
 
 class Command(UserCommand):
-    args = "user email"
     help = "Update user email address"
 
-    def handle(self, *args, **kwargs):
-        super(Command, self).handle(*args, **kwargs)
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "user",
+            help="Username of account",
+        )
+        parser.add_argument(
+            "email",
+            help="New email address",
+        )
+
+    def handle(self, **options):
+        super(Command, self).handle(**options)
         try:
-            accounts.utils.update_user_email(self.get_user(args[0]), args[1])
+            accounts.utils.update_user_email(self.get_user(options['user']),
+                                             options['email'])
         except ValidationError as e:
             raise CommandError(e)
-        self.stdout.write("Email updated: %s, %s\n" % args)
+        self.stdout.write("Email updated: %s, %s\n" % (options['user'],
+                                                       options['email']))
--- a/pootle/apps/accounts/management/commands/update_user_email.py
+++ b/pootle/apps/accounts/management/commands/update_user_email.py
@@ -33,7 +33,6 @@ class Command(UserCommand):
         )
 
     def handle(self, **options):
-        super(Command, self).handle(**options)
         try:
             accounts.utils.update_user_email(self.get_user(options['user']),
                                              options['email'])
--- a/tests/commands/merge_user.py
+++ b/tests/commands/merge_user.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+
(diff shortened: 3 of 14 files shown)


pootle/apps/accounts/management/commands/purge_user.py
```diff
@@ -11,9 +11,8 @@
...
 
 class Command(UserCommand):
-    args = "user"
     help = "Delete user and all related objects"
 
-    def handle(self, *args, **kwargs):
-        super(Command, self).handle(*args, **kwargs)
-        self.get_user(args[0]).delete(purge=True)
+    def handle(self, **options):
+        for user in options['user']:
```
julen: Does this now accept multiple username arguments? If so, docs need to be updated accordingly. The same would apply to `verify_user`.

dwaynebailey: @julen yes indeed they did change, ironically to do what I think the code originally intended.  I've adjusted the docs as needed.  ``argparse`` already makes the needed changes for the commands ``--help``.

pootle/apps/accounts/management/commands/update_user_email.py
```diff
@@ -20,13 +20,21 @@
...
     help = "Update user email address"
 
-    def handle(self, *args, **kwargs):
-        super(Command, self).handle(*args, **kwargs)
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "user",
+        )
+        parser.add_argument(
+            "email"
```
unho: Shouldn't these use `nargs=1` ?

tests/commands/purge_user.py
```diff
@@ -0,0 +1,27 @@
+import pytest
```
unho: Missing license.

tests/commands/merge_user.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_merge_user_nofrom(capfd, nobody, member2):
```
unho: Are the `capfd` and `nobody` fixtures necessary?

tests/commands/purge_user.py
```diff
@@ -0,0 +1,27 @@
...
+        call_command('purge_user')
+    assert "too few arguments" in str(e)
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_purge_user(capfd, evil_member):
+    call_command('purge_user', 'evil_member')
+    out, err = capfd.readouterr()
+    assert "User purged: evil_member" in out
```
unho: Perhaps missing test about passing several users to purge?

tests/commands/merge_user.py
```diff
@@ -0,0 +1,46 @@
...
+    with pytest.raises(CommandError) as e:
+        call_command('merge_user', 'not_a_user', 'member2')
+    assert "User not_a_user does not exist" in str(e)
```
unho: Missing tests for not passing enough arguments or no arguments at all to `merge_user`?

tests/commands/update_user_email.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_update_user_email_nouser(capfd):
```
unho: I am pretty sure that for checking the exception is not necessary to use `capfd`.

tests/commands/verify_user.py
```diff
@@ -0,0 +1,55 @@
...
+    call_command('verify_user', 'member_with_email')
+    out, err = capfd.readouterr()
+    assert "User 'member_with_email' has been verified" in out
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_verify_user_all(capfd, member_with_email, member2):
+    member2.email = "member2@this.test"
+    member2.save()
```
unho: Perhaps it is cleaner to add a new fixture for a second user with email.

tests/commands/verify_user.py
```diff
@@ -0,0 +1,55 @@
...
+    call_command('verify_user', '--all')
+    out, err = capfd.readouterr()
+    assert "Verified user 'member_with_email'" in out
+    assert "Verified user 'member2'" in out
```
unho: It would be nice to have a test for several users but not using `--all`.

pootle/apps/accounts/management/commands/purge_user.py
unho: This looks like have been addressed.

unho:
Review finished.

tests/commands/verify_user.py
phlax: im not sure this is necessary as the verify_user functionality is tested - we just need to test the command

pootle/apps/accounts/management/commands/update_user_email.py
dwaynebailey: Technically not needed, but I added it to be explicit.  Also added some ``help=`` texts.

dwaynebailey: Seems adding them messes with expected output in errors, so taking it away. You're still limited to one anyway.

tests/commands/merge_user.py
dwaynebailey: Nope, removed

tests/commands/purge_user.py
dwaynebailey: Added

tests/commands/merge_user.py
dwaynebailey: Added

tests/commands/update_user_email.py
dwaynebailey: Fixed

tests/commands/verify_user.py
dwaynebailey: Maybe overboard, but did it since I wanted to reuse in another test

tests/commands/verify_user.py
dwaynebailey: Yip probably overkill.  But it does test a slightly different aspect of the command.

I really though don't want to go into such depth with the more complex commands.  For this simple reasons; do we want absolutely full test coverage or do we want to move to Django 1.9, can't have both and I'm sure we actually want the later.  I want the former also but not if its going to take days to make these tests.

tests/commands/purge_user.py
dwaynebailey: Added

PR #4408, opened 2016-01-21
Optimize vfolder query on directory context

phlax:
This finds out which subdirs have vfolders in a single db call.

My db timings for this optimization:

/af/firefox/

OLD: 22 queries run, total 0.022 seconds
NEW: 13 queries run, total 0.013 seconds

/af/firefox/browser/

OLD: 20 queries run, total 0.02 seconds
NEW: 12 queries run, total 0.012 seconds

### DIFF
--- a/pootle/apps/pootle_translationproject/views.py
+++ b/pootle/apps/pootle_translationproject/views.py
@@ -19,7 +19,7 @@ from django.utils.lru_cache import lru_cache
 
 from import_export.views import handle_upload_form
 from pootle.core.browser import (
-    get_children, get_parent, get_table_headings)
+    get_parent, get_table_headings, make_directory_item, make_store_item)
 from pootle.core.decorators import (
     get_path_obj, permission_required)
 from pootle.core.helpers import (
@@ -32,7 +32,7 @@ from pootle_app.models.permissions import (
 from pootle_app.views.admin.permissions import admin_permissions as admin_perms
 from pootle_store.models import Store
 from virtualfolder.helpers import (
-    extract_vfolder_from_path, make_vfolder_treeitem_dict)
+    extract_vfolder_from_path, make_vfolder_treeitem_dict, vftis_for_child_dirs)
 from virtualfolder.models import VirtualFolderTreeItem
 
 from .models import TranslationProject
@@ -233,7 +233,25 @@ class TPBrowseView(TPDirectoryMixin, TPBrowseBaseView):
 
     @cached_property
     def items(self):
-        return get_children(self.object)
+        if 'virtualfolder' in settings.INSTALLED_APPS:
+            dirs_with_vfolders = set(
+                vftis_for_child_dirs(self.object).values_list(
+                    "directory__pk", flat=True))
+        else:
+            dirs_with_vfolders = []
+        directories = [
+            make_directory_item(
+                child,
+                **(dict(sort="priority")
+                   if child.pk in dirs_with_vfolders
+                   else {}))
+            for child in self.object.children
+            if isinstance(child, Directory)]
+        stores = [
+            make_store_item(child)
+            for child in self.object.children
+            if isinstance(child, Store)]
+        return directories + stores
 
     def get_upload_widget(self, project):
         ctx = {}
--- a/pootle/apps/virtualfolder/helpers.py
+++ b/pootle/apps/virtualfolder/helpers.py
@@ -7,6 +7,8 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from pootle_app.models import Directory
+
 from .models import VirtualFolderTreeItem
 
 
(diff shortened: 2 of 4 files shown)


pootle/apps/virtualfolder/helpers.py
```diff
@@ -46,3 +46,11 @@ def extract_vfolder_from_path(request_path):
...
     else:
         return vfti.vfolder, vfti.directory.pootle_path
+
+
+def vftis_for_children(directory):
+    """
+    Returns the vfoldertreeitems for a directory's children
+    """
+    child_pks = [child.pk for child in directory.children]
+    return VirtualFolderTreeItem.objects.filter(directory__pk__in=child_pks)
```
unho: A directory can have children (stores and other dirs) but if they are in a `VirtualFolderTreeItem` (or several of them), those vftis will have it directory to be the parent directory for those children. So why not just use `directory.vf_treeitems.all()`?

phlax: > So why not just use directory.vf_treeitems.all()?

that is what this PR replaces

unho: With the code above it seems that you are retrieving the PKs for all children (stores and subdirs) but then retrieving the vftis assuming they are only for directories?

tests/views/tp.py
```diff
@@ -85,14 +86,30 @@ def _test_browse_view(tp, request, response, kwargs):
     if vfolders:
         filters['sort'] = 'priority'
 
+    dirs_with_vfolders = vftis_for_children(ob).values_list(
+        "directory__pk", flat=True)
+    directories = [
+        make_directory_item(
+            child,
+            **(dict(sort="priority")
```
unho: `**({'sort': 'priority'}` looks cleaner to me.

phlax: i think thats a matter of style - its pretty explicit as is

pootle/apps/virtualfolder/helpers.py
```diff
@@ -46,3 +48,15 @@ def extract_vfolder_from_path(request_path):
...
+def vftis_for_children(directory):
+    """
+    Returns the vfoldertreeitems for a directory's children
+    """
+    child_pks = [
+        child.pk
+        for child
+        in directory.children
+        if isinstance(child, Directory)]
+    return VirtualFolderTreeItem.objects.filter(directory__pk__in=child_pks)
```
unho: I would remove in this function any reference to *children* or *child* and use **subdirectories** instead. Just to avoid further confusion.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -233,7 +233,22 @@ class TPBrowseView(TPDirectoryMixin, TPBrowseBaseView):
 
     @cached_property
     def items(self):
-        return get_children(self.object)
+        dirs_with_vfolders = set(
+            vftis_for_children(self.object).values_list(
```
ta2-1: @phlax, do we query even if we don't use virtual folders?

tests/views/tp.py
unho: ok

pootle/apps/pootle_translationproject/views.py
phlax: @ta2-1 good point ill change

pootle/apps/pootle_translationproject/views.py
```diff
@@ -233,7 +233,22 @@ class TPBrowseView(TPDirectoryMixin, TPBrowseBaseView):
 
     @cached_property
     def items(self):
-        return get_children(self.object)
+        dirs_with_vfolders = set(
```
ta2-1: is it worth to create a function or method for it since you you the same code in tests?

pootle/apps/virtualfolder/helpers.py
phlax: i think child is common with treeitem so im gonna change to vftis_for_child_dirs

unho: As long as it is not just `child` or `children` which also comprise stores, works for me.

pootle/apps/pootle_translationproject/views.py
```diff
@@ -233,7 +233,25 @@ class TPBrowseView(TPDirectoryMixin, TPBrowseBaseView):
 
     @cached_property
     def items(self):
-        return get_children(self.object)
+        if 'virtualfolder' not in settings.INSTALLED_APPS:
```
unho: Shouldn't it be just `in`?

phlax: good spot!

tests/views/tp.py
```diff
@@ -85,14 +86,30 @@ def _test_browse_view(tp, request, response, kwargs):
     if vfolders:
         filters['sort'] = 'priority'
 
+    dirs_with_vfolders = vftis_for_child_dirs(ob).values_list(
```
unho: Shouldn't check for 'virtualfolder' in settings.INSTALLED_APPS as well?

ta2-1:
lgtm

unho:
lgtm

PR #4409, opened 2016-01-21
Optimization: use values_list to calculate vfolder paths

phlax:
In my tests this optimization almost halves the db time to run the refresh_stats command on the sync side.

timings:

pootle refresh_stats --project=firefox --language=af

OLD: 564 queries run, total 0.573 seconds
NEW: 320 queries run, total 0.326 seconds

**can i please emphasize that this is for a single TP**

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -39,7 +39,8 @@ from pootle.core.mixins import CachedMethods, CachedTreeItem
 from pootle.core.models import Revision
 from pootle.core.search import SearchBroker
 from pootle.core.storage import PootleFileSystemStorage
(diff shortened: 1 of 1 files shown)


unho:
I am not particularly happy with duplicating code that is already on the VirtualFolderTreeItem model. But I guess it is for the speedup.

phlax:
@unho well im thinking maybe we take it off the model as im not really seeing a way to do this with a values_list there

unho:
That was what I was expecting for you to say. I prefer to do as it is on this PR.

lgtm.

ta2-1:
@phlax I reckon, this is a great optimisation for those who uses virtual folders. I have checked that dirty counts works well. lgtm.

PR #4410, opened 2016-01-21
Optimization: select_related when extracting vfolder from path

phlax:
This makes a small difference, but nevertheless reduces load on db - which is a good thing.

Timings for /af/firefox/translate/user1/browser/#filter=incomplete:

OLD: 10 queries run, total 0.015 seconds
NEW: 8 queries run, total 0.008 seconds

also small reduction in the db time of get_units if vfolder in effect:

timings for /xhr/units/?path=%2Faf%2Ffirefox%2Fuser1%2Fbrowser%2F&initial=true&filter=incomplete:

OLD: 9 queries run, total 0.021 seconds
NEW: 7 queries run, total 0.018 seconds

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -39,6 +39,7 @@ from pootle_misc.forms import make_search_form
 from pootle_misc.util import ajax_required, get_date_interval, to_int
 from pootle_statistics.models import (Submission, SubmissionFields,
                                       SubmissionTypes)
+from virtualfolder.models import VirtualFolderTreeItem
 
 from .decorators import get_unit_context
 from .fields import to_python
@@ -452,7 +453,10 @@ def get_units(request):
     if 'virtualfolder' in settings.INSTALLED_APPS:
         from virtualfolder.helpers import extract_vfolder_from_path
 
-        vfolder, pootle_path = extract_vfolder_from_path(pootle_path)
+        vfolder, pootle_path = extract_vfolder_from_path(
+            pootle_path,
+            vfti=VirtualFolderTreeItem.objects.select_related(
+                "directory", "vfolder"))
 
     path_keys = [
         "project_code", "language_code", "dir_path", "filename"]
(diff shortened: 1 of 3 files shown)


pootle/apps/virtualfolder/helpers.py
```diff
@@ -27,7 +27,7 @@ def make_vfolder_treeitem_dict(vfolder_treeitem):
         'icon': 'vfolder'}
 
 
-def extract_vfolder_from_path(request_path):
+def extract_vfolder_from_path(request_path, vfti=None):
```
unho: Update the docstring below to include the new argument. Don't forget to specify it should be a VirtualFolderTreeItem queryset.

phlax: yep k

pootle/apps/pootle_store/views.py
```diff
@@ -452,7 +453,10 @@ def get_units(request):
     if 'virtualfolder' in settings.INSTALLED_APPS:
         from virtualfolder.helpers import extract_vfolder_from_path
 
-        vfolder, pootle_path = extract_vfolder_from_path(pootle_path)
+        vfolder, pootle_path = extract_vfolder_from_path(
+            pootle_path,
+            vfti=VirtualFolderTreeItem.objects.select_related(
+                "directory", "vfolder"))
```
unho: Just wondering if instead of calling here `select_related` it would be better to pass a boolean flag to `extract_vfolder_from_path` and call `select_related` there.

unho:
Reviewed. There are some nits.

pootle/apps/pootle_store/views.py
phlax: well im thinking that we should make this function as extensible as possible - even though my tests seem to suggest that in these 2 circumstances the optimization is the same.

unho: ok, lets keep it as is now.

ta2-1:
lgtm after addressing nits.

pootle/apps/virtualfolder/helpers.py
unho: This is now fixed.

PR #4412, opened 2016-01-21
Migrate BaseCommand derived commands to argparse

dwaynebailey:
This focuses on all commands that inherit directly from BaseCommand

### DIFF
--- a/tests/commands/import.py
+++ b/tests/commands/import.py
@@ -1,0 +1,107 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+import pytest
+
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+
+@pytest.mark.cmd
+def test_import_user_nofile():
+    """Missing 'file' positional argument."""
+    with pytest.raises(CommandError) as e:
+        call_command('import')
+    assert "too few arguments" in str(e)
+
+
+@pytest.mark.cmd
+def test_import_user_non_existant_file():
+    """No file on filesystem."""
+    with pytest.raises(CommandError) as e:
+        call_command('import', 'nofile.po')
+    assert "No such file" in str(e)
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_import_emptyfile(capfd, afrikaans_tutorial, tmpdir):
+    """Load an empty PO file"""
+    p = tmpdir.mkdir("sub").join("empty.po")
+    p.write("")
+    with pytest.raises(CommandError) as e:
+        call_command('import', os.path.join(p.dirname, p.basename))
+    assert "missing X-Pootle-Path header" in str(e)
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_import_onefile(capfd, afrikaans_tutorial, tmpdir):
+    """Load an one unit PO file"""
+    p = tmpdir.mkdir("sub").join("tutorial.po")
+    p.write("""msgid ""
+msgstr ""
+"X-Pootle-Path: /af/tutorial/tutorial.po\\n"
+"X-Pootle-Revision: 12\\n"
+
+msgid "rest"
+msgstr "test"
+           """)
+    call_command('import', os.path.join(p.dirname, p.basename))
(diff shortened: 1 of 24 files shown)


dwaynebailey:
@translate/core this is ready for review.

- I've kept all command changes in one commit
- Bugs found through testing are next each on their own
- Lastly, each command is tested.  I will likely squash these test commits once reviewed.

tests/commands/webpack.py
```diff
@@ -0,0 +1,18 @@
...
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from django.core.management import call_command
+
+
+@pytest.mark.cmd
+def test_webpack(capfd):
```
julen: In its current shape, this test doesn't add any value, as it's not asserting anything but just running the command.

julen:
The code changes look good to me overall.

tests/commands/webpack.py
dwaynebailey: @julen yes wanted to ping you on this one.  Wasn't too sure how best to test this or if we should test it.

tests/commands/initdb.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.mark.django_db
+def test_initdb_noprojects(capfd):
+    """Initialise the database with initdb
+
+    Testing without --no-projects would take too long
+    """
+    call_command('initdb', '--no-projects')
+    out, err = capfd.readouterr()
+    assert "Successfully populated the database." in out
+    assert "Successfully populated the database." in out
```
unho: Repeated assert.

tests/commands/test_checks.py
```diff
@@ -0,0 +1,88 @@
...
+    with pytest.raises(CommandError) as e:
+        call_command('test_checks', '--unit=100000')
+    assert 'Unit matching query does not exist' in str(e)
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_test_checks_unit(capfd, site_matrix):
+    """Check a --unit"""
+    # Unit 3/104 is /language0/project0/store0.po Translated Target
```
unho: If this comment refers to 3 and 196 below, the numbers don't match.

tests/commands/test_checks.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@pytest.mark.cmd
+def test_test_checks_srctgt_missing_args():
+    """Check a --source --target with incomplete options."""
+    with pytest.raises(CommandError) as e:
+        call_command('test_checks', '--source="%s files')
+    assert 'Use a pair of --source and --target' in str(e)
+    with pytest.raises(CommandError) as e:
+        call_command('test_checks', '--target="%s leers')
```
unho: I think that using the `%s` placeholder in the test string might confuse anyone reading the code. I suggest using just `Files` and `Leers`.

tests/commands/import.py
```diff
@@ -0,0 +1,101 @@
...
+msgstr ""
+"X-Pootle-Path: /af/tutorial/tutorial.po\\n"
+"X-Pootle-Revision: 12\\n"
+
+msgid "rest"
+msgstr "test"
+           """)
+    call_command('import', os.path.join(p.dirname, p.basename))
+    out, err = capfd.readouterr()
+    assert "Updating /af/tutorial/tutorial.po" in err
```
unho: Shouldn't this be in `out`?

tests/commands/import.py
```diff
@@ -0,0 +1,101 @@
...
+msgstr ""
+"X-Pootle-Path: /af/tutorial/tutorial.po\\n"
+"X-Pootle-Revision: 12\\n"
+
+msgid "rest"
+msgstr "test"
+           """)
+    call_command('import', '--user=member', os.path.join(p.dirname, p.basename))
+    out, err = capfd.readouterr()
+    assert "Updating /af/tutorial/tutorial.po" in err
```
unho: Same as in test above.

unho:
Reviewed. lgtm besides Travis failures and a bunch of tests just running commands. Also the tests could be a bit more extensive, but perhaps it is not worth the effort.

tests/commands/import.py
dwaynebailey: Does the test fail?

unho: This was one of the three tests that I saw that failed on Travis.

tests/commands/webpack.py
dwaynebailey: I ended up just droping the webpack test.

PR #4413, opened 2016-01-22
Optimization: filter project object before using regex

phlax:
in /project/ views.

This significantly improves (~40-70% db speedup) the query made to find ProjectResources.

Timings:

/projects/firefox/browser/

NEW: 8 queries run, total 0.094 seconds
OLD: 8 queries run, total 0.138 seconds

/projects/firefox/browser/chrome/browser/customizableui/customizableWidgets.properties.po

NEW: 8 queries run, total 0.286 seconds
OLD: 8 queries run, total 0.722 seconds

### DIFF
--- a/pootle/apps/pootle_project/views.py
+++ b/pootle/apps/pootle_project/views.py
@@ -64,21 +64,24 @@ class ProjectMixin(object):
         if not (self.kwargs["dir_path"] or self.kwargs["filename"]):
             return self.project
 
-        regex = (
-            r"^/.*/%s/%s"
+        project_path = (
+            "/%s/%s%s"
             % (self.project.code,
-               self.kwargs['dir_path']))
-
+               self.kwargs['dir_path'],
+               self.kwargs['filename']))
+        regex = r"^/[^/]*%s$" % project_path
         if not self.kwargs["filename"]:
             dirs = Directory.objects.live().select_related("parent__pootle_path")
             resources = (
-                dirs.filter(pootle_path__regex="%s$" % regex)
-                    .exclude(pootle_path__startswith="/templates"))
+                dirs.exclude(pootle_path__startswith="/templates")
+                    .filter(pootle_path__endswith=project_path)
+                    .filter(pootle_path__regex=regex))
         else:
-            regex = "%s%s$" % (regex, self.kwargs["filename"])
             resources = (
                 Store.objects.live()
                              .select_related("translation_project__language")
+                             .filter(translation_project__project=self.project)
+                             .filter(pootle_path__endswith=project_path)
                              .filter(pootle_path__regex=regex))
         if resources:
             return ProjectResource(


pootle/apps/pootle_project/views.py
```diff
@@ -64,21 +64,24 @@ def get_object(self):
...
-        regex = (
-            r"^/.*/%s/%s"
+        project_path = (
+            "/%s/%s%s"
             % (self.project.code,
-               self.kwargs['dir_path']))
-
+               self.kwargs['dir_path'],
+               self.kwargs['filename']))
+        regex = r"^/[^/]*%s$" % project_path
```
unho: Is the `*` necessary. I think `[^/]` will occur once and only once.

pootle/apps/pootle_project/views.py
```diff
@@ -64,21 +64,24 @@ def get_object(self):
...
 
-        regex = (
-            r"^/.*/%s/%s"
+        project_path = (
+            "/%s/%s%s"
             % (self.project.code,
-               self.kwargs['dir_path']))
-
+               self.kwargs['dir_path'],
+               self.kwargs['filename']))
```
unho: I am bit concerned about what would happen if filename is not provided. Is it `None` or empty string?

phlax: always an empty string - i believe - tests would fail if not

pootle/apps/pootle_project/views.py
phlax: you are probably right

pootle/apps/pootle_project/views.py
unho: k

julen:
Thank you, @phlax, having numbers really helps.

Here are my results for a project with ~2.5M words, in the avg. of 5 runs always starting with an empty cache.

 | /projects/website2_evernote/ | /projects/website2_evernote/business/certified-consultants/inc/right-column.php.po
---- | ---- | ----
master | 18 queries, 0.0682s | 18 queries, 0.1348s
PR/4413 | 18 queries, 0.079s | 18 queries, 0.1034s

There seems to be a small regression when accessing the root of a project.

With the cache already filled, also the avg. of 5 runs:

 | /projects/website2_evernote/ | /projects/website2_evernote/business/certified-consultants/inc/right-column.php.po
---- | ---- | ----
master | 9 queries, 0.0034s | 9 queries, 0.072s
PR/4413 | 9 queries, 0.0028s | 9 queries, 0.0396s

The improvement doesn't look as visible as in your tests, however I assume your difference comes from the fact of having a project almost 4x bigger. I'm curious to know in which conditions you have measured your numbers though.

phlax:
hi @julen im thinking that the benefit here will depend alot on the db.

ie  - if the first filter reduces the dataset a lot then there will be more benefit.

in terms of test infrastructure - im happy to PM you the details/provide access if you ping me on gitter

phlax:
> There seems to be a small regression when accessing the root of a project.

Ive seen similar results in other tests but not in this particular case.

Im not quite sure why this would be my thought is that the regex is the more expensive query and therefore would always be better last, but in a very small minority of cases this doesnt seem to be true.

Mostly though by adding this query pattern (or similar) ive seen quite significant performance improvements.

phlax:
@julen - just realising the point about the root of the /projects/ dir - we probs need to not filter __endwith in this case as its an empty string.

phlax:
@julen i have investigated this further. This commit makes *no* changes to either the sql or python for views of the sort /projects/PROJECT/ - as the method does/did return the project without building any resources if dir_path and filename are empty - https://github.com/phlax/pootle/blob/view_optimizations_regex/pootle/apps/pootle_project/views.py#L64

so any differences in timings on these views are not related to this PR afaict

pootle/apps/pootle_project/views.py
unho: What is the outcome of this?

pootle/apps/pootle_project/views.py
```diff
@@ -64,21 +64,24 @@ def get_object(self):
...
+        regex = r"^/[^/]*%s$" % project_path
         if not self.kwargs["filename"]:
             dirs = Directory.objects.live().select_related("parent__pootle_path")
             resources = (
-                dirs.filter(pootle_path__regex="%s$" % regex)
-                    .exclude(pootle_path__startswith="/templates"))
+                dirs.exclude(pootle_path__startswith="/templates")
+                    .filter(pootle_path__endswith=project_path)
+                    .filter(pootle_path__regex=regex))
```
unho: I read https://thingsilearn.wordpress.com/2008/02/28/mysql-query-speed-regexp-vs-like/ but sadly it seems that we can't get rid of the regex altogether.

pootle/apps/pootle_project/views.py
phlax: works with your improvement

ta2-1: @unho why do you think it will occur only once?

ta2-1:
+1 to https://github.com/translate/pootle/pull/4413#issuecomment-174157751.  There are no changes for `/projects/PROJECT_NAME` pages.
And there are obvious improvements since we check less rows by `pootle_path` via regex. 
lgtm.

pootle/apps/pootle_project/views.py
unho: I wasn't thinking about chars, but about the group for the language code, which can appear only once.

unho:
lgtm

julen:
> This commit makes no changes to either the sql or python for views of the sort /projects/PROJECT/

Thanks for the explanation, @phlax!

PR #4414, opened 2016-01-22
Optimization: move select_relateds to creation of ProjectResource

phlax:
this removes an expensive query on /project/ views for directories

/projects/firefox/browser/

NEW: 6 queries run, total 0.071 seconds
OLD: 7 queries run, total 0.139 seconds

/projects/firefox/browser/chrome/browser/safebrowsing/

NEW: 6 queries run, total 0.052 seconds
OLD: 7 queries run, total 0.096 seconds

### DIFF
--- a/pootle/apps/pootle_project/views.py
+++ b/pootle/apps/pootle_project/views.py
@@ -70,12 +70,18 @@ class ProjectMixin(object):
                self.kwargs['dir_path']))
 
         if not self.kwargs["filename"]:
-            dirs = Directory.objects.live().select_related("parent__pootle_path")
+            dirs = Directory.objects.live()
+            if self.kwargs['dir_path'].count("/"):
+                tp_prefix = "parent__" * self.kwargs['dir_path'].count("/")
+                dirs = dirs.select_related(
+                    "%stranslationproject" % tp_prefix,
+                    "%stranslationproject__language" % tp_prefix)
             resources = (
                 dirs.filter(pootle_path__regex="%s$" % regex)
                     .exclude(pootle_path__startswith="/templates"))
         else:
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_project/views.py
```diff
@@ -70,12 +70,18 @@ def get_object(self):
                self.kwargs['dir_path']))
 
         if not self.kwargs["filename"]:
-            dirs = Directory.objects.live().select_related("parent__pootle_path")
+            dirs = Directory.objects.live()
+            if self.kwargs['dir_path'].count("/"):
+                tp_prefix = "parent__" * self.kwargs['dir_path'].count("/")
```
unho: Instead of this perhaps it is better to include a TP field on Directory.

pootle/apps/pootle_project/views.py
```diff
@@ -70,12 +70,18 @@ def get_object(self):
                self.kwargs['dir_path']))
 
         if not self.kwargs["filename"]:
-            dirs = Directory.objects.live().select_related("parent__pootle_path")
+            dirs = Directory.objects.live()
+            if self.kwargs['dir_path'].count("/"):
```
unho: Perhaps I am wrong, but no class inheriting from `ProjectMixin` actually accepts any call that provides a filename neither a `dir_path`, so is this code actually being run? Is it necessary the `if not self.kwargs["filename"]:` above?

unho:
Reviewed.

pootle/apps/pootle_project/views.py
phlax: @unho yep its used

phlax:
this has moved to #4423

PR #4420, opened 2016-01-23
Convert PootleCommand derived commands to argparse

dwaynebailey:
Next batch for conversions and tests.  This focuses on PootleCommand which drive commands that process Stores and have options like ``--language`` and ``--project``.

### DIFF
--- a/pootle/apps/pootle_app/management/commands/dump.py
+++ b/pootle/apps/pootle_app/management/commands/dump.py
@@ -9,7 +9,6 @@
 
 
 import os
-
 import sys
 reload(sys)
 sys.setdefaultencoding('utf-8')
@@ -20,7 +19,10 @@ from django.core.management.base import CommandError
 
 from pootle_app.management.commands import PootleCommand
 from pootle_app.models import Directory
+from pootle_language.models import Language
 from pootle_project.models import Project
+from pootle_translationproject.models import TranslationProject
+
 
 DUMPED = {
     'TranslationProject': ('pootle_path', 'real_path', 'disabled'),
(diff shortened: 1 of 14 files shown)


dwaynebailey:
@translate/core this is ready for review.

The commits are arranged as follows:

- The main argparse conversion
- Any changes needed for issues found in tests
- The test themselves (these can be squashed if needed)

Coverage is quite good:

- There are some issues with ``dump --data``, I'm not sure my fixes make it do what it was intended to do, but it doesn't break now
- ``update_store`` and ``export`` have relatively low coverage compared to the others.  But it's mostly covering the corner use cases to flesh out coverage.

pootle/apps/pootle_app/management/commands/dump.py
```diff
@@ -42,30 +42,30 @@
...
+        parser.add_argument(
             '--stop-level',
             action='store',
             dest='stop_level',
             default=-1,
             type=int,
-        ),
-    )
-    option_list = PootleCommand.option_list + shared_option_list
+            help="Depth of data to retreive",
```
julen: Typo: _retrieve_

julen:
Nice one @dwaynebailey :+1:

The code switched over to argparse looks good and I checked out the PR branch and commands work fine.

One thing I notice though is there are a few tests with commented code, which I wouldn't be very keen on committing. Unless someone goes and fixes these, how about marking such parts as `xfail`? Otherwise one might get the wrong impression a test already covers a command.

Use your best judgment for the point above; other than that this is GTM from my side.

Issue #4424, opened 2016-01-25
[Regression] AttributeError: 'list' object has no attribute 'values_list'


julen:
In deployments where `virtualfolders` is not installed, submissions break Pootle currently. 
_(For now I have commented out the buggy piece of code in production so translators can work)_

```python
AttributeError: 'list' object has no attribute 'values_list'
  File "django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "pootle_misc/util.py", line 58, in wrapper
    return f(request, *args, **kwargs)
  File "pootle_store/decorators.py", line 66, in decorated_f
    return f(request, unit, *args, **kwargs)
  File "pootle_store/views.py", line 924, in submit
    form.save()
  File "django/forms/models.py", line 459, in save
    construct=False)
  File "django/forms/models.py", line 105, in save_instance
    instance.save()
  File "pootle_store/models.py", line 548, in save
    self.store.update_dirty_cache()
  File "pootle/core/mixins/treeitem.py", line 513, in update_dirty_cache
    self.register_all_dirty()
  File "pootle/core/mixins/treeitem.py", line 471, in register_all_dirty
    for p in self.all_pootle_paths():
  File "pootle_store/models.py", line 2253, in all_pootle_paths
    vftis = self.parent_vfolder_treeitems.values_list(
```

On the one hand, `Store.parent_vfolder_items` needs to return `Store.objects.none()` instead of `[]`. On the other hand, using [`values_list()` with vfolder-related fields](https://github.com/translate/pootle/blob/814f5f0db7df03c47d375d213a2e9f213c35a51e/pootle/apps/pootle_store/models.py#L2253-L2254) will break too.

cc @phlax.

phlax:
@julen i dont think this should be called at all if virtualfolders are not installed - ill PR a fix now

julen:
I agree with that. Still I think the return value of `Store.parent_vfolder_items` should be consistent and return `Store.objects.none()` as mentioned above.

PR #4426, opened 2016-01-25
Fix: all_pootle_paths ignore vfolders if not installed

phlax:
Fixes: #4424


phlax:
@julen can you confirm this fixes the problem

unho:
lgtm

julen:
Thank you for taking care of this @phlax, works fine here. GTM.

PR #4430, opened 2016-01-27
Move update_tmserver to argparse

dwaynebailey:
This one was a little more complicate to migrate and test.  So here on its own.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,3 +51,4 @@ notifications:
     on_start: never
 services:
   - redis-server
+  - elasticsearch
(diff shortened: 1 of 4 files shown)


unho:
I suppose that tests for the command could be more extensive, but given tests longer I think it is fine with these.

I will trust that the change to aggregations works fine.

There is a typo in one of the commit messages "Curent".

unho:
Found https://www.elastic.co/guide/en/elasticsearch/reference/2.x/search-aggregations-metrics-stats-aggregation.html which convinced me that the change to aggregations is fine. As I said, lgtm.

dwaynebailey:
@unho aggregation was actually needed to get it to work on ES in Ubuntu which is older than the version we're running on mlo.  But glad you found that other ref.

PR #4431, opened 2016-01-27
Issue/4406

unho:


docs/server/installation.rst
```diff
@@ -214,7 +214,7 @@ run it, just issue:
 
 .. code-block:: bash
 
-   (env) $ pootle runserver
+   (env) $ pootle runserver --insecure
```
dwaynebailey: Needs a big red warning about running with ``--insecure`` just as is in the Django docs.  Or at least point to Django's concern.

unho: Added a warning with a reference to Django docs.

dwaynebailey: "Never use it __in__ production."

Otherwise GTM

phlax:
lgtm - +1 to nit

PR #4433, opened 2016-01-28
Test --help across all commands

dwaynebailey:
This is to tease out any potential config issues in current and future argparse commands.

### DIFF
--- a/tests/commands/minus_minus_help.py
+++ b/tests/commands/minus_minus_help.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from django.core.management import call_command, get_commands
+
+
+@pytest.mark.cmd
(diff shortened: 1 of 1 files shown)


julen:
I'm not sure I understand the motivation behind this; mind elaborating?

dwaynebailey:
@julen simple really.  This picked up a number of issues with optparse to argparse migration.  But as a default will run the config setup for any commands we land.  While it doesn't exercise the meat of any command it makes sure we're at least trying to run them and keep them working.

tests/commands/minus_minus_help.py
```diff
@@ -0,0 +1,26 @@
...
+import pytest
+
+from django.core.management import call_command, get_commands
+
+
+@pytest.mark.cmd
+@pytest.mark.parametrize("command,app", [
+    (command, app)
+    for command, app in get_commands().iteritems()
+    if not app.startswith("django.") and not app.startswith("django_")
```
unho: Is there any reason not to use just `not app.startswith("django")`?

tests/commands/minus_minus_help.py
```diff
@@ -0,0 +1,26 @@
...
+])
+def test_initdb_help(capfd, command, app):
+    """Catch any simple command issues"""
+    print "Command: %s, App: %s" % (command, app)
```
unho: Nit, use print as a function.

PR #4435, opened 2016-01-29
Django 1.8: enable SessionAuthenticationMiddleware

dwaynebailey:
SessionAuthenticationMiddleware will be enabled by default in Django 1.10 and it provided by default for new projects since Django 1.7

**Note:** this is blocked by wanting an upstream release, see [comment below](https://github.com/translate/pootle/pull/4435#issuecomment-178612506) for details.

See
https://docs.djangoproject.com/en/1.9/releases/1.8/#using-authenticationmiddleware-without-sessionauthenticationmiddleware

And
https://docs.djangoproject.com/en/1.9/topics/auth/default/#session-invalidation-on-password-change

RFC:
* I need to understand how this impacts our auth workflow
* Opinions on how to document/manage this for upgrades

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,7 +3,7 @@
 Django>=1.8.12,<1.9
 
 # Django apps
-django-allauth>=0.24.1
+django-allauth>=0.25.2
 django-assets>=0.11
 django-contact-form==1.0
 django-overextends>=0.4.1


dwaynebailey:
@translate/core would love some input on how best we should handle this one.  The fix is easy, but I'd love some thoughts on how we get this deployed.

unho:
According to my reading of https://docs.djangoproject.com/en/1.9/topics/auth/default/#session-invalidation-on-password-change it seems that we are using a `User` class inheriting from `AbstractBaseUser` https://github.com/translate/pootle/blob/master/pootle/apps/accounts/models.py#L52 and after checking which django-allauth version we are using https://github.com/translate/pootle/blob/master/requirements/base.txt#L4 I looked into django-allauth code and it seems that the required changes are already present on django-allauth https://github.com/pennersr/django-allauth/blob/0.24.1/allauth/account/views.py#L486

So no extra changes are required re code. Regarding deployment since we are taking the site down to upgrade I would just change the `SECRET_KEY` to force all sessions to be invalidated.

unho:
lgtm

julen:
Enabling this is fine, however I'd wait until a new version of allauth is out which contains the fix for pennersr/django-allauth#1262.

@dwaynebailey realize [this middleware won't be needed in 1.10+](https://docs.djangoproject.com/en/1.8/releases/1.8/#using-authenticationmiddleware-without-sessionauthenticationmiddleware), as opposed to the _enabled by default_ you mentioned:
> In Django 1.10, session verification will be enabled regardless of whether or not SessionAuthenticationMiddleware is enabled (at which point SessionAuthenticationMiddleware will have no significance

So I'd change the comment to indicate the middleware can go away once in 1.10.

julen:
> Regarding deployment since we are taking the site down to upgrade I would just change the SECRET_KEY to force all sessions to be invalidated.

I wonder why would you want to force-logout your users when it's not necessary.

dwaynebailey:
Thanks @julen.  Yes my wording was bad.  I guess my query though is that with it enabled by default we need to transition from 'not enabled by default'.  Sounds from the comments that we don't really need to worry about this, though we can point users to the docs if needed where they can read the instructions themselves.

dwaynebailey:
With regards to version of allauth

1262 has landed, it claims to be part of 0.24.0 but it isn't.  Currently [0.24.1 is the latest release](https://github.com/pennersr/django-allauth/releases) released on 9 Nov 2015.  Anything from now 0.25 or 0.24.2 will contain the needed changes.

julen:
Note it doesn't claim to be part of 0.24, but any version newer than 0.24.x. In this regards, version 0.25 was released a few days ago and includes the fix.

unho:
lgtm

phlax:
lgtm

dwaynebailey:
Landed in 36f3350d370ab0dcc0c3aab8c7e917f8ad7c1cc6

PR #4436, opened 2016-01-29
Webpack: add --progress and use --no-color

dwaynebailey:
* Add --progress to allow progress noise to be silenced.  Useful for Travis builds
* Use Django's ``--no-color`` option to control webpack's ``--colors``, though switching off colours doesn't seem to actually work.


unho:
lgtm. Check the Travis failures.

julen:
Could we please bring back `--progress` at least when `--dev` is specified? (I use this a lot).

dwaynebailey:
@julen so --progress to on when you have --dev? Go for it :)

julen:
Done in e449215.

unho:
@julen I don't think that is the commit.

PR #4446, opened 2016-02-03
Travis: Cache assets

unho:
This might provide noticeable speed up of 3 minutes when running Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
 cache:
   directories:
     - pootle/static/js/node_modules
+    - pootle/assets
     - $HOME/.cache/pip
     - docs/_build
     - .tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ env:
 matrix:
   allow_failures:
     - env: TOXENV=isort
+before_cache:
+  - rm -f $HOME/.cache/pip/log/debug.log
 cache:
   directories:
     - pootle/static/js/node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,11 @@
 sudo: false
 language: python
 env:
-  - TOXENV=py27-django18-sqlite
-  - TOXENV=py27-django18-mysql
-  - TOXENV=py27-django18-postgres
+  - TOXENV=py27-django18-sqlite PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
+  - TOXENV=py27-django18-mysql PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
+  - TOXENV=py27-django18-postgres PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
   # Meta
-  - TOXENV=sdist
-  - TOXENV=eslint
-  - TOXENV=docs
-  - TOXENV=mo
-  - TOXENV=isort
-  - TOXENV=pep8_flake8
-  - TOXENV=pep257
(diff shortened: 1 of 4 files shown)


phlax:
@unho - gtm - absolutely brilliant - this will reduce our carbon footprint and dev wait time considerably - great!

Makefile
```diff
@@ -27,6 +27,20 @@ assets:
...
 	chmod 664 ${ASSETS_DIR}.webassets-cache/*
 
+travis-assets:
+	if [ -d "${ASSETS_DIR}.webassets-cache/" ]; then \
+		echo "eating cache - yum!"; \
+	else \
+		cd ${JS_DIR} && \
+		npm install && \
+		cd ${CWD}; \
+		python manage.py webpack --extra=--display-error-details --extra=--dev --extra=--nowatch; \
```
julen: As already mention, this has to read `python manage.py webpack --dev --nowatch --extra=--display-error-details`

unho: I think that it is enough `python manage.py webpack --dev --nowatch` with the changes you landed.

tox.ini
```diff
@@ -38,71 +34,24 @@ commands=
     #sleep 10
     #kill -2 $TESTPID
     py.test --cov-report= --cov=. -v --duration=10
-    python {toxinidir}/run_coveralls.py
+    # python {toxinidir}/run_coveralls.py
```
phlax: @unho this line is wrong - we deal with this better in master

tox.ini
```diff
@@ -38,71 +34,24 @@ commands=
...
 
-[testenv:sdist]
-basepython=python
-usedevelop=False
-skipsdist=True
+[testenv:project]
 whitelist_externals=
     make
+    npm
+    /bin/cd
```
phlax: this line is not needed

.travis.yml
```diff
@@ -2,34 +2,34 @@
...
   - pip install --upgrade "pip>=8.0" # FIXME remove once Travis support pip8 by default
   - pip install --upgrade "tox>=2.3"
+  - pip install
+      -r requirements/base.txt
+      -r requirements/tests.txt
+      -e .
+
 script:
-  - tox -e $TOXENV
+  - tox --sitepackages -e $TOXENV
```
phlax: not sure if we need sitepackages as its set in tox.ini - the problem was with the site-packages folder - so this _might_ be unnessary

unho: Then I guess we *might* leave it stay.

unho: Removing it.

phlax:
apart from tiny nit, and once we know tests pass 100% - GTM

dwaynebailey:
lgtm, some comments but mostly just flagging some things.

.travis.yml
```diff
@@ -2,32 +2,32 @@
 sudo: false
 language: python
 env:
-  - TOXENV=py27-django18-sqlite
-  - TOXENV=py27-django18-mysql
-  - TOXENV=py27-django18-postgres
+  - TOXENV=py27-django18-sqlite PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
```
julen: It'd be good to document with a comment why these envvars need to be set as one would think tox already takes care of the `PYTHONPATH`.

phlax: hi @julen what this PR does is create a python env outside of tox - ie in the default travis env - and then make this available to the tox venvs.

there are some implications to the changes that we are making to speed up test runs though (although not entirely from this PR)

I will write up something more substantial, but worth noting:

- python libs are only installed when you run uncached - so if you want to update upstream you need to clear cache
- static resources are only cooked into cache - so again you need to clear cache to get changes - this may error anyway without clearing cache - not sure
- when pushing to master you should clear and test PR uncached to ensure it works with latest upstream etc

julen: Thanks for explanation, @phlax! As said, mind adding a comment in the file so everyone reading it is aware of it? TIA.

PR #4452, opened 2016-02-04
TM: leverage Levenshtein distance on ES hits

### LINKED ISSUE #3971: Local TM results are a large edit distance from the source text
Current Local TM is giving matches based on Elasticsearches Levenshtein distance.  The matches do have similarity but in quite a few cases the distance is so great that its not a useful suggestion.  There are cases where substrings within the match prove useful (and indicate the likely reason the match was found).

It does seems that ES's idea of Levenshtein is different from what we are used to.  So we either need delve deaper to find some setting to alter or apply our own Levenshtein to the matches proposed.

julen:
This leverages the Levenshtein distance on top of the hits returned by ES, considering the whole text, not individual terms.

In my testing the irrelevant results are now gone. As mentioned in the issue, we set a threshold to allow changes up to a third of the text contents. YMMV so feel free to propose any adjustments to the threshold.

The PR also removes the `MIN_SCORE` setting because it results irrelevant (check the commit message for further reasoning).

Fixes #3971.

### DIFF
--- a/pootle/core/search/backends/elasticsearch.py
+++ b/pootle/core/search/backends/elasticsearch.py
@@ -11,6 +11,8 @@ from __future__ import absolute_import
 
 import logging
 
+import Levenshtein
+
 try:
     from elasticsearch import Elasticsearch
     from elasticsearch.exceptions import ElasticsearchException
@@ -24,6 +26,36 @@ __all__ = ('ElasticSearchBackend',)
 
 
 logger = logging.getLogger(__name__)
+
+
+DEFAULT_MIN_SIMILARITY = 0.7
+
+
+def filter_hits_by_distance(hits, source_text,
+                            min_similarity=DEFAULT_MIN_SIMILARITY):
+    """Returns ES `hits` filtered according to their Levenshtein distance
+    to the `source_text`.
+
+    Any hits with a similarity value (0..1) lower than `min_similarity` will be
+    discarded. It's assumed that `hits` is already sorted from higher to lower
+    score.
+    """
+    if min_similarity <= 0 or min_similarity >= 1:
+        min_similarity = DEFAULT_MIN_SIMILARITY
+
+    filtered_hits = []
+    for hit in hits:
+        hit_source_text = hit['_source']['source']
+        distance = Levenshtein.distance(source_text, hit_source_text)
+        similarity = (
+            1 - distance / float(max(len(source_text), len(hit_source_text)))
+        )
+        if similarity < min_similarity:
+            break
+
+        filtered_hits.append(hit)
+
+    return filtered_hits
 
 
 class ElasticSearchBackend(SearchBackend):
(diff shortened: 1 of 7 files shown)


unho:
Shouldn't removal of `MIN_SCORE` be recorded in release notes?

unho:
Shouldn't removal of MIN_SCORE be recorded in release notes?

pootle/core/search/backends/elasticsearch.py
```diff
@@ -26,6 +28,26 @@
 logger = logging.getLogger(__name__)
 
 
+def filter_hits_by_distance(hits, source_text, max_distance=0.3):
+    """Returns ES `hits` filtered according to their Levenshtein distance
+    to the `source_text`.
+
+    Any hits with a normalized distance value (0..1) higher than
+    `max_distance` will be discarded. It's assumed that `hits` is already
+    sorted from higher to lower score.
```
unho: You shouldn't assume that. Even more, [score is altered below in this file by using the TM weight](https://github.com/translate/pootle/blob/master/pootle/core/search/backends/elasticsearch.py#L113).

unho: Leaving that post-processing change it seems that we can rely on [elasticsearch returning the results sorted by relevance](https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting.html).

dwaynebailey:
@julen So my reading is we just do ``AUTO`` all the time now and use LS to filter the results.  So there is now no way to actually set a threshold score.  Would it not be better to allow the threshold to be set in the config?

pootle/core/search/backends/elasticsearch.py
julen: It's perfectly fine to assume that. Even if you multiply the score by a weight factor later, those relevance scores are still a) sorted, b) refer to irrelevant results.

julen:
I have no problem to allow setting the threshold value externally. How did testing go with the PR by the way?

julen:
I updated the PR with some of the proposed fixes, and the threshold can now be configured with `MAX_DISTANCE`. Looking forward to hear about your test results!

dwaynebailey:
@julen ran some tests, and tried to fake some LS distances.  Pretty hard to determine which should match.  But the matching works correctly as far as I can tell. Would be helpful if the ``data-score`` was actually a levenshtein value.  I've got values of 10.NNN for my tests so hard to know if this is even filtering edit distance values of e.g. 0.3

Otherwise lgtm.

julen:
Oh, good catch, actually I should remove the `data-score` values from the DOM — they were meant to help debug scores, but we've seen these are irrelevant.

dwaynebailey:
@julen might be useful if they actually where LS distances.  Certainly helps quickly confirm that things are working as expected.  So I'd support changing them to LS and keeping them.

julen:
I added an extra commit which changes this to LS distances, I'd appreciate if you can double-check now. Still I'd be in favor of not transmitting such _debugging_ information over the wire. We need to add some tests for the backend code definitely, although at the moment I'm pretty confident all results over the threshold are being discarded.

dwaynebailey:
@julen retested.  TBH having the data there is helpful so I wouldn't want to remove it.  I ended up adding my own debug statements so that I could see what was accepted and what was rejected.  Please add some if you feel it may be useful, but not critical.

GTM for the code.

My only concern is this.  AUTO in ES was meant to adjust the acceptable distance depending on string length.  While this didn't work I did notice that a fake string of 'One' won't match 'Ones', i.e. for really short strings we won't get matches that we might want.  I'm not 100% sure its worth worrying about though.

julen:
I updated the PR taking into account @friedelwolff's suggestion to use similarity rather than distance for consistency with the rest of our tools. I also added debug information via logging so anyone can check what's going on behind the scenes. I consider this is good to go from my side.

dwaynebailey:
lgtm.  Didn't test, but was happy with my last batch of testing.

PR #4453, opened 2016-02-04
Travis cleanups

phlax:
- Cache python libs in Travis
- Improve tox runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,27 +7,18 @@ env:
   - TOXENV=py27-django18-postgres PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
   # Meta
   - TOXENV=project PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
-matrix:
-  allow_failures:
-    - env: TOXENV=isort
-before_cache:
-  - rm -f $HOME/.cache/pip/log/debug.log
 cache:
   directories:
     - pootle/static/js/node_modules
     - pootle/assets
-    - $HOME/.cache/pip
-before_install:
-  # Error as early as possible if we can't compile
-  - python -m compileall -q -f -x ".*[.]tox.*" .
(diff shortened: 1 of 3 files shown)


unho:
Is this speeding up things?

.travis.yml
```diff
@@ -11,19 +11,20 @@ matrix:
...
 before_cache:
-  - rm -f $HOME/.cache/pip/log/debug.log
+  - pyclean $HOME/virtualenv/python2.7.9/lib
 cache:
   directories:
     - pootle/static/js/node_modules
     - pootle/assets
-    - $HOME/.cache/pip
+    - $HOME/virtualenv/python2.7.9/lib
+    - $HOME/virtualenv/python2.7/bin
```
dwaynebailey: Really?  This venv comes with the python Travis build so we'll be reverting any fixes that they make.  Is that a good thing?

.travis.yml
```diff
@@ -11,19 +11,20 @@ matrix:
...
+
 before_install:
   # Error as early as possible if we can't compile
   - python -m compileall -q -f -x ".*[.]tox.*" .
 install:
   - pip install --upgrade "pip>=8.0" # FIXME remove once Travis support pip8 by default
-  - pip install --upgrade "tox>=2.3"
-  - pip install
+  - pip install --upgrade
```
dwaynebailey: I think you want skip upgrade, since that would requires pings to pypi which would happen anyway if the local requirement didn't meet ``requirements/*``.  Having said that removing ``--upgrade`` from pip and tox upgrade didn't show any real speedup on my tests.

.travis.yml
phlax: we are not testing upstream

any time the cache is cleared we will pull in the libs

we dont need to be rebuilding python libs on every test run - would you do that on your machine?

phlax: not sure i understand the point about reverting - im just freezing the python deps at pre-cache build time

phlax: also this PR has gone a bit WIP temporarily i have some further changes to backport

unho: I am concerned about hardcoding version numbers.

phlax:
> Is this speeding up things?

@unho yep

unho:
lgtm

dwaynebailey:
lgtm

PR #4456, opened 2016-02-05
[RFC] Fix TM missing fullname

dwaynebailey:
Tentative fix for 'someone' always appearing as the TM user who contributed the TM when we don't have the fullname but only the username.

@julen would you be able to verify this?

Not sure if we have a case where both fullname and username are missing?


dwaynebailey:
Nicely fixed in e7398e5ae6a886ba5ff73a8bb4dfeec0607daa46

julen:
> Not sure if we have a case where both fullname and username are missing?

Yes, when importing TM units from files (and IIRC also in some cases with some ES indexes created long time ago).

PR #4458, opened 2016-02-06
Test cleanups

phlax:
These tests attempt to write to the testing systems pootle.conf

### DIFF
--- a/tests/commands/pootle_runner.py
+++ b/tests/commands/pootle_runner.py
@@ -38,8 +38,9 @@ def test_pootle_init(capfd):
 
 
 @pytest.mark.cmd
-def test_pootle_init_db_sqlite(capfd):
+def test_pootle_init_db_sqlite(capfd, tmpdir):
     """pootle init --help"""
-    call(['pootle', 'init', '--db=sqlite'])
+    test_conf_file = tmpdir.join("pootle.conf")
+    call(['pootle', 'init', '--db=sqlite', '--config=%s' % test_conf_file])
     out, err = capfd.readouterr()
     assert "Configuration file created" in out
--- a/pytest_pootle/factories.py
+++ b/pytest_pootle/factories.py
@@ -209,3 +209,9 @@ class AnnouncementFactory(factory.django.DjangoModelFactory):
     class Meta(object):
         model = 'staticpages.StaticPage'
     active = True
+
+
+class SuggestionFactory(factory.django.DjangoModelFactory):
+
+    class Meta(object):
+        model = 'pootle_store.Suggestion'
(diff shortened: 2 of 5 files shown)


tests/commands/pootle_runner.py
```diff
@@ -30,7 +30,8 @@ def test_pootle_version(capfd):
 
 
 @pytest.mark.cmd
-def test_pootle_init(capfd):
+@pytest.mark.django_db
+def __test_pootle_init(capfd):
     """pootle init --help"""
     call(['pootle', 'init', '--help'])
```
dwaynebailey: This doesn't write out a ``pootle.conf`` file. So not 100% sure why its needs disabling

tests/commands/pootle_runner.py
```diff
@@ -38,7 +39,7 @@ def test_pootle_init(capfd):
 
 
 @pytest.mark.cmd
-def test_pootle_init_db_sqlite(capfd):
+def __test_pootle_init_db_sqlite(capfd):
     """pootle init --help"""
     call(['pootle', 'init', '--db=sqlite'])
```
dwaynebailey: We can write a temporary files using ``--config``

tests/commands/pootle_runner.py
phlax: apologies - i probs disabled it in error

tests/commands/pootle_runner.py
dwaynebailey:
```diff
diff --git a/tests/commands/pootle_runner.py b/tests/commands/pootle_runner.py
index 2dba648..f4258f8 100644
--- a/tests/commands/pootle_runner.py
+++ b/tests/commands/pootle_runner.py
@@ -38,8 +38,9 @@ def test_pootle_init(capfd):
 
 
 @pytest.mark.cmd
-def test_pootle_init_db_sqlite(capfd):
+def test_pootle_init_db_sqlite(capfd, tmpdir):
     """pootle init --help"""
-    call(['pootle', 'init', '--db=sqlite'])
+    test_conf_file = tmpdir.join("pootle.conf")
+    call(['pootle', 'init', '--db=sqlite', '--config=%s' % test_conf_file])
     out, err = capfd.readouterr()
     assert "Configuration file created" in out
```

dwaynebailey:
I'm no expert, but lgtm

phlax:
> Mmm I really prefer if email is not None

it really doesnt matter in this context. testing for is/not ``None`` is essential in a certain type of situation - ie where the var has the potential to have a zero value other than none - and you do/not want to use the var in that situation.

PR #4460, opened 2016-02-07
Test cleanups (continued)

phlax:


tests/commands/list_languages.py
```diff
@@ -40,9 +42,11 @@ def test_list_languages_modified_since(capfd, afrikaans_tutorial,
                                        spanish_tutorial, french_tutorial):
     """Languages modified since a revision
 
-    We expect fr, as its is updated on revision 6 and 7.
```
unho: Since you remove this, there is no reason to keep the multiline docstring.

tests/commands/changed_languages.py
```diff
@@ -10,23 +10,28 @@
...
+         "%d (inclusive)"
+         % (revision, revision))
+        in err)
 
 
 @pytest.mark.cmd
 @pytest.mark.django_db
 def test_changed_languages_since_revision(capfd, afrikaans_tutorial,
-                                          french_tutorial, revision):
+                                          french_tutorial):
```
ta2-1: I had to open the source file to confirm my initial understanding that this change doesn't match the commit message.

tests/commands/list_languages.py
phlax: fair enough - ill clean up

tests/commands/changed_languages.py
phlax: kind of. its unnecessary as ``revision`` is set to autouse - also i wanted to use  ``revision`` as a local var

tests/models/store.py
```diff
@@ -233,8 +236,11 @@ def test_update_set_last_sync_revision(ru_update_set_last_sync_revision_po):
 
 @pytest.mark.django_db
 def test_update_upload_defaults(en_tutorial_po, system):
-    _update_from_upload_file(en_tutorial_po,
-                             "tests/data/po/tutorial/en/tutorial_update.po")
+    update_store(
+        en_tutorial_po,
+        [("Hello, world", "Hello, world UPDATED")],
```
ta2-1: @phlax, does `en_tutorial_po` remain a standard fixture? And it contains one `Hello, world` unit. This is a kind of hardcode. Are we looking for a unit for updating by its source string here?

unho: Since we now only pass the units, I think we are not properly testing reading the uploaded file revision. Is it not possible to pass here the full uploaded file content to avoid IO delay, and thus keep using old code?

phlax: en_tutorial_po is still there and is used - eg by this test.

It still has the same unit in.

The point about updating Stores from files in tests is that previously the ``revision`` was hardcoded.

As the underlying db has changed the hardcoded revision is no longer correct and nothing gets updated etc.

These changes address this.

tests/commands/changed_languages.py
unho: I think the commit mixes two separate changes. One of those looks like removing unnecessary usage of `revision` fixture which surely can be directly merged once confirmed tests work fine without it.

phlax: @unho - i can confirm this for you and point you quickly to the reason why - this is not something that requires testing (these are tests)

tests/models/store.py
ta2-1: @phlax but then we should add a separate test for reading `revision` from a file, yes?

unho:
lgtm once docstring nit addressed.

ta2-1:
ltgm

tests/models/store.py
phlax: perhaps - although my feeling is that this is more of a TTK thing

PR #4469, opened 2016-02-09
Tests: skip pootle runner tests if not installed via setup.py

julen:
Allows to omit some tests when developing directly in a virtualenv, where `setup.py` extras are not available (the `pootle` command is an entry point provided by `setup.py`).


dwaynebailey:
lgtm, wasn't sure how to test this as I couldn't get the non dev install PR to work.

unho:
lgtm

PR #4473, opened 2016-02-11
Get units search filters

phlax:


pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+
+    def filter_my_suggestions(self):
+        if not self.user:
+            return self.qs.none()
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.PENDING,
+            suggestion__user=self.user).distinct()
+
+    def filter_user_suggestions(self):
+        return self.filter_my_suggestions()
```
unho: Nit: I would invert this, make `filter_my_suggestions` call `filter_user_suggestions`.

phlax: k - ill change

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+
+    def filter(self, qs, unit_filter, *args, **kwargs):
+        for search_filter in self.filters:
+            # try each of the filter classes to find one with a method to handle
+            # `unit_filter`
+            try:
+                return search_filter(qs, *args, **kwargs).filter(unit_filter)
+            except FilterNotFound:
+                pass
+        # if none match then return the empty qs
```
unho: Nit: Use capitalization and periods in all comments.

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+from pootle_statistics.models import SubmissionTypes
+from pootle_store.models import SuggestionStates
+from pootle_store.util import FUZZY, TRANSLATED, UNTRANSLATED
+
+
+class FilterNotFound(Exception):
+    pass
+
+
+class UnitFilter(object):
```
unho: Nit: I would rename it to `UnitFilterBase`.

phlax: k

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+                qualitycheck__name__in=self.checks).distinct()
+        elif self.category:
+            return self.qs.filter(
+                qualitycheck__false_positive=False,
+                qualitycheck__category=self.category).distinct()
+        return self.qs.none()
+
+
+class UnitStateFilter(UnitFilter):
+    """Filter a Unit qs based on unit state"""
```
unho: There is no `filter_all` method so when providing a `'all'` unit_filter an empty queryset will be returned, am I wrong?

pootle/apps/pootle_store/unit/filters.py
unho: Or `BaseUnitFilter`.

ta2-1: It seems `Base` is added in the beginning?

pootle/apps/pootle_store/unit/filters.py
phlax: you are right - good spot - ill fix

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+        return qs.exclude(submitted_by=self.user).distinct()
+
+    def filter_user_submissions_overwritten(self):
+        return self.filter_my_submissions_overwritten()
+
+
+class UnitSearchFilter(object):
+
+    filters = (
+        UnitChecksFilter, UnitStateFilter, UnitContributionFilter)
```
unho: Is there any reason why we filter by checks first when in the original code we filtered by checks last?

phlax: we dont really

it filters on one thing or the other atm - and in this code.

i think we may want to revisit this functionality but not in this PR

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+    def filter_suggestions(self):
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.PENDING).distinct()
+
+    def filter_my_suggestions(self):
+        if not self.user:
+            return self.qs.none()
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.PENDING,
+            suggestion__user=self.user).distinct()
```
unho: For making it clearer, I would this to be a filter by provided user on top of the queryset returned by `filter_suggestions`.

phlax: im sorry - i dont follow

pootle/apps/pootle_store/views.py
```diff
@@ -131,71 +131,20 @@ def get_step_query(request, units_queryset):
...
-                )
-            elif unit_filter == 'suggestions':
-                match_queryset = units_queryset.filter(
-                    suggestion__state=SuggestionStates.PENDING).distinct()
-            elif unit_filter in ('my-suggestions', 'user-suggestions'):
-                match_queryset = units_queryset.filter(
-                    suggestion__state=SuggestionStates.PENDING,
-                    suggestion__user=user,
-                ).distinct()
-                sort_on = 'suggestions'
```
unho: With the new code this is no longer set. Same for `'my-submissions', 'user-submissions'` below.

phlax: again really good spot!

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.ACCEPTED,
+            suggestion__user=self.user).distinct()
+
+    def filter_user_suggestions_rejected(self):
+        if not self.user:
+            return self.qs.none()
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.REJECTED,
+            suggestion__user=self.user).distinct()
```
unho: There is a lot of filters on suggestions by user. I wonder if we can refactor that out.

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+
+    def filter_my_submissions(self):
+        if not self.user:
+            return self.qs.none()
```
unho: This is also repeated a lot in the class methods.

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+        return self.qs.filter(
+            submission__submitter=self.user,
+            submission__type__in=SubmissionTypes.EDIT_TYPES).distinct()
+
+    def filter_user_submissions(self):
+        return self.filter_my_submissions()
```
unho: As above it makes sense that `filter_my_submissions` is the one that calls the more generic `filter_user_submissions`.

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,139 @@
...
+    def filter_my_submissions_overwritten(self):
+        if not self.user:
+            return self.qs.none()
+        qs = self.qs.filter(
+            submission__submitter=self.user,
+            submission__type__in=SubmissionTypes.EDIT_TYPES)
+        return qs.exclude(submitted_by=self.user).distinct()
+
+    def filter_user_submissions_overwritten(self):
+        return self.filter_my_submissions_overwritten()
```
unho: As above, invert who calls who.

pootle/apps/pootle_store/unit/filters.py
phlax: k

pootle/apps/pootle_store/unit/filters.py
unho: Sorry, there was a typo on my message. And perhaps even now it is not clear enough so lets try with some code:

```python
        return self.filter_suggestions().filter(suggestion__user=self.user)
```

I am now wondering if all those `distinct()` calls are actually necessary.

pootle/apps/pootle_store/unit/filters.py
phlax: done

pootle/apps/pootle_store/unit/filters.py
unho: This was me finding harder to compare previous code with newer code since the order was altered. Once this lands it makes no sense which is the order in where stuff is. So this is not actually about this specific line but about all the code above.

pootle/apps/pootle_store/unit/filters.py
phlax: that would work - but doesnt matter so much now as i followed your other suggestion

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,145 @@
...
+        return self.qs.filter(
+            suggestion__state=SuggestionStates.PENDING).distinct()
+
+    def filter_user_suggestions(self):
+        if not self.user:
+            return self.qs.none()
+        return self.user_suggested_units.filter(
+            suggestion__state=SuggestionStates.PENDING).distinct()
+
+    def filter_my_suggestions(self):
```
ta2-1: where do we use methods like `filter_my_*`?

ta2-1: @phlax seems we don't?

pootle/apps/pootle_store/unit/filters.py
```diff
@@ -6,6 +6,145 @@
...
+        return self.user_submitted_units.filter(
+            submission__type__in=SubmissionTypes.EDIT_TYPES).distinct()
+
+    def filter_my_submissions(self):
+        return self.filter_user_submissions()
+
+    def filter_user_submissions_overwritten(self):
+        if not self.user:
+            return self.qs.none()
+        qs = self.user_submitted_units.filter(
```
unho: Nit: I find this variable unnecessary.

pootle/apps/pootle_store/unit/filters.py
phlax: its kinda automagic - see here @ta2-1 

https://github.com/translate/pootle/pull/4473/files#diff-ff1b62296234e6d098a91dc441a11ee1R28

pootle/apps/pootle_store/unit/filters.py
unho: I guess there is no easy way to get rid of all of these? Perhaps some decorator that bails out soon?

pootle/apps/pootle_store/unit/filters.py
unho: Perhaps there is some remnant code here.

pootle/apps/pootle_store/unit/filters.py
phlax: perhaps in the future - for now this is good enough

pootle/apps/pootle_store/unit/filters.py
phlax: im not sure i follow

ta2-1: ops... tbh I'm happy with this automagic.

pootle/apps/pootle_store/unit/filters.py
phlax: it was pref in terms of formatting - ill see if its still necessary

pootle/apps/pootle_store/unit/filters.py
unho: No, these are all used.

pootle/apps/pootle_store/views.py
unho: This is now fixed.

pootle/apps/pootle_store/unit/filters.py
ta2-1: I'm feeling like we have very string dependency on names of filters which don't even defined constants but stings specified within code.

pytest_pootle/fixtures/search.py
```diff
@@ -53,6 +53,71 @@
...
+    if check_type == "category":
+        return check_type, get_category_id(check_data)
+    return check_type, check_data.split(",")
+
+
+@pytest.fixture(params=UNITS_CONTRIB_SEARCH_TESTS)
+def units_contributor_searches(request):
+    from django.contrib.auth import get_user_model
+
+    User = get_user_model()
```
unho: Nit: Move the three lines above inside the `if`.

pootle/apps/pootle_store/unit/filters.py
phlax: they match request params

pytest_pootle/fixtures/search.py
```diff
@@ -53,6 +53,71 @@
...
+
+
+@pytest.fixture(params=UNITS_STATE_SEARCH_TESTS)
+def units_state_searches(request):
+    return request.param
+
+
+@pytest.fixture(params=UNITS_CHECKS_SEARCH_TESTS)
+def units_checks_searches(request):
+    from pootle_misc.checks import get_category_id
```
unho: Nit: move the import within the `if`.

tests/search.py
```diff
@@ -105,3 +206,60 @@ def test_get_units_text_search(units_text_searches):
...
+        _test_units_checks_filter(_qs, check_type, check_data)
+
+
+@pytest.mark.django_db
+def test_units_checks_filter_bad():
+    qs = Unit.objects.all()
+    with pytest.raises(FilterNotFound):
+        UnitChecksFilter(qs).filter("BAD")
+    # if you dont supply check/category you get empty qs
+    assert not UnitChecksFilter(qs).filter("checks").count()
```
unho: How is no category being tested here?

phlax: neither are being tested - its testing what happens if you dont specify which type of search - ie checks or category

its confusing because the the flag is also called ``checks``

tests/search.py
```diff
@@ -76,6 +152,31 @@ def _test_unit_text_search(qs, text, sfields, exact, empty=True):
             assert searchword_found
 
 
+def _test_units_state_filter(qs, unit_filter):
+    result = UnitStateFilter(qs).filter(unit_filter)
+    for item in result:
+        assert item in qs
+    assert result.count() == result.distinct().count()
```
ta2-1: what do you check here?

ta2-1: @phlax, can you tell me when results differ here?

phlax: the result would differ if the query didnt give back a unique qs - in this case thats unlikely but not impossible.

unho:
lgtm

ta2-1:
gtm

PR #4475, opened 2016-02-11
Tests: restore the way we used to load fixtures

julen:
Makes fixtures easier to maintain by removing a big chunk of the boilerplate.


phlax:
great @julen this is much cleaner, lgtm

ta2-1:
thanks @julen, I like it :+1:

PR #4476, opened 2016-02-11
Tests: parametrize `tp_views` fixture

julen:
Refs. https://github.com/translate/pootle/pull/4471#discussion-diff-52581021. Let me know what you think of the approach.

### DIFF
--- a/pytest_pootle/fixtures/views.py
+++ b/pytest_pootle/fixtures/views.py
@@ -134,20 +134,17 @@ def project_views(request, client):
     return test_type, project, response.wsgi_request, response, kwargs
 
 
-@pytest.fixture(params=TP_VIEW_TESTS.keys())
-def tp_view_test_names(request):
-    return request.param
-
-
 @pytest.fixture(params=(None, 'member', 'admin'))
-def tp_view_usernames(request):
+def username(request):
     return request.param
 
 
-@pytest.fixture
-def tp_views(tp_view_test_names, tp_view_usernames, client):
+@pytest.fixture(params=TP_VIEW_TESTS.keys())
+def tp_views(request, client, username):
     from pootle.core.helpers import SIDEBAR_COOKIE_NAME
     from pootle_translationproject.models import TranslationProject
+
+    tp_view_test_names = request.param
 
     test_type = tp_view_test_names.split("_")[0]
     tp = TranslationProject.objects.all()[0]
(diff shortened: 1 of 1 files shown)


phlax:
@julen tbh i prefer the current implementation as it feels simpler.

The other advantage is that i think we are likely to reuse at least the usernames fixture to param some of the other view tests,

ta2-1:
@julen, look at this https://github.com/ta2-1/pootle/commit/1509415aabce29b145a260be5a3e0411b49645e5
You have added `ids` to create proper test names. I didn't know that it's possible.
But I think it's better to avoid using of `itertools` here.

unho:
It looks cleaner. Other than that I am not sure if it is better or not.

julen:
I agree on the usernames fixtures being potentially reusable; there's not much use in making the test names a separate fixture though. I made an update to reflect this.

@ta2-1 funny we came up with the same thing :smile:. And yeah it's possible to [customize the IDs that py.test will report](
https://github.com/pytest-dev/pytest/commit/4e35c00ab005da542e91a488da58fa3638618168) by passing in a string or a callable. Note Travis has failed here because it's using an older version from its cache (2.6) which doesn't support this yet (2.7+ does).

pytest_pootle/fixtures/views.py
```diff
@@ -134,21 +134,18 @@ def project_views(request, client):
...
 
 
-@pytest.fixture(params=TP_VIEW_TESTS.keys())
-def tp_view_test_names(request):
-    return request.param
-
-
 @pytest.fixture(params=(None, 'member', 'admin'))
-def tp_view_usernames(request):
+def username(request):
```
phlax: can we make this plural - so its clear its params - also ``site_users`` or similar might be useful - im thinking of adding the password into the fixture

julen: The point is you always use it in singular, i.e. `client.login(username=site_users, password=password)` doesn't make much sense IMHO. Also username and user are conceptually different (str vs object).

phlax: yep - but if you make the fixture singular its not clear that the test is parametrized

my point about ``usernames`` is that im going to change it to include passwords - ie dict not string - but i can happily rename it when i do

phlax: hi @julen i also need parametrized site_users in tests and have added a commit in a PR for this - https://github.com/translate/pootle/pull/4483. Perhaps this could be useful for this test too

ta2-1:
@julen I landed your change partly within PR #4497. So I'm closing this one.

PR #4477, opened 2016-02-11
Fix: default to StandardChecker for TP.checkers

phlax:


unho:
A brief explanation of our journey through the rabbit hole might help, isn't it?

ta2-1:
merge it faster!

unho:
A possible explanation:

The absence of 'standard' key in translate.filters.check.projectcheckers (fixed in https://github.com/translate/translate/commit/87a5e3db) caused projects with 'standard' checkstyle to default to `StandardUnitChecker` (4 checks) instead of `StandardChecker` (dozens of checks) thus rendering it useless.

unho:
lgtm

PR #4480, opened 2016-02-12
Test env checks

phlax:


pytest_pootle/env.py
```diff
@@ -428,7 +428,7 @@ def _add_submissions(self, unit, created):
 
         # add another suggestion as different user 7 days later
         suggestion2, _ = unit.add_suggestion(
-            "Suggestion 2 for %s" % unit.source,
+            "Suggestion 2 for %s" % (unit.target or unit.source),
```
unho: Is this change actually necessary?

phlax: yes - because otherwise the (erroneous) target string is overwritten with (not-erroneous) source

unho:
lgtm

PR #4481, opened 2016-02-12
Tests: simple manage.py test

dwaynebailey:
Simple effort to ensure ``./manage.py`` is exercised as part of out Pootle commands testing.


unho:
lgtm, but it probably can be improved to actually check that it outputs the right revision.

PR #4484, opened 2016-02-13
Travis: always build assets for 'master'

dwaynebailey:
Force rebuild of assets whenever we commit to the master branch.

Note:
- We should expand this to include ``stable/*``

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,9 @@ notifications:
     on_success: change
     on_failure: always
     on_start: never
+before_cache:
+  # Force rebuilds by removing cache for 'master' builds
+  - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then rm -rf pootle/static/js/node_modules/* pootle/assets/*; fi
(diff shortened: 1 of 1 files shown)


dwaynebailey:
Done various checks and this works now.

1. We force a rebuild on master by always deleting the assets cache at the end of a master build
2. Branches and PRs will keep their cache, you can force new cache builds by removing their cache.
3. Even in cases where PRs and branches revert to getting master cache, they will be an empty asset cache and thus will rebuild assets.

So seems we now have the best of both worlds.

1. master is always rebuilding assets
2. Branches and PRs keep their cache unless you remove it.

So looking for a gtm

phlax:
@dwaynebailey this lgtm, but my one concern is re python upstream deps.

I think if we force a rebuild of the assets cache on every master test run, without also forcing an upgrade of python deps, we are creating a false sense of security that master will build fresh.

phlax:
The other thing that im thinking is that this doesnt negate the need to clear the cache for your PR before pushing to master, if you want to know with 100% certainty that master wont break when you do.

PR #4485, opened 2016-02-13
commands: replace print with self.std{out,err}

dwaynebailey:
We shouldn't be using ``print`` in commands, use self.std* instead.


unho:
gtm

PR #4486, opened 2016-02-13
Omit settings file and script from coverage report

phlax:


dwaynebailey:
You might want to leave out ``pootle/strings.py`` also.  gtm

PR #4488, opened 2016-02-14
Fix: use pootle_path properly when ordering non-unique qs in get_units

### LINKED ISSUE #4079: Priority sorting has random file order where it should be alphabetical
In a server with vfolders, when clicking on an incomplete count, it automatically sets the priority sorting to the editor view. The sorting, in turn, exposes units in a very illogical way: sorted by priority but scattered across files. This results in totally confusing surrounding units.

See an example in the following screenshot — note all units in there have *1.0* priority.
weird_priority

phlax:
most likely fixes #4079

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -168,7 +168,8 @@ def get_step_query(request, units_queryset):
             sort_by = ALLOWED_SORTS[sort_on].get(sort_by_param, None)
             if sort_by is not None:
                 if sort_on in SIMPLY_SORTED:
-                    match_queryset = match_queryset.order_by(sort_by)
+                    match_queryset = match_queryset.order_by(
+                        sort_by, "store__pootle_path", "index")
                 else:
                     # Omit leading `-` sign
                     if sort_by[0] == '-':
@@ -183,7 +184,7 @@ def get_step_query(request, units_queryset):
                     # (unless PostreSQL is used and `distinct(field_name)`)
                     match_queryset = match_queryset \
                         .annotate(sort_by_field=Max(max_field)) \
-                        .order_by(sort_order)
+                        .order_by(sort_order, "store__pootle_path", "index")
 
             units_queryset = match_queryset
 
@@ -377,7 +378,7 @@ def get_units(request):
 
     if is_initial_request:
         sort_by_field = None
-        if len(step_queryset.query.order_by) == 1:
+        if len(step_queryset.query.order_by) > 2:
             sort_by_field = step_queryset.query.order_by[0]
 
         sort_on = None
@@ -390,8 +391,7 @@ def get_units(request):
             # Since `extra()` has been used before, it's necessary to
             # explicitly request the `store__pootle_path` field. This is a
             # subtetly in Django's ORM.
-            uid_list = [u['id'] for u
-                        in step_queryset.values('id', 'store__pootle_path')]
+            uid_list = list(step_queryset.values_list('id', flat=True))
         else:
             # Not using `values_list()` here because it doesn't know about all
             # existing relations when `extra()` has been used before in the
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_store/views.py
```diff
@@ -390,8 +391,7 @@ def get_units(request):
             # Since `extra()` has been used before, it's necessary to
             # explicitly request the `store__pootle_path` field. This is a
             # subtetly in Django's ORM.
-            uid_list = [u['id'] for u
-                        in step_queryset.values('id', 'store__pootle_path')]
+            uid_list = list(step_queryset.values_list('id', flat=True))
```
unho: `extra` seems not to be used anymore, so perhaps also adjust comment above.

pootle/apps/pootle_store/views.py
```diff
@@ -403,9 +403,7 @@ def get_units(request):
             # Django looks for `sort_by_field` field in the initial table.
             # https://code.djangoproject.com/ticket/19434
             uid_list = [u['id'] for u
-                        in step_queryset.values('id', 'sort_by_field',
-                                                'store__pootle_path')]
-
+                        in step_queryset.values('id', 'sort_by_field')]
```
unho: `extra` seems not to be used anymore, so perhaps also adjust comment above and just retrieve the `id` with a `values_list('id', flat=True)`.

pootle/apps/pootle_store/views.py
phlax: its been like that for a while - and in a later PR all of this is replace with a simple values_list as none of it is necessary any more

pootle/apps/pootle_store/views.py
phlax: see above comment

pootle/apps/pootle_store/views.py
```diff
@@ -183,7 +184,7 @@ def get_step_query(request, units_queryset):
                     # (unless PostreSQL is used and `distinct(field_name)`)
                     match_queryset = match_queryset \
                         .annotate(sort_by_field=Max(max_field)) \
-                        .order_by(sort_order)
+                        .order_by(sort_order, "store__pootle_path", "pk")
```
unho: Why `pk`?

pootle/apps/pootle_store/views.py
```diff
@@ -413,6 +411,7 @@ def get_units(request):
                 begin = max(index - chunk_size, 0)
                 end = min(index + chunk_size + 1, len(uid_list))
                 uids = uid_list[begin:end]
+                units = step_queryset[begin:end]
```
unho: Is this to avoid calling `filter` a few lines below?

pootle/apps/pootle_store/views.py
phlax: in mysql and postgres they will not give back a determinate order, unless we specify something unique - ie pk

pootle/apps/pootle_store/views.py
phlax: no - calling filter a few lines below is *wrong* and is partly to blame for random results

when you filter pk__in=somelist - you will not get the results back in the order of somelist from a real db (altough you do from the in-mem test db)

pootle/apps/pootle_store/views.py
unho: Sure, but since you are removing the `store__pootle_path` field now, it makes sense to adjust the comment now. Otherwise when reviewing in the future we will have to check history to know what is the comment talking about. So I would kindly request this to be replaced with a simple `values_list` in this PR, and also properly adjust or remove these comments.

pootle/apps/pootle_store/views.py
unho: ok

pootle/apps/pootle_store/views.py
unho: ok

pootle/apps/pootle_store/views.py
phlax: cool - ill remove the comment

pootle/apps/pootle_store/views.py
ta2-1: @phlax it seems that you are happy with any determinate order, are you? probably `index` suits better here?

phlax: fine

unho:
lgtm

ta2-1:
gtm

PR #4489, opened 2016-02-14
Get units form

phlax:


pootle/apps/pootle_store/form_fields.py
```diff
@@ -0,0 +1,59 @@
...
+    def value_from_datadict(self, data, files, name):
+        if hasattr(data, "getlist"):
+            return data.getlist(name)
+        return []
+
+
+class MultipleArgsField(forms.Field):
+    widget = MultipleValueWidget
+
+    def __init__(self, *la, **kwa):
```
unho: `*la, **kwa`? Not `*args, **kwargs`?

pootle/apps/pootle_store/views.py
```diff
@@ -321,56 +321,52 @@ def get_units(request):
...
+    month = search_form.cleaned_data['month']
+    pootle_path = search_form.cleaned_data['pootle_path']
+    unit_filter = search_form.cleaned_data['filter']
+    search = search_form.cleaned_data['search']
+    sfields = search_form.cleaned_data['sfields']
+    sort_by = search_form.cleaned_data["sort_by"]
+    sort_on = search_form.cleaned_data["sort_on"]
+    uids = search_form.cleaned_data["uids"]
+    user = search_form.cleaned_data['user']
+    vfolder = search_form.cleaned_data["vfolder"]
```
unho: We have a lot of inconsistent quoting being used in this lines.

pootle/apps/pootle_store/form_fields.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def clean(self, value):
+        if len(value) == 1 and "," in value[0]:
+            value = value[0].split(",")
+        return [self.field.clean(x) for x in value]
+
+
+class ISODateTimeField(forms.DateTimeField):
+
+    def to_python(self, v):
```
unho: I don't particularly like using abstruse variables. What does `v` mean? Perhaps `value`?

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+
+
+class UnitSearchForm(forms.Form):
+
+    initial = forms.BooleanField(required=False)
+    count = forms.IntegerField(required=False)
+    path = forms.CharField(
+        max_length=2048,
+        required=True)
+    uids = MultipleArgsField(
```
unho: To be honest I still don't get why we need to pass `uids`, in plural. But obviously it is not part of this PR to change this.

pootle/apps/pootle_store/form_fields.py
dwaynebailey: This has bothered me for some time.  There doesn't seem to be any good reason not to follow Python conventions on this.  Yes it just a convention, but then so is self and cls and we follow those to make it easier for anyone reading the code.

pootle/apps/pootle_store/form_fields.py
dwaynebailey: [It's value](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.Field.to_python) but it makes no sense not to use the Django signature.

pootle/apps/pootle_store/form_fields.py
phlax: 2 reasons:
- pdb has ``args`` so it conflicts with locals() which is a massive pita!
- kwargs has 3 extra chars that i hate having to type like all the time

that said - i try and clean this up before submitting - so apologies - ill fix now

pootle/apps/pootle_store/views.py
```diff
@@ -321,56 +321,52 @@ def get_units(request):
...
+                elif error.code == "required":
+                    raise Http400(_('Arguments missing.'))
+        raise Http404(forms.ValidationError(search_form.errors).messages)
+
+    category = search_form.cleaned_data['category']
+    checks = search_form.cleaned_data['checks']
+    chunk_size = search_form.cleaned_data['count']
+    exact = 'exact' in search_form.cleaned_data['soptions']
+    is_initial_request = search_form.cleaned_data['initial']
+    modified_since = search_form.cleaned_data['modified-since']
```
unho: There is no such field in the form.

unho: I see it now.

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+        queryset=get_user_model().objects.all(),
+        required=False,
+        to_field_name="username")
+
+    search = forms.CharField(
+        required=False,
+        widget=forms.TextInput(attrs={
+            'size': '15',
+            'placeholder': _('Search'),
+            'title': _("Search (Ctrl+Shift+S)<br/>Type and press Enter to "
```
unho: We are not displaying this in the UI, so why provide a widget with placeholder and such?

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+            ('source', _('Source Text')),
+            ('target', _('Target Text')),
+            ('notes', _('Comments')),
+            ('locations', _('Locations'))),
+        initial=['source', 'target'])
+
+    def __init__(self, *la, **kwa):
+        self.request_user = kwa.pop("user")
+        super(UnitSearchForm, self).__init__(*la, **kwa)
+        self.fields["modified-since"] = ISODateTimeField(required=False)
```
unho: Any reason to inject the field here? If yes please put some comment in code explaining why.

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+        sort_by_param = self.cleaned_data["sort"]
+
+        if sort_by_param in ALLOWED_SORTS[sort_on]:
+            self.cleaned_data["sort_by"] = ALLOWED_SORTS[sort_on][sort_by_param]
+        else:
+            self.cleaned_data["sort_by"] = None
+
+        self.cleaned_data["sort_on"] = sort_on
+
+    def clean_month(self):
```
unho: This is a stupid nit, but I don't like having a three lines method on the bottom because it is easy to miss. Also it kinda belongs with the other `clean_FIELD` methods above.

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+        if "checks" in self.errors:
+            del self.errors["checks"]
+            self.cleaned_data["checks"] = None
+        if "user" in self.errors:
+            User = get_user_model()
+            del self.errors["user"]
+            self.cleaned_data["user"] = User.get(self.request_user)
+        if self.errors:
+            return
+        if self.cleaned_data['count'] is None and not self.errors:
```
unho: I think `not self.errors` is redundant.

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+            self.cleaned_data["count"] = (
+                self.cleaned_data["user"].get_unit_rows())
+
+        self.cleaned_data["vfolder"] = None
```
unho: Setting a default on the field doesn't work?

pootle/apps/pootle_store/views.py
```diff
@@ -321,56 +321,52 @@ def get_units(request):
...
-    path_keys = [
-        "project_code", "language_code", "dir_path", "filename"]
-    try:
-        path_kwargs = {
-            k: v
-            for k, v in resolve(pootle_path).kwargs.items()
-            if k in path_keys}
-    except Resolver404:
-        raise Http404('Unrecognised path')
+    from pootle_store.forms import UnitSearchForm
```
ta2-1: this can be moved to the top of the module.

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+            if unit_filter in ('suggestions', 'user-suggestions'):
+                sort_on = 'suggestions'
+            elif unit_filter in ('user-submissions', ):
+                sort_on = 'submissions'
+        sort_by_param = self.cleaned_data["sort"]
+
+        if sort_by_param in ALLOWED_SORTS[sort_on]:
+            self.cleaned_data["sort_by"] = ALLOWED_SORTS[sort_on][sort_by_param]
+        else:
+            self.cleaned_data["sort_by"] = None
```
unho: What about just using `self.cleaned_data["sort_by"] = ALLOWED_SORTS[sort_on].get(sort_by_param)` instead of this whole `if`?

pootle/apps/pootle_store/forms.py
phlax: because it contains a hyphen

pootle/apps/pootle_store/forms.py
phlax: no its not

if there are errors then it will throw non-ValidationErrors parsing the rest of the vars (potentially)

anyhow - job is done - why continue?

pootle/apps/pootle_store/forms.py
phlax: there is no field

pootle/apps/pootle_store/forms.py
phlax: yep - ill change - it was doing something else previously - its factored this far

pootle/apps/pootle_store/forms.py
phlax: ill move clean_user down to make it alpha

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+        self.cleaned_data["vfolder"] = None
+        pootle_path = self.cleaned_data.get("path")
+        if 'virtualfolder' in settings.INSTALLED_APPS:
+            vfolder, pootle_path = extract_vfolder_from_path(
+                pootle_path,
+                vfti=VirtualFolderTreeItem.objects.select_related(
+                    "directory", "vfolder"))
+            self.cleaned_data["vfolder"] = vfolder
+            self.cleaned_data["pootle_path"] = pootle_path
+        path_keys = [
```
unho: I would put some blank line before this line since this a different logic block.

pootle/apps/pootle_store/views.py
phlax: well - this PR is a step - its moving very soon - but i can update

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+            "project_code", "language_code", "dir_path", "filename"]
+        try:
+            path_kwargs = {
+                k: v
+                for k, v in resolve(pootle_path).kwargs.items()
+                if k in path_keys}
+        except Resolver404:
+            raise forms.ValidationError('Unrecognised path')
+
+        self.cleaned_data.update(path_kwargs)
```
unho: A comment explaining that `path_kwargs`, and thus which fields can be updated here, can be really helpful.

pootle/apps/pootle_store/form_fields.py
phlax: hmm - i think you guys dont use pdb so much 8/

pootle/apps/pootle_store/forms.py
unho: I see two lines above "if errors then return" so if we get to this line we obviously have no errors, so `not self.errors` is always `True`. Thus redundant. Or am I missing something?

pootle/apps/pootle_store/forms.py
phlax: hmm - i seem to have some mem of it being necessary - but i can check and remove if ok - i cant think why it would matter off hand

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+
+    def __init__(self, *la, **kwa):
+        self.request_user = kwa.pop("user")
+        super(UnitSearchForm, self).__init__(*la, **kwa)
+        self.fields["modified-since"] = ISODateTimeField(required=False)
+
+    def clean_user(self):
+        if self.cleaned_data["user"] is None:
+            if self.request_user.is_anonymous():
+                return get_user_model().get(self.request_user)
```
ta2-1: seems we can leave this line only and remove one lines above and below.

pootle/apps/pootle_store/forms.py
unho: As said, please add this in a comment in code.

pootle/apps/pootle_store/forms.py
unho: ok

pootle/apps/pootle_store/forms.py
unho: If the form is just used for validation nothing of this is actually necessary.

pootle/apps/pootle_store/form_fields.py
unho: Probably I don't.

pootle/apps/pootle_store/forms.py
phlax: k - remmed - i would like to get rid of pootle_misc.forms.SearchForm - but ill add this when that is done

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,129 @@ def save(self, **kwargs):
...
+        choices=CATEGORY_CODES.items())
+    month = forms.DateField(
+        required=False,
+        input_formats=['%Y-%m'])
+    sort = forms.ChoiceField(
+        required=False,
+        choices=UNIT_SEARCH_SORT_CHOICES)
+
+    user = forms.ModelChoiceField(
+        queryset=get_user_model().objects.all(),
```
ta2-1: I was thinking if we need here `default`, `nobody`. But seems they are necessary.

phlax: i believe so

pootle/apps/pootle_store/forms.py
phlax: this is a pretty standard pattern - not something we should be concerning ourselves with documenting

pootle/apps/pootle_store/forms.py
phlax: the previous lines remove certain types of errors

this line returns if there are errors

ta2-1: `self.errors == False` always here. So checking it has no sense.

phlax: this is not true

pootle/apps/pootle_store/forms.py
```diff
@@ -373,3 +407,110 @@ def save(self, **kwargs):
...
+        required=False,
+        widget=CommaSeparatedCheckboxSelectMultiple,
+        choices=(
+            ('source', _('Source Text')),
+            ('target', _('Target Text')),
+            ('notes', _('Comments')),
+            ('locations', _('Locations'))),
+        initial=['source', 'target'])
+
+    def __init__(self, *la, **kwa):
```
unho: @phlax You forgot to change this.

phlax: apologies - i changed it in form_fields - ill add cleanup now

PR #4495, opened 2016-02-17
Get units tests (continued) and assorted test cleanups/fixes

phlax:
- Improves handling of uids and initial params
- Adds tests(s) for non-initial get_units requests

### DIFF
--- a/pytest_pootle/env.py
+++ b/pytest_pootle/env.py
@@ -392,6 +392,7 @@ class PootleTestEnv(object):
         submissions.update(creation_time=update_time)
 
     def _add_submissions(self, unit, created):
+        from pootle_statistics.models import SubmissionTypes
         from pootle_store.models import UNTRANSLATED, FUZZY, OBSOLETE, Unit
 
         from django.contrib.auth import get_user_model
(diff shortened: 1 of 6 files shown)


pytest_pootle/env.py
```diff
@@ -440,5 +409,18 @@ def _add_submissions(self, unit, created):
             unit.markfuzzy()
 
         # mark OBSOLETE
-        if original_state == OBSOLETE:
+        elif original_state == OBSOLETE:
             unit.makeobsolete()
+
+        elif unit.target:
+            # edit unit again
```
unho: So this is changing translation for already translated unit, isn't it?

phlax: yep

pytest_pootle/search.py
```diff
@@ -23,18 +23,22 @@
...
 
 def calculate_search_results(kwargs, user):
-    initial = kwargs["initial"]
     pootle_path = kwargs["pootle_path"]
 
     category = kwargs.get("category")
     checks = kwargs.get("checks")
+    initial = kwargs.get("initial", False)
     limit = kwargs.get("count", 9)
     modified_since = kwargs.get("modified_since")
```
unho: Is it not necessary to alter this too to `modified-since`?

phlax: hmm - we could do it instead i think

phlax: and do the parsing here too, actually

pytest_pootle/env.py
unho: Then I would rewrite this particular comment as it is not 100% clear to me.

unho:
lgtm

ta2-1:
gtm

PR #4498, opened 2016-02-17
UnitForm: no need to handle id and index

julen:
ID is passed around in the URL when making the request, and in no case it needs
to be updated, same as for index. These were needed when the editor wasn't
XHR-based.

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -204,9 +204,8 @@ def unit_form_factory(language, snplurals=None, request=None):
     class UnitForm(forms.ModelForm):
         class Meta(object):
             model = Unit
-            fields = ('id', 'index', 'target_f', 'state',)
+            fields = ('target_f', 'state',)
 
-        id = forms.IntegerField(required=False)
         target_f = MultiStringFormField(
             nplurals=tnplurals,
             required=False,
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/forms.py
```diff
@@ -204,9 +204,8 @@ def unit_form_factory(language, snplurals=None, request=None):
     class UnitForm(forms.ModelForm):
         class Meta(object):
             model = Unit
-            fields = ('id', 'index', 'target_f', 'state',)
```
phlax: im not sure if i understand correctly, but do you need to keep 'index' ?

ta2-1:
@julen is it worth to know why we don't need it any more?

pootle/apps/pootle_store/forms.py
julen: There's no use in keeping it, that's why it's being removed.

phlax: apologies - i see now

unho:
Is the Travis failure related?

julen:
@ta2-1 as mentioned in the description/commit message, there's no scenario from the web UI one would ever want to alter id/indexes, hence it makes no sense to have them listed in the `ModelForm`.

phlax:
>  Is the Travis failure related?

@unho - no unrelated - its an ES server failure

phlax:
lgtm

ta2-1:
lgtm

Issue #3459, opened 2015-01-09
Completely get rid of profile references


unho:
Like `request.profile`.

jleclanche:
cc @julen 
This is something I addressed over several commits in master, including:

d355b9a76a8fece705ab4cce31e892fdec851520
967925231e24d837590cc08051207b61c351feef
a4a0ee6688c84ca57c501ec7b751a9e24e0a8096
29aca838fe8276fcefa020d63b13c613ba7ab162
a433c7f4e1ef82a347c97a1ee74742e1b29317cb
eca2333d07da5a7168b98153c7499a7944137f1d
18cf1f826d4e90e0928e9d4e00eff75e9e901619

Are you ok with this approach?

julen:
Being from an era where `PootleProfile` existed, I don't see *much* relevance in those commits. Most of the references were already changed when migrating away from `PootleProfile`.  #3455 is relevant here.

Another thing to note is that blindly renaming `request.profile` → `request.user` doesn't work, because when anonymous users are involved the *nobody* user has to be retrieved (Django otherwise returns its own `AnonymousUser`).

jleclanche:
Noted. Will see what can safely be moved and write down the rest.

julen:
My initial idea for this was overriding `request.user` in a middleware, doing `request.user = User.get(request.user)` — the class method takes care of returning the *real* user. From that point onwards it'd be safe to use `request.user` everywhere, plus there's no need to handle this special-casing in views.

I didn't have time to address (and test) this because there were other priorities at the time, so I'm not sure if this has other unwanted consequences or conflicts with other middlewares in the pipe. Also note that `request.user` is a proxy (`SimpleLazyObject`) to the actual objects.

unho:
I fail to see the relevance of https://github.com/translate/pootle/issues/3455 here, but I agree with the rest.

julen:
The title says *Completely get rid of profile references* and #3455 is just another (type of) reference to *profile*.

PR #4500, opened 2016-02-18
Optimization: only override request.user if user.is_anonymous

phlax:
This sets request.profile to request.user to maintain compatibility

### DIFF
--- a/pootle/core/views.py
+++ b/pootle/core/views.py
@@ -74,9 +74,11 @@ def set_permissions(f):
     @functools.wraps(f)
     def method_wrapper(self, request, *args, **kwargs):
         User = get_user_model()
-        request.profile = User.get(self.request.user)
+        if request.user.is_anonymous():
+            request.user = User.get(request.user)
+        request.profile = request.user
         request.permissions = get_matching_permissions(
-            request.profile,
+            request.user,
             self.permission_context) or []
         return f(self, request, *args, **kwargs)
     return method_wrapper


ta2-1:
@phlax can you recall me why we'll get rid of `request.profile` and use `request.user` instead?

phlax:
@ta2-1 overriding the request.user is bad from a security pov, but also performance

using obscure vars ie request.profile is just another thing to learn/code/test - when django takes care of this already

what django also does is optimize anon views by *avoiding* a db call

unfortunately this PR doesnt take care of that - but i think we should asap

pootle/core/views.py
```diff
@@ -74,9 +74,11 @@ def set_permissions(f):
     @functools.wraps(f)
     def method_wrapper(self, request, *args, **kwargs):
         User = get_user_model()
-        request.profile = User.get(self.request.user)
+        if request.user.is_anonymous():
+            request.user = User.get(request.user)
```
unho: Perhaps we should be updating `self.request.user` as well.

phlax: its the same object

unho: I should check before asking.

phlax: also this allows the deco to be used if self.request has not been set

unho:
Ideally we should be getting rid of `request.profile` (see #3459).

lgtm

phlax:
> Ideally we should be getting rid of request.profile (see #3459).

agreed.

pootle/core/views.py
```diff
@@ -74,9 +74,11 @@ def set_permissions(f):
     @functools.wraps(f)
     def method_wrapper(self, request, *args, **kwargs):
         User = get_user_model()
-        request.profile = User.get(self.request.user)
```
ta2-1: @julen, as I understand we made extra query here. Now we make this query for anonymous only. This is optimisation, is this?

ta2-1: hmm... I checked now and I understand your comment
> user.is_anonymous() makes no DB calls and user.is_authenticated()
so I'm not sure now if this change was necessary. I'll check this in `pdb`.

julen: From my inspection and understanding it's not — I raised #4508 addressing it.

PR #4501, opened 2016-02-18
Test cleaups

phlax:
removing unused code and fixtures from pytest_pootle and tests

### DIFF
--- a/pytest_pootle/fixtures/models/store.py
+++ b/pytest_pootle/fixtures/models/store.py
@@ -205,25 +205,19 @@ def _require_store(tp, po_dir, name):
 
 
 def _create_submission_and_suggestion(store, user,
-                                      filename=None,
                                       units=None,
                                       suggestion="SUGGESTION"):
 
     from pootle.core.models import Revision
 
-    from tests.models.store import _update_from_upload_file
-
     # Update store as user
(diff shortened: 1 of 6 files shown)


pytest_pootle/fixtures/models/store.py
```diff
@@ -205,25 +205,19 @@ def _require_store(tp, po_dir, name):
...
                                       units=None,
                                       suggestion="SUGGESTION"):
 
     from pootle.core.models import Revision
 
-    from tests.models.store import _update_from_upload_file
-
     # Update store as user
-    if filename:
-        _update_from_upload_file(store, filename, user=user)
```
ta2-1: don't we need test submissions with `UPLOAD` type?

phlax: @ta2-1 see tests in ``tests/models/store.py`` it is tested

unho: Perhaps it is cleaner to have a separate test for that.

unho: Now I see that there is a test for that.

unho:
lgtm

ta2-1:
gtm

PR #4503, opened 2016-02-18
Further export command tests

dwaynebailey:
Gets us close to 100% for export command

### DIFF
--- a/tests/commands/export.py
+++ b/tests/commands/export.py
@@ -13,27 +13,27 @@ from django.core.management import call_command
 
 @pytest.mark.cmd
 @pytest.mark.django_db
-def test_export_noargs(capfd, en_tutorial_po_member_updated):
+def test_export_noargs(capfd):
     """Export whole site"""
     call_command('export')
     out, err = capfd.readouterr()
-    assert 'Created tutorial-en.zip' in out
+    assert 'Created project0-language0.zip' in out
 
 
 @pytest.mark.cmd
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


tests/commands/export.py
```diff
@@ -31,9 +32,52 @@ def test_export_project(capfd, afrikaans_tutorial, french_tutorial):
 
 @pytest.mark.cmd
 @pytest.mark.django_db
+def test_export_project_and_lang(capfd, afrikaans_tutorial, french_tutorial):
```
phlax: no need to use these fixtures use ``project0`` and ``language0``

tests/commands/export.py
```diff
@@ -31,9 +32,52 @@ def test_export_project(capfd, afrikaans_tutorial, french_tutorial):
...
     """Export a language"""
     call_command('export', '--language=af')
     out, err = capfd.readouterr()
     assert 'af.zip' in out
     assert 'fr.zip' not in out
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_export_path(capfd, afrikaans_tutorial, french_tutorial):
```
phlax: same as above

unho:
lgtm, but mind comments by others.

tests/commands/export.py
dwaynebailey: @phlax can you give me a quick lesson then, I'm not sure I understand the new fixtures

phlax: yep np

PR #4509, opened 2016-02-19
Get rid of profile references and use request.user everywhere

### LINKED ISSUE #3459: Completely get rid of profile references

julen:
I had to duplicate a _minimal_ part of `django.contrib.auth.middleware` but I think it's an acceptable trade-off. Now one can simply use `request.user` and forget about `request.profile` altogether.

Fixes #3459.

### DIFF
--- a/pootle/middleware/auth.py
+++ b/pootle/middleware/auth.py
@@ -22,8 +22,9 @@ from django.utils.functional import SimpleLazyObject
 
 def get_user(request):
     if not hasattr(request, '_cached_user'):
-        current_user = auth.get_user(request)
-        request._cached_user = auth.get_user_model().get(current_user)
+        user = auth.get_user(request)
+        request._cached_user = (user if user.is_authenticated() else
+                                auth.get_user_model().objects.get_nobody_user())
     return request._cached_user
 
 
--- a/pootle/middleware/auth.py
+++ b/pootle/middleware/auth.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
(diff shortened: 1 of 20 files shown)


phlax:
lgtm - havent done in clicky testing - nice one for cleaning this up!

unho:
Spoke too soon on #4508.

pootle/middleware/auth.py
```diff
@@ -0,0 +1,38 @@
...
+# AUTHORS file for copyright and authorship information.
+
+"""
+Custom Pootle authentication middleware that takes care of returning Pootle's
+special `nobody` user instead of Django's own `AnonymousUser`.
+
+The code has been slightly adapted from django.contrib.auth.middleware.
+
+Note this customization would probably be unnecessary if there was a fix for
+https://code.djangoproject.com/ticket/20313
```
unho: This ticket being fixed also will render unnecessary `User.get` so perhaps it is a good idea to put some fixme note there as well pointing to this ticket.

unho:
I would also replace `profile` references in https://github.com/translate/pootle/blob/master/pootle/apps/pootle_profile/templatetags/profile_tags.py#L17-L18

Did some click testing and seems to work fine.

lgtm once nits addressed.

pootle/middleware/auth.py
julen: Thanks for the pointer @unho. In fact, I ended up removing the `User.get` method as it has a single and very specific use which is fine to do directly in the middleware, as opposed to the previous scenario where it was necessary to do it manually in view code. So fortunately there's no need to be concerned about it anymore :smile:

unho: Nice!

PR #4510, opened 2016-02-19
Get units backend

phlax:
- Allows for a configurable SearchBackend
-  used by get_units
-  implements DBSearchBackend


ta2-1:
gtm

PR #4511, opened 2016-02-20
Cleanup: remove unused utils

phlax:
these functions dont appear to be used anywhere

in order to find these i have looked through the untested lines on coveralls, found functions and methods that are not used at all, and then grepped the code to ascertain if they are being used and just not being tested.

afaict all of these are unused.

there may be some mistakes or i may have misunderstood how some things work - apologies in advance if that is the case.


julen:
Thank you @phlax, these look good to me, and it's nice to see them go!

Before merging the commit removing the lock though I'd prefer to have at least a couple pair of eyes check and test that. Note there's also the `LOCKED` _constant_ at the top of the module which can be removed.

phlax:
yep - ill remove - and also add a WIP header to this PR - i should have done so already

phlax:
@julen if you had 10 mins for a gitter chat that would be really handy

julen:
Note the `dateformat` template tag is not dead code (at least for the time being), so you should only remove `locale_reverse_align` in the last commit.

dwaynebailey:
Did a quick test.  GTM with the revert the @julen raised above.

PR #4514, opened 2016-02-21
Get units values

phlax:
Adds classes for parsing Unit.values_lists from get_units searches


pootle/apps/pootle_store/unit/results.py
```diff
@@ -0,0 +1,121 @@
...
+    @property
+    def target_lang(self):
+        return self.unit[
+            "store__translation_project__language__code"]
+
+    @property
+    def translate_url(self):
+        store_url = u''.join(
+            [reverse("pootle-tp-store-translate",
+                     args=split_pootle_path(self.pootle_path))])
```
unho: Nit: Why not just `str()` it?

phlax: agreed - i coped out this code - ill update

pootle/apps/pootle_store/unit/results.py
```diff
@@ -0,0 +1,121 @@
...
+        self.units = units
+
+    @property
+    def data(self):
+        meta = None
+        units_list = []
+
+        for unit in iter(self.units):
+            unit = UnitResult(unit)
+            if meta is None:
```
unho: Doing this for all units looks like a bit inefficient, so better do it once before `for`.

unho:
lgtm once nits addressed.

pootle/apps/pootle_store/unit/results.py
phlax: no - at least in the current implementation it needs the unit.

pootle/apps/pootle_store/views.py
```diff
@@ -240,38 +241,6 @@ def _prepare_unit(unit):
...
-            meta = {
-                'source_lang': project.source_language.code,
-                'source_dir': project.source_language.direction,
-                'target_lang': tp.language.code,
-                'target_dir': tp.language.direction,
-                'project_code': project.code,
-                'project_style': project.checkstyle,
-            }
-
-        units_list.append(_prepare_unit(unit))
```
ta2-1: @phlax, will we get rid of _prepare_unit?

phlax: i think so - but its still used elsewhere atm

ta2-1:
lgtm

PR #4516, opened 2016-02-21
Denormalize Unit

phlax:
Denormalize to Unit:
- Store.pootle_path

This allows us to add an index on pootle_path,index for correct/fast sort ordering

### DIFF
--- a/pootle/apps/pootle_store/migrations/0008_unit_pootle_path.py
+++ b/pootle/apps/pootle_store/migrations/0008_unit_pootle_path.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_store', '0007_case_sensitive_schema'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='unit',
+            name='pootle_path',
+            field=models.CharField(default='', unique=False, max_length=255, verbose_name='Path', db_index=True),
+            preserve_default=False,
+        ),
+    ]
--- a/pootle/apps/pootle_store/migrations/0009_denormalize_pootle_path.py
+++ b/pootle/apps/pootle_store/migrations/0009_denormalize_pootle_path.py
@@ -1,0 +1,24 @@
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def denormalize_pootle_path(apps, schema_editor):
+
+    Store = apps.get_model("pootle_store", "Store")
+    Unit = apps.get_model("pootle_store", "Unit")
+    paths = Store.objects.values_list("pootle_path", flat=True)
+    for pootle_path in paths.iterator():
+        Unit.objects.filter(
+            store__pootle_path=pootle_path).update(pootle_path=pootle_path)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_store', '0008_unit_pootle_path'),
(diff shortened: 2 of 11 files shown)


pootle/apps/pootle_store/receivers.py
```diff
@@ -0,0 +1,26 @@
...
+# AUTHORS file for copyright and authorship information.
+
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+from pootle_store.models import Store
+
+
+@receiver(pre_save, sender=Store)
+def store_unit_pre_save_handler(sender, instance, **kwargs):
```
ta2-1: @phlax, I'm not sure how `store.pootle_path` can be changed. I think it can't be atm.

phlax: weve had issues in the past when projects have been renamed or similar - i was kinda hoping this would catch that.

ta2-1: I think if we want to catch Project renaming (I mean a change of a project code) then we have to add such receiver for the Project class first.

pootle/apps/pootle_store/migrations/0008_unit_pootle_path.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
unho: Nit: I would rename this migration to `add_unit_pootle_path_field`.

pootle/apps/pootle_store/migrations/0009_denormalize_pootle_path.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import unicode_literals
```
unho: Nit: I would rename this migration to `populate_denormalized_unit_pootle_path`.

pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.utils.db import set_mysql_collation_for_column
+
+
+def make_unit_paths_cs(apps, schema_editor):
```
unho: Please rename to `make_unit_paths_case_sensitive`.

pootle/apps/pootle_store/receivers.py
unho: I agree that this is unnecessary for now.

pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
phlax: no - that is not all it does

pootle/apps/pootle_store/receivers.py
phlax: sorry - this is about the relationship between store and unit. I agree we need to look at what happens when a Project code is renamed but its outside of the scope of this PR, this is not.

pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
unho: `cs` is difficult to understand.

phlax: sorry - i understand - you mean the function - rather than the filename

unho: I mean the function name.

phlax:
I have consolidated this PR in #4527

PR #4519, opened 2016-02-23
Move PROJECT_CHECKERS constant from sim named var in ttk

### LINKED ISSUE #4517: Move projectcheckers from ttk to pootle
currently the `Project.checkers` field uses `translate.filters.checks.project_checkers` to set its choices.

while this is not exactly "dynamic" it creates a problem if ttk/pootle versions dont match and pootle wants to create a migration.

i would propose that the best solution is to move projectcheckers into `pootle_project`, and remove the validation step here:
 https://github.com/translate/translate/blob/master/translate/storage/poheader.py#L356

...on the basis that its not ttk's job to verify what is a valid projectchecker for this particular Pootle Project.

phlax:
Fixes #4517

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -41,6 +41,17 @@ from staticpages.models import StaticPage
 
 
 RESERVED_PROJECT_CODES = ('admin', 'translate', 'settings')
+PROJECT_CHECKERS = {
+    "standard": checks.StandardChecker,
+    "openoffice": checks.OpenOfficeChecker,
+    "libreoffice": checks.LibreOfficeChecker,
+    "mozilla": checks.MozillaChecker,
+    "kde": checks.KdeChecker,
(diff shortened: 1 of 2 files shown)


unho:
This fails because you removed a var referenced in https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/views/admin/projects.py#L38

phlax:
@unho - good spot - cool ill fix

ta2-1:
@phlax I think then [this](https://github.com/translate/pootle/blob/37fd7e5fb299d8a86093aac9359c6f2964fafc3e/pytest_pootle/fixtures/models/translation_project.py#L83) should be adjusted also.

phlax:
> @phlax I think then this should be adjusted also.

yep - you are right!

phlax:
@ta2-1 @unho updated - shall i land?

unho:
lgtm

PR #4520, opened 2016-02-23
Fix: UnitResult gets nplural data and passes that pluralize_target util

phlax:
prevents pluralize_target from trying to fetch nplural data for a UnitResult

### DIFF
--- a/pootle/apps/pootle_store/unit/results.py
+++ b/pootle/apps/pootle_store/unit/results.py
@@ -22,6 +22,11 @@ from pootle_store.unit.proxy import UnitProxy
 class UnitResult(UnitProxy):
 
     @property
+    def nplurals(self):
+        return self.unit[
+            "store__translation_project__language__nplurals"] or 0
+
+    @property
     def pootle_path(self):
         return self.unit["store__pootle_path"]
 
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_store/unit/results.py
```diff
@@ -22,6 +22,11 @@
 class UnitResult(UnitProxy):
 
     @property
+    def nplurals(self):
+        return self.unit[
+            "store__translation_project__language__nplurals"] or 0
```
unho: This default surely is not the best choice. Perhaps it is better to use 1 or 2.

phlax: i think the default is correct - im just trying prevent ``None`` from being passed

phlax: ie - this is the default on the model

unho: You are right. Sorry.

julen: I think this should fallback to `1`.

julen: Oh, didn't get your messages up to the point. If it's tied to the model, then `0` is good.

julen:
Tested and works fine, thank you for the quick fix!

pootle/apps/pootle_store/unit/results.py
phlax: @julen im not really sure what it should be but im reluctant to make it different to the default on the model - https://github.com/translate/pootle/blob/master/pootle/apps/pootle_language/models.py#L92

PR #4521, opened 2016-02-23
Docs: Use latest Sphinx version and other cleanups

dwaynebailey:
* Removes the upper bound of Sphinx - by fixing the ``:options`` markers.
* Other cleanups


docs/developers/styleguide.rst
```diff
@@ -324,11 +324,13 @@ Quoting
...
 
-
         <!-- Bad -->
+        <a href='{% url "whatever" %}' class='highlight'>
+
+    .. code-block:: text
+
         <a href="{% url "whatever" %}" class="highlight">
         <a href='{% url 'whatever' %}' class='highlight'>
-        <a href='{% url "whatever" %}' class='highlight'>
```
unho: What was wrong with the previous way?

docs/server/commands.rst
```diff
@@ -685,9 +745,8 @@ collectstatic
 Running the Django admin :djadmin:`django:collectstatic` command finds and
 extracts static content such as images, CSS and JavaScript files used by the
 Pootle server, so that they can be served separately from a static webserver.
-Typically, this is run with the :option:`--clear` :option:`--noinput` options,
-to flush any existing static data and use default answers for the content
-finders.
+Typically, this is run with the ``--clear`` ``--noinput`` options, to flush any
```
unho: I think it is ok to put both in the same highlight since the idea is to specify both at the same time.

docs/releases/2.7.3b2.rst
```diff
@@ -20,6 +20,8 @@ Details of changes
   usernames.
 - :djadmin:`refresh_scores` now recalculates user scores and accept
   multiple usernames.
+- :djadmin:`refresh_stats` option ``--calculate-checks`` was removed.  Use
```
unho: Actually several other options were dropped when removing old `refresh_stats` https://github.com/translate/pootle/commit/def9bf0029b4a35b73ad07e5cea094f61d6feb86#diff-083f311bbadfbeedbdbc2c70b180f198L46 But this change was already included in 2.7.1.

docs/developers/styleguide.rst
dwaynebailey: The HTML parser rejects it.  And when it rejects it the parsing fails and you don't get highlighting. Is the commit message not clear enough?

> Docs: invalid HTML cannot be highlighted
Place the invalid HTML in a plain text rendering section.

unho: Then please put all bad examples in the text code-block instead of spreading them in two separate code-blocks.

PR #4522, opened 2016-02-24
[RFC] Cleanup requirements

dwaynebailey:
Would like some input with these proposed changes.  Summary here:

* Requirement listed in one place - they get out of sync and some of ours where already out of sync between dev and travis.
* Distinguish between private ``_x.txt`` and target ``x.txt`` build requirements.  Target are those you're likely to want, private are those on which various target requirements depend.  i.e. ``_lint.txt`` vs ``dev.txt``

### DIFF
--- a/requirements/_docs.txt
+++ b/requirements/_docs.txt
@@ -1,0 +1,3 @@
+# Documentation
+
+sphinx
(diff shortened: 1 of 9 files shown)


phlax:
lgtm - although i would say that ``tests`` should be a public target in its own right

PR #4527, opened 2016-02-25
[WIP] Denormalize Unit pootle_path, project and lang and add index

phlax:
Denormalize project to unit to allow fast filtering and indexing together with language.
BEFORE: [version](https://github.com/translate/pootle/tree/2c525a484c1d81512639641adb784a095ae6f41c) before any get_units refactoring.
AFTER: current PR.
The timings are specifically for the get_units db call - with the uIds call switched off.

### /projects/translate
AFTER: ~~.004s~~ .002s
BEFORE: ∞ 

### /projects/firefox/translate
AFTER: ~~.004s~~ .002s
BEFORE: ~~154s~~, >5s

### /ru/firefox/translate
AFTER: ~~.29s~~ ~~.31s~~ .002s
BEFORE: 0.02s

### /ru/translate/
AFTER: ~~.29s~~ ~~.31s~~ .012s
BEFORE: 0.05s

### /projects/firefox/translate/browser/branding/official/brand.dtd.po
AFTER: ~~.26s~~ ~~.248s~~ .23s
BEFORE: 0.261s

### /ru/firefox/translate/browser/branding/official/brand.dtd.po
AFTER: ~~.001s~~ .001s
BEFORE: .001s

### /projects/firefox/translate/browser/branding/
AFTER: ~~.114s~~ ~~.104s~~ .099s
BEFORE: 0.2s

### /ru/firefox/translate/browser/branding/
AFTER: ~~137s~~ .001s
BEFORE: .001s

### DIFF
--- a/pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
+++ b/pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.utils.db import set_mysql_collation_for_column
+
+
+def make_unit_paths_cs(apps, schema_editor):
+    cursor = schema_editor.connection.cursor()
+    set_mysql_collation_for_column(
+        apps,
+        cursor,
+        "pootle_store.Unit",
+        "pootle_path",
+        "utf8_bin",
+        "varchar(255)")
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_store', '0009_denormalize_pootle_path'),
+    ]
+
+    operations = [
+        migrations.RunPython(make_unit_paths_cs),
+        migrations.AlterIndexTogether(
+            name='unit',
+            index_together=set([('pootle_path', 'index')]),
+        ),
+        migrations.AlterModelOptions(
+            name='unit',
+            options={'ordering': ['pootle_path', 'index'], 'get_latest_by': 'mtime'},
+        ),
+    ]
--- a/pootle/apps/pootle_store/migrations/0016_denormalize_project.py
+++ b/pootle/apps/pootle_store/migrations/0016_denormalize_project.py
@@ -1,0 +1,31 @@
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def denormalize_project(apps, schema_editor):
+    Unit = apps.get_model("pootle_store", "Unit")
+    Store = apps.get_model("pootle_store", "Store")
+    stores = Store.objects.values_list(
+        "pk", "translation_project__project_id")
+
+    for store_id, project_id in stores.iterator():
+        if not project_id:
+            continue 
+        units = Unit.objects.filter(
+            store_id=store_id).exclude(project_id=project_id)
+        if units.count() == 0:
+            continue
+        units.update(project_id=project_id)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_translationproject', '0002_remove_translationproject_disabled'),
+        ('pootle_store', '0015_unit_project')
+    ]
+
+    operations = [
+        migrations.RunPython(denormalize_project),
+    ]
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -511,8 +511,8 @@ class Unit(models.Model, base.TranslationUnit):
 
     def delete(self, *args, **kwargs):
         action_log(user='system', action=UNIT_DELETED,
-                   lang=self.store.translation_project.language.code,
-                   unit=self.id, translation='', path=self.store.pootle_path)
+                   lang=self.language.code,
+                   unit=self.id, translation='', path=self.pootle_path)
 
         self.flag_store_before_going_away()
 
@@ -577,9 +577,9 @@ class Unit(models.Model, base.TranslationUnit):
 
         if not created and hasattr(self, '_save_action'):
             action_log(user=self._log_user, action=self._save_action,
-                       lang=self.store.translation_project.language.code,
+                       lang=self.language.code,
                        unit=self.id, translation=self.target_f,
-                       path=self.store.pootle_path)
+                       path=self.pootle_path)
 
         if (self._state_updated and self.state == TRANSLATED and
             self._save_action == TRANSLATION_CHANGED and
@@ -608,9 +608,9 @@ class Unit(models.Model, base.TranslationUnit):
                 self.store.mark_dirty(CachedMethods.WORDCOUNT_STATS)
 
             action_log(user=self._log_user, action=self._save_action,
-                       lang=self.store.translation_project.language.code,
+                       lang=self.language.code,
                        unit=self.id, translation=self.target_f,
-                       path=self.store.pootle_path)
+                       path=self.pootle_path)
 
             self.add_initial_submission(user=user)
 
@@ -642,7 +642,7 @@ class Unit(models.Model, base.TranslationUnit):
                '#unit=%s' % unicode(self.id)))
 
     def get_search_locations_url(self, **kwargs):
-        lang, proj, dir, fn = split_pootle_path(self.store.pootle_path)
+        lang, proj, dir, fn = split_pootle_path(self.pootle_path)
 
         return u''.join([
             reverse('pootle-project-translate', args=[proj, dir, fn]),
@@ -650,8 +650,7 @@ class Unit(models.Model, base.TranslationUnit):
         ])
 
     def get_screenshot_url(self):
-        prefix = self.store.translation_project.\
-            project.screenshot_search_prefix
+        prefix = self.project.screenshot_search_prefix
         if prefix:
             return prefix + urlquote(self.source_f)
 
@@ -665,7 +664,7 @@ class Unit(models.Model, base.TranslationUnit):
 
         from pootle_project.models import Project
         user_projects = Project.accessible_by_user(user)
-        return self.store.translation_project.project.code in user_projects
+        return self.project.code in user_projects
 
     def add_initial_submission(self, user=None):
         if self.istranslated() or self.isfuzzy():
@@ -758,7 +757,7 @@ class Unit(models.Model, base.TranslationUnit):
 
         if unit.target != self.target:
             if unit.hasplural():
-                nplurals = self.store.translation_project.language.nplurals
+                nplurals = self.language.nplurals
                 target_plurals = len(self.target.strings)
                 strings = self.target.strings
                 if target_plurals < nplurals:
@@ -1001,8 +1000,8 @@ class Unit(models.Model, base.TranslationUnit):
             'id': self.id,
             # 'revision' must be an integer for statistical queries to work
             'revision': self.revision,
-            'project': self.store.translation_project.project.fullname,
-            'path': self.store.pootle_path,
+            'project': self.project.fullname,
+            'path': self.pootle_path,
             'source': self.source,
             'target': self.target,
             'username': '',
@@ -1023,7 +1022,7 @@ class Unit(models.Model, base.TranslationUnit):
                 'email_md5': md5(self.submitted_by.email).hexdigest(),
             })
 
-        get_tm_broker().update(self.store.translation_project.language.code,
+        get_tm_broker().update(self.language.code,
                                obj)
 
     def get_tm_suggestions(self):
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -302,6 +302,13 @@ class Unit(models.Model, base.TranslationUnit):
     unitid_hash = models.CharField(max_length=32, db_index=True,
                                    editable=False)
 
+    # Denormalized from Store.pootle_path
+    # any changes to the `pootle_path` field may require updating the schema
+    # see migration 0007_case_sensitive_schema.py
+    pootle_path = models.CharField(
+        max_length=255, null=False, unique=False,
+        db_index=True, verbose_name=_("Path"))
+
     source_f = MultiStringField(null=True)
     source_hash = models.CharField(max_length=32, db_index=True,
                                    editable=False)
@@ -442,6 +449,8 @@ class Unit(models.Model, base.TranslationUnit):
             User = get_user_model()
             self._log_user = User.objects.get_system_user()
         user = kwargs.pop("user", self._log_user)
+
+        self.pootle_path = self.store.pootle_path
 
         if created:
             self._save_action = UNIT_ADDED
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -318,8 +318,7 @@ class Unit(models.Model, base.TranslationUnit):
     # denormalized TP.project
     project = models.ForeignKey(
         "pootle_project.Project",
-        null=True,
-        blank=True,
+        null=False,
         db_index=True,
         related_name='units')
 
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -262,7 +262,7 @@ class UnitManager(models.Manager):
                 store__translation_project__language__code=language_code)
         else:
             units_qs = units_qs.exclude(
-                store__pootle_path__startswith="/templates/")
+                pootle_path__startswith="/templates/")
 
         if project_code:
             units_qs = units_qs.filter(
@@ -282,17 +282,17 @@ class UnitManager(models.Manager):
         if language_code and project_code:
             if filename:
                 return units_qs.filter(
-                    store__pootle_path=pootle_path)
+                    pootle_path=pootle_path)
             else:
                 return units_qs.filter(
-                    store__pootle_path__startswith=pootle_path)
+                    pootle_path__startswith=pootle_path)
         else:
             # we need to use a regex in this case as lang or proj are not
             # set
             if filename:
                 pootle_path = "%s$" % pootle_path
             return units_qs.filter(
-                store__pootle_path__regex=pootle_path)
+                pootle_path__regex=pootle_path)
 
 
 class Unit(models.Model, base.TranslationUnit):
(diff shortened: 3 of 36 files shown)


pootle/apps/pootle_store/migrations/0010_unit_path_schema.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.utils.db import set_mysql_collation_for_column
+
+
+def make_unit_paths_cs(apps, schema_editor):
```
unho: Please rename this function, since `cs` doesn't mean anything.

pootle/apps/pootle_store/models.py
```diff
@@ -302,6 +324,26 @@ class Unit(models.Model, base.TranslationUnit):
...
+    # denormalized TP.language
+    language = models.ForeignKey(
+        "pootle_language.Language",
+        null=False,
+        db_index=True,
+        related_name='units')
+    # denormalized TP.project
+    project = models.ForeignKey(
+        "pootle_project.Project",
+        null=False,
```
unho: This doesn't allow `NULL` values, but in `016_denormalize_project` migration you are explicitly skipping stores with no project, thus if any of those stores has some unit, the project field for those units will be NULL. Is this not a problem?

unho: Just saw `0011_cleanup_orphaned_units` migration. So not an issue.

pootle/apps/pootle_store/migrations/0016_denormalize_project.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def denormalize_project(apps, schema_editor):
+    Unit = apps.get_model("pootle_store", "Unit")
+    Store = apps.get_model("pootle_store", "Store")
+    stores = Store.objects.values_list(
+        "pk", "translation_project__project_id")
+
+    for store_id, project_id in stores.iterator():
+        if not project_id:
```
unho: Is this a real possibility? If yes we might end with units without project. See https://github.com/translate/pootle/pull/4527/files#r54554808 below.

pootle/core/receivers.py
```diff
@@ -0,0 +1,31 @@
...
+from pootle.core.site import pootle_site
+from pootle_language.models import Language
+from pootle_project.models import Project
+
+
+@receiver(pre_save, sender=Language)
+def language_cache_expiry_pre_save_handler(sender, instance, **kwargs):
+    """Expire pootle.core.site.pootle_site Languages
+    """
+    # import pdb; pdb.set_trace()
```
unho: Nit: Remove commented code.

pootle/apps/pootle_store/models.py
```diff
@@ -248,51 +270,51 @@ def get_translatable(self, user, project_code=None, language_code=None,
...
-            units_qs = units_qs.exclude(
-                store__pootle_path__startswith="/templates/")
-
-        if project_code:
+                language_id=pootle_site.languages[language_code]["pk"])
+        elif project_code:
             units_qs = units_qs.filter(
-                store__translation_project__project__code=project_code)
-        elif not user.is_superuser:
+                language_id__in=pootle_site.projects[project_code]["languages"])
```
unho: Shouldn't this be filtering by `project_id__in`?

phlax: no - its filtering on the languages that have TPs for this project

pootle/apps/pootle_store/migrations/0016_denormalize_project.py
phlax: i think not actually - it was defensive coding before adding the cleanup migration - ill remove and hope it works when i rerun migration 8/

pootle/apps/pootle_store/models.py
unho: Is it not overcomplicating it? I mean, we are filtering units here, so why not:

```python
            units_qs = units_qs.filter(project_id=project_code)
``` 

Perhaps `project_id__in` in my first comment was the root of the confusion. My apologies.

julen:
After several rounds of testing here are the numbers I got, and while the site-wide and TP root query timings greatly improve, there looks to be quite a negative impact in the cross-language and cross-project views.

I was also testing locally on MySQL 5.7 first (it had [some gotchas when migrating](http://blog.wl0.org/2015/12/a-couple-of-mysql-5-7-gotchas-to-be-aware-of/)), although I had to give up because it seems the default configuration changed quite a bit and the overall test runs were taking forever.

<table>
<thead>
<tr>
<th><em>path</em></th>
<th><em>branch</em></th>
<th>MySQL 5.1 (stage)</th>
<th>MySQL 5.6 (local)</th>
</tr>
</thead>
<tbody>
<tr>
<td rowspan="3">/projects/translate/</td>
<td>en-release</td>
<td>74.364s (7q)</td>
<td>97.044s (8q)</td>
</tr>
<tr>
<td>master</td>
<td>129.148s (4q)</td>
<td>118.856s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>4.432s (74q)</td>
<td>4.255s (74q)</td>
</tr>
<tr>
<td rowspan="3">/projects/web_evernote/translate/</td>
<td>en-release</td>
<td>7.909s (5q)</td>
<td>11.551s (6q)</td>
</tr>
<tr>
<td>master</td>
<td>8.334s (5q)</td>
<td>8.764s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>57.837s (4q)</td>
<td>32.598s (5q)</td>
</tr>
<tr>
<td rowspan="3">/ru/web_evernote/translate/</td>
<td>en-release</td>
<td>0.378s (5q)</td>
<td>0.218s (6q)</td>
</tr>
<tr>
<td>master</td>
<td>0.501s (4q)</td>
<td>0.37s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>0.062s (4q)</td>
<td>0.053s (5q)</td>
</tr>
<tr>
<td rowspan="3">/ru/translate/</td>
<td>en-release</td>
<td>2.504s (7q)</td>
<td>4.062s (8q)</td>
</tr>
<tr>
<td>master</td>
<td>4.435s (4q)</td>
<td>2.91s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>19.941s (4q)</td>
<td>9.712s (5q)</td>
</tr>
<tr>
<td rowspan="3">/projects/web_evernote/translate/<br/>gwt/com/evernote/web/gwt/<br/>client/v2/i18n/<br/>FocusMessages.properties.po</td>
<td>en-release</td>
<td>0.404s (5q)</td>
<td>0.253s (6q)</td>
</tr>
<tr>
<td>master</td>
<td>0.609s (4q)</td>
<td>0.521s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>16.664s (4q)</td>
<td>10.148s (5q)</td>
</tr>
<tr>
<td rowspan="3">/ru/web_evernote/translate/<br/>gwt/com/evernote/web/gwt/<br/>client/v2/i18n/<br/>FocusMessages.properties.po</td>
<td>en-release</td>
<td>0.014s (5q)</td>
<td>0.009s (6q)</td>
</tr>
<tr>
<td>master</td>
<td>0.022s (4q)</td>
<td>0.006s (5q)</td>
</tr>
<tr>
<td>PR</td>
<td>0.129s (4q)</td>
<td>0.121s (5q)</td>
</tr>
<tr>
<td rowspan="3">/projects/web_evernote/translate/<br/>gwt/com/evernote/web/<br/>gwt/client/v2/i18n/</td>
<td>en-release</td>
<td>0.851s (6q)</td>
<td>0.618s (7q)</td>
</tr>
<tr>
<td>master</td>
<td>1.399s (5q)</td>
<td>0.945s (6q)</td>
</tr>
<tr>
<td>PR</td>
<td>23.601s (4q)</td>
<td>13.297s (6q)</td>
</tr>
<tr>
<td rowspan="3">/ru/web_evernote/translate/<br/>gwt/com/evernote/web/<br/>gwt/client/v2/i18n/</td>
<td>en-release</td>
<td>0.025s (6q)</td>
<td>0.016s (7q)</td>
</tr>
<tr>
<td>master</td>
<td>0.06s (5q)</td>
<td>0.017s (6q)</td>
</tr>
<tr>
<td>PR</td>
<td>0.017s (4q)</td>
<td>0.041s (6q)</td>
</tr>
</tbody>
</table>

phlax:
@julen can you come on gitter to discuss

iafan:
I know that you guys have put a good deal of time and energy into this PR, but I personally have mixed feelings about this approach. I absolutely agree that we need to improve the performance when working with large installations, but this particular way of addressing the problem seems like a brute-force one: it increases the code complexity and DB size while not giving any reasonable reproducible speedups that everybody would agree with. The most frustrating thing would be for all of us to rush this into production, do the costly migrations only to realize several months later that there's a better approach and we need to undo the code and migrations and implement this in a more scalable fashion. 

I would suggest us to test other approaches before sticking with this PR. For example, we now use ElasticSearch for TM lookups, and it proved to be a great solution in this particular area. Did you consider using ES for indexing and retrieving units as well? I have a feeling that it has a potential of  drastically improving the search speed in all scenarios, without complicating the main codebase much. I bet writing a script that just populates ES index from Pootle units will take an hour or so to implement, and then we could all experiment with it, and run queries that return needed pages of data from the entire data set sorted by various fields (store path + unit index, creation date, etc.). I would strongly suggest doing at least this experiment before deciding what to do with this PR. I might be wrong in my assumptions, but if I am, this will only give extra credibility to this PR and will make all of us feel more easy about pushing it into production.

Again, I don't want to insult anyone with this comment, and you @phlax specifically do a really amazing work in pushing the project forward. I just want to make sure everybody who works on the project agrees on the important architectural decisions, fully understands why they are necessary and what pros/cons these changes bring.

phlax:
hi @iafan thanks for taking the time to comment

I would like to make a few points in respect to this

Current problems irrespective of performance
---

- the current code doesnt order correctly ie it groups units by (pootle_path, index) that havent been ordered that way in the total dataset - leading to various problems/inconsistencies
- some of the ordering is also not what you might expect - in particular oldest/newest is neither the earliest/most recent change/creation time - but when unit was translated (mostly)

Current expectation/desired functionality
---

 - imho and afaict most peoples expectations we *should* order on pootle_path,index as this is basically by_store, by_order in store
- we also expect (for good reason) ordering by other metrics - ie mtime, submitted_time, priority etc


Performance issues
---

- current setup requires complex sorts on not only very large Unit tables, but very large unit sets within that table 
- this is causing *massive* performance issues for us before/after this PR - but defo less so with this PR in place if we expect proper pootle_path,index ordering other issues also need to be resolved to really fix the problem).
 
Performance fixes
---

- Denormalizing fields where bottlenecks occur is good practice
- Adding multi-column indeces for complex/common sorts/filters with MySQL is necessary for large tables and good practice
- Getting the optimal/*minimal* denormalization is defo good practice and im working really hard to try and achieve that
- The performance issues that we are seeing on large datasets are going to be reflected in smaller installs of Pootle also so my feeling is that bad performance is highlighting to us deeper issues rather than just some bad performance on specific TPs on MLO/EN

Using Elasticsearch or other Indexers
---

- Using a (distributed/~async) indexer bring its own complexity - particularly where Units are being edited in quick succession - so this requires some testing
- Making ElasticSearch/OtherIndexer a hard dependency is less conducive to someone installing Pootle easily so i would really rather avoid this if possible (for now at least)
- Local TM is currently not a *hard*-dep so while its useful that we already have those docs/install path etc its not the same as forcing everyone who installs Pootle to install ElasticSearch/OtherIndexer
- Optimizing Indexes/Facets within WhateverIndexers is going to take some time (like lots of very disk IO-intensive time)
- For the record I absolutely see this as the way forward for large Pootle installations in case that is not clear already

iafan:
> Making ElasticSearch/OtherIndexer a hard dependency is less conducive to someone installing Pootle easily so i would really rather avoid this if possible (for now at least)

For small installations, Pootle is doing just fine without denormalizing the data, IMO. For huge installations, there needs to be some indexing in place. I believe we both agree on that.

I think that testing whether ES brings significant improvements is pretty cheap for us, and it really does, no one would ever complain about making ES a dependency. I think that it will potentially be relatively easy to add a code that does ES-based searches if ES is configured while defaulting to the current behavior otherwise to make it a non-hard dependency. But all this optimization and discussion is a bit premature; the point I'm trying to make is: let's spend a day or two trying to get the data into ES and playing with the queries, and see how these compare to your current results. Using any proper index, be it SQL DB-based or ES/whatever-based, should equally address all the sorting issues you outlined above. The question is what approach is best for Pootle in the long run. Of course I wish we did this comparison much earlier, but even now, when the PR is ready, I think its critical for us to evaluate both approaches before making a commitment.

I can spend my time writing such a test code, the only problem would be that it would be in Perl. :)

phlax:
> I think that testing whether ES brings significant improvements is pretty cheap for us

I have spent many, many hours/days/weeks waiting for ES to finish indexing 8/

:bitter experience face:

iafan:
Are you talking about aggregated lifetime experience with ES, or one-time DB indexing frustration? ;)

I don't think any indexing solution would do any magic: you have to spend some time indexing data, and you can either do that in realtime at the expense of insert operation speed, or in a delayed fashion. Obviously, reindexing gigabytes of data from scratch is not something anyone would do often. The query speed is what matters in day-to-day use.

phlax:
@iafan the point about ES is that you as a sysadmin/dev on large datasets get to spend a large part of your life testing/creating indexes to make everyone elses life quicker - its a kind of penance - and no its not a penance ive suffered on a isolated occasion - or only with ElasticSearch for that matter

iafan:
I prefer to consider these things as challenges, not a penance I have to pay for being a software engineer. Makes my life a tad easier.

So back to my proposal: would you guys try to experiment with creating an index creation script + ES queries to test search timing? Or would you prefer me to do that on my side? In any case, I would kindly ask to wait for these test results before merging this PR.

phlax:
@iafan to be 100% clear its not currently on our radar - whereas fixing the db queries is

iafan:
Fixing the queries (for proper sorting) and denormalizing the DB for speed are two separate things. I'm ok with the former and concerned with the latter. The problem with this PR is that it mixes several things.

phlax:
its really not possible to fix one without addressing the other - this PR needs some rebasing/separation - but essentially these are the things (with the caveat that its still being tested) that we have found to be necessary.

iafan:
Ok, I did my homework with ElasticSearch-based indexing: https://github.com/translate/pootle/wiki/Using-ElasticSearch-as-Unit-Index

TL;DR: it is extremely fast on my DB of **2.4M+ units** (for such DB size, you can typically expect retrieving results **under 70 ms** for all units on a decent server hardware; in the majority of real-life cases the results will be served in **under 10 ms**).

Note that I was taking real-life use cases and real production queries to get first 20 sorted results from the entire data set and to get a page of results starting from a large offset. Also note that if we use ES, the unit insert/update speed won't be affected much, as compared to synchronous index updates in MySQL (and this could be a problem with any bulk unit update operations during `update_stores` or file upload).

Though not strictly required in such cases, I feel that ES would give a performance boost even for small Pootle instances, spitting results almost instantly.

@julen, can you please run this on your local machine and your DB so we could compare apples to apples?

phlax:
> TL;DR: it is extremely fast on my DB of 2.4M+ units (for such DB size, you can typically expect retrieving results under 70 ms for all units on a decent server hardware; in the majority of real-life cases the results will be served in under 10 ms).

i think you may not have understood

Yes, ES and other indexers **are extremely fast at retrieving data** its their point

Creating a strategy for what should be indexed is another matter, and requires time, thought, a good understanding of what is happening, and even better understanding of the types of query that it will be used for. Indexing is denormalizing - you still have to decide what to denormalize.

Setting up those queries, making use of faceting or other optimizations/strategies that takes time/expertise.

Waiting for the indexer to index (ie **create** the indeces) - that takes *a lot* of time.

If Evernote are up for contributing this work i would be overjoyed - ive made a start on it here https://github.com/translate/pootle_index, and can happily give pointers on eg using django-haystack.

From a Translate House perspective this is not currently our focus

phlax:
@iafan can also make the suggestion that if Evernote want to contribute to a faster resolution to the deep problems that we currently have, and are causing such massive issues for EN and TH customers - then fixing the uId problem would be far far more helpful

iafan:
> Waiting for the indexer to index (ie create the indeces) - that takes a lot of time.

You can query ES index right after it gets populated. Which is, as I stated in the report, happens in under 6 minutes for the entire database of 2.4M+ units (and you don't have to reindex the entire DB every now and then).

> Creating a strategy for what should be indexed is another matter, and requires time, thought, a good understanding of what is happening, and even better understanding of the types of query that it will be used for. Indexing is denormalizing - you still have to decide what to denormalize.

With ES, implementing this is much easier than doing the same things in SQL databases. Adding a new denormalized field now or in the future is easy and requires no expensive migrations. I also assume that you have good understanding of what is happening and the types of queries you're going to execute, otherwise you wouldn't be working on low-level DB optimizations in the first place. I can definitely help you with the ES strategy part if you need it. We simply weren't in the loop on any architectural decisions recently, and this is not our fault we couldn't provide our expertise in time.

> From a Translate House perspective this is not currently our focus

I'm sorry, but this is not an approach to the problem I would expect, Ryan. This PR (and I'm not the only one who shares the same concerns) potentially **worsens** the problem in certain scenarios (as per Julen's real-life tests, not the synthetic tests you provide at the very beginning); you don't test this thoroughly against important metrics (like getting units starting with some large offset, or speed degradation of insert operations). I'm trying to prove that there is a solution that is guaranteed to fix the performance issues without complicating the core and DB state much and will be easier to maintain in the long run. Our focus is to make sure we deliver quality product, and I hope everybody shares the same vision.

julen:
> @julen, can you please run this on your local machine and your DB so we could compare apples to apples?

I tested your script and queries against the current DB I have (2973954 units) and I can confirm the indexing time stays below 6 minutes. Needless to say query performance is great (it even outperforms your numbers), but that's beyond the point.

phlax:
hi @julen it would be really useful to have a quick chat about this PR, or more importantly what it was trying to achieve - if you have the chance.

phlax:
thankfully this PR was superseded by a different approach

julen:
Hey @phlax, for future readers it'd be good to summarize why the approach was abandoned.

PR #4529, opened 2016-02-29
Provide URLs for project/TP admin pages

### LINKED ISSUE #4524: Can’t navigate to browse/translate from admin pages
1. Go to a project
2. Go to TP admin using the dropdown.  Browse->Languages
3. Return to the Browse or Translate view.  Nothing happens
4. Move to the Permissions view, we move
5. Try move to Browse or Translate view again.  Nothing happens.

julen:
Fixes #4524.

### DIFF
--- a/pootle/apps/pootle_language/views.py
+++ b/pootle/apps/pootle_language/views.py
@@ -7,6 +7,7 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from django.core.urlresolvers import reverse
 from django.http import Http404
 from django.shortcuts import redirect, render
 from django.utils.functional import cached_property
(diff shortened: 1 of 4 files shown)


unho:
What about `pootle-projects-export`, `pootle-projects-translate`, `pootle-projects-browse`, `pootle-tp-export`, `pootle-project-export`, `pootle-language-export` removed in https://github.com/translate/pootle/commit/8fe18be59a1409ce21bce77c474cb6b61669f290.

julen:
Export view is only accessible from the editor, which is not affected by the issue, and I don't see where the other `pootle-projects-*` patterns would apply either. Feel free to test and report if that's not the case though.

unho:
@julen If while in a TP browse I go to "Terminology" and then try to go back it doesn't work.


> Export view is only accessible from the editor, which is not affected by the issue, and I don't see where the other `pootle-projects-*` patterns would apply either. Feel free to test and report if that's not the case though.

Tested and these are not necessary as you say.

julen:
Good catch, that should now be fixed, it'd be cool if you could confirm.

unho:
@julen It is fixed now. gtm.

PR #4532, opened 2016-03-02
Cleanup: dont test for debugging messages

phlax:
The code here is testing for messages produced in logging.debug

This relies on debug being True and isnt what we should be testing for (if we really want to test for debugging messages then caplog is the way)


unho:
lgtm

dwaynebailey:
lgtm, interestingly the messages we are now capturing are also coming via logging.

PR #4536, opened 2016-03-02
Docs: Fix instructions on how to run in a subdirectory

### LINKED ISSUE #4397: Issues when pootle is running in a subdirectory
I've recently upgraded to pootle 2.7.3b1 and stumbled upon some issues when pootle is installed in a subdirectory.

Our pootle is running on https://www.dream-multimedia-tv.de/pootle
The first Place where i came across the issue is the password reset mechanism.

The POST request is then trying to send the data to /pootle/pootle/... instead of /pootle/...

The second is that the "last action" in /projects is missing the /pootle and points to e.g. /de/... instead of /pootle/de/...

I've worked around the password reset issue using an internal rewrite rule. The second one is not pressing at all but as our pootle is accessed through a reverse proxy it's not fixable in using a simple redirect or rewrite

unho:
I was unable to verify if this actually works.

Fixes #4397.

### DIFF
--- a/docs/server/apache-virtualhost.conf
+++ b/docs/server/apache-virtualhost.conf
@@ -14,8 +14,12 @@ WSGIPythonOptimize 1
     #   not end with a slash):
     #   WSGIScriptAlias /your-pootle /var/www/pootle/wsgi.py
     # - Change the Alias directive for 'assets' to include the '/your-pootle'.
-    # - Include the following setting in your custom Pootle settings:
+    # - Include the following settings in your custom Pootle settings:
     #   STATIC_URL = '/your-pootle/assets/'
+    #   FORCE_SCRIPT_NAME = '/your-pootle'
+    # - If you have previously calculated the stats:
+    #   - Restart the RQ workers.
+    #   - Run refresh_stats to recalculate the stats data.
     ServerName my-pootle.example.com
 
     # Set the 'POOTLE_SETTINGS' environment variable pointing at your custom


julen:
Thank you for looking at this @unho. For verifying this you'll wanna check the configuration of your local web server.

Note `FORCE_SCRIPT_NAME` is only needed when Pootle is deployed in non-root locations. Personally I'd explicitly mention that in a separate section for each web server configuration along with a link to the [Django docs on the matter](https://docs.djangoproject.com/en/1.8/ref/settings/#force-script-name), as this is not the common setup.

Also note that running `refresh_stats` is only needed in case stats were calculated prior to properly configuring the non-root setup.

unho:
@julen Did minor changes to make sure `refresh_stats` is only run if there were previous stats in place. I am not moving `FORCE_SCRIPT_NAME` to any other place since the place where it is now is about setting with Apache in non-root location, and the docs on using other web servers are outdated and incomplete, which is a separate issue that IMHO should be handled separatedly (see #4551).

dwaynebailey:
For what we want to achieve here this is lgtm.  I'd have preferred to break out the deployment instructions from the config.  But since the notes in the config are clearly marked as being specific to this use case I'm fine with that.

unho:
>  I'd have preferred to break out the deployment instructions from the config.

I usually find comments in provided config useful when I am editing it.

I would like to have a lgtm from @julen since he is the one that raised concerns.

julen:
No objections form my side @unho.

PR #4552, opened 2016-03-09
Cleanup: pootle runner handles unicode in pootle.conf template

phlax:
Currently the ``pootle init`` command fails if there is unicode in the template file


unho:
lgtm

PR #4554, opened 2016-03-09
ENChecker tests: use parametrization

julen:
Since we are already using pytest this is more natural and we get support for `xfail` and finer-grained reporting for free.


ta2-1:
lgtm

unho:
lgtm

PR #4555, opened 2016-03-09
Add Plurr syntax checker

julen:
This PR adds a new checker to alert when Plurr-formatted violate the expected syntax. Syntax checking is implemented via the [Plurr library](https://github.com/iafan/Plurr) itself.

### DIFF
--- a/pootle/apps/pootle_misc/checks.py
+++ b/pootle/apps/pootle_misc/checks.py
@@ -115,6 +115,7 @@ check_names = {
     'potential_unwanted_placeholders': _(u"Potential unwanted placeholders"),
     'double_quotes_in_tags': _(u"Double quotes in tags"),
     'percent_brace_placeholders': _(u"Percent brace placeholders"),
+    'plurr_format': _(u'Plurr format'),
 }
 
 excluded_filters = ['hassuggestion', 'spellcheck', 'isfuzzy',
@@ -960,6 +961,32 @@ class ENChecker(checks.UnitChecker):
                 return True
 
         raise checks.FilterFailure(u"Double quotes in tags mismatch")
+
+    @critical
+    def plurr_format(self, str1, str2, **kwargs):
+        """For plurr-formatted strings, checks the syntax is correct."""
+        # Ignore check for non Plurr-formatted strings
+        if not plurr_format_regex.search(str1):
+            return True
+
+        # Ignore check if library is missing
+        try:
+            from plurr import Plurr
+        except ImportError:
+            return True
+
+        plurr = Plurr()
+
+        try:
+            plurr.format(str2, {}, {
+                'locale': kwargs['language_code'],
+                'strict': False,
+                'callback': lambda x: '',
+            })
+        except SyntaxError as e:
+            raise checks.FilterFailure(e.message)
+
+        return True
 
 
 def run_given_filters(checker, unit, check_names=None):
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_misc/checks.py
```diff
@@ -266,56 +267,82 @@ class SkipCheck(Exception):
...
+
+    def __init__(self, checkerconfig=None, excludefilters=None,
+                 limitfilters=None, errorhandler=None):
+        super(ENChecker, self).__init__(checkerconfig, excludefilters,
+                                        limitfilters, errorhandler)
+
+    def run_test(self, test, unit):
+        """Runs the given test on the given unit.
+
+        Overrides `TranslationChecker`'s method to avoid extra plural handling
```
unho: `TranslationChecker` is no longer used, so it is not being overridden. Please adjust this docstring.

pootle/apps/pootle_misc/checks.py
```diff
@@ -266,56 +267,82 @@ class SkipCheck(Exception):
     pass
 
 
-class ENChecker(checks.TranslationChecker):
+class ENChecker(checks.UnitChecker):
+
+    def __init__(self, checkerconfig=None, excludefilters=None,
```
unho: Is it actually necessary to put the `__init__`? I don't see any change in reference to its parent.

pootle/apps/pootle_misc/checks.py
```diff
@@ -266,56 +267,82 @@ class SkipCheck(Exception):
...
+
+    def run_test(self, test, unit):
+        """Runs the given test on the given unit.
+
+        Overrides `TranslationChecker`'s method to avoid extra plural handling
+        (there are no gettext-style plural units in Evernote PO files), and
+        makes language information available to checks that want to make use of
+        it.
+        """
+        return test(self.str1, self.str2, language_code=self.language_code)
```
unho: If I am not wrong, `test` is a method from this class, so shouldn't it be able to access to `self.language_code` within the method instead of passing it as an argument?

pootle/apps/pootle_misc/checks.py
julen: Thank you for spotting this — it was a leftover of the initial customizations I made on top of `TranslationChecker`.

pootle/apps/pootle_misc/checks.py
julen: Yes, it is able to do so, however such implicit contract/internal state makes it harder to test checks independently, as opposed to passing an explicit argument.

unho: In TTK tests we initialize the checker by passing the language, for example https://github.com/translate/translate/blob/master/translate/filters/test_checks.py#L164-L165

unho:
Nice renaming `str1` and `str2` to `source_string` and `target_string`. Perhaps it is a good idea to do the same thing in the method signatures.

lgtm

julen:
> Nice renaming str1 and str2 to source_string and target_string. Perhaps it is a good idea to do the same thing in the method signatures.

Yes, I'm planning to do but didn't want to add extra noise to the PR besides what's strictly needed.

dwaynebailey:
lgtm.  Wondered if this wants any documentation? Its pretty hidden without it.  Does it want the Plurr dependency at least mentioned somewhere in requirements/*.txt?

unho:
Yes, at least the new checker should be documented in https://github.com/translate/pootle/blob/master/pootle/templates/help/_pootle_quality_checks.html

julen:
Added the mention in the help page. Re. adding the dependency in requirements, I'm happy to add it there commented out once we add a setup.py for Plurr and upload it to PyPI (should happen soon).

PR #4556, opened 2016-03-09
Contributors command improvements

unho:


pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -87,3 +139,13 @@ def list_contributions(self, contribs, sort_by):
 
         for line in out:
             self.stdout.write(line)
+
+    def list_only_emails(self, units):
+        nobody_user = User.objects.get_system_user()
```
julen: I presume this should be `nobody_user = User.objects.get_nobody_user()` instead.

docs/server/commands.rst
```diff
@@ -324,6 +324,9 @@ Available options:
 
 .. django-admin-option:: --from-revision
 
+  .. versionchanged:: 2.8.0 You cannot specify this option and
```
julen: `--since` being a new option, not sure if it's worth adding extra noise with `versionadded` here. I'd personally remove the directive and leave it as a normal paragraph.

docs/server/commands.rst
```diff
@@ -339,6 +342,25 @@ Available options:
 
   Default: ``name``.
 
+.. django-admin-option:: --only-emails
+
+  .. versionadded:: 2.8.0
+
+  Specifies to only output user names and emails. Users with no email are
+  skipped, thus anonymous users are not listed.
```
julen: Note there's only one anonymous user, `nobody` — leaving it as _Users with no email are skipped._ feels enough to me.

pootle/apps/pootle_app/management/commands/contributors.py
unho: Sure.

docs/server/commands.rst
unho: This is `versionchanged`.

docs/server/commands.rst
unho: Will change.

docs/server/commands.rst
julen: Sorry, I meant `versionchanged` instead — the comment applies anyway.

docs/releases/2.7.3b2.rst
```diff
@@ -20,6 +20,9 @@ Details of changes
   usernames.
 - :djadmin:`refresh_scores` now recalculates user scores and accept
   multiple usernames.
+- :djadmin:`contributors` has to new options
```
dwaynebailey: two

docs/server/commands.rst
dwaynebailey: TBH from a users perspective it is way more important what it does, then the fact that it is mutually exclusive.  That is lost with the noise that @julen pointed out.

docs/server/commands.rst
```diff
@@ -339,6 +342,25 @@ Available options:
...
+
+.. django-admin-option:: --since
+
+  .. versionadded:: 2.8.0
+
+  Tells to only take into account contributions newer than the specified
```
dwaynebailey: Only consider contributions since the specified date.

Avoid the programmer phrase Datetime, there is no such word.  Its just a date and the example makes it clear that you can specify a time.

docs/server/commands.rst
```diff
@@ -339,6 +342,25 @@ Available options:
...
+  datetime. Datetime must be in ISO 8601 format (``2016-01-24T23:15:22+0000``)
+  or in a string formatted like ``"2016-01-24 23:15:22 +0000"`` (quotes
+  included).
+
+  You cannot specify this option and :option:`--from-revision <contributors
+  --from-revision>` at the same time.
```
dwaynebailey: ``--since`` and ``--from-revision`` are mutually exclusive.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -42,6 +52,41 @@ def add_arguments(self, parser):
             help="Sort by specified item. Accepts: %(choices)s. "
                  "Default: %(default)s",
         )
+        parser.add_argument(
+            "--only-emails",
+            action="store_true",
+            dest="only_emails",
+            help="If provided only emails are output, no contribution counts",
```
dwaynebailey: Output only names and email addresses.  Contributions counts are excluded.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -87,3 +139,13 @@ def list_contributions(self, contribs, sort_by):
 
         for line in out:
             self.stdout.write(line)
+
+    def list_only_emails(self, units):
```
dwaynebailey: Don't need an answer but wondered why this couldn't be folded into the other output format.  It seems to just be an extra exclusion and a different format.  Seems overkill and confusing to have its own method.  Just smells a bit.

dwaynebailey:
Note comments, otherwise lgtm

unho:
Addressed all raised concerns.

PR #4557, opened 2016-03-09
Tests: set up site-wide DB only if tests requested the DB

julen:
This is an attempt to avoid setting up the site-wide DB when the set of selected tests don't make use of the DB at all.

There shouldn't be any effects when running the entire suite via `py.test`, however running something like `py.test -k misc/checks` should be immediate, as the selected tests makes use of no DB and hence the expensive setup can be avoided.

Probably there is a better way to approach this and I'm happy to improve it based on feedback, although at its current state this worked for me perfectly.

### DIFF
--- a/pytest_pootle/fixtures/site.py
+++ b/pytest_pootle/fixtures/site.py
@@ -12,7 +12,21 @@ from pytest_pootle.env import PootleTestEnv
 
 
 @pytest.fixture(autouse=True, scope='session')
+def setup_db_if_needed(request):
+    """Sets up the site DB only if tests requested to use the DB (autouse)."""
+    is_db_marker_set = [
+        item for item in request.node.items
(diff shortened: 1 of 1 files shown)


unho:
I don't understand how this works, but lgtm.

julen:
@unho at the beginning of the test session, it automatically goes through all collected test functions (`request.node.items`) and [checks whether they have the mark](https://pytest.org/latest/writing_plugins.html#_pytest.main.Node.get_marker) which requests DB access set (this is pytest-django's `django_db` marker). Only in case some test function set that, it [dynamically retrieves](https://pytest.org/latest/builtin.html#_pytest.python.FixtureRequest.getfuncargvalue) the `post_db_setup` fixture, which does all the DB-related setup. HTH.

PR #4560, opened 2016-03-11
Remove uid lists

phlax:
This PR removes the uid_list of full search results in the translation editor.

Instead the units are paginated forwards/backwards using offsets.

You can jump to a given index ~as before.

When paginating forwards the editor sends the previous units to adjust the offsets for units that may have been removed eg from being edited.

You can jump to a unit in any results set if you know the offset. If the uid is not in the result set at that offset - you will be shown the units currently at that offset.

For the Store translation view you can also search by uid of a unit without knowing the offset beforehand, in which case the backend will calculate the offset for the client

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -480,9 +480,9 @@ class UnitSearchForm(forms.Form):
             self.cleaned_data["user"] = self.request_user
         if self.errors:
             return
-        if self.cleaned_data['count'] is None:
-            self.cleaned_data["count"] = (
-                self.cleaned_data["user"].get_unit_rows())
+        self.cleaned_data['count'] = min(
+            self.cleaned_data.get("count", 10) or 10,
+            self.cleaned_data["user"].get_unit_rows() or 10)
         self.cleaned_data["vfolder"] = None
         pootle_path = self.cleaned_data.get("path")
         if 'virtualfolder' in settings.INSTALLED_APPS:
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -413,11 +413,14 @@ def unit_comment_form_factory(language):
 
 class UnitSearchForm(forms.Form):
 
-    initial = forms.BooleanField(required=False)
     count = forms.IntegerField(required=False)
+    offset = forms.IntegerField(required=False)
     path = forms.CharField(
         max_length=2048,
         required=True)
+    previous_uids = MultipleArgsField(
+        field=forms.IntegerField(),
+        required=False)
     uids = MultipleArgsField(
         field=forms.IntegerField(),
         required=False)
--- a/pootle/apps/pootle_store/unit/search.py
+++ b/pootle/apps/pootle_store/unit/search.py
@@ -17,7 +17,7 @@ from pootle_store.views import SIMPLY_SORTED
 
 class DBSearchBackend(object):
 
-    default_chunk_size = 9
+    default_chunk_size = 10
     default_order = "store__pootle_path", "index"
     select_related = (
         'store__translation_project__project',
--- a/pootle/apps/pootle_store/unit/search.py
+++ b/pootle/apps/pootle_store/unit/search.py
@@ -34,10 +34,6 @@ class DBSearchBackend(object):
             self.default_chunk_size)
 
     @property
-    def initial(self):
-        return self.kwargs.get('initial', True)
-
-    @property
     def project_code(self):
         return self.kwargs.get("project_code")
 
@@ -56,6 +52,14 @@ class DBSearchBackend(object):
     @property
     def unit_filter(self):
         return self.kwargs.get("filter")
+
+    @property
+    def offset(self):
+        return self.kwargs.get("offset", None)
+
+    @property
+    def previous_uids(self):
+        return self.kwargs.get("previous_uids", []) or []
 
     @property
     def sort_by(self):
@@ -145,22 +149,39 @@ class DBSearchBackend(object):
         return self.sort_qs(self.filter_qs(self.units_qs))
 
     def search(self):
-        uid_list = []
-        begin = 0
-        end = 2 * self.chunk_size
-        if self.initial:
-            results = self.results
-            uid_list = list(results.values_list('id', flat=True))
-            if len(self.uids) == 1:
-                if self.uids[0] not in uid_list:
-                    return [], results.none()
-                index = uid_list.index(self.uids[0])
-                begin = max(index - self.chunk_size, 0)
-                end = min(index + self.chunk_size + 1, len(uid_list))
-        elif self.uids:
-            results = self.results.filter(id__in=self.uids)
-        else:
-            # used by export view
-            uid_list = list(self.results.values_list('id', flat=True))
-            results = self.results[:self.chunk_size]
-        return uid_list, results[begin:end]
+        total = self.results.count()
+        start = self.offset
+
+        if start > total:
+            return total, total, total, self.results.none()
+
+        find_unit = (
+            self.language_code
+            and self.project_code
+            and self.filename
+            and self.uids)
+        find_next_slice = (
+            self.previous_uids
+            and self.offset)
+
+        if find_unit:
+            # find the uid in the Store
+            uid_list = list(self.results.values_list("pk", flat=True))
+            if self.uids[0] in uid_list:
+                unit_index = uid_list.index(self.uids[0])
+                start = (
+                    int(unit_index / (2 * self.chunk_size))
+                    * (2 * self.chunk_size))
+        elif find_next_slice:
+            # if both previous_uids and offset are set then try to ensure
+            # that the results we are returning start from the end of previous
+            # result set
+            _start = start = max(self.offset - len(self.previous_uids), 0)
+            end = max(self.offset + (2 * self.chunk_size), total)
+            uid_list = self.results[start:end].values_list("pk", flat=True)
+            for i, uid in enumerate(uid_list):
+                if uid in self.previous_uids:
+                    start = _start + i + 1
+        start = start or 0
+        end = min(start + (2 * self.chunk_size), total)
+        return total, start, end, self.results[start:end]
--- a/pootle/apps/pootle_store/unit/search.py
+++ b/pootle/apps/pootle_store/unit/search.py
@@ -18,7 +18,7 @@ from pootle_store.views import SIMPLY_SORTED
 class DBSearchBackend(object):
 
     default_chunk_size = 9
-    default_order = "store", "index"
+    default_order = "store__pootle_path", "index"
     select_related = (
         'store__translation_project__project',
         'store__translation_project__language')
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -210,20 +210,13 @@ def get_units(request):
                     raise Http400(_('Arguments missing.'))
         raise Http404(forms.ValidationError(search_form.errors).messages)
 
-    uid_list, units_qs = get_search_backend()(
+    total, start, end, units_qs = get_search_backend()(
         request.user, **search_form.cleaned_data).search()
-
-    bad_uid = (
-        search_form.cleaned_data["initial"]
-        and len(search_form.cleaned_data["uids"]) == 1
-        and search_form.cleaned_data["uids"][0] not in uid_list)
-    if bad_uid:
-        raise Http404
-
-    response = {'unitGroups': GroupedResults(units_qs).data}
-    if uid_list:
-        response['uIds'] = uid_list
-    return JsonResponse(response)
+    return JsonResponse(
+        {'start': start,
+         'end': end,
+         'total': total,
+         'unitGroups': GroupedResults(units_qs).data})
 
 
 @ajax_required
--- a/pootle/core/views.py
+++ b/pootle/core/views.py
@@ -685,14 +685,14 @@ class PootleExportView(PootleDetailView):
             raise Http404(
                 ValidationError(search_form.errors).messages)
 
-        uid_list, units_qs = get_search_backend()(
+        total, start, end, units_qs = get_search_backend()(
             self.request.user, **search_form.cleaned_data).search()
 
         units_qs = units_qs.select_related('store')
 
-        if len(uid_list) > search_form.cleaned_data["count"]:
+        if total > search_form.cleaned_data["count"]:
             ctx.update(
-                {'unit_total_count': len(uid_list),
+                {'unit_total_count': total,
                  'displayed_unit_count': search_form.cleaned_data["count"]})
 
         unit_groups = [
--- a/pootle/static/js/browser.js
+++ b/pootle/static/js/browser.js
@@ -78,7 +78,9 @@ function navigateTo(languageCode, projectCode, resource) {
 
   const PTL = window.PTL || {};
   if (PTL.hasOwnProperty('editor')) {
-    const hash = utils.getHash().replace(/&?unit=\d+/, '');
+    const hash = utils.getHash()
+          .replace(/&?unit=\d+/, '')
+          .replace(/&?offset=\d+/, '');
     if (hash !== '') {
       newUrl = [newUrl, hash].join('#');
     }
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -100,6 +100,7 @@ PTL.editor = {
     this.$navPrev = $('#js-nav-prev');
     this.unitCountEl = document.querySelector('.js-unit-count');
     this.unitIndexEl = document.querySelector('.js-unit-index');
+    this.offsetRequested = 0;
 
     /* Initialize variables */
     this.units = new UnitSet([], {
@@ -120,6 +121,8 @@ PTL.editor = {
 
     this.isLoading = true;
     this.showActivity();
+
+    this.fetchingOffsets = [];
 
     /* Regular expressions */
     this.cpRE = /^(<[^>]+>|\[n\|t]|\W$^\n)*(\b|$)/gm;
@@ -343,6 +346,7 @@ PTL.editor = {
       const params = utils.getParsedHash(hash);
       let isInitial = true;
       let uId = 0;
+      let initialOffset = 0;
 
       // Walk through known filtering criterias and apply them to the editor object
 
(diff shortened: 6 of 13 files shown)


pootle/apps/pootle_store/unit/search.py
```diff
@@ -145,22 +149,36 @@ def results(self):
...
-            results = self.results.filter(id__in=self.uids)
-        else:
-            # used by export view
-            uid_list = list(self.results.values_list('id', flat=True))
-            results = self.results[:self.chunk_size]
-        return uid_list, results[begin:end]
+        total = self.results.count()
+        start = self.offset
+        if start > total:
+            total, total, total, self.results.none()
```
unho: `return`?

pootle/apps/pootle_store/unit/search.py
```diff
@@ -145,22 +149,36 @@ def results(self):
...
+
+        find_unit = (
+            not self.offset
+            and self.language_code
+            and self.project_code
+            and self.filename
+            and self.uids)
+        find_next_slice = (
+            self.previous_uids
+            and self.offset) and True or False
```
unho: ` (WHATEVER) and True or False` == `(WHATEVER) and True` == `bool(WHATEVER)`

pootle/apps/pootle_store/unit/search.py
```diff
@@ -58,6 +54,14 @@ def unit_filter(self):
         return self.kwargs.get("filter")
 
     @property
+    def offset(self):
+        return self.kwargs.get("offset", None)
```
unho: Is it not better if it defaults to `0`?

pootle/apps/pootle_store/unit/search.py
```diff
@@ -145,22 +149,36 @@ def results(self):
...
+            and self.uids)
+        find_next_slice = (
+            self.previous_uids
+            and self.offset) and True or False
+
+        if find_unit:
+            # find the uid in the Store
+            uid_list = list(self.results.values_list("pk", flat=True))
+            unit_index = uid_list.index(self.uids[0])
+            start = int(unit_index / (2 * self.chunk_size)) * (2 * self.chunk_size)
```
unho: Is this not killing the ability to return a list of uids where the uid we are looking for is exactly in the middle?

phlax: we dont return a list of uids any more.

this finds the offset for the client from the uid - in the specific case of translating a store

pootle/apps/pootle_store/unit/search.py
```diff
@@ -145,22 +149,36 @@ def results(self):
...
+            and self.filename
+            and self.uids)
+        find_next_slice = (
+            self.previous_uids
+            and self.offset) and True or False
+
+        if find_unit:
+            # find the uid in the Store
+            uid_list = list(self.results.values_list("pk", flat=True))
+            unit_index = uid_list.index(self.uids[0])
```
unho: This throws an exception if the uid is not on the list. How do we catch it?

pootle/apps/pootle_store/unit/search.py
unho: I know we don't return slices anymore. But previous code did return a slice centered around the unit we wanted. Now we return the offset for a slice, being the offset a multiple of `(2 * self.chunk_size)`, thus the offset is for a virtual slice not centered around the unit anymore. For example we want start for unit 13 and chunk_size is 5, then `start = int(13 /20) / 10` so `start = 10` where previously it was `8` (5 units before and 5 units after the one we looked for).

But perhaps I am mistaken.

julen:
Did some quick testing (haven't looked at the code yet), and here are the issues I found:

* When navigating down through units, there's some point where the editing widget doesn't remain at the center — it acts as if it was finishing to go through all units, even if there are pending units, and it looks like the unit fetching is being done too late, only when opening the _last_ unit in the current screen.
* Opening a unit via URL doesn't display it at the center.
* Jumping to a unit by specifying the number in the result set is broken, and it doesn't display the editing widget at the center.
* Switching languages via the drop-down is broken, as it passes the `offset` which is irrelevant in the new view. I also wonder why such query parameter is used — it makes copying and pasting URLs less friendly.
* Accessing a URL with a filter and a uid on it when the unit doesn't meet the filtering criteria anymore breaks the server instead of displaying a _"No results"_ text as it does now. Relevant traceback:
```python
  File ".virtualenvs/pootle/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
  File "pootle/apps/pootle_misc/util.py", line 58, in wrapper
    return f(request, *args, **kwargs)
  File "pootle/apps/pootle_store/views.py", line 214, in get_units
    request.user, **search_form.cleaned_data).search()
  File "pootle/apps/pootle_store/unit/search.py", line 170, in search
    unit_index = uid_list.index(self.uids[0])
ValueError: 4443894 is not in list
```

phlax:
> - When navigating down through units, there's some point where the editing widget doesn't remain at the center — it acts as if it was finishing to go through all units, even if there are pending units, and it looks like the unit fetching is being done too late, only when opening the last unit in the current screen.
- Opening a unit via URL doesn't display it at the center.

ill look into this

phlax:
> Jumping to a unit by specifying the number in the result set is broken,

k, its almost there in that it jumps to the beginning of that offset slice - but ill add some code to make it go to the correct place.

> and it doesn't display the editing widget at the center.

as above

phlax:
> Switching languages via the drop-down is broken, as it passes the offset which is irrelevant in the new view.

yep. ill look into fixing this

> I also wonder why such query parameter is used — it makes copying and pasting URLs less friendly.

its necessary to maintain the correct offset/unit focus if you reload the page or paste a link to that unit

phlax:
> Accessing a URL with a filter and a uid on it when the unit doesn't meet the filtering criteria anymore breaks the server instead of displaying a "No results" text as it does now. Relevant traceback

yep - this is essentially the effect of the bug that @unho pointed out above

unho:
@phlax Actually I think I raised most of the issues @julen points out on his testing while doing my code review in the PR. The one about not displaying current unit always surrounded by the same amount of units before and after it (except when it is closer to the begin of the end of list of units) will be really annoying for users.

iafan:
I want the following use case to be supported (and I believe it is currently not):

1) I click on any "Continue translation" link to start translating.
2) As I go through units, the `#unit=xxxxxx` part of the URL is changed
3) Reloading the page (or sending the link over to someone else) should restore the context: the unit with the given uid (not the index!) needs to appear in the editor, surrounded by an appropriate list of surrounding units (based on current translation scope/filters/sorting).

If selecting by uid works, then the next logical step would be to remove the redundant `index=xxxx` parameter in URL. You can obviously use indexes internally in XHR requests.

iafan:
The other thing that concerns me is that the index (position) is more volatile as compared to uid: is there any documentation/overview on how the UI will behave when some new strings are inserted or deleted due to update_stores?

Common use case: I as a translator decided to continue translation through the entire language. In other words, I'm going through all untranslated strings in all translation projects. Now let's assume I am at unit with `index=100` (somewhere in the middle of translation projects). Now update_stores kicks in and adds a bunch of strings in the projects I already went through. What will happen in terms of the UI when the user navigates to the next unit (or when the UI loads the next chunk of units)? What will happen if the user goes through units and some units are deleted (obsoleted) due to update_stores?

phlax:
> What will happen if the user goes through units and some units are deleted (obsoleted) due to update_stores?

it will be a lot better than the current implementation because atm the list is frozen when you first load the page, which doesnt accomodate for the result set changing at all.

With the new implementation you will be able to see that the total number of units have changed, and there is some fault tolerance that will catch the 99% case.

tbh though as with the other related questions it really depends on what has changed. Using unitids is **more unreliable** in result sets that are subject to change, for the simple reason that they may no longer be in the result set - with an offset you are going to get back roughly to where you were.

With the new implementation - in a Store you get the best of both worlds - it will find the unit for you *if* its still in the result set - if not it will take you to the offset where it last appeared. Afaict its this view that most translators use - and certainly **all** of the current links to translate pages with a uid link to the unit's store.

With either implementation for the list to remain "true" it would also be necessary to remove or at least remove from the index, any units that no longer qualify for that result set - eg because they have been translated, suggestion accepted/rejected, or have had their failing check addressed.

This is not something that i intend to address now - and will be much easer to do in a good way once we have websockets - even then it will still not allow the iteration of huge result sets in search of a possible uid though.

What must be addressed in the current implementation:
1) it sends back ridiculously huge (>100Ms) of uid_lists, pointlessly. Not only does this lock up massive amounts of server resources, application threads, and network bandwidth it also crashes all but the most modern of computers that receive these lists (unexpectedly)
2) it allows users to iterate those same huge lists in search of a uid that may or may not even be there - this is a huge DOS vector - and would be extremely trivial to exploit.
3) there are brightly coloured buttons that not only make it possible that all of the above happens, but extremely probable.

phlax:
@julen i believe that all of the issues that you raised have been addressed - it would be great if you were up for testing again, ta.

dwaynebailey:
@phlax Did a review of the functionality on the staging server.  Much snappier.  Behaves as I expected.  LGTM.

@iafan I looked at some of the issues you raised.  Reloading works (obviously as long as the underlying result set hasn't actually changed). Reporting issues works.  Copying the link for the breadcrumbs works.

phlax:
> Reloading works (obviously as long as the underlying result set hasn't actually changed)

at least if it hasnt changed by more than 20 units in the preceeding results since you last requested some results.

iafan:
> at least if it hasnt changed by more than 20 units in the preceeding results since you last requested some results.

...which happens pretty frequently in our case when we do continuous localization. And in such cases UI behavior will look broken.

phlax:
@iafan do you understand how serious this problem is, and do you understand the 2 underlying issues that cause it?

i think its worth stressing that in no shape or form is the current implementation meeting expectations - ie in the case you are giving it would continue showing you units that had gone from the result set.

Apart from anything else, until a few weeks ago it was ordering in the db by one criteria and grouping by another - causing wierd duplication etc etc - not just occasionally - but frequently.

The new implementation is not only far far more performant - and removes a huge and trivially exploitable DOS vector, and makes the translation editor work in a reasonable way as opposed to the broken and sluggish way it works now, and gives far more accurate results - im really wondering where the argument is.

I would be more than happy for you to see the new implementation on our staging server if that helps.

dwaynebailey:
@iafan I'm still not sure I understand how this is actually broken for you.

Here is an example of how the current UI is broken.

1. The person is translating a list that does not actually represent reality.  That most certainly in my mind does not reflect the reality of continuous localisation.  The new approach would allow new units to be added to the list while you are translating.  Current approach pretends it's a static list when it isn't.
2. If I decide to translate "Needs work" units.  Then actually translate one of those units.  Then share that URL or reload it later.  Recall it now does not actually reflect the filter stored in it.  I.e. this URL https://mozilla.locamotion.org/en_ZA/firefox/translate/dom/#filter=incomplete&sort=priority&unit=20968525 This is now broken in that that units is no longer incomplete.  In fact now it is gone.  This is on the old UI and the new UI.  So this never worked as you think it does.

So I'm slightly at a loss as to what broken UI you are really talking about here.  I realise now that you are on a break.  But some real examples would be really helpful.  Or if you could actually test this on a staging server and give feedback compared to current.

iafan:
> The person is translating a list that does not actually represent reality. That most certainly in my mind does not reflect the reality of continuous localisation. The new approach would allow new units to be added to the list while you are translating. Current approach pretends it's a static list when it isn't.

By continuous localization I mean the system that constantly adds or removes units to reflect the state of localizable resources, not the "continuous translation". Current system, while being obviously broken from the technical side, was at least able to provide meaningful UX: the user sees that they need to translate N words, they click on this task, the system essentially makes the snapshot of the task and the user performs the translation. Their translation context is not affected if new units are added or deleted by a continuous localization process. Once they go through the entire task, they are are redirected back to the stats. At this point they may see that there are some more units requiring the translation (which is ok, since they were added after the user started translation).

The current system, while fixing the technical side of things, makes UX more fragile: when the user translates something and the system changes the units in the background, indexes get changed, and the result is undetermined: from the user perspective, when the new page of units is loaded, it can be something he already translated, or he will jump over some untranslated units.

The URLs that this PR introduces have both uid and index, but in the majority of cases the `unit=xxxxxx` parameter is simply ignored, and no indication is given that the user is actually seeing the wrong unit. So if someone sends over a URL to someone else, there's a good chance they will be looking at different URLs. You could at least think of how to deal with this situation gracefully. For example, providing a notice, offering the user a permalink to a unit.

> with an offset you are going to get back roughly to where you were.

"roughly where you are" is something that I would like to avoid at all costs. That's an example of broken UX.

> in a Store you get the best of both worlds - it will find the unit for you if its still in the result set - if not it will take you to the offset where it last appeared. Afaict its this view that most translators use

Definitely not in our case: translators translate either a project, or entire language, but not individual stores.

I don't try to diminish the work you do to resolve the technical issues, I just want to make sure we don't break UX in between.

dwaynebailey:
@iafan 

This is what the PR does:

1. A translator can work through a list efficiently. Without adversely affecting their platform or the server.
2. Translators can resume approximately where they last were in an arbitrary result set.
3. You can link to specific units within a Store.  Which is the way the UI shares links in all places; unit navbar, string error reporting.

This change has been on the cards for some time, the need is very clear as to why and what needs to be addressed.  I'm yet to see an issue raised that actually impacts the way translators are working, or that directly impacts on something baked into the UI design.

iafan:
I understand what the PR does and why it does that. I saw how it works on our staging server. I provided my input on why I think it breaks the "actual way translators are working" (use cases where the system behavior is worse for the user than it was previously). I also provided some input on how the UI could be improved so that end users would understand the system behavior better.

The item 3 on your list doesn't mention the URLs the UI generates while the user goes through the units. These are the URLs which behavior is now undetermined. Unfortunately, these are the URL users use for sharing as well, because they are easier to use. I guess you don't consider this an issue, since you mention "resume approximately where they there" as if it is an acceptable behavior. I don't consider it as such. You may need to test this on real users to see if they are ok with this approach, but the problem is that since this lands on master, we will have no choice but test the results on our users as well.

ta2-1:
There are different `unitIds` for current unit in URL and in the editor.
image
@phlax is it possible to remove `#unitid=NNN` part if we open different unit in the editor?

pootle/apps/pootle_store/unit/search.py
unho: I still think `0` is a better default.

pootle/apps/pootle_store/forms.py
```diff
@@ -477,9 +480,9 @@ def clean(self):
             self.cleaned_data["user"] = self.request_user
         if self.errors:
             return
-        if self.cleaned_data['count'] is None:
-            self.cleaned_data["count"] = (
-                self.cleaned_data["user"].get_unit_rows())
+        self.cleaned_data['count'] = min(
+            self.cleaned_data.get("count", 10) or 10,
+            self.cleaned_data["user"].get_unit_rows() or 10)
```
unho: I don't particularly like hardcoding `10`. But I am ok with making it a future cleanup.

phlax: bear in mind this is a default - rather than a hardcoded setting

unho: ok. But it is a hardcoded default all over the place.

I thought I left this comment on the files tab, not in a commit. Weird.

phlax: in the tests (iirc) they use a non-default value to ensure that its not hardcoding anything.

pootle/apps/pootle_store/unit/search.py
phlax: i dont really agree - mostly because you have no way then of telling whether no offset was set or an offset of 0 was set (not that it matters in the current implementation)

unho:
gtm. Tested and I found some problem when the user has "Number of rows" set to 50, something that is not quite common.

pootle/apps/pootle_store/unit/results.py
```diff
@@ -88,7 +73,6 @@ def data(self):
                     'project_style': unit.project_style}
             units_list.append(
                 {'id': unit.id,
-                 'url': unit.translate_url,
```
julen: Notice how this [is not dead code](https://github.com/translate/pootle/blob/master/pootle/templates/editor/units/xhr_view.html#L2), and moreover the change breaks opening editing rows in new tabs.

phlax: thanks for pointing this out - it was showing in coveralls as not being used, after i had made a change.

Ill add it back in.

julen:
I've done another round of testing with the latest changes and the following results:

* I can confirm the issues I reported earlier have been addressed.

* With the last issue I reported last time (accessing a URL with filter + uid, the one which produced the traceback), now there's this issue @ta2-1 already reported: the uid is kept in the URL, but the result set doesn't contain it, hence displaying a unit which doesn't match the URL. 
An option with this would be to remove the unit id from the URL (this was already mentioned in the group chat). Still, the previous behavior was explicit in the sense of X was requested matching Y, and nothing was found so _No results_ was displayed. I don't have strong reasons to defend one or the other, although I find more intuitive the latter. I'd like to hear @iafan's thoughts on this one.

* Jumping to a unit by specifying the position in the result set doesn't update/display the `unitid` hash as it does now.

* While translating with a filter, sometimes the _No results_ message flashes.

* Translating and going through units doesn't work as expected. Taking the most common use case, i.e. going through `filter=incomplete` at any level, one experiences that after consuming a batch of units and being close to the end, no more units are loaded, leaving the editor in a dumb state. This can easily be reproduced with having say 20ish incomplete units, and translating them.
The issue seems to be caused because the code assumes the passed `previous_id`s are part of the queryset, but these are dynamic by their nature, so the assumption is wrong in the first place. From the user's standpoint this is somehow similar to evernote/pootle#48, which ended up adding the `uid_list` :wink:

phlax:
> Jumping to a unit by specifying the position in the result set doesn't update/display the unitid hash as it does now.

this worksforme unless im misunderstanding something - the only difference is that now uses offsets in addition to the uid.

> While translating with a filter, sometimes the No results message flashes.

This is now fixed i believe

> 5....

not sure i entirely follow here - but it corrects to a margin of 20 units - as the editor ordinarily fetches 20-40 units for a given view - its unlikely afaict that this will not suffice in most use cases

phlax:
> 2 ....snip...An option with this would be to remove the unit id from the URL (this was already mentioned in the group chat). 

it now updates the URL with uid of the unit that is selected

ta2-1:
gtm

PR #4561, opened 2016-03-11
Project views

phlax:
This PR restricts the very expensive cross language Translation views to administrators.

It will also restrict the paths for which get_units can be called.

It would be possible to make the permissions configurable, but this would add a layer of complexity to testing both the view and the xhr call so it would be far easier to use the "administrate" permission

~~**currently rebased on #4585**~~


pootle/templates/browser/_actions.html
```diff
@@ -1,25 +1,25 @@
 {% load i18n %}
 <ul>
   <li id="js-action-continue">
-    <a class="continue-translation" href="{{ url_action_continue }}">
+    <a class="continue-translation"{% if url_action_continue %} href="{{ url_action_continue }}"{% endif %}>
```
dwaynebailey: My understanding this is making the button not have a href. Yet it continues to behave like it is clickable?

Can we make sure that we turn the button into an info only display and make sure it clearly doesn't appear to be clickable.

phlax: yep - will do

dwaynebailey:
@phlax did a UI review.  I think the disabling works.  The muted colours, missing highlighting and pointer convey that the link isn't clickable anymore.

The one issue identified in that TP links on the `projects/$project` page are also disabled, they shouldn't be as they're just an exposure of links that would appear in `$lang/$project`

@iafan we have a staging server up if you can have a look at the changes.  That would be appreciated.

iafan:
I'm not at the computer (on a vacation till the end of next week), so can't fully test this. What will happen with cross-language search?

dwaynebailey:
@iafan If you are not logged in (i.e. don't have perms) then:

* `project/$project` - you get Permissions Denied page
* `$lang` - works as expected

phlax:
k - ill hide the search widget in the x-lang views without necessary perms

iafan:
FYI: We have cross-language search links in each unit (to see the same unit in all languages) — something we consider helpful not only to admins, but to translators as well, and we advertise them as such). I admit this is a non-standard feature, but this is something that the new behavior will break, unless it allows cross-language search in a store (I bet this is pretty cheap, though; at least we didn't see serious delay in getting the results back). @julen, do you mind providing additional details/links on this feature?

dwaynebailey:
@iafan are you referring to the linked Location comments?

iafan:
Yes. Here's an example of the unit: https://translate.evernote.com/ru/ol365_clipper_evernote/translate/strings.json.po#unit=5350198

In Locations: field, there's a link to other units across all languages: https://translate.evernote.com/projects/ol365_clipper_evernote/translate/strings.json.po#search=File%3A%0Astrings.json%0AID%3A%0A01f84e64b95394e03a4551fe8d2651b5&sfields=locations

iafan:
Yes. Here's an example of the unit: https://translate.evernote.com/ru/ol365_clipper_evernote/translate/strings.json.po#unit=5350198

In Locations: field, there's a link to other units across all languages: https://translate.evernote.com/projects/ol365_clipper_evernote/translate/strings.json.po#search=File%3A%0Astrings.json%0AID%3A%0A01f84e64b95394e03a4551fe8d2651b5&sfields=locations

iafan:
Yes. Here's an example of the unit: https://translate.evernote.com/ru/ol365_clipper_evernote/translate/strings.json.po#unit=5350198

In Locations: field, there's a link to other units across all languages: https://translate.evernote.com/projects/ol365_clipper_evernote/translate/strings.json.po#search=File%3A%0Astrings.json%0AID%3A%0A01f84e64b95394e03a4551fe8d2651b5&sfields=locations

phlax:
@iafan i havent updated PR yet but this is not shown as a link any more if you are not able to see that view

iafan:
That's exactly what I'm talking about: we consider this useful for non-admins as well, and this is something this PR will degrade (though we don't track who actually uses these links). Also: having non-clickable links is not the best UX. If people can't use the link, they shouldn't be presented it at all (or fall back to the original rendering of the location field). This also applies to links in `/projects/` — probably it was simply better to restrict this entire URL to admins instead of making links muted/non-clickable, hiding search bar, etc.

dwaynebailey:
@iafan @phlax mentioned on t/dev channel that we'd need to bring that back in a new view. But lets do that after landing these changes first.

Re Location rendering.  I assume removing the clickability should turn it back to normal rendering, we can check once @phlax has pushed changes to the PR.  But removing is not the correct approach.  The location information is useful in most cases as in most uses it isn't a hash but a human parseable key, and that might be valuable information to the translator.

Re `/projects/` I think the muted UI works fine now.  I'm worried that hiding the whole view has a lot of spillover in terms of the UI usage.  I really recommend that you look at the staging instance and give feedback based on that.

iafan:
> I assume removing the clickability should turn it back to normal rendering, we can check once @phlax has pushed changes to the PR. But removing is not the correct approach. The location information is useful in most cases as in most uses it isn't a hash but a human parseable key, and that might be valuable information to the translator.

Yes, displaying the plain value of the location field would be the best option.

> I really recommend that you look at the staging instance and give feedback based on that.

What's your staging site URL?

dwaynebailey:
> Yes, displaying the plain value of the location field would be the best option.

Checked, it works as such.  I.e. like it did before it become a clickable link

> What's your staging site URL?

I'll email

julen:
Checked, and overall feature-wise works as advertised. Nice that the languages drop-down has also been adjusted :+1:

By looking at deeper details:
* Note the failing checks are still clickable at any level when expanding the statistics in the browsing pages.
* The search box is still available in `/projects/` and `/project/<code>/`. I know this is known but just a heads-up in case it was forgotten.
* I'm a bit confused with the coloring of the pills, it makes me think they are clickable even if the pointer is regular and the colors are a bit darkened. It's also confusing because the text contains commands/actions to perform. Would love to hear more opinions on this.

I also want to add my voice to the request for bringing back the the cross-language view of a specific unit for non-admins, be it in this or other shape, but I'm fine to treat that separately.

pootle/static/js/stats.js
```diff
@@ -145,13 +145,16 @@ const stats = {
   },
 
   updateItemStats($td, count) {
+    // console.log("updating stat item")
```
julen: Please don't forget to remove these.

pootle/templates/browser/_table.html
```diff
@@ -42,7 +42,11 @@
 
       {% if 'total' in table.fields %}
       <td id="total-words-{{ item.code|cssid }}" class="stats-number total not-inited">
-        <a class="non-zero" href="{{ item.href_all }}"></a>
+	{% if can_translate %}
```
julen: Please do not use tabs for indenting, all our HTML files contain 2-space indents.

phlax: np - ill try and rem to untabify before saving

pootle/static/js/stats.js
phlax: yep - will do

pootle/core/views.py
```diff
@@ -599,18 +599,28 @@ def get(self, *args, **kwargs):
...
+                state='suggestions',
+                **filters)
+            url_action_view_all = self.object.get_translate_url(state='all')
+        else:
+            if self.project:
+                can_translate = True
+            (url_action_continue,
+             url_action_fixcritical,
+             url_action_review,
+             url_action_view_all) = [None] * 4
```
julen: This definitely works, however these being just 4 variables, I'd suggest to avoid the fanciness and use regular assignments.

phlax: k

pootle/apps/pootle_store/forms.py
```diff
@@ -517,6 +519,27 @@ def clean_month(self):
     def clean_user(self):
         return self.cleaned_data["user"] or self.request_user
 
+    def clean_path(self):
+        if self.cleaned_data["path"].strip("/") in ["", "projects"]:
+            permission_context = Directory.objects.get(pootle_path="/projects/")
```
julen: It's easier to get this from `Directory.objects.projects`.

pootle/apps/pootle_store/forms.py
```diff
@@ -517,6 +519,27 @@ def clean_month(self):
     def clean_user(self):
         return self.cleaned_data["user"] or self.request_user
 
+    def clean_path(self):
+        if self.cleaned_data["path"].strip("/") in ["", "projects"]:
```
julen: `split_pootle_path()` already provides the stripping being done here — as I see only the two first items it returns are needed in this context.
```python
language_code, project_code = split_pootle_path(self.cleaned_data['path'])[:2]
```

pootle/apps/pootle_misc/templatetags/search.py
```diff
@@ -17,6 +17,16 @@
...
 def render_search(context):
+    search_form = make_search_form(request=context['request'])
+    if context["page"] != "translate" and not context["can_translate_stats"]:
+        search_form.fields["search"].widget.attrs["readonly"] = "readonly"
+        search_form.fields["search"].widget.attrs["disabled"] = True
+        search_form.fields["search"].widget.attrs["title"] = ""
+        search_form.fields["search"].widget.attrs[
+            "placeholder"] = "Search unavailable"
+        search_form.fields["search"].widget.attrs[
+            "style"] = "background: #eee; font-style; italics"
```
unho: IMHO we should set some CSS class and keep CSS styling in stylesheets, instead of here.

pootle/apps/pootle_store/forms.py
```diff
@@ -520,6 +523,28 @@ def clean_month(self):
...
+            try:
+                permission_context = Project.objects.select_related(
+                    "directory").get(code=project_code).directory
+            except Project.DoesNotExist:
+                raise forms.ValidationError("Unrecognized path")
+        else:
+            # no permission checking on lang translate views
+            return self.cleaned_data["path"]
+        if self.request_user.is_superuser:
+            return self.cleaned_data["path"]
```
unho: Is it not better to put this `if` at the top?

pootle/apps/pootle_misc/templatetags/search.py
ta2-1: I've thought the same.

phlax: k 8/

pootle/apps/pootle_misc/templatetags/search.py
```diff
@@ -17,6 +17,15 @@
 
 @register.inclusion_tag('core/search.html', takes_context=True)
 def render_search(context):
+    search_form = make_search_form(request=context['request'])
+    if context["page"] != "translate" and not context["can_translate_stats"]:
+        search_form.fields["search"].widget.attrs["readonly"] = "readonly"
+        search_form.fields["search"].widget.attrs["disabled"] = True
+        search_form.fields["search"].widget.attrs["title"] = ""
+        search_form.fields["search"].widget.attrs[
+            "placeholder"] = "Search unavailable"
```
unho: Is this not translatable?

PR #4566, opened 2016-03-12
Tidy get_edit_unit

phlax:
- add tests
- refactor to class
- add select_related or other performance optimizations

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -16,11 +16,13 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.core.urlresolvers import reverse
-from django.db.models import Max, Q
+from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import redirect
 from django.template import RequestContext, loader
 from django.utils import timezone
+from django.utils.functional import cached_property
+from django.utils.lru_cache import lru_cache
 from django.utils.safestring import mark_safe
 from django.utils.translation import to_locale, ugettext as _
 from django.utils.translation.trans_real import parse_accept_lang_header
@@ -31,6 +33,7 @@ from pootle.core.decorators import (get_path_obj, get_resource,
                                     permission_required)
 from pootle.core.exceptions import Http400
 from pootle.core.http import JsonResponse, JsonResponseBadRequest
+from pootle.core.views import PootleJSON
 from pootle_app.models.directory import Directory
 from pootle_app.models.permissions import (check_permission,
                                            check_user_permission)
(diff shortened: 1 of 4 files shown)


tests/views/get_edit_unit.py
```diff
@@ -0,0 +1,89 @@
...
+            except VirtualFolder.DoesNotExist:
+                pass
+
+        if priority is None:
+            # Retrieve the unit top priority, if any. This can happen if we are
+            # not in a virtual folder or if the passed virtual folder pk is
+            # invalid.
+            priority = unit.vfolders.aggregate(
+                priority=Max('priority')
+            )['priority']
```
unho: If I am not wrong, the priority when no virtualfolders exist for this unit is `1`, so it is safe to default to it instead of `None` and get rid of this `if`.

phlax: well - actually - i realised that none of this is code is actually necessary any more because the priority is to denormed to unit already.

I kinda left it as it was a good test that the denorm was working correctly but i think i should just use Unit.priority

unho: Indeed.

tests/views/get_edit_unit.py
```diff
@@ -0,0 +1,89 @@
...
+    project = translation_project.project
+    language = translation_project.language
+
+    response = client.get(
+        "/xhr/units/%s/edit/" % unit.id,
+        HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+    request = response.wsgi_request
+    result = json.loads(response.content)
+
+    priority = None
```
unho: We are not checking for priority anywhere in the response.

pootle/apps/pootle_store/views.py
```diff
@@ -407,102 +410,122 @@ def save_comment(request, unit):
...
+            "store__translation_project__language")
+
+    def get_sources(self):
+        sources = {
+            unit.store.translation_project.language.code: unit.target_f.strings
+            for unit in self.get_alt_srcs()}
+        sources[self.source_language.code] = self.object.source_f.strings
+        return sources
+
+    def get_context_data(self, *la, **kwa):
```
unho: Nit: Use `*args, **kwargs`.

pootle/apps/pootle_store/views.py
```diff
@@ -407,102 +410,122 @@ def save_comment(request, unit):
...
+    @cached_property
+    def directory(self):
+        return self.store.parent
+
+    def get_edit_template(self):
+        if self.project.is_terminology or self.store.is_terminology:
+            t = loader.get_template('editor/units/term_edit.html')
+        else:
+            t = loader.get_template('editor/units/edit.html')
+        return t
```
unho: Unnecessarily complex:

```python
        if self.project.is_terminology or self.store.is_terminology:
            return loader.get_template('editor/units/term_edit.html')
        return loader.get_template('editor/units/edit.html')
```

pootle/apps/pootle_store/views.py
```diff
@@ -407,102 +410,122 @@ def save_comment(request, unit):
...
+                                               "review",
+                                               self.directory),
+            'is_admin': check_user_permission(self.request.user,
+                                              'administrate',
+                                              self.directory),
+            'altsrcs': self.get_alt_srcs()}
+
+    def get_response_data(self, context):
+        return {
+            'editor': self.render_edit_template(context),
```
unho: `context` shouldn't instead be `self.get_context_data()`?

pootle/apps/pootle_store/views.py
```diff
@@ -407,102 +410,122 @@ def save_comment(request, unit):
...
+
+    def render_edit_template(self, context):
+        return self.get_edit_template().render(
+            RequestContext(self.request, context))
+
+    def get_unit_edit_form(self):
+        if self.object.hasplural():
+            snplurals = len(self.object.source.strings)
+        else:
+            snplurals = None
```
unho: Nit: simpler:

```python
        snplurals = None
        if self.object.hasplural():
            snplurals = len(self.object.source.strings)
```

pootle/apps/pootle_store/views.py
```diff
@@ -407,102 +410,122 @@ def save_comment(request, unit):
...
-        'tm_suggestions': unit.get_tm_suggestions(),
-        'is_obsolete': unit.isobsolete(),
-        'sources': sources,
-    })
-
-    return JsonResponse(json)
+class UnitEditJSON(PootleJSON):
+
+    model = Unit
+    pk_url_kwarg = "uid"
```
unho: These two look like are unused.

pootle/apps/pootle_store/views.py
phlax: no its already been created by this point

pootle/apps/pootle_store/views.py
phlax: see here to understand their use https://docs.djangoproject.com/en/1.9/ref/class-based-views/generic-display/#detailview

pootle/apps/pootle_store/views.py
unho: And how do you pass it to `self.render_edit_template()`?

pootle/core/views.py
```diff
@@ -509,6 +512,31 @@ def get_context_data(self, *args, **kwargs):
...
+        return super(PootleDetailView, self).dispatch(request, *args, **kwargs)
+
+    def get_response_data(self, context):
+        """Override this method if you need to render a template with the context object
+        but wish to include the rendered output within a JSON response
+        """
+        return context
+
+    def render_to_response(self, context, **response_kwargs):
+        response_kwargs.setdefault('content_type', self.content_type)
```
phlax: > And how do you pass it to self.render_edit_template()?

@unho this method is overridden to allow setting `response_data` separately from `context_data`

ta2-1:
Timing was not so bad. But it's nice to see decreasing number of queries. 
```
"GET /xhr/units/19685276/edit/ HTTP/1.1" 200 9283
21 queries run, total 0.008 seconds
```

```
"GET /xhr/units/19685276/edit/ HTTP/1.1" 200 9283
11 queries run, total 0.005 seconds
```

phlax:
> Timing was not so bad

I have run timing tests against a large number of uids - and saw a consistent ~3% speedup. Not huge - but yep its good to reduce the db load.

This was not so much focused on performance, more just refactoring the code to make it easier to change/test - but there are some kinda obvious performance boosts that we can put in place

pootle/core/views.py
```diff
@@ -464,7 +465,8 @@ def is_admin(self):
 
     @property
     def language(self):
-        return None
+        if self.tp:
+            return self.tp.language
```
ta2-1: @phlax, I wonder why didn't you put this to `UnitEditJSON` class. What have I missed?

phlax: because it implements the basic functionality - which most if not all child classes share i believe.

ta2-1: it looks implicit. can we comment which views are going to use this (not overwritten) method?

phlax: its not at all "implicit" this is the parent class - and this is a method of the all of the descendants

unho: It was returning `None` because the other descendants overwrite it when necessary, even when they don't have a TP.

PR #4569, opened 2016-03-14
ProjectForm: ensure `clean_<field>` methods are used

julen:
An indenting mistake (cd323b3) was making the field-specific clean methods
useless. At the same time, duplicated validation added in the past (4c15390e)
can now be removed.


unho:
lgtm

Issue #4273, opened 2015-12-04
calculate_checks uses a huge amount of app memory on large sites


phlax:
It used up ~4GB before my test server ran out of mem/swap

phlax:
@ta2-1 maybe the solution is to remove the tool - trying to rem if there is anything that calculate_checks is still needed for

julen:
`calculate_checks` is needed when willing to (selectively) calculate checks without refreshing all stats. This usually happens when a quality check has been refined/updated.

ta2-1:
When you add new checks or change existing one you'll need to recalculate these checks for all units:
- update `QualityCheck` objects;
- refresh stats cache;

`refresh_stats` command addresses the second only. So I'd not get rid of this.

PR #4572, opened 2016-03-14
update_qualitychecks: remove dead code

julen:
The `check_names` and `existing` parameters are dead code since #4273 was
addressed.


unho:
lgtm

Issue #4553, opened 2016-03-09
Use a single code path to run quality check filters


julen:
Pootle is currently making use of [different (+_duplicated_) code paths](https://github.com/translate/pootle/blob/e53ea85f934689df735f09e57614e1bd0ea12e47/pootle/apps/pootle_store/models.py#L846-L849) to run quality check filters. The only difference is being able to selectively choose which checks to run:
```python
       if check_names is None:
            qc_failures = checker.run_filters(self, categorised=True)
        else:
            qc_failures = run_given_filters(checker, self, check_names)
```

Here, [`run_given_filters()`](https://github.com/translate/pootle/blob/e53ea85f934689df735f09e57614e1bd0ea12e47/pootle/apps/pootle_misc/checks.py#L949-L1008) is duplicating code from TTK's [`UnitChecker.run_filters()`](https://github.com/translate/translate/blob/981a3579de702e5fd2165553479b7eac791b4793/translate/filters/checks.py#L419-L478), but also changes its API, removes functionalities such as ignoring checks for specific languages, and mutates the checker from the outside [to accommodate `TranslationChecker.run_filters()`' code](https://github.com/translate/translate/blob/981a3579de702e5fd2165553479b7eac791b4793/translate/filters/checks.py#L525-L534).

For the time being, IMHO it'd be much better to:
* Enhance TTK checkers' API to support specifying a list of explicit checks to run
* Remove `run_given_filters()` from Pootle, and make it use a single code path

unho:
+1

julen:
Here's an extra concern: `calculate_checks` now doesn't make use of `Unit.update_qualitychecks()` but calls `run_given_filters()` directly by passing a trimmed-down version of a `Unit` object (for performance reasons). 
However this at the same time couples checkers with the implementation details of such lightweight `Unit` object: checkers can always make free use of TTK unit class capabilities, and failure to provide so can inadvertently render checks broken in Pootle.

PR #4573, opened 2016-03-14
ENChecker: fix running checks via `calculate_checks`

julen:
Due to the confusing double code-path for running checks, checkers are locked
into the setup being done in `run_given_filters()` if they want to remain
compatible with running individual checks from the CLI.

At the same time, note `calculate_checks` never passes an actual unit to the
check runners for performance reasons, but a trimmed-down version of it, and
checkers need to therefore implicitly rely on a CLI command's implementation
details.

These extra concerns should be addressed when dealing with #4553.


unho:
lgtm

unho:
@julen Right now #4553 only has a reference to this PR, so it would be better if comment there about the extra concerns that should be addressed.

PR #4574, opened 2016-03-14
LegalPages: do not perform checking for XHR views

julen:
The check is only relevant for views rendering `layout.html`.


unho:
lgtm

PR #4579, opened 2016-03-16
Edit widget: display priority only if using vfolders

### LINKED ISSUE #4577: Priority is displayed even if no virtual folders are being used
This is a regression: in a server with no virtual folders, units still display a default priority of _1.0_.

julen:
Fixes #4577.


unho:
lgtm

PR #4582, opened 2016-03-17
Add tests for admin view adding TP to Project

phlax:
tests permissions and context for the admin view adding and deleting TPs to/from Projects

~~need to add some POST tests and refactor a little~~

~~**currently rebased on #4585**~~

### DIFF
--- a/tests/views/admin.py
+++ b/tests/views/admin.py
@@ -9,10 +9,80 @@
 
 import pytest
 
-from django.core.urlresolvers import reverse_lazy
+from django.core.urlresolvers import reverse_lazy, reverse
+from django.utils.safestring import mark_safe
+
+from pytest_pootle.env import TEST_USERS
+from pytest_pootle.factories import LanguageFactory
+
+from pootle.core.paginator import paginate
+from pootle.core.url_helpers import split_pootle_path
+from pootle_app.views.admin.util import form_set_as_table
+from pootle_project.models import Project
+from pootle_translationproject.models import TranslationProject
 
 
 ADMIN_URL = reverse_lazy('pootle-admin')
+
+
+def _test_admin_view(response, project):
+    request = response.wsgi_request
+    qs = TranslationProject.objects.filter(
+        project=project).order_by('pootle_path')
+    page = paginate(request, qs)
+
+    url_kwargs = {
+        'project_code': project.code,
+        'dir_path': '',
+        'filename': ''}
+
+    assert page.number == response.context["objects"].number
+    assert page.start_index() == response.context["objects"].start_index()
+    assert page.end_index() == response.context["objects"].end_index()
+    assert (
+        list(response.context["objects"].object_list.values_list("pk", flat=True))
+        == list(qs.values_list("pk", flat=True)))
+
+    assert (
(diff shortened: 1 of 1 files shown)


tests/views/admin.py
```diff
@@ -9,12 +9,83 @@
...
+    qs = TranslationProject.objects.filter(
+        project=project).order_by('pootle_path')
+    page = paginate(request, qs)
+
+    url_kwargs = {
+        'project_code': project.code,
+        'dir_path': '',
+        'filename': ''}
+
+    # bit wierd this one
```
ta2-1: what is wierd?

tests/views/admin.py
```diff
@@ -9,12 +9,83 @@
...
+    assert page.number == response.context["objects"].number
+    assert page.start_index() == response.context["objects"].start_index()
+    assert page.end_index() == response.context["objects"].end_index()
+    assert (
```
ta2-1: is the ordering always defined here?

tests/views/admin.py
phlax: ill remove the comment - but the ``page`` being ``objects`` was what i was referring to mostly because the page has an ``objects_list`` var like in a django list view - so it was a bit confusing at first

tests/views/admin.py
```diff
@@ -36,3 +107,111 @@ def test_admin_access(client):
...
+def test_admin_view_projects_add_tp(english, client, admin):
+
+    user = admin
+    project = Project.objects.get(code="project0")
+
+    new_language = LanguageFactory()
+
+    templates_tp = TranslationProject.objects.create(
+        language=english, project=project)
+    templates_tp.save()
```
ta2-1: why? fs stuff?

phlax: not sure - it may not be necessary

tests/views/admin.py
ta2-1: I didn't want you to remove the comment. Why can't you put this explanation there?

tests/views/admin.py
phlax: ill remove

tests/views/admin.py
phlax: hmm - i would rather change the var name - than add a comment that i thought the name was wierd - but tbh thats defo outside the scope of this PR

tests/views/admin.py
phlax: yep, by pootle_path, both on the test side and in the implementation

tests/views/admin.py
```diff
@@ -36,3 +106,110 @@ def test_admin_access(client):
...
+
+@pytest.mark.django_db
+def test_admin_view_projects_delete_tp(english, client, admin):
+
+    user = admin
+    project = Project.objects.get(code="project0")
+
+    templates_tp = TranslationProject.objects.create(
+        language=english, project=project)
+    templates_tp.save()
```
ta2-1: @phlax I missed this.

phlax: cool - ill remove now - and check if i did it anywhere else

ta2-1:
gtm

PR #4585, opened 2016-03-18
Clean fs for every test

phlax:
This PR:

- adds a session-scoped fixture for use of PootleEnv
- copies out the contents of PootleEnvs translation_dir and set the POOTLE setting
- updates tests/fixtures to make use of new fs env


pytest_pootle/fixtures/models/store.py
```diff
@@ -275,35 +273,6 @@ def _make_member_updates(store, member):
     _mark_unit_fuzzy(store.units[0], member)
 
 
-@pytest.fixture(scope='session')
```
ta2-1: @phlax can we put this change to https://github.com/translate/pootle/pull/4585/commits/3738cca454a24f2163d7e38ae42a1a0ffaf5bb2b commit to show that it was moved?

pytest_pootle/fixtures/models/translation_project.py
```diff
@@ -23,7 +23,8 @@ def _require_tp(language, project):
     """Helper to get/create a new translation project."""
     from pootle_translationproject.models import create_translation_project
 
-    return create_translation_project(language, project)
+    tp = create_translation_project(language, project)
+    return tp
```
ta2-1: is it necessary?

pytest_pootle/fixtures/models/store.py
phlax: yep

pytest_pootle/fixtures/import_export_fixtures.py
```diff
@@ -31,40 +31,32 @@ def file_import_failure(request):
 
 
 @pytest.fixture
-def ts_directory(request, tmpdir):
-    """Sets up a tmp directory with test ts files."""
+def ts_directory(po_directory, request, tmpdir, settings):
```
ta2-1: @phlax I dropped `po_directory` fixture from here and all tests passed.

phlax: cool - we can remove i think i just added to be sure that the env was to true to what would happen - ie the po files would already be there when the ts projects were set up.

phlax: ...but i dont see it making much diff either way

phlax: cool as agreed - kept - but docstring improved

ta2-1:
gtm

PR #4587, opened 2016-03-20
Add view tests

phlax:
Adds tests for remaining untested views:

- tests the output of the xhr timeline view
- get_stats xhr view

### DIFF
--- a/tests/views/timeline.py
+++ b/tests/views/timeline.py
@@ -1,0 +1,285 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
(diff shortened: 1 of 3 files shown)


tests/views/timeline.py
```diff
@@ -0,0 +1,285 @@
...
+    user = request_user["user"]
+    if user.username != "nobody":
+        client.login(
+            username=user.username,
+            password=request_user["password"])
+    response = client.get(url)
+
+    no_permission = (
+        not user.is_superuser
+        and unit not in Unit.objects.get_translatable(user))
```
unho: Shouldn't be `or` instead of `and`?

unho: It is correct as it is, sorry.

unho:
lgtm

PR #4588, opened 2016-03-20
Refactor timeline to class based view

phlax:
This PR refactors the very costly timeline view into a class based view.

It also uses qs.values to miminize the cost of the call.

### DIFF
--- a/pootle/apps/pootle_store/unit/timeline.py
+++ b/pootle/apps/pootle_store/unit/timeline.py
@@ -1,0 +1,261 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from hashlib import md5
+from itertools import groupby
+
+from django.contrib.auth import get_user_model
+from django.core.urlresolvers import reverse
+from django.db.models import Q
+from django.utils.functional import cached_property
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext as _
+
+from pootle_misc.checks import check_names
+from pootle_statistics.models import (
+    Submission, SubmissionFields, SubmissionTypes)
+
+from pootle_store.fields import to_python
+from pootle_store.util import STATES_MAP
+
+
+class DisplayUser(object):
+
+    def __init__(self, username, full_name, email=None):
+        self.username = username
+        self.full_name = full_name
+        self.email = email
+
+    @property
+    def author_link(self):
+        return format_html(
+            u'<a href="{}">{}</a>',
+            self.get_absolute_url(),
+            self.display_name)
+
+    @property
+    def display_name(self):
+        return (
+            self.full_name.strip()
+            if self.full_name.strip()
+            else self.username)
+
+    def get_absolute_url(self):
+        return reverse(
+            'pootle-user-profile',
+            args=[self.username])
+
+    def gravatar_url(self, size=80):
+        email_hash = md5(self.email).hexdigest()
+        return (
+            'https://secure.gravatar.com/avatar/%s?s=%d&d=mm'
+            % (email_hash, size))
+
+
+class SuggestionEvent(object):
+
+    def __init__(self, submission_type, username, full_name):
+        self.submission_type = submission_type
+        self.username = username
+        self.full_name = full_name
+
+    @cached_property
+    def user(self):
+        return DisplayUser(self.username, self.full_name)
+
+    @property
+    def description(self):
+        author_link = self.user.author_link
+        description_dict = {
+            SubmissionTypes.SUGG_ADD: _(u'Added suggestion'),
+            SubmissionTypes.SUGG_ACCEPT: _(u'Accepted suggestion from %s',
+                                           author_link),
+            SubmissionTypes.SUGG_REJECT: _(u'Rejected suggestion from %s',
+                                           author_link)}
+        return description_dict.get(self.submission_type, None)
+
+
+class ProxySubmission(object):
+
+    def __init__(self, values):
+        self.values = values
+
+    @property
+    def old_value(self):
+        return self.values['old_value']
+
+    @property
+    def new_value(self):
+        return self.values['new_value']
+
+    @property
+    def field(self):
+        return self.values['field']
+
+    @property
+    def field_name(self):
+        return SubmissionFields.NAMES_MAP.get(self.field, None)
+
+    @property
+    def type(self):
+        return self.values['type']
+
+    @property
+    def qc_name(self):
+        return self.values['quality_check__name']
+
+    @property
+    def suggestion(self):
+        return self.values['suggestion_id']
+
+    @property
+    def suggestion_full_name(self):
+        return self.values['suggestion__user__full_name']
+
+    @property
+    def suggestion_username(self):
+        return self.values['suggestion__user__username']
+
+    @property
+    def suggestion_target(self):
+        return self.values['suggestion__target_f']
+
+
+class TimelineEntry(object):
+
+    def __init__(self, submission):
+        self.submission = submission
+
+    @property
+    def entry_dict(self):
+        return {
+            'field': self.submission.field,
+            'field_name': self.submission.field_name,
+            'type': self.submission.type}
+
+    def base_entry(self):
+        entry = self.entry_dict
+        entry['new_value'] = to_python(self.submission.new_value)
+        return entry
+
+    def qc_entry(self):
+        entry = self.entry_dict
+        check_name = self.submission.qc_name
+        check_url = (
+            u''.join(
+                [reverse('pootle-checks-descriptions'),
+                 '#', check_name]))
+        entry.update(
+            {'check_name': check_name,
+             'check_display_name': check_names[check_name],
+             'checks_url': check_url})
+        return entry
+
+    def state_change_entry(self):
+        entry = self.entry_dict
+        entry['old_value'] = STATES_MAP[int(to_python(self.submission.old_value))]
+        entry['new_value'] = STATES_MAP[int(to_python(self.submission.new_value))]
+        return entry
+
+    def suggestion_entry(self):
+        entry = self.entry_dict
+        suggestion_description = mark_safe(
+            SuggestionEvent(
+                self.submission.type,
+                self.submission.suggestion_username,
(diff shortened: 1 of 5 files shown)


pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+        if (has_creation_entry):
+            grouped_entries[0]['created'] = True
+        else:
+            created = {
+                'created': True,
+                'submitter': User.objects.get_system_user()}
+            if self.object.creation_time:
+                created['datetime'] = self.object.creation_time
+            grouped_entries[:0] = [created]
+        return grouped_entries
```
unho: All this piece of code looks like it comes from the possible absence of a creation submission. My guess is that it would be cleaner to add a migration that create those missing submissions so we can simplify this bit of code to something like:

```python
    def add_creation_entry(self, grouped_entries):
        grouped_entries[0]['created'] = True  # There is always some submission.
        return grouped_entries
```

But I am fine with keeping this as is and get the PR landed as soon as possible. This improvement can be seen as some further future post-landing cleanup.

pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+                              .exclude(field=SubmissionFields.COMMENT,
+                                       creation_time=self.object.commented_on)
+                              .order_by("id"))
+
+    @property
+    def language(self):
+        return self.object.store.translation_project.language
+
+    def add_creation_entry(self, grouped_entries):
+        User = get_user_model()
```
unho: Perhaps this is better to have at module level.

unho:
Reviewed and tested. lgtm

pootle/apps/pootle_store/views.py
phlax: @unho yep agreed its a kludge, but one that was already there so lets clean up after

ta2-1:
@phlax this code can be removed https://github.com/translate/pootle/blob/1891669d9a6babf2843c481e9d02386871c72717/pootle/apps/pootle_statistics/models.py#L327-L346.

pootle/apps/pootle_store/views.py
ta2-1: @unho IIRC we discussed such migration and it would add ~30000000 rows to submission table but all necessary information is already in DB. So we decided not to add such migration.

pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+        has_creation_entry = (
+            len(grouped_entries) > 0
+            and grouped_entries[0]['datetime'] == self.object.creation_time)
+        if (has_creation_entry):
```
ta2-1: unnecessary brackets

pootle/apps/pootle_store/views.py
phlax: @ta2-1 that sounds like a pretty valid reason not to migrate!

pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+        elif item['quality_check__name']:
+            check_name = item['quality_check__name']
+            check_url = (
+                u''.join(
+                    [reverse('pootle-checks-descriptions'),
+                     '#', check_name]))
+            entry.update(
+                {'check_name': check_name,
+                 'check_display_name': check_names[check_name],
+                 'checks_url': check_url})
```
ta2-1: I've not paid attention to such things before but just realised that I agree with @julen here re readability style:
```
entry.update({
    'check_name': check_name,
    'check_display_name': check_names[check_name],
    'checks_url': check_url
})
```

phlax: im sorry but i dont - and i really dont want to go back to the situation we had before

pootle/apps/pootle_store/views.py
unho: I can only agree.

pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+        return Unit.objects.get_translatable(self.request.user).select_related(
+            "store__parent",
+            "store__translation_project__project",
+            "store__translation_project__language",
+            "store__translation_project__directory")
+
+    def _get_suggestion_description(self, submission):
+        user_url = reverse(
+            'pootle-user-profile',
+            args=[submission["suggestion__user__username"]])
```
phlax: i would be happy for you to add it, not sure if it should be ProxySuggestion or ProxySubmission or both if you are going down that road

phlax:
...also thinking about it @ta2-1 im wondering if we should create a Timeline(unit) class and move the timeline logic into there

pootle/apps/pootle_store/views.py
```diff
@@ -437,6 +333,181 @@ def source_language(self):
...
+    @property
+    def project(self):
+        return self.object.store.translation_project.project
+
+    def get_queryset(self):
+        return Unit.objects.get_translatable(self.request.user).select_related(
+            "store__parent",
+            "store__translation_project__project",
+            "store__translation_project__language",
+            "store__translation_project__directory")
```
ta2-1: @phlax I checked SQL and didn't notice any diff if we leave 
```
"store__translation_project__project",
"store__translation_project__language",
```
only. Have I missed anything?

ta2-1: and if I put `__directory` to `project` it will be less for 1 query.
```select_related(
    "store__translation_project__project__directory",
    "store__translation_project__language")
```

ta2-1:
lgtm. there are some nits.

pootle/apps/pootle_store/views.py
phlax: hmm yep, i think i follow ill play around with the SRs

pootle/apps/pootle_store/unit/timeline.py
```diff
@@ -0,0 +1,170 @@
...
+            if self.submission["submitter__full_name"].strip()
+            else self.submission["submitter__username"])
+
+    def get_absolute_url(self):
+        return reverse(
+            'pootle-user-profile',
+            args=[self.submission["submitter__username"]])
+
+    def gravatar_url(self, size=80):
+        email_hash = md5(self.submission['submitter__email']).hexdigest()
```
julen: I understand these proxy classes are for performance reasons (and to avoid filling up the memory with heavy `User` models), although it'd be cool to avoid duplicated logic with regards to the `User` model by externalizing the logic as functions and having both the proxy classes and models use these. Thoughts?

phlax: i dont really like the style of lots helper functions. As the code had become quite complex im refactoring it out to make more sense into ~reusable classes - just about to push some further updates on this

phlax: also i should add i try to avoid adding methods to models, because it obliges you to instantiate the model, to make use of it. My preference would be that code that currently uses User.gravatar_url etc should use something like one of these proxy classes that can potentially work from just a values dict or from a full object.

pootle/apps/pootle_store/unit/timeline.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @property
+    def grouped_entries(self):
+        grouped_entries = self.get_grouped_timeline_entries()
+        grouped_entries = self.add_creation_entry(grouped_entries)
+        grouped_entries.reverse()
+        return grouped_entries
+
+    @property
+    def timeline(self):
```
julen: Being `Timeline` the name of the class, it's probably redundant to have `timeline*` in contained attributes and methods. For this specific case, what about naming the method `events`?

pootle/apps/pootle_store/unit/timeline.py
julen: Cool 👍 looking forward to see how duplication is avoided.

pootle/apps/pootle_store/unit/timeline.py
phlax: yep, was wondering about this. As it returns a queryset, im erring towards calling `qs` or `queryset`

pootle/apps/pootle_store/unit/timeline.py
phlax: its not quite finished, but please see latest push for an example of how it can.

pootle/apps/pootle_store/unit/timeline.py
phlax: ...or perhaps submissions as that is what the qs is

julen:
I tested the PR at its current shape and didn't experience any changes with respect to master in terms of queries at least, although as I read the comments this is still in flux and am looking forward to re-check it once it's been stabilized. Didn't try with really long timelines, but I believe the memory footprint should notably reduce there compared to master.

pootle/apps/pootle_store/views.py
julen: To reiterate what has been said before, this is about keeping consistency throughout the codebase and sticking to a common practice used throughout years in all the major projects under the translate umbrella (TTK, virtaal, pootle). It's true that the previous lack of linting unfortunately contributed to nitpicky reviews and no one wants or is suggesting to go back there. Nevertheless here we're facing a major inconsistency in multi-line dicts, lists and calls, which is a recent thing.

Needless to say it's in the project's best interest and it's everyone's responsibility to keep the code consistently readable, and not contribute to a mess which becomes uncomfortable.

pootle/apps/pootle_store/unit/timeline.py
```diff
@@ -0,0 +1,214 @@
...
+
+class SuggestionEvent(object):
+
+    def __init__(self, submission_type, username, full_name):
+        self.submission_type = submission_type
+        self.username = username
+        self.full_name = full_name
+
+    @cached_property
+    def author(self):
```
unho: This looks like it belongs to `DisplayUser` rather that this class.

pootle/apps/pootle_store/views.py
unho: Not that I want to reopen this fight, but the following:

```python
entry.update({
    'check_name': check_name,
    'check_display_name': check_names[check_name],
    'checks_url': check_url,
})
```

is preferable to:

```python
entry.update(
    {'check_name': check_name,
     'check_display_name': check_names[check_name],
     'checks_url': check_url})
```

for two reasons:

- It is more readable
- It always implies same length or shorter diffs if the code is changed

ta2-1:
@phlax thanks for your work here. All new updates look good.
GTM!

PR #4596, opened 2016-03-24
Don't show unaccesible projects

francescortiz:


ta2-1:
Hi @francescortiz, thanks for your PR. In my review I noticed that there where other cases like the one your change addressed. So I've created #4616 which addresses all cases where we need to filter by accessible projects. Thanks again for the PR, if you are able to give any input or test the new PR then that would be appreciated.

ta2-1:
#4616 is landed. So I'm closing this PR.

PR #4598, opened 2016-03-25
Add checkerstyle migration

dwaynebailey:
This is a nuisance migration, but is needed to prevent the error

```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

from appearing after running `pootle migrate` a second time.

This PR includes #4595 which was PRs to show that the check in Travis actually does catch unlanded migrations.


phlax:
lgtm

PR #4601, opened 2016-03-26
Editor js fixes and cleanups

phlax:
- Fix: editor js fetches batch at last offset correctly


unho:
lgtm, but I would split the "Fix: remove autosize event listeners to..."  commit in two separate commits, and add detailed commit messages explaining the issue they are fixing (IMHO it is better to explain in commit message than referencing the issue).

ta2-1:
gtm. tested.

PR #4604, opened 2016-03-27
Pootle plugin 1: add decorators for providers and getters

phlax:
This PR adds signal hooks to allow views and other functionality to be extended by other apps

### DIFF
--- a/pootle/core/plugin/__init__.py
+++ b/pootle/core/plugin/__init__.py
@@ -1,0 +1,12 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from delegate import getter, provider
+
+
+__all__ = ("getter", "provider")
--- a/pootle/core/plugin/delegate.py
+++ b/pootle/core/plugin/delegate.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.dispatch import Signal
+from django.dispatch.dispatcher import NO_RECEIVERS
+
+from .exceptions import StopProviding
+from .results import GatheredDict
+
+
+class Provider(Signal):
+
+    result_class = GatheredDict
+
+    def __init__(self, *args, **kwargs):
+        self.result_class = kwargs.pop("result_class", self.result_class)
+        super(Provider, self).__init__(*args, **kwargs)
+
+    def gather(self, sender=None, **named):
+        gathered = self.result_class(self)
+        named["gathered"] = gathered
+        no_receivers = (
+            not self.receivers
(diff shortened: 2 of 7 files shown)


pootle/apps/pootle_project/views.py
```diff
@@ -171,6 +171,10 @@ def dispatch(self, request, *args, **kwargs):
     def pootle_path(self):
         return self.object.pootle_path
 
+    @property
+    def request_path(self):
+        return "/projects/%(project_code)s/%(dir_path)s" % self.kwargs
```
unho: Is the `dir_path` actually necessary when translating a project?

phlax: this allows you to eg get_stats for subdirs

unho: This is `ProjectTranslateView` so I don't think it is going to retrieve `get_stats` in any case.

pootle/core/views.py
```diff
@@ -523,6 +524,20 @@ def get_context_data(self, *args, **kwargs):
             'export_url': self.export_url,
             'browse_url': self.browse_url}
 
+    def gather_context_data(self, context):
```
unho: Doesn't this clash with `gather_context_data` signal imported above?

pootle/apps/pootle_project/views.py
phlax: id need to debug - but im pretty sure its necessary

pootle/core/views.py
phlax: no

pootle/core/views.py
```diff
@@ -646,26 +660,26 @@ def get_context_data(self, *args, **kwargs):
...
+        return self.request_path, "", {}
 
     @property
-    def display_vfolder_priority(self):
-        return False
+    def resource_path(self):
+        return (
+            "%s%s"
+            % (self.extracted_path[1],
+               self.request_path[len(self.ctx_path):]))
```
phlax: @unho the request_path is used to here do determine the resource_path (ie pootle_path without the tp part or /project(s) part) - in the case of vfolders it also uses the extracted_path[1] which is basically any additional path elements - in this case ``$virtualfoldername/``

pootle/core/views.py
```diff
@@ -646,26 +660,26 @@ def get_context_data(self, *args, **kwargs):
 class PootleTranslateView(PootleDetailView):
     template_name = "editor/main.html"
 
-    @property
-    def ctx_path(self):
-        return self.pootle_path
```
unho: Why is `ctx_path` being removed?

phlax: because it was always being subclassed and coveralls showed it as not being used.

I think perhaps it should be copied out from one of the subclasses tho

ta2-1:
@phlax lgtm, I'm glad that you starter working on pluggable system.
I realise that documentation is required here but it's worth to add it later (not now) when we have ready examples how we are using it.

phlax:
@ta2-1 thanks, for ref i started a wiki page, and hope that we once the interfaces are somewhat stabilised we can use as basis for dev docs - wiki page is here https://github.com/translate/pootle/wiki/PootlePlugin

julen:
As @ta2-1 said, 👍 for taking lead on this one. Also many thanks for the link to the wiki page, I believe that's a key piece I was missing here, and will try to have a look at it in the next few days.

The PR looks good to me, you will need to adjust the commit messages now that this is in `pootle.core.plugin` though.

phlax:
> The PR looks good to me, you will need to adjust the commit messages now that this is in pootle.core.plugin though.

was pushing the update as you posted 8)

PR #4617, opened 2016-04-01
Cleanup: avoid extra nesting by aborting early

julen:
Addresses the _problematic_ case first.
(Related read: https://philikon.wordpress.com/2010/01/26/flat-is-better-than-nested-bail-out-early/).


unho:
lgtm

PR #4624, opened 2016-04-05
Factories: add a few subfactories

julen:
This allows to generate such factory models without passing existing related
models.

(Note that moving factories around is necessary due to inter-dependencies.)


julen:
Closing this (based on #4627); plus specific use-cases will bring any necessary factories.

PR #4627, opened 2016-04-05
Factories: adjust naming to be explicit when they require DB access

julen:
It's surprising the fact that factories depend directly on the DB, hence
mutilating the option to use a "build" strategy (no DB writes) over a "create"
strategy (writes to the DB).

These resulted to be footguns when using them as subfactories in other
factories, so for now it'll be better to explicitly mention that they require DB
access.


unho:
lgtm

PR #4632, opened 2016-04-05
Test env: use `get_or_create()`

julen:
While this commit on its own doesn't still enable reusing the DB across test
runs with pytest-django's `--reuse-db`, it definitely aims to be a step in that
direction.


unho:
lgtm

PR #4634, opened 2016-04-06
Cleanup: move functions to manager methods

julen:
This PR converts the `get_tasks()` and `get_scores()` functions from scores and paid tasks into manager methods because that's basically what they are.


unho:
lgtm.

ta2-1:
lgtm

PR #4642, opened 2016-04-12
Add pootle comments app

phlax:
This PR adds a pootle_comments app to:

- add django_comments (previously ka django.contrib.comments)
- customize comment forms
- create custom comment model - this is not used but would make it much easier in the event that we do want to customize the model


ta2-1:
The code looks good.
Using of `django_contrib_comments` allows us to externalise comments through the whole system (as inject comments from Submission model). 
Also it's worth to mention that there is a first example how `pootle_plugin.core.getter` can be used.
GTM!

phlax:
thanks @ta2-1 - ill merge

PR #4646, opened 2016-04-14
Pootle plugin 2

phlax:
Move virtualfolder code out of other apps


tests/core/plugin/results.py
```diff
@@ -84,3 +84,34 @@ def test_gathered_list():
...
+    members = [
+        [2, 3, 4],
+        [3, 4, 5],
+        [4, 4, 7],
+        ["foo", "bar"],
+        None]
+    memberset = set()
+    [memberset.update(member) for member in members if member]
+    [gs.add_result(None, member)
+     for member in members]
```
unho: Is it really necessary to wrap this?

tests/core/plugin/results.py
```diff
@@ -84,3 +84,34 @@ def test_gathered_list():
...
+    provider_test = Provider()
+    gs = GatheredSet(provider_test)
+    members = [
+        [2, 3, 4],
+        [3, 4, 5],
+        [4, 4, 7],
+        ["foo", "bar"],
+        None]
+    memberset = set()
+    [memberset.update(member) for member in members if member]
```
unho: Using list comprehension looks a bit weird. Is it really necessary?

tests/core/plugin/results.py
phlax: i dont think i understand. This is testing what this class does

tests/core/plugin/results.py
phlax: some kind of loop is necessary.

tests/core/plugin/results.py
unho: Doesn't it fit in a single line? But forget about that, this looks saner like:

```python
    for member in members:
        gs.add_result(None, member)
```

tests/core/plugin/results.py
unho: What about a regular `for` loop?

```python
    for member in members:
        if member:
            memberset.update(member)
```

tests/core/plugin/results.py
phlax: fine ill update

tests/core/plugin/results.py
phlax: ill update with something similar

PR #4648, opened 2016-04-15
Integrate invoices

### LINKED ISSUE #4619: Integrate existing command to generate invoices
At Evernote we have developed and have been actively using an invoicing solution integrated with Pootle, which with the help of the existing reporting code in Pootle, allows deployments to generate HTML and PDF invoices and have them sent via e-mail.

We believe this can be beneficial for any Pootle instance running in a context similar to ours (with paid translators and/or agencies involved), and we are happy to make this code available for everyone, including the necessary instructions and documentation.

At first we thought we could make a separate package out of it, although after checking how reports currently work internally, it'd be best to merge this in Pootle. This way anyone in a situation like ours can start using invoicing features right away. We are happy to look towards decoupling both reporting and invoicing features in the future, though.

We'd love to get a :+1: before creating a PR for this.

julen:
This PR adds a command for being able to generate invoices out of existing user activity. It includes the command, default templates, docs and tests.

Some commits from our internal repo have been preserved in order to understand better how we reached the current implementation, hence due to the number of commits I'd suggest to check the entire diff only.

Fixes #4619.


ta2-1:
Hi @julen, you did a great work 👍 !

Some comments below:
- I wonder why settings are called `POOTLE_INVOICES...` when `invoices` is a part of `reports` app.
- `pootle generate_invoices` doesn't output anything (if you change nothing). It's confusing.
- `py.test` generates `pootle/log/invoices/...` when it shouldn't.

Reading this PR I'm realising that it really should be a plugin containing `PaidTask` model.

I'll appreciate if you pick your set of questions regarding how this plugin could be created.
As I mentioned before I could see non-technical (not about implementation) questions:
- where to keep plugins?
- how to advertise plugin infrastructure?
- how to pin Pootle versions for different plugins?
I could be wrong and there are more important questions we have to answer before. But we need to prioritise all this stuff. And I hope we can create a plugin development roadmap.

julen:
Thanks for your time to have a look a this @ta2-1.

> I wonder why settings are called POOTLE_INVOICES... when invoices is a part of reports app.

As you know, we were initially working towards another direction (again, not all of the commits could be imported from our internal repo), besides that there's no specific reason. I'll get to rename this quickly.

> `pootle generate_invoices` doesn't output anything (if you change nothing). It's confusing.

Would displaying something like:
_"No invoicing configuration found, nothing to be done.
Please read the docs at XXX to learn more about how to use this feature."_
work for you? Do you have any specific suggestion?

> `py.test` generates `pootle/log/invoices/...` when it shouldn't.

This is created by the `generate_invoices` command when it's run (by the tests), and it's expected behavior. Do you perhaps mean it should be created in a temporary location?

Re. plugin I'm happy to think through further but I'm afraid this PR won't be the appropriate forum to do so.

pootle/settings/40-apps.conf
```diff
@@ -131,6 +131,23 @@ POOTLE_SIGNUP_ENABLED = True
 # chart with events
 POOTLE_REPORTS_MARK_FUNC = ''
 
+# Invoicing
+
+# Path to PhantomJS binary (needed to generate PDF invoices)
+POOTLE_INVOICES_PHANTOMJS_BIN = None
+
+# Base directory where PDF invoices will be created
+POOTLE_INVOICES_DIRECTORY = working_path('invoices')
```
unho: Won't this create problems if the directory is missing? I suggest creating the directory with a README file within to ensure the directory is always present, like we did for `dbs` directory: https://github.com/translate/pootle/tree/master/pootle/dbs

pootle/apps/reports/models/invoice.py
```diff
@@ -0,0 +1,376 @@
...
+    return timezone.now().replace(day=1) - timedelta(days=1)
+
+
+class Invoice(object):
+
+    required_config_fields = (
+        ('name', 'paid_by', 'wire_info'),  # Core fields
+        ('email', 'accounting_email'),  # Email-related fields
+    )
+    generators = (HTMLGenerator, PDFGenerator)
```
unho: This is `GENERATOR_MODULES`.

pootle/settings/40-apps.conf
julen: The configured directory will be created if missing, provided the command has enough permissions to do so: I believe it's safe to assume so as this should be the case if `working_path('invoices')` wants to be used.

pootle/apps/reports/models/invoice.py
julen: Good catch, this one is dead code.

julen:
Thanks for the comments folks; I consider that the raised concerns have been addressed.

dwaynebailey:
@julen A few points

1. Raised concerns - I don't think most of us have time available, motivation or energy to review to 60+ commits that make up this PR.  So any concerns raised from a technical perspective are very shallow reviews.  With 2.8.0 on our radar this is way down the list of priority focus areas.
2.  2.8.0 is on our radar, this can't land as part of that I'm afraid. There is just too much at stake in 2.8.0 to carry this bag of code with it and too much risk of issues related to stability and testing.
3. And the **most critical issues**.  This should be a plugin.  The justification to not put any effort into breaking this out into a plugins is still to mind very very week.  The code, I'm afraid, does raise maintenance considerably for a feature likely to be used by a small subset of users and not understood or used by most of the core devs.  A plugin deals with it elegantly and we agreed in future that we'd move towards plugins.  The virtual folder code got plugin treatment as this code should also. So since we've agreed, and vfolders showed its possible, I'm left feeling a bit in the dark why there is this really hard push to land this in core and not even try to move it to a plugin.  Is there a backstory that we're just missing here?

julen:
As said on IRC, feel free to raise any politics on #4619, which was opened to avoid this type of problems in case the PR was created.

Re. planning we understand, and this can definitely rest for a bit (we'd aim to use this code at the end of May); not aiming at a bit-rotten PR though.

Also please realize that no one asked to review 60+ individual commits, but rather the contrary. Note these are 3 modules (2 very simple) and the command. The rest are docs and tests.

docs/features/invoices.rst
```diff
@@ -0,0 +1,182 @@
...
+
+Out of users' activity and with the help of some extra configuration, Pootle is
+able to generate monthly-based invoices and optionally send them via e-mail.
+This is useful when paid contractors and/or agencies are involved in the
+translation process.
+
+
+.. _invoices#calculation:
+
+Calculation of Amounts
```
iafan: Let's call this `Calculation of Payments`

docs/features/invoices.rst
```diff
@@ -0,0 +1,182 @@
...
+----------------------
+
+Total amounts are calculated based on the :ref:`rates set for users
```
iafan: "Total amounts" => "Payments"

docs/features/invoices.rst
```diff
@@ -0,0 +1,182 @@
...
+too. These tasks allow adding a description to easily identify the type of work
+being reported.
+
+In order to manually add paid tasks, administrators can go to the *Reports*
+section of the administration site, select a user from the drop-down and add
+tasks below.
+
+Alternatively this can also be done by translators themselves by going to their
+statistics page and adding the tasks right below their daily activity graph.
+This option is only available to them once an administrator have set payment
```
iafan: typo: "have" => "has"

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+  The user configuration accepts the following keys:
+
+  ``name`` (required)
+    Full name as displayed in the invoice. The official "full" person name can
+    be different from the public *shortened* version available in the Pootle
+    profile, hence this name is used instead as it's dealing with financial
+    documents.
+
+  ``wire_info`` (required)
+    Wire information which includes payment details.
```
iafan: I think we can mention that this can be a multi-line string, and that such a multi-line string will properly be rendered as a multi-line one in the invoice. Also, to allow better formatting of the configuration files, we will trim leading and trailing spaces on each line of this multi-line value.

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+
+  ``paid_by`` (required)
+    Official name of the company/person paying the invoice.
+
+  ``invoice_prefix``
+    Prefix used to generate invoice IDs. This will be prepended to the current
+    month in `YYYY-MM` format as a string.
```
iafan: Worth recommending the "Language-Initials" format. For example, if there's a translator named "John Doe" translating into Spanish (ES), the prefix might look like 'ES-JD-' and the invoices generated for this person will be consequently named like 'ES-JD-2016-06', 'ES-JD-2016-07' and so on. If one wants these IDS to be globally unique, they need to specify a unique prefix for each of their paid translators.

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+
+  ``minimal_payment``
+    Minimal amount required to process the invoice. If the summed up amount is
```
iafan: "Minimal amount required to process the invoice." => "Minimal payment threshold value to be reached for the payment to be sent out during the monthly pay cycle."

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+
+  ``subcontractors``
+    A list of usernames which will be considered as subcontractors of the main
+    user. This way, subcontractors' work is consolidated in a single invoice.
+
+  When sending emails with ``--send-emails``, the following keys are available
+  as well:
+
+  ``email`` (required)
+    E-mail address where the invoice will be sent to. It can contain multiple
```
iafan: E-mail address __of a translator__ where the invoice will be sent to.

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+    Official name of the company/person paying the invoice.
+
+  ``invoice_prefix``
+    Prefix used to generate invoice IDs. This will be prepended to the current
+    month in `YYYY-MM` format as a string.
+
+  ``minimal_payment``
+    Minimal amount required to process the invoice. If the summed up amount is
+    lower than ``minimal_payment``, the amount for the month will be carried
+    over to the next month.
```
iafan: Need to mention that this amount is in the currency set along with the rate in the Pootle user reports UI.

docs/server/settings.rst
```diff
@@ -250,6 +250,147 @@ Configuration settings for applications used by Pootle.
...
+
+  ``extra_add``
+    Fixed amount to be added on top of the total accrued amounts. Use this e.g.
+    to reimburse transaction fees.
```
iafan: Need to mention that this amount is in the currency set along with the rate in the Pootle user reports UI.

iafan:
@julen, I went through the documentation and provided the feedback. I think that overall the documentation is pretty solid. Great job!

docs/server/settings.rst
julen: I have added a note about this, however note the original code only removes initial leading whitespace; I won't change the original behavior in this PR, we can do that later if needed.

docs/server/settings.rst
julen: Thanks for this. I added your explanation almost verbatim.

docs/server/settings.rst
julen: I have adjusted the text to use your suggestion but it'd be great to have extra feedback on this one: in my mind _sent out_ can be confused with sending emails (even if there's a explicit section below for that), which is an independent function, i.e. `minimal_payment` will be taken into account regardless of sending emails.

docs/server/settings.rst
```diff
@@ -250,6 +250,164 @@ Configuration settings for applications used by Pootle.
...
+    A recommended approach for prefixing invoices is the *"Language-Initials"*
+    format. For example, if there's a translator named *"John Doe"* translating
+    into Spanish (ES), the prefix might look like ``'ES-JD-'`` and the invoices
+    generated for this person will be consequently named like *'ES-JD-2016-06'*,
+    *'ES-JD-2016-07'* and so on. If there is a need for these IDs to be globally
+    unique, one needs to specify a unique prefix for each of the paid
+    translators.
+
+  ``minimal_payment``
+    Minimal payment threshold value to be reached for the payment to be sent out
```
iafan: @julen Let's change "sent out" to "initiated" to avoid the potential confusion you mentioned in the previous review cycle.

PR #4652, opened 2016-04-16
Add search_backend getter

phlax:
Replaces POOTLE_SEARCH_BACKEND settings with a search_backend getter.

To add a custom search backend, an app with the relevant getter and backend should be added before pootle_store in INSTALLED_APPS

### DIFF
--- a/pootle/apps/pootle_store/apps.py
+++ b/pootle/apps/pootle_store/apps.py
@@ -6,4 +6,15 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-default_app_config = 'pootle_store.apps.PootleStoreConfig'
+import importlib
+
+from django.apps import AppConfig
+
+
+class PootleStoreConfig(AppConfig):
+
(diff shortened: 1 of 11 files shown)


julen:
Changes look good to me and testing didn't show any issues.

> To add a custom search backend, an app with the relevant getter and backend should be added before pootle_store in INSTALLED_APPS

In general, do I understand it correctly that the first getter defined/registered (in the sense of Django initialization, i.e. `INSTALLED_APPS`) for a delegate wins?
I know this is still in the early stages although as things stabilize we should definitely document information like this somewhere.

phlax:
> In general, do I understand it correctly that the first getter defined/registered (in the sense of Django initialization, i.e. INSTALLED_APPS) for a delegate wins?

yep, well, the first one that returns a non-None result

> I know this is still in the early stages although as things stabilize we should definitely document information like this somewhere. 

Agreed, I was hoping to land the next 2 plugin PRs which will add some hooks, and separate out the vfolder code, and then i was intending to add some dev docs that outlined the delegation points

PR #4653, opened 2016-04-16
Get units fix export view

### LINKED ISSUE #4618: Export view only displays 18 units
When opening up the export view, I noticed that when there are results, it always
* Displays the first 18 units
* The message at the bottom says _We displayed the first 9 units out of NN. Please refine your search._

After inspecting with git bisect, 8ec8f5fe1 seems to have introduced this.

phlax:
Allows export view to use a full queryset

Fixes #4618

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -467,6 +467,8 @@ class UnitSearchForm(forms.Form):
             ('locations', _('Locations'))),
         initial=['source', 'target'])
 
+    default_count = 10
+
     def __init__(self, *args, **kwargs):
         self.request_user = kwargs.pop("user")
         super(UnitSearchForm, self).__init__(*args, **kwargs)
@@ -483,9 +485,14 @@ class UnitSearchForm(forms.Form):
             self.cleaned_data["user"] = self.request_user
         if self.errors:
             return
-        self.cleaned_data['count'] = min(
-            self.cleaned_data.get("count", 10) or 10,
-            self.cleaned_data["user"].get_unit_rows() or 10)
+        if self.default_count:
+            count = (
+                self.cleaned_data.get("count", self.default_count)
+                or self.default_count)
+            user_count = (
+                self.cleaned_data["user"].get_unit_rows()
+                or self.default_count)
+            self.cleaned_data['count'] = min(count, user_count)
         self.cleaned_data["vfolder"] = None
         pootle_path = self.cleaned_data.get("path")
         if 'virtualfolder' in settings.INSTALLED_APPS:
@@ -552,8 +559,7 @@ class UnitExportForm(UnitSearchForm):
         max_length=2048,
         required=False)
 
+    default_count = None
+
     def clean_path(self):
         return self.cleaned_data.get("path", "/") or "/"
-
-    def clean_count(self):
-        return EXPORT_VIEW_QUERY_LIMIT
--- a/pootle/apps/pootle_store/unit/search.py
+++ b/pootle/apps/pootle_store/unit/search.py
@@ -17,7 +17,7 @@ from pootle_store.views import SIMPLY_SORTED
 
 class DBSearchBackend(object):
 
-    default_chunk_size = 10
+    default_chunk_size = None
     default_order = "store__pootle_path", "index"
     select_related = (
         'store__translation_project__project',
@@ -167,7 +167,7 @@ class DBSearchBackend(object):
         if find_unit:
             # find the uid in the Store
             uid_list = list(self.results.values_list("pk", flat=True))
-            if self.uids[0] in uid_list:
+            if self.chunk_size and self.uids[0] in uid_list:
                 unit_index = uid_list.index(self.uids[0])
                 start = (
                     int(unit_index / (2 * self.chunk_size))
@@ -182,6 +182,8 @@ class DBSearchBackend(object):
             for i, uid in enumerate(uid_list):
                 if uid in self.previous_uids:
                     start = _start + i + 1
+        if self.chunk_size is None:
+            return total, 0, total, self.results
         start = start or 0
         end = min(start + (2 * self.chunk_size), total)
         return total, start, end, self.results[start:end]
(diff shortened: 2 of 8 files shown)


unho:
This renders useless the ability to set a "Number of Rows" in user's settings, isn't it?

phlax:
@unho yep

julen:
Mind elaborating on that? Is there something preventing from keeping the current behavior in the editor while fixing the export view? Note at least a dozen of our active translators have a custom setting, and we'd like to keep it that way for the time being.

phlax:
@julen because the editor doesnt really use it any more afaict

julen:
It has always been used, check out https://github.com/translate/pootle/blob/1891669d9a6babf2843c481e9d02386871c72717/pootle/templates/editor/_scripts.html#L25 as a reference.

phlax:
my point is more that the ui itself doesnt make use of it - ie it just shows you the visible units.

as the editor requests a minimum of 20 units in either direction, i reckon this covers even the largest of view screens

julen:
I have the feeling we are not on the same page: I'm talking about the _Number of Rows_ user setting @unho mentioned earlier, which the UI is using at the moment (including in this PR): try changing your user setting and you'll see it in action.

Besides, I agree that having some upper and lower boundaries in this setting makes sense, although I consider that's beyond the scope of fixing #4618.

julen:
From what I can tell this is still showing exactly the same issues as originally reported (I triple-checked I was running the updated code).

phlax:
@julen ill check

phlax:
@julen apologies for not testing this properly previously. I believe this now works as before.

pootle/apps/pootle_store/unit/search.py
```diff
@@ -183,5 +183,8 @@ def search(self):
                 if uid in self.previous_uids:
                     start = _start + i + 1
         start = start or 0
-        end = min(start + (2 * self.chunk_size), total)
-        return total, start, end, self.results[start:end]
+        if self.chunk_size is None:
+            return total, 0, total, self.results
+        else:
+            end = min(start + (2 * self.chunk_size), total)
+            return total, start, end, self.results[start:end]
```
unho: This looks cleaner like:

```python
        if self.chunk_size is None:
            return total, 0, total, self.results

        start = start or 0
        end = min(start + (2 * self.chunk_size), total)
        return total, start, end, self.results[start:end]
```

phlax: cool, ill update

pootle/apps/pootle_store/forms.py
```diff
@@ -483,9 +485,14 @@ def clean(self):
             self.cleaned_data["user"] = self.request_user
         if self.errors:
             return
-        self.cleaned_data['count'] = min(
-            self.cleaned_data.get("count", 10) or 10,
-            self.cleaned_data["user"].get_unit_rows() or 10)
+        if self.default_count:
```
unho: This is always `10`, so this evaluates as `True`, thus is unnecessary.

phlax: no - not in subclasses its not

unho:
Tested and this now correctly returns all the units while previously it was limited to the number of rows set for the editor. Perhaps correctly it is not the best word since this can have a massive impact on performance, but it returns all the units matched by the filtering.

pootle/core/views.py
```diff
@@ -48,6 +48,9 @@
 from .utils.json import PootleJSONEncoder, jsonify
 
 
+EXPORT_VIEW_QUERY_LIMIT = 10000
```
dwaynebailey: Wondering if this should be a POOTLE_* config variable so its more easily adjustable?

phlax: agreed, that would also make it testable 8)

unho:
Also my tests show that I am now seeing 24 units after and 24 before current unit in editor if I set 50 as Number of Rows in my user's settings. gtm

pootle/apps/pootle_store/forms.py
unho: Yes, you are right.

julen:
Cool, the export view now works as expected.

To avoid confusion with import/export, the `POOTLE_EXPORT_LIMIT` could probably be renamed to `POOTLE_EXPORT_VIEW_LIMIT`, and at the same time docs for it need to be added, as they are missing at the moment. Some of the commits referencing it can probably be squashed, too, and perhaps the default limit set to something lower like `1000`.

As already commented earlier, since there's a change on the maximum number of unit rows allowed for users, this should be addressed on its own (currently there's no validation in user settings form data and users with values higher than the max would need to be migrated). Could you please file a case on that?

phlax:
@julen re naming of settings - yep fine - sounds good

re get_unit_rows, there is some validation, but it currently does the wrong thing, so for sure one way or another in needs a solution.

The reason for not filing a ticket yet, is im not sure what ticket to file. My own preference would be to do away with get_unit_rows, but i need to do some clicky testing so i am clearer what effect it has.

unho:
@phlax I do have my personal opinion on `get_unit_rows` but it belongs to that issue that should be created, rather than this PR.

julen:
> As already commented earlier, since there's a change on the maximum number of unit rows allowed for users, this should be addressed on its own (currently there's no validation in user settings form data and users with values higher than the max would need to be migrated). Could you please file a case on that?

Hm, just read @unho's previous comment and noticed an inconsistent behavior: opening up the editor for the first time displays 20 units at max. From there onwards, user's setting is taken into account as-is: I have 50 set there, and it loads 24-25 before/after. Seeing this, I'd suggest to back out this change from the PR, focus on the fix for the export view (which is working and can merge quickly) and address the rest separately.

phlax:
@julen this PR doesnt affect that behaviour afaict, this is an issue on master already

phlax:
@julen also there is a bug as it uses `self.cleaned_data["user"].get_unit_rows()` and not `self.request_user.get_unit_rows()` in the form.

but again this is not related to fixing the export views afaict

julen:
Yeah, I figured that now re-reading the code. In any case there's a bug/inconsistency we should address separately. Mind filing that one? TIA.

phlax:
@julen yep i certainly will (its been on my scrum log for the last couple of days), but as said its more of a question about what the bug is. Let me do some testing and then i can give a more informed opinion, and hopefully open a better ticket/PR.

Issue #3656, opened 2015-04-01
Source directory layout reorganization


julen:
In terms of Django-specific code, Pootle's directory layout is a mix of years of work with no established pattern and legacy code. There's a **common will to reorganize** that.
## Current situation

Taking a general overview, the situation can be summarized as follows:
- Almost every model/view lives in a separate app: `pootle_language`, `pootle_project`, `pootle_translationproject`, etc.
- Some models/views live in a general `pootle_app` app: `Directory`, `Permission`.

Historically, the `pootle_*` apps were aimed to be reusable, but continuous development has shown it to be a dead end: in most of the cases the models from these apps have constant inter-dependencies, and are tightly coupled one with each other (`Store` → `Submission`, `TranslationProject` → `Directory`, `Store` → `User` etc.). Having them spread across applications has no mutual benefit.

There are legitimate cases for having models in separate apps: apps that are focused in a single specific task and are decoupled from Pootle's core functionalities. In other words, Pootle can run without these apps just fine because they provide **extra functionality** on top of the base set by the core.
## Goals

Reorganization should strive for:
1. **Removing excessive split** in applications by gathering core Pootle components in a single location: these are tightly coupled.
2. Keeping applications that provide **extra functionality explicitly separate**.
3. **Restructuring** project's layout to better reflect the previous changes and the nature of its components.
#### Should be gathered as part of core
- `pootle_app` is something similar to a Pootle core app. Actually models and views from here were taken out piece by piece (5b2541638f) in order to make them _"reusable apps"_.
- `pootle_language` has the `Language` model, and some language-related views, but not necessarily language-specific (overview, editor, export view, permissions).
- `pootle_project` has the `Project` model, and some project-related views, but not necessarily project-specific (overview, editor, export view, permissions).
- `pootle_translationproject` has the `TranslationProject` model, and some translation project-related views, but not necessarily translation project-specific (overview, editor, export view, permissions).
- `pootle_statistics` includes the `Submission` and the `ScoreLog` models that track the events and score changes for `Unit` models.
- `accounts` includes the `User` model.
- `pootle_profile` contains views that deal with the `User` model: user's public profile, user's settings, and XHR API to deal with it. This used to be home for the now gone `PootleProfile` model.
- `pootle_store`: has the `Unit`, `Store`, `Suggestion` and `QualityCheck` models. This is Pootle's real data store. It also includes views for the internal XHR API that is consumed by the overview pages and the translation editor.
[...]

PR #4661, opened 2016-04-22
Move a few bits away from pootle_misc

julen:
This PR moves some utilities away from `pootle_misc` into `pootle.core`.

Refs. #3656.


unho:
lgtm

PR #4679, opened 2016-04-26
Test caching

phlax:
This PR clears the redis cache between tests

### DIFF
--- a/pytest_pootle/fixtures/revision.py
+++ b/pytest_pootle/fixtures/revision.py
@@ -11,8 +11,12 @@ import pytest
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


pytest_pootle/fixtures/revision.py
```diff
@@ -11,8 +11,9 @@
 
 
 @pytest.fixture(autouse=True)
-def revision():
+def revision(clear_cache):
```
ta2-1: `clear_cache` is not necessary here if you use `Revision.set(Unit.max_revision())`.
And revision is saved in `redis` cache.

ta2-1:
lgtm, after travis pass.

PR #4684, opened 2016-04-27
utils: add helper to retrieve online docs URLs

julen:
Passing a path relative to the docs directory will build the absolute URL for
it, including the version string.

Example usage: `get_docs_url('server/commands.html')`


unho:
lgtm

julen:
FTR: I'll be using this right away in the [invoices PR](https://github.com/translate/pootle/pull/4648), although it's useful in many other cases.

phlax:
no tests?

PR #4685, opened 2016-04-27
Cleanup: remove unneeded `bin/env python` preambles

julen:
Note these are only relevant for standalone scripts.


phlax:
nice one, lgtm

PR #4691, opened 2016-04-28
Fix RemovedInDjango19Warning

### LINKED ISSUE #4657: RemovedInDjango19Warning with new commenting code
When firing up the server, Django 1.9-related warnings are printed a few times:
```
django_comments/models.py:170: RemovedInDjango19Warning: Model class django_comments.models.Comment doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class Comment(CommentAbstractModel):

django_comments/models.py:176: RemovedInDjango19Warning: Model class django_comments.models.CommentFlag doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class CommentFlag(models.Model):
```

julen:
django_comments needs to be added to INSTALLED_APPS.

Fixes #4657.

### DIFF
--- a/pootle/apps/pootle_comment/migrations/0002_use_abstract_module.py
+++ b/pootle/apps/pootle_comment/migrations/0002_use_abstract_module.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_comment', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='comment',
+            name='user',
+            field=models.ForeignKey(related_name='comment_comments', on_delete=django.db.models.deletion.SET_NULL, verbose_name='user', blank=True, to=settings.AUTH_USER_MODEL, null=True),
+        ),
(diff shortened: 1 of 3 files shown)


phlax:
@julen i think this will force people to install the default django Comment models - no?

julen:
That's expected (and now enforced) behavior from a django app with a models module. If that's not desired, then you will have to make django-contrib-apps move their abstract base models to a non-*models.py* module (this is what they did with [`django.contrib.auth`'s base models](https://docs.djangoproject.com/en/dev/releases/1.9/#django-contrib-auth)).

For the time being this is the only approach I know to get rid of the warning, which potentially can contribute to hiding other more important issues.

phlax:
@julen what breaks if we dont hook this up?

julen:
The point is not breaking/not breaking, but getting rid of the warning, which is doable and more than desirable as expressed by many of us. This PR is just a suggestion in that direction, but I've proposed another one too.

julen:
PR updated: had the abstract models move to a separate module in django-contrib-comments (thanks to the prompt action of Claude), and now this is importing them from there, which makes the warnings go away.

unho:
Tested and works. gtm

phlax:
nice one

PR #4693, opened 2016-04-28
Add docs page explaining how to create a plugin package

phlax:
This is the most basic documentation at the moment, but at least a start.

It would be really useful to land as then i can ref the sections here in other docs im writing.

### DIFF
--- a/docs/developers/index.rst
+++ b/docs/developers/index.rst
@@ -21,3 +21,4 @@ other way, make sure to read through this part.
    styleguide
    deprecation
    releasing
+   plugins
--- a/docs/developers/plugins.rst
+++ b/docs/developers/plugins.rst
@@ -1,0 +1,177 @@
+.. _plugins:
+
+Plugins
+=======
+
+.. warning::
+
+   Pootle's plugin system is currently in an early stage of development, and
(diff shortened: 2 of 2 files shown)


dwaynebailey:
Did a read of plugin.rst.  I think there is enough here to get the gist of what we're doing with plugins.  The code examples help but we might want to create some more concrete examples in future.  The only area that feels unclear to me is that we talk of Signals but have receivers.  While plugins and getters match in terms of naming.  So might help to clarify the Signal/Receiver relationship, but not critical.

lgtm, this will get tighter as we flesh out the framework.  I'd rather not spend hours polishing something that needs refresh when we finalise.  Put up a warning that its in flux and direct people to gitter for help would help.

phlax:
>  The only area that feels unclear to me is that we talk of Signals but have receivers. While plugins and getters match in terms of naming. So might help to clarify the Signal/Receiver relationship, but not critical.

Yep, ill update to explain a little better. I think my writing it in that way reflects the direction of the data, in that with a signal you just emit, whereas the others receive data

> Put up a warning that its in flux and direct people to gitter for help would help.

ill add now

PR #4694, opened 2016-04-28
Add useful test fixtures

phlax:
This provides 2 fixtures:

- store_po - an empty Store in /language0/project0
- test_fs - the open method of this fixture can be
  called to open files in the test data
  eg `with test_fs.open("path/to/my.po") as f`


dwaynebailey:
lgtm.  You might want to transfer some of the PR comments into the fixture comments.  From what I see we're not actually using complex.po are we?

phlax:
> we're not actually using complex.po are we?

no but the very next PR i want to land will

ta2-1:
lgtm

PR #4697, opened 2016-05-01
Pootle config

phlax:
This PR adds a generic configuration system for Pootle.

This is useful so that plugins can persist data without having to make schema changes or add additional models.

There are 3 types of config objects:
- config with content_type and object_pk - these are related to specific model instances
- config with only content_type - these are related to a particular model
- config with neither content_type, nor object_pk - these are for site configurations

keys are max 255 chars
values can be any json serializable value

example usage
---
https://github.com/translate/pootle/wiki/DocsDevConfig

### DIFF
--- a/pootle/apps/pootle_config/abstracts.py
+++ b/pootle/apps/pootle_config/abstracts.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import collections
+
+from jsonfield.fields import JSONField
+
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from .managers import ConfigManager, ConfigQuerySet
+
+
+class AbstractConfig(models.Model):
+    content_type = models.ForeignKey(
+        ContentType,
+        blank=True,
+        null=True,
+        db_index=True,
+        verbose_name=_('content type'),
+        related_name="content_type_set_for_%(class)s",
+        on_delete=models.CASCADE)
+    object_pk = models.CharField(
+        _('object ID'),
+        max_length=255,
+        db_index=True,
+        blank=True,
+        null=True)
+    content_object = GenericForeignKey(
+        ct_field="content_type",
+        fk_field="object_pk")
+    key = models.CharField(
+        _("Configuration key"),
+        max_length=255,
+        blank=False,
+        null=False,
+        db_index=True)
+    value = JSONField(
+        _("Configuration value"),
+        default="",
+        blank=True,
+        null=False,
+        load_kwargs={'object_pairs_hook': collections.OrderedDict})
+
+    objects = ConfigManager.from_queryset(ConfigQuerySet)()
+
+    class Meta(object):
+        abstract = True
+        ordering = ['pk']
+        index_together = ["content_type", "object_pk"]
+
+    def save(self, **kwargs):
+        if not self.key:
+            raise ValidationError("Config object must have a key")
+        if self.object_pk and not self.content_type:
+            raise ValidationError(
+                "Config object must have content_type when object_pk is set")
+        super(AbstractConfig, self).save(**kwargs)
--- a/pootle/apps/pootle_config/abstracts.py
+++ b/pootle/apps/pootle_config/abstracts.py
@@ -31,7 +31,6 @@ class AbstractConfig(models.Model):
     object_pk = models.CharField(
         _('object ID'),
         max_length=255,
-        db_index=True,
         blank=True,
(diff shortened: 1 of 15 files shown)


pootle/apps/pootle_config/models.py
```diff
@@ -0,0 +1,60 @@
...
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from .managers import ConfigManager, ConfigQuerySet
+
+
+class Config(models.Model):
```
ta2-1: @phlax should we add `db_index=True` to some fields?

phlax: yep

pootle/apps/pootle_config/managers.py
```diff
@@ -0,0 +1,136 @@
...
+        if not model:
+            if getattr(self, "query_model", None):
+                conf = self
+            else:
+                conf = self.site_config()
+        else:
+            conf = self.config_for(model)
+            if not isinstance(model, models.Model):
+                conf = conf.filter(object_pk__isnull=True)
+        return conf
```
unho: Nit: Use `return` directly and avoid having a variable and also avoid nesting

```python
        if model:
            if isinstance(model, models.Model):
                return self.config_for(model)

            return conf.filter(object_pk__isnull=True)

        if getattr(self, "query_model", None):
            return self

        return self.site_config()
```

pootle/apps/pootle_config/managers.py
```diff
@@ -0,0 +1,136 @@
...
+            no_set = config_should_not_be_set.get(
+                model.__class__,
+                instance=model,
+                key=key, value=value)
+        else:
+            no_set = config_should_not_be_set.get(
+                model,
+                key=key, value=value)
+        if no_set:
+            raise ConfigurationError("Unable to save: %s" % no_set)
```
unho: I don't think this error message is explanatory enough.

phlax: im actually gonna get rid of the message altogether, ie
```
ConfigurationError(e)
```
That way it will get the upstream error message properly

phlax: on that note, im thinking that it should be a convention that validating getters and providers should return Exceptions if necessary (as opposed to just returning a message or raising the error)

this is similar to what django forms does

pootle/apps/pootle_config/getters.py
```diff
@@ -0,0 +1,44 @@
...
+        if instance is not None and not isinstance(instance, sender):
+            raise ConfigurationError(
+                _("'instance' must be an instance of sender, when specified"))
+        if instance:
+            conf = Config.objects.config_for(instance)
+        else:
+            conf = Config.objects.config_for(sender).filter(object_pk__isnull=True)
+    elif instance:
+        raise ConfigurationError(
+            _("'instance' must be an instance of sender, when specified"))
```
ta2-1: sender must be defined when instance is specified

phlax: wierd, i was just looking at the same bit of code. I think this could be refactored a bit 8/

phlax: but yep, good spot, ill update the message

pootle/apps/pootle_config/managers.py
```diff
@@ -0,0 +1,149 @@
...
+    @property
+    def base_qs(self):
+        qs = self.__class__(self.model)
+        if hasattr(self, "query_model"):
+            qs.query_model = self.query_model
+        return qs
+
+    def config_for(self, model):
+        """
+        QuerySet for all comments for a particular model (either an instance or
```
ta2-1: comments?

pootle/apps/pootle_config/managers.py
```diff
@@ -0,0 +1,149 @@
...
+        self.get_config_queryset(model).filter(key=key).delete()
+
+    def set_config(self, key, value="", model=None):
+        model = self.get_query_model(model)
+        conf = self.get_config_queryset(model)
+        if self.should_set_config(key, value, model):
+            try:
+                conf = conf.get(key=key)
+            except conf.model.DoesNotExist:
+                conf = self.create(
```
ta2-1: can we create config with value here?

ta2-1: @phlax it seemed that you called `self.objects.create()` and I thought there is extra query. So I was wrong I see now that you just create an object here.
But probably if you call  `self.objects.create()` with ready value the code would be less branchy.

pootle/apps/pootle_config/abstracts.py
```diff
@@ -0,0 +1,66 @@
...
+        blank=True,
+        null=True,
+        db_index=True,
+        verbose_name=_('content type'),
+        related_name="content_type_set_for_%(class)s",
+        on_delete=models.CASCADE)
+    object_pk = models.CharField(
+        _('object ID'),
+        max_length=255,
+        db_index=True,
```
ta2-1: do we keep (obejct_pk) and (content_type, object_pk) both ?

ta2-1:
@phlax thanks for bringing good patterns. lgtm.

PR #4709, opened 2016-05-05
Pootle config fix

phlax:
this PR adjust the config api slightly so that conf.get_config("DOESNOTEXIST") returns None rather than raising error


unho:
lgtm

ta2-1:
@phlax have you thought about adding default value to `get_config()` method?
`config.get().get_config("foo", None)`

phlax:
@ta2-1 im going to add a `ConfigDict` object which i think will make this api a bit more typing friendly and easier to use.

ta2-1:
lgtm

PR #4710, opened 2016-05-06
Add Config utility classes

phlax:
Wraps pootle_config and allows Site, Model, and Object config to be represented as dictionaries, eg

```python
>>> from pootle_config.utils import SiteConfig, ModelConfig, ObjectConfig
>>> site_conf = SiteConfig()
>>> site_conf["foo"] = "bar"
>>> site_conf["foo"]
"bar"
>>> SiteConfig()["foo"]
"bar"
>>> from pootle_project.models import Project
>>> model_conf = ModelConfig(Project)
>>> model_conf["foo"] = "bar2"
>>> ModelConfig(Project)["foo"]
"bar2"
>>> from pootle_project.models import Project
>>> object_conf = ObjectConfig(Project.objects.first())
>>> object_conf["foo"] = "bar3"
>>> ObjectConfig(Project.objects.first())["foo"]
"bar3"
```

### DIFF
--- a/pootle/apps/pootle_config/managers.py
+++ b/pootle/apps/pootle_config/managers.py
@@ -93,7 +93,7 @@ class ConfigQuerySet(models.QuerySet):
                 conf = conf.filter(key__in=key)
         elif key is not None:
             conf = conf.filter(key=key)
-        for item in conf:
+        for item in conf.order_by("key", "pk"):
             # dont use values_list to trigger to_python
             conf_list.append((item.key, item.value))
         return conf_list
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_config/managers.py
```diff
@@ -93,7 +93,7 @@ def list_config(self, key=None, model=None):
                 conf = conf.filter(key__in=key)
         elif key is not None:
             conf = conf.filter(key=key)
-        for item in conf:
+        for item in conf.order_by("key", "pk"):
```
ta2-1: why is this necessary?

phlax: because there may be multiple k/v pairs with same k, this ensures it gets the last - which is at least predictable

phlax: but i certainly thought about making this configurable - because most times you dont care

ta2-1:
lgtm

PR #4711, opened 2016-05-06
Lang mapper

phlax:
Adds a language mapper with presets for Projects

At the moment its not hooked up (other than in pootle_fs), but im thinking that Store.update/sync should also respect any configured lang mappings.

initial docs here https://github.com/translate/pootle/wiki/DocsDevLangMapper

### DIFF
--- a/pootle/apps/pootle_project/lang_mapper.py
+++ b/pootle/apps/pootle_project/lang_mapper.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+from collections import OrderedDict
+
+from django.utils.functional import cached_property
+from django.utils.lru_cache import lru_cache
+
+from pootle_config.utils import SiteConfig
+from pootle_language.models import Language
+
+
+logger = logging.getLogger(__name__)
+
+
+class ProjectLanguageMapper(object):
+    """For a given code find the relevant Pootle lang taking account of any
+    lang mapping configuration.
+
+    Presets can be defined with a dictionary of named mappings on the site_config
+    `pootle.core.lang_mapping_presets`.
+
+    Presets are enabled with a list on the project_config
+    `pootle.core.use_lang_mapping_presets`.
+
+    Lang mappings can also be configured per-project using the config setting
+    `pootle.core.lang_mapping`, which should be a k, v dictionary of
+    upstream_code, pootle_code. This will override any mappings in configured
+    presets.
+    """
+
+    def __init__(self, project):
+        self.project = project
+
+    def __getitem__(self, k):
+        return self.get_lang(k)
+
+    def __contains__(self, k):
+        return self.get_lang(k) and True or False
+
+    @cached_property
+    def site_config(self):
+        return SiteConfig()
+
+    @property
+    def site_presets(self):
+        """Get the pret mappings from the site config"""
+        return OrderedDict(
+            self.site_config.get(
+                "pootle.core.lang_mapping_presets", {}))
+
+    @property
+    def project_presets(self):
+        """Get the projects mapping from the Project.config"""
+        return self.project.config.get(
+            "pootle.core.use_lang_mapping_presets", [])
+
+    @property
+    def project_mappings(self):
+        """Get the projects mapping from the Project.config"""
+        return OrderedDict(
+            self.project.config.get(
+                "pootle.core.lang_mapping", {}))
+
+    @property
+    def mappings_from_presets(self):
+        mappings = OrderedDict()
+        for preset_name in self.project_presets:
+            if preset_name not in self.site_presets:
+                logger.warning(
+                    "Unrecognised lang mapping preset: %s"
+                    % preset_name)
+                continue
+            mappings.update(self.site_presets[preset_name])
+        return mappings
+
+    @cached_property
+    def lang_mappings(self):
+        """Language mappings after Project.config and presets are parsed"""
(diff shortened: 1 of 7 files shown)


pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,96 @@
...
+    def _parse_mappings(self):
+        self._mapping = OrderedDict()
+
+        for k, v in self.project_mapping.items():
+            if k.startswith("$"):
+                preset = self.site_presets.get(v, None)
+                if not preset:
+                    logger.warning(
+                        "Unrecognised lang mapping preset: %s" % preset)
+                else:
```
unho: Nit: We can avoid the `else` by using `continue`.

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,96 @@
...
+            return Language.objects.get(
+                code=self.get_pootle_code(lang_code))
+        except Language.DoesNotExist:
+            return None
+
+    def get_pootle_code(self, lang_code):
+        return self.lang_mappings.get(lang_code, lang_code)
+
+    def _add_lang_to_mapping(self, k, v):
+        # as its a 1-2-1 mapping remove any previous items with
```
unho: "1 to 1"?

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @lru_cache()
+    def get_lang(self, lang_code):
+        try:
+            return Language.objects.get(
+                code=self.get_pootle_code(lang_code))
+        except Language.DoesNotExist:
+            return None
+
+    def get_pootle_code(self, lang_code):
```
unho: Nit: I would put `get_pootle_code` and `get_upstream_code` one after the other.

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,96 @@
...
+            for _k, _v in self._mapping.items():
+                if v == _v:
+                    del self._mapping[_k]
+                    break
+        self._mapping[k] = v
+
+    def _parse_mappings(self):
+        self._mapping = OrderedDict()
+
+        for k, v in self.project_mapping.items():
```
unho: `k` and `v` do not convey any information. But I see that above `fs_code` and `pootle_code` are used, and those are easier to understand.

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,96 @@
...
+from pootle_language.models import Language
+
+
+logger = logging.getLogger(__name__)
+
+
+class LanguageMapper(object):
+    """For a given code find the relevant pootle lang taking account of any
+    lang mapping configuration
+
```
unho: Unneeded blank line.

unho:
I guess that the purpose of this code is to match Pootle language code (for example `ne`) to the code used in the filesystem (for example `ne_NP`), isn't it? I am asking because I don't feel this is explained at all. It would be nice to have at least some docstrings explaining it.

phlax:
@unho you are right on both accounts.

As this is configured with `pootle config` and currently is not hooked into anything (other than pootle_fs) i would like to land as is, with some docs and a PR to hook in properly to follow shortly after.

adding some docstrings now...

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,110 @@
...
+                preset = self.site_presets.get(pootle_code, None)
+                if not preset:
+                    logger.warning(
+                        "Unrecognised lang mapping preset: %s" % preset)
+                    continue
+                for preset_upstream_code, preset_pootle_code in preset:
+                    self._add_lang_to_mapping(
+                        preset_upstream_code,
+                        preset_pootle_code)
+            else:
```
unho: This `else` can also be avoided by using `continue`.

unho:
I checked docstrings and I still don't see any real example. I think a real example might help understanding the purpose.

pootle/apps/pootle_project/lang_mapper.py
phlax: im not so sure that is more readable tbh

pootle/apps/pootle_project/lang_mapper.py
```diff
@@ -0,0 +1,110 @@
...
+
+from pootle.core.delegate import config
+from pootle_language.models import Language
+
+
+logger = logging.getLogger(__name__)
+
+
+class LanguageMapper(object):
+    """For a given code find the relevant pootle lang taking account of any
```
unho: Pootle language

pootle/apps/pootle_project/lang_mapper.py
unho: k

unho:
lgtm

tests/misc/lang_mapper.py
```diff
@@ -0,0 +1,201 @@
...
+     ["bar",
+      [["testbar1", "en"],
+       ["testbar2", "language1"],
+       ["testbar3", "language0"],
+       ["testbar4", "bar"]]],
+     ["baz",
+      [["testbaz1", "en"],
+       ["testbaz2", "language1"],
+       ["testbaz3", "language0"],
+       ["testbaz4", "baz"]]]])
```
dwaynebailey: foo, bar, baz makes it frustratingly hard to understand what exactly is being tested.  Real codes for real languages would have helped way way more.

phlax: ill rename test[foo|bar|baz] to "upstream_code_n"

phlax: and for the preset names `preset_n`

ta2-1:
@phlax great work. thanks. Looks very good after you added `pootle.core.use_lang_mapping_presets`. GTM.

PR #4713, opened 2016-05-06
fix update_stores not changing fuzzy state

cloph:
regression from 80d35df2b2c3987470873c541c288dc95c0ec79e
ignoring change in state makes it impossible to update the fuzzy state
of a translatin unit by modifying the files on disk, and running
update_stores. With state ignored in the comparison, those changes are
silently ignored (and thenget overridden again on next sync_stores)


phlax:
@cloph thanks for this, its really helpful.

Ill take a look at why the test is failing and merge

cloph:
thx for finishing up properly :)

PR #4714, opened 2016-05-07
Pull fuzzy update stores

phlax:
If a units state has changed, but there are no other changes then the Store is not updated

This PR fixes that.

I think we need to ensure that *all* unit changes are properly registered regardless of source/target changes, and probs add some more tests around this. (ref #4715)

thanks to @cloph for spotting this and submitting initial fix #4713.

### DIFF
--- a/pootle/apps/pootle_store/diff.py
+++ b/pootle/apps/pootle_store/diff.py
@@ -20,7 +20,7 @@ class UnitDiffProxy(UnitProxy):
     """Wraps File/DB Unit dicts used by StoreDiff for equality comparison"""
 
     match_attrs = ["context", "developer_comment", "locations",
-                   "source", "target", "translator_comment"]
+                   "source", "state", "target", "translator_comment"]
 
     def __eq__(self, other):
         return all(getattr(self, k) == getattr(other, k)
--- a/pootle/apps/pootle_store/diff.py
+++ b/pootle/apps/pootle_store/diff.py
@@ -13,7 +13,7 @@ from django.utils.functional import cached_property
 
 from .fields import to_python as multistring_to_python
 from .unit import UnitProxy
-from .util import OBSOLETE
+from .util import FUZZY, OBSOLETE, TRANSLATED, UNTRANSLATED
 
 
 class UnitDiffProxy(UnitProxy):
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_store/util.py
```diff
@@ -36,6 +38,12 @@
     FUZZY: "fuzzy",
     TRANSLATED: "translated"}
 
+TTK_STATES = {
+    pounit.S_FUZZY: FUZZY,
+    pounit.S_OBSOLETE: OBSOLETE,
+    pounit.S_TRANSLATED: TRANSLATED,
+    pounit.S_UNTRANSLATED: UNTRANSLATED}
```
unho: This is PO centric. How this will affect projects in another formats, for example XLIFF?

phlax: hmm, fair point ill check ttk, and see if its different

pootle/apps/pootle_store/diff.py
```diff
@@ -32,7 +32,9 @@ def __ne__(self, other):
 
 class DBUnit(UnitDiffProxy):
 
-    pass
+    @property
+    def state(self):
+        return self.unit["state"] or 0
```
ta2-1: 0 or UNTRANSLATED?

phlax: ill update

pootle/apps/pootle_store/diff.py
```diff
@@ -80,13 +79,20 @@ def file_units(self):
...
+                state = TRANSLATED
+            elif unit.isfuzzy():
+                state = FUZZY
             file_units[unit.getid()] = {
                 "unitid": unit.getid(),
                 "context": unit.getcontext(),
                 "locations": unit.getlocations(),
                 "source": unit.source,
                 "target": unit.target,
-                "state": unit.get_state_n(),
```
unho: Why is not this method changed instead to return the right results?

Also some explanation on commit message would help.

phlax: becuase it returns different things for different formats. Changing this would be a significant change for ttk. Ill update the commit message

unho: Ok. Thanks.

ta2-1:
gtm!

PR #4716, opened 2016-05-07
Add documentation for pootle_fs

phlax:
Separated documentation from the code PR for pootle_fs.

Ill copy these pages to the wiki so we can edit and copy back when we are happy


docs/developers/components/finder.rst
```diff
@@ -0,0 +1,328 @@
+#doctest: +ELLIPSIS
+.. _translation_file_finder:
+
+
+Translation File Finder
+=======================
+
+This tool provides a way of mapping translation files in a filesystem to pootle codes for
```
ta2-1: pootle codes? did you mean paths?

phlax: i did mean codes, but not sure what is clearest - ie it doesnt return a `pootle_path` more like the contents of ``split_pootle_path``

PR #4720, opened 2016-05-09
Pootle fs finder

phlax:
Add a TranslationFileFinder for pootle_fs

https://github.com/translate/pootle/wiki/DocsTranslationFileFinder

more recent docs here

https://github.com/translate/pootle/pull/4716

### DIFF
--- a/pootle/apps/pootle_fs/finder.py
+++ b/pootle/apps/pootle_fs/finder.py
@@ -1,0 +1,187 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import fnmatch
+import os
+import re
+
+import scandir
+
+from django.utils.functional import cached_property
+from django.utils.lru_cache import lru_cache
+
+
+PATH_MAPPING = (
+    (".", "\."),
+    ("<language_code>", "(?P<language_code>[\w\-\.]*)"),
+    ("<filename>", "(?P<filename>[\w\-\.]*)"),
+    ("/<dir_path>/", "/<dir_path>"),
+    ("<dir_path>", "(?P<dir_path>[\w\/\-]*?)"))
+
+DEFAULT_EXTENSIONS = ("po", "pot")
+
+
+class TranslationFileFinder(object):
+
+    extensions = DEFAULT_EXTENSIONS
+    path_mapping = PATH_MAPPING
+
+    def __init__(self, translation_path, path_filters=None, extensions=None):
+        TranslationPathValidator(translation_path).validate()
+        self.translation_path = translation_path
+        if extensions:
+            self.extensions = extensions
+        self.path_filters = path_filters
+
+    @cached_property
+    def regex(self):
+        return re.compile(self._parse_path())
+
+    @cached_property
+    def file_root(self):
+        """The deepest directory that can be used from the translation path
+        before the <tags>.
+        """
+        file_root = self.translation_path.split("<")[0]
+        if not file_root.endswith("/"):
+            file_root = "/".join(file_root.split("/")[:-1])
+        return file_root.rstrip("/")
+
+    def match(self, file_path):
+        """For a given file_path find translation_path matches.
+        If a match is found `file_path`, `matchdata` is returned.
+        """
+        match = self.regex.match(file_path)
+        if not match:
+            return
+        matched = match.groupdict()
+        matched["dir_path"] = (
+            matched.get("dir_path", "").strip("/"))
+        if not matched.get("filename"):
+            matched["filename"] = os.path.splitext(
+                os.path.basename(file_path))[0]
+        return file_path, matched
+
+    def walk(self):
+        """Walk a filesystem"""
+        for root, dirs, files in scandir.walk(self.file_root):
+            for filename in files:
+                yield os.path.join(root, filename)
+
+    def find(self):
+        """Find matching files anywhere in file_root"""
+        for filepath in self.walk():
+            match = self.match(filepath)
+            if match:
+                yield match
+
+    @lru_cache(maxsize=None)
+    def reverse_match(self, language_code, filename=None,
+                      extension=None, dir_path=None):
+        """For given matchdata return the file path that would be
+        matched.
+        """
+        if extension is None:
+            extension = self.extensions[0]
+        if extension not in self.extensions:
+            raise ValueError("ext must be in the list of possible extensions")
+        if not filename:
+            filename = language_code
+        extension = extension.strip(".")
+        path = (
+            "%s.%s"
+            % (os.path.splitext(self.translation_path)[0], extension))
+        if dir_path and "<dir_path>" not in path:
+            return
+        path = (path.replace("<language_code>", language_code)
+                    .replace("<filename>", filename))
+        if "<dir_path>" in path:
+            if dir_path and dir_path.strip("/"):
+                path = path.replace(
+                    "<dir_path>", "/%s/" % dir_path.strip("/"))
+            else:
+                path = path.replace("<dir_path>", "")
+        local_path = path.replace(self.file_root, "")
+        if "//" in local_path:
+            path = os.path.join(
+                self.file_root,
+                local_path.replace("//", "/").lstrip("/"))
+        return path
+
+    def _ext_re(self):
+        return (
+            r".(?P<ext>(%s))"
+            % "|".join(
+                ("%s$" % x)
+                for x in set(self.extensions)))
+
+    def _parse_path_regex(self):
+        path = self.translation_path
+        for k, v in self.path_mapping:
+            path = path.replace(k, v)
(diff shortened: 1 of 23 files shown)


pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,180 @@
...
+    ("<filename>", "(?P<filename>[\w\-\.]*)"),
+    ("/<dir_path>/", "/<dir_path>"),
+    ("<dir_path>", "(?P<dir_path>[\w\/\-]*?)"))
+
+
+class TranslationFileFinder(object):
+
+    path_mapping = PATH_MAPPING
+
+    def __init__(self, translation_path, path_filters=None, ext=["po", "pot"]):
```
phlax: ext should be None or immutable type

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+DEFAULT_EXTS = ("po", "pot")
+
+
+class TranslationFileFinder(object):
+
+    path_mapping = PATH_MAPPING
+
+    def __init__(self, translation_path, path_filters=None, ext=None):
+        self.translation_path = translation_path
+        self.ext = ext or DEFAULT_EXTS
```
unho: ext? Better use extensions.

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        if "<" in self.stripped_path or ">" in self.stripped_path:
+            raise ValueError(
+                "Only <language_code>, <dir_path>, <filename> and <ext> are valid "
+                "patterns to match in the translation path")
+
+    def validate_path(self):
+        if re.search("[^\w\/\-\.]+", self.stripped_path):
+            raise ValueError(
+                "Translation paths can only contain alpha-numeric characters, "
+                "_ or -: '%s'" % self.path)
```
unho: What about `/`?

phlax: i think i might just change to `Invalid translation_path  character '%s'`

pootle/apps/pootle_fs/finder.py
phlax: yep

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __init__(self, path):
+        self.path = path
+
+    def validate_no_dots(self):
+        links_to_parent = (
+            self.path == ".."
+            or self.path.startswith("../")
+            or "/../" in self.path
+            or self.path.endswith("/.."))
```
unho: Is it not enough to check `".." in self.path`?

phlax: `foo..bar` is a valid filepath. I suppose (and this relates to valid chars issue too) it deps on how much we want to support of any potential filesystems

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+
+class TranslationFileFinder(object):
+
+    path_mapping = PATH_MAPPING
+
+    def __init__(self, translation_path, path_filters=None, ext=None):
+        self.translation_path = translation_path
+        self.ext = ext or DEFAULT_EXTS
+        self.path_filters = path_filters
+        TranslationPathValidator(translation_path).validate()
```
unho: Perhaps it is better to put this at the beginning of the block to fail early.

phlax: yep

pootle/apps/pootle_fs/finder.py
phlax: hmm @unho as it requires absolute paths, im wondering if a better check is

```self.path == os.path.abspath(self.path)```

File: pootle/apps/pootle_fs/finder.py [hunk 6]
```diff
@@ -0,0 +1,182 @@
...
+                self.file_root,
+                local_path.replace("//", "/").lstrip("/"))
+        return path
+
+    def _ext_re(self):
+        return (
+            r".(?P<ext>(%s))"
+            % "|".join(
+                ("%s$" % x)
+                for x in set(self.ext)))
```
unho: Is it not necessary to use list comprehension?

phlax: as opposed to a generator expression?

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        self.ext = ext or DEFAULT_EXTS
+        self.path_filters = path_filters
+        TranslationPathValidator(translation_path).validate()
+
+    @cached_property
+    def regex(self):
+        return re.compile(self._parse_path())
+
+    @cached_property
+    def file_root(self):
```
unho: Is `file_root` the directory in the path?

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        file_root = self.translation_path.split("<")[0]
+        if not file_root.endswith("/"):
+            file_root = "/".join(file_root.split("/")[:-1])
+        return file_root.rstrip("/")
+
+    def match_filepath(self, file_path):
```
unho: As always it would be nice to have some docstrings that provide a brief description of what it does. Not nice to always have to read the code to understand what it does.

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        filename = os.path.basename(file_path)
+        match = self.match(file_path)
+        if not match:
+            return
```
unho: `filename` two lines above is not necessary in case we return here

pootle/apps/pootle_fs/finder.py
phlax: its the deepest directory path it can get without using any of the tags

pootle/apps/pootle_fs/finder.py
phlax: fair point, ill add some - also in relation to this method - i think i will merge with `match` - doesnt seem much point having both

pootle/apps/pootle_fs/finder.py
unho: If you can find out which are the offending characters it looks like a better approach.

pootle/apps/pootle_fs/finder.py
unho:
```python
In [6]: os.path.abspath("..")
Out[6]: '/home'
```

?¿?

pootle/apps/pootle_fs/finder.py
unho: Please add explanatory docstring.

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        matched = match.groupdict()
+        matched["dir_path"] = (
+            matched.get("dir_path", "").strip("/"))
+        if not matched.get("filename"):
+            matched["filename"] = os.path.splitext(
+                os.path.basename(filename))[0]
+        return file_path, matched
+
+    def walk(self):
+        # print("Walking the FS: %s" % self.file_root)
```
unho: I guess you would drop commented code before merging.

pootle/apps/pootle_fs/finder.py
phlax:
```
>>> ".." == os.path.abspath("..")
False
```

pootle/apps/pootle_fs/finder.py
```diff
@@ -0,0 +1,182 @@
...
+        return self.regex.match(file_path)
+
+    @lru_cache(maxsize=None)
+    def reverse_match(self, language_code, filename=None, ext=None, dir_path=None):
+        if not filename:
+            filename = language_code
+        if ext is None:
+            ext = self.ext[0]
+        if ext not in self.ext:
+            raise ValueError("ext must be in the list of possible extensions")
```
unho: Please move bail out early code as close as possible to beginning of block.

pootle/apps/pootle_fs/finder.py
unho: Current code is fine. I am just not used to generator expressions. My apologies.

pytest_pootle/fixtures/pootle_fs/finder.py
```diff
@@ -0,0 +1,199 @@
...
+    "foo/@<language_code>/bar.<ext>"]
+
+ROOT_PATHS = OrderedDict()
+ROOT_PATHS["<language_code>.<ext>"] = ""
+ROOT_PATHS["foo/<language_code>.<ext>"] = "foo"
+ROOT_PATHS["foo/bar/baz-<filename>-<language_code>.<ext>"] = "foo/bar"
+
+MATCHES = OrderedDict()
+MATCHES["po/<language_code>.<ext>"] = (
+    ["en.po", "foo/bar/en.po"],
```
ta2-1: I didn't expect that this is not_matching part.

ta2-1:
@phlax thanks. it's done very well. lgtm.

PR #4721, opened 2016-05-10
More RTL language candidates

dwaynebailey:
Found anther source of RTL languages.  It doesn't have references that
make me more confident that they're right, so document their source for
future debuggers.

### DIFF
--- a/pootle/i18n/gettext.py
+++ b/pootle/i18n/gettext.py
@@ -61,8 +61,16 @@ def language_dir(language_code):
     RTL_LANGS = [
         # Helpful list of RTL codes:
         # https://en.wikipedia.org/wiki/Right-to-left#RTL_Wikipedia_languages
-        "ar", "arc", "bcc", "bqi", "ckb", "dv", "fa", "glk", "he", "ks", "lrc",
-        "mzn", "pnb", "ps", "sd", "ug", "ur", "yi", "nqo",
+        'ar', 'arc', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'glk', 'he', 'ks', 'lrc',
+        'mzn', 'pnb', 'ps', 'sd', 'ug', 'ur', 'yi', 'nqo',
+        # and
+        # https://github.com/i18next/i18next/blob/ee3afd8e5d958e8d703a208194e59fa5228165fd/src/i18next.js#L257-L262
+        'shu', 'sqr', 'ssh', 'xaa', 'yhd', 'yud', 'aao', 'abh', 'abv', 'acm',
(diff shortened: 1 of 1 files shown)


pootle/i18n/gettext.py
```diff
@@ -61,8 +61,16 @@ def language_dir(language_code):
     RTL_LANGS = [
         # Helpful list of RTL codes:
         # https://en.wikipedia.org/wiki/Right-to-left#RTL_Wikipedia_languages
-        "ar", "arc", "bcc", "bqi", "ckb", "dv", "fa", "glk", "he", "ks", "lrc",
-        "mzn", "pnb", "ps", "sd", "ug", "ur", "yi", "nqo",
+        'ar', 'arc', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'glk', 'he', 'ks', 'lrc',
+        'mzn', 'pnb', 'ps', 'sd', 'ug', 'ur', 'yi', 'nqo',
```
unho: Since you are changing this I suggest putting one per line.

pootle/i18n/gettext.py
```diff
@@ -61,8 +61,16 @@ def language_dir(language_code):
...
+        # and
+        # https://github.com/i18next/i18next/blob/ee3afd8e5d958e8d703a208194e59fa5228165fd/src/i18next.js#L257-L262
+        'shu', 'sqr', 'ssh', 'xaa', 'yhd', 'yud', 'aao', 'abh', 'abv', 'acm',
+        'acq', 'acw', 'acx', 'acy', 'adf', 'ads', 'aeb', 'aec', 'afb', 'ajp',
+        'apc', 'apd', 'arb', 'arq', 'ars', 'ary', 'arz', 'auz', 'avl', 'ayh',
+        'ayl', 'ayn', 'ayp', 'bbz', 'pga', 'iw', 'pbt', 'pbu', 'pst', 'prp',
+        'prd', 'ydd', 'yds', 'yih', 'ji', 'hbo', 'men', 'xmn', 'jpr', 'peo',
+        'pes', 'prs', 'sam',
```
unho: It would be nicer to have the list in alphabetical order to make it easier to check if the language is missing.

unho:
I will trust that all of those codes correspond to RTL languages. I only checked a few of them, all variants of Arabic.

pootle/i18n/gettext.py
dwaynebailey: Way too painful to list one per line, the file isn't meant to be taken up by a list of RTL languages.  And way to painful to sort, that's what search is for in your favourite editor.

unho:
lgtm

phlax:
lgtm

ta2-1:
lgtm

PR #4728, opened 2016-05-12
Fix: serializers gather correctly and pass store

phlax:
This PR contains a cleanup and a couple of fixes

- Config.set_config does not return a Config object
- Config.none().get_config was not working correctly on ModelConfig
- Fix serializers should be called with Store, and gather on Project

and adjusts tests accordingly

### DIFF
--- a/tests/models/config.py
+++ b/tests/models/config.py
@@ -36,9 +36,11 @@ def _test_config_bad_create(create_func):
             key="foo4", value="bar", object_pk="23").save()
 
 
-def _test_config_ob(conf, **kwargs):
-    conf.save()
-    conf = Config.objects.get(id=conf.id)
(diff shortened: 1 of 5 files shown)


unho:
lgtm

tests/models/config.py
```diff
@@ -37,8 +37,10 @@ def _test_config_bad_create(create_func):
 
 
 def _test_config_ob(conf, **kwargs):
-    conf.save()
```
ta2-1: it overrides `conf` fuction parameter.

ta2-1:
lgtm after comment is addressed.

tests/models/config.py
phlax: cool, ill remove

PR #4733, opened 2016-05-13
Cleanup: rename views' `is_admin` to `has_admin_access`

julen:
Renames views' `is_admin` property to a more natural `has_admin_access` (also in the template context).


unho:
lgtm

PR #4735, opened 2016-05-13
Browser: hide disabled items and allow showing them to admins

### LINKED ISSUE #4609: Add an ability to show/hide disabled projects
Now  that we only disable projects, not delete them, the list of disabled projects can grow and distract from the active projects. Let's add a toggle that would allow to show /hide disabled projects (and have them hidden by default). The toggle would say either "Hide disabled" or "Show disabled" and would only appear for admins and only when there's at least one disabled project (or translation project):

screenclip 4

By default on every page load the disabled projects would be hidden, and the link would say "Show disabled".

Implementation-wise, this can be done in CSS by toggling a global class on a table that would hide or show the disabled rows. This, however, will require the change in how we highlight odd/even rows. See e.g. http://stackoverflow.com/questions/6629576/make-css-nth-child-only-affect-visible

julen:
When disabled elements add up, these produce too much noise for administrators,
therefore now they are hidden by default. At the same time, a new "Show
disabled" toggle has been added so administrators can have these items displayed
at their own will.

Implementation-wise, since CSS's `nth-child` takes any elements in the DOM
into account regardless of their visibility/display properties, some ugly hacks
are needed, and `nth-child` has been replaced by explicit CSS classes. These are
controlled ad-hoc at the moment; it's ugly but it's a short-term compromise to
have this working without having to rewrite the entire table.

Looking forward, the table would become a component purely powered by data, so
we could decide what to render at any point in time, and rows that shouldn't be
displayed would be away from the DOM, hence allowing us to put back the
`nth-child` rule and removing any ugly code added here.

Fixes #4609.

### DIFF
--- a/pootle/static/js/browser/components/VisibilityToggle.js
+++ b/pootle/static/js/browser/components/VisibilityToggle.js
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) Pootle contributors.
+ *
+ * This file is a part of the Pootle project. It is distributed under the GPL3
+ * or later license. See the LICENSE file for a copy of the license and the
+ * AUTHORS file for copyright and authorship information.
+ */
+
+import React, { PropTypes } from 'react';
+
+import TextToggle from 'components/TextToggle';
+
+
+const VisibilityToggle = React.createClass({
+
+  propTypes: {
+    // FIXME: read from context when it becomes available
+    uiLocaleDir: PropTypes.string.isRequired,
+  },
+
+  componentDidMount() {
+    this.handleVisibility({ isActive: false });
+  },
+
+  /**
+   * Note this is coupled with the DOM and how the browsing table is rendered.
+   * The implementation must change once the table is rendered as a component.
+   */
+  handleVisibility({ isActive: showDisabled }) {
+    const rows = [...document.querySelectorAll('tr.item')];
+    rows.map(row => row.classList.remove('odd'));
+    rows
+      .filter(row => row.classList.contains('is-disabled'))
+      .forEach(row => row.classList.toggle('is-visible', showDisabled));
+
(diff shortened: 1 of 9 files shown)


phlax:
lgtm

pootle/static/js/browser/components/VisibilityToggle.js
```diff
@@ -0,0 +1,63 @@
...
+
+    [...document.querySelectorAll('tr.is-visible')].forEach((row, i) => {
+      row.classList.toggle('odd', i % 2 === 0);
+    });
+  },
+
+  render() {
+    const style = {
+      position: 'absolute',
+      [this.props.uiLocaleDir === 'ltr' ? 'right' : 'left']: '2em',
```
ta2-1: @julen As we are getting rid of CSS in favour of components. Is it temporary solution?

unho: From what I see at PR description, yes.

unho:
lgtm

pootle/static/js/browser/components/VisibilityToggle.js
unho: Well, actually part of the CSS will be brought back.

julen: CSS is not being removed by any means — it's what browsers need! Component styling should ideally be encapsulated, and CSS doesn't allow this by its global nature. This one is a very simple component that's why I decided to use some inline styles, although I would be happy to experiment with CSS modules, which are local by default.

julen:
I'd appreciate some basic testing, at least to ensure this is doing the right thing not just for me 😎.

ta2-1:
Tested. GTM.

Issue #4737, opened 2016-05-18
Django RQ 0.9.1 RQ_SHOW_ADMIN_LINK option breaks new Pootle installs


theflow:
I'm trying to follow the installation docs and I can't get it to work. I tried it both on my Mac and on a Ubuntu 12.04 machine. It always fails like this:

```pytb
(env) pootle@pootle0:~/pootle$ pootle --version
Traceback (most recent call last):
  File "/podio/pootle/pootle/env/bin/pootle", line 7, in <module>
    from pootle.runner import main
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/pootle/runner.py", line 19, in <module>
    from .core.utils.redis_rq import rq_workers_are_running
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/pootle/core/utils/redis_rq.py", line 12, in <module>
    from django_rq.queues import get_queue
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/django_rq/__init__.py", line 5, in <module>
    from .workers import get_worker
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/django_rq/workers.py", line 5, in <module>
    from .settings import EXCEPTION_HANDLERS
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/django_rq/settings.py", line 8, in <module>
    SHOW_ADMIN_LINK = getattr(settings, 'RQ_SHOW_ADMIN_LINK', False)
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/django/conf/__init__.py", line 46, in __getattr__
    self._setup(name)
  File "/podio/pootle/pootle/env/local/lib/python2.7/site-packages/django/conf/__init__.py", line 40, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
django.core.exceptions.ImproperlyConfigured: Requested setting RQ_SHOW_ADMIN_LINK, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

If I set the settings, I get another error:

```pytb
(env) pootle@pootle0:~/pootle$ DJANGO_SETTINGS_MODULE="pootle.settings" pootle --version
Traceback (most recent call last):
  File "/podio/pootle/pootle/env/bin/pootle", line 7, in <module>
    from pootle.runner import main
[...]

PR #4738, opened 2016-05-18
A fix and a cleanup

phlax:
Allows compatibility with django-rq > 0.9

- move django rq import in pootle.runner - Fixes: #4737
- make factories tz aware


julen:
Looks good 👍 couldn't test though.

PR #4741, opened 2016-05-19
Pootle fs state and response

phlax:
Adds state and response objects to be used by pootle_fs, and tests

### DIFF
--- a/pootle/core/response.py
+++ b/pootle/core/response.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class ItemResponse(object):
+
+    def __init__(self, response, action_type, complete=True, msg=None, **kwargs):
+        self.response = response
+        self.action_type = action_type
+        self.complete = complete
+        self.msg = msg or ""
+        self.kwargs = kwargs
+
+    def __str__(self):
+        extra = ""
+        if self.failed:
+            extra = " FAILED"
+        if self.msg:
+            extra = "%s %s" % (extra, self.msg)
+        if self.kwargs:
+            extra = "%s %s" % (extra, str(self.kwargs))
+        return (
+            "<%s(%s): %s%s>"
+            % (self.__class__.__name__,
+               self.response,
+               self.action_type,
+               extra))
+
+    @property
+    def failed(self):
+        return not self.complete
+
+
+class Response(object):
+
+    __responses__ = None
+    response_class = ItemResponse
+
+    def __init__(self, context):
+        self.context = context
+        self.__responses__ = {}
+        self.clear_cache()
+
+    def __getitem__(self, k):
+        return self.__responses__[k]
+
+    def __iter__(self):
+        for k in self.__responses__:
+            if self.__responses__[k]:
+                yield k
+
+    def __len__(self):
+        return len([x for x in self.__iter__()])
+
+    def __str__(self):
+        failed = ""
+        if self.has_failed:
+            failed = "FAIL "
+        if self.made_changes:
+            return (
+                "<%s(%s): %s%s>"
+                % (self.__class__.__name__,
+                   self.context,
+                   failed,
+                   self.success))
+        return (
+            "<%s(%s): %sNo changes made>"
+            % (self.__class__.__name__,
+               self.context,
+               failed))
+
+    @property
+    def has_failed(self):
+        return len(list(self.failed())) > 0
+
+    @property
+    def made_changes(self):
+        return len(list(self.completed())) > 0
+
+    @property
+    def response_types(self):
+        return self.__responses__.keys()
+
+    @property
+    def success(self):
+        return ', '.join(
+            ["%s: %s" % (k, len(list(self.completed(k))))
+             for k in self.response_types
+             if len(list(self.completed(k)))])
+
+    def add(self, response_type, complete=True, msg=None, **kwargs):
+        response = self.response_class(
+            self, response_type, complete=complete, msg=msg, **kwargs)
+        if response_type not in self.__responses__:
+            self.__responses__[response_type] = []
+        self.__responses__[response_type].append(response)
+        return response
+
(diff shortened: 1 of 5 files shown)


pootle/apps/pootle_fs/response.py
```diff
@@ -0,0 +1,197 @@
...
+    "description":
+        ("Merged - Pootle won where unit updates conflicted")}
+
+
+class ActionStatus(Status):
+
+    def __init__(self, action_type, original_status, complete=True, msg=None):
+        self.action_type = action_type
+        self.complete = complete
+        self.original_status = original_status
```
unho: Nit, assign attributes in the same order as parameters.

pootle/apps/pootle_fs/response.py
```diff
@@ -0,0 +1,197 @@
...
+        ("Merged - FS won where unit updates conflicted")}
+FS_ACTION["merged_from_pootle"] = {
+    "title": "Merged from pootle",
+    "description":
+        ("Merged - Pootle won where unit updates conflicted")}
+
+
+class ActionStatus(Status):
+
+    def __init__(self, action_type, original_status, complete=True, msg=None):
```
unho: Which is the point on accepting `msg`? Is `ActionStatus` meant to be subclassed?

pootle/core/state.py
```diff
@@ -0,0 +1,111 @@
...
+    @property
+    def has_changed(self):
+        return any(self.__state__.values())
+
+    def add(self, k, v):
+        if k in self.__state__:
+            self.__state__[k].append(v)
+
+    @property
+    def states(self):
```
unho: Please put the properties together, following convention for order in models (despite this is not a model) http://docs.translatehouse.org/projects/pootle/en/latest/developers/styleguide.html#order-in-models

tests/misc/state.py
```diff
@@ -0,0 +1,214 @@
...
+        def state_foo(self, **kwargs):
+            for x in [1, 2, 3]:
+                yield {str(x): x}
+
+        def state_bar(self, **kwargs):
+            for x in [4, 5, 6, 7, 8, 9]:
+                yield {str(x): x}
+
+        def state_baz(self, **kwargs):
+            yield dict(never="called")
```
unho: Nit: `yield {'never': "called"}`

tests/misc/state.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@pytest.mark.django_db
+def test_state_states():
+
+    class ContextualState(State):
+
+        @property
+        def states(self):
+            return ["foo", "bar", "empty"]
```
unho: Nit: Use alphabetical order as below.

unho: nvm. I see that it is intended.

pootle/core/response.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class ItemResponse(object):
+
+    def __init__(self, response, action_type, complete=True, msg=None, **kwargs):
+        self.response = response
+        self.action_type = action_type
+        self.complete = complete
+        self.kwargs = kwargs
+        self.msg = msg or ""
```
unho: Nit: Set in the same order as in arguments.

pootle/core/response.py
```diff
@@ -0,0 +1,127 @@
...
+                yield k
+
+    def __len__(self):
+        return len([x for x in self.__iter__()])
+
+    def __str__(self):
+        failed = ""
+        if self.has_failed:
+            failed = "FAIL "
+        if not self.made_changes:
```
unho: Nit: In case like this I find cleaner to invert the order so `not` is not used.

pootle/core/response.py
```diff
@@ -0,0 +1,127 @@
...
+               self.context,
+               failed,
+               self.success))
+
+    @property
+    def success(self):
+        return ', '.join(
+            ["%s: %s" % (k, len(list(self.completed(k))))
+             for k in self.response_types
+             if (len(list(self.completed(k))))])
```
unho: Nit: is the parentheses around `len` actually necessary. It already has a lot of them.

pootle/core/response.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @property
+    def response_types(self):
+        return self.__responses__.keys()
+
+    @property
+    def has_failed(self):
+        return len(list(self.failed())) and True or False
```
unho: `return len(list(self.failed())) > 0`

pootle/core/response.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @property
+    def made_changes(self):
+        for complete in self.completed():
+            return True
+        return False
```
unho: `return len(self.completed()) > 0`?

unho:
lgtm nits aside.

tests/misc/state.py
phlax: im happy to use dict(), esp in a test

there are situations when an explicit `{}` is better (ie when you need to use vars for keys), but on the whole, i prefer using the `dict` builtin

pootle/core/state.py
```diff
@@ -0,0 +1,111 @@
...
+    def __str__(self):
+        return (
+            "<%s(%s): %s %s>"
+            % (self.__class__.__name__,
+               self.state.context,
+               self.state_type,
+               self.kwargs))
+
+    def __getattr__(self, k):
+        if k in self.kwargs:
```
ta2-1: is this uncovered?

phlax: it gets covered in the next PR, but yep, it should be tested in its own right - ill add

tests/misc/state.py
ta2-1: is it necessary? I think that `State.states` should have the same result.

phlax: no the default algo would include `baz` that is what this is testing

phlax: ill add a docstring

pootle/core/state.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __iter__(self):
+        for k in self.__state__:
+            if self.__state__[k]:
+                yield k
+
+    def __str__(self):
+        if not self.has_changed:
+            return (
+                "<%s(%s): Everything up-to-date>"
```
ta2-1: "Everything up-to-date" is default str() and we put `State` to `pootle.core.state`. I feel that `State` is going to be used for tracking update changes but there is no infomation here about that.

ta2-1:
lgtm

PR #4744, opened 2016-05-21
Travis: Use nodejs 0.12

unho:
This allows us to get rid of most of the warnings in build process.


unho:
Closed since the usage of `nvm` is not strictly necessary and adds another layer.

Issue #4633, opened 2016-04-06
Ensure tests can be run with `--reuse-db`


julen:
When iterating over a test that require DB access (or a few of them), currently a site-wide setup is made which in such scenario ends up being relatively time-consuming and tedious.

Ideally one could use [pytest-django's `--reuse-db` flag](http://pytest-django.readthedocs.org/en/latest/database.html#reuse-db-reuse-the-testing-database-between-test-runs) to considerably reduce setup time on test iterations, however at the current state of things such feature cannot be used due to the way the Pootle test DB environment is setup.

Let's try to fix that so we can benefit from `--reuse-db`.

PR #4745, opened 2016-05-21
Cleanup: dont add vfolders in test setup

phlax:
this doesnt speed things up as much as i had hoped, but does seem to make test runs a little  faster.

this PR also has the benefit that it speeds up setup time significantly, making it quicker when running tests locally, repeatedly


unho:
lgtm

julen:
> making it quicker when running tests locally, repeatedly

cf. #4633.

PR #4747, opened 2016-05-22
Add method to StoreManager to create a store by path

phlax:
also, by default, will create the tp and directory if required

includes tests

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -45,6 +45,7 @@ from pootle.core.url_helpers import (
 from pootle.core.utils import dateformat
 from pootle.core.utils.aggregate import max_column
 from pootle.core.utils.timezone import datetime_min, make_aware
+from pootle_app.models import Directory
 from pootle_misc.checks import check_names, get_checker
 from pootle_misc.util import import_func
 from pootle_statistics.models import (Submission, SubmissionFields,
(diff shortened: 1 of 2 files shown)


tests/models/store.py
```diff
@@ -15,22 +15,25 @@
 
 import pytest
 
+from pytest_pootle.factories import LanguageDBFactory
 from pytest_pootle.utils import update_store
 
 from translate.storage.factory import getclass
 
 from django.core.files.uploadedfile import SimpleUploadedFile
-
```
unho: Keep this blank line.

tests/models/store.py
```diff
@@ -723,3 +726,102 @@ def provide_serializers(**kwargs):
...
+@pytest.mark.django_db
+def test_store_create_by_path():
+
+    # create in tp
+    path = '/language0/project0/path.po'
+    store = Store.objects.create_by_path(path)
+    assert store.pootle_path == path
+
+    # create in tp again - get existing store
+    path = '/language0/project0/path.po'
```
unho: Is it really necessary to redefine the same path?

pootle/apps/pootle_store/models.py
```diff
@@ -1273,6 +1274,50 @@ def live(self):
         """Filters non-obsolete stores."""
         return self.filter(obsolete=False)
 
+    def create_by_path(self, pootle_path, project=None,
+                       create_tp=True, create_directory=True):
+        from pootle_language.models import Language
+        from pootle_project.models import Project
+
+        lang, proj, dir, fn = split_pootle_path(pootle_path)
```
unho: Despite there are already other places where `dir` is used like this, IMHO it is better to use `dir_path` instead (or something similar) to avoid redefining `dir`.

pootle/apps/pootle_store/models.py
```diff
@@ -1273,6 +1274,50 @@ def live(self):
...
+
+        ext = fn.split(".")[-1]
+
+        if not project:
```
unho: `project is None`

pootle/apps/pootle_store/models.py
```diff
@@ -1273,6 +1274,50 @@ def live(self):
...
+        if dir:
+            dir_path = "/".join(["", lang, proj, dir])
+            if not create_directory:
+                parent = Directory.objects.get(pootle_path=dir_path)
+            else:
+                parent = tp.directory
+                for child in dir.strip("/").split("/"):
+                    parent, created = Directory.objects.get_or_create(
+                        name=child, parent=parent)
+        if not dir:
```
unho: `else`?

unho:
Nits aside, lgtm.

pootle/apps/pootle_store/models.py
phlax: good catch!

unho: Please note `dir_path` is already being used below.

ta2-1:
`StoreManager.create_by_path()` method is going to be used by `pootle_fs`. 
Having to say that it touches very important topic: DB objects created via pootle_fs are decoupled from `POOTLE_TRANSLATION_DIRECTORY` content. That wasn't obviuos from this PR.
I'm happy that #4753 is created for addressing of that issue.
So it's my GTM.

phlax:
merged in 46a28fccbadf2139b11797667cf68d882cd8985b

PR #4750, opened 2016-05-23
Travis: don't cache builds on stable/* branches

dwaynebailey:
We don't cache builds on `master` that are not pull requests.  We shouldn't be caching on `stable/*` either.


unho:
lgtm

PR #4754, opened 2016-05-23
Add "none" option to Project.treestyle

phlax:
This PR allows a Project to have its treestyle set to "none" to disable scan/mkdir behaviour

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -181,10 +181,10 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
 
     treestyle_choices = (
         # TODO: check that the None is stored and handled correctly
-        ('auto', _('Automatic detection (slower)')),
+        ('auto', _('Automatic detection of gnu/non-gnu file layouts (slower)')),
         ('gnu', _('GNU style: files named by language code')),
         ('nongnu', _('Non-GNU: Each language in its own directory')),
-    )
+        ('none', _('Allow pootle_fs to manage filesystems')))
     treestyle = models.CharField(max_length=20, default='auto',
                                  choices=treestyle_choices,
                                  verbose_name=_('Project Tree Style'))
@@ -364,8 +364,11 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         return self.fullname
 
     def save(self, *args, **kwargs):
-        # Create file system directory if needed
-        if not self.directory_exists_on_disk() and not self.disabled:
+        requires_translation_directory = (
+            not self.treestyle == "none"
+            and not self.disabled
+            and not self.directory_exists_on_disk())
+        if requires_translation_directory:
             os.makedirs(self.get_real_path())
 
         from pootle_app.models.directory import Directory
--- a/pootle/apps/pootle_translationproject/models.py
+++ b/pootle/apps/pootle_translationproject/models.py
@@ -158,7 +158,7 @@ class TranslationProject(models.Model, CachedTreeItem):
 
     language = models.ForeignKey(Language, db_index=True)
     project = models.ForeignKey(Project, db_index=True)
-    real_path = models.FilePathField(editable=False)
+    real_path = models.FilePathField(editable=False, null=True, blank=True)
     directory = models.OneToOneField(Directory, db_index=True, editable=False)
     pootle_path = models.CharField(max_length=255, null=False, unique=True,
                                    db_index=True, editable=False)
@@ -191,11 +191,15 @@ class TranslationProject(models.Model, CachedTreeItem):
 
     @property
     def abs_real_path(self):
-        return absolute_real_path(self.real_path)
+        if self.real_path is not None:
+            return absolute_real_path(self.real_path)
 
     @abs_real_path.setter
     def abs_real_path(self, value):
-        self.real_path = relative_real_path(value)
+        if value is not None:
+            self.real_path = relative_real_path(value)
+        else:
+            self.real_path = None
 
     @property
     def file_style(self):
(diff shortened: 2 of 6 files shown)


pootle/apps/pootle_translationproject/models.py
```diff
@@ -191,11 +191,15 @@ def fullname(self):
 
     @property
     def abs_real_path(self):
-        return absolute_real_path(self.real_path)
+        if self.real_path is not None:
+            return absolute_real_path(self.real_path)
```
unho: Shouldn't this be returning something always?

pootle/apps/pootle_translationproject/models.py
```diff
@@ -158,7 +158,7 @@ class TranslationProject(models.Model, CachedTreeItem):
 
     language = models.ForeignKey(Language, db_index=True)
     project = models.ForeignKey(Project, db_index=True)
-    real_path = models.FilePathField(editable=False)
+    real_path = models.FilePathField(editable=False, null=True, blank=True)
```
unho: This is used in other parts of the codebase. I unsure if those other places require adjustments to handle the `None` value.

pootle/apps/pootle_translationproject/models.py
```diff
@@ -257,12 +261,14 @@ def save(self, *args, **kwargs):
                                       .get_or_make_subdir(self.project.code)
         self.pootle_path = self.directory.pootle_path
 
-        project_dir = self.project.get_real_path()
-        from pootle_app.project_tree import get_translation_project_dir
-        self.abs_real_path = get_translation_project_dir(
-            self.language, project_dir, self.file_style, make_dirs=not
-            self.directory.obsolete)
-
+        if not self.project.treestyle == "none":
```
unho: Nit: I find cleaner not to negate a condition when I can just switch the order in the branches.

pootle/apps/pootle_translationproject/models.py
phlax: no, i want it to return none if no project_tree is set up, that way there can be no possibility of anything on disk interfering with Stores

pootle/apps/pootle_translationproject/models.py
unho: Also it would be `!=` instead of `not ==`

pootle/apps/pootle_translationproject/models.py
phlax: well, ive grepped and checked, and run all the pootle_fs tests with treestyle set to None, and there doesnt seem to be any issues.

afaict the relevant bits of code are only ever triggered from Project/TP.save()

pootle/apps/pootle_translationproject/models.py
dwaynebailey: or `if self.project.treestyle != "none":` surely?

phlax: ill update

tests/models/translationproject.py
```diff
@@ -143,3 +143,31 @@ def test_tp_checker(tp_checker_tests):
...
+        source_language=english,
+        treestyle="none")
+    language = LanguageDBFactory()
+    TranslationProjectFactory(
+        language=templates, project=project)
+
+    tp = TranslationProject.objects.create(
+        project=project, language=language)
+
+    assert not tp.abs_real_path
```
unho: Just for me learning something, what does return a property when it returns nothing?

phlax: `None`

pootle/apps/pootle_translationproject/models.py
unho: Then make it explicit. In order words, return something always.

phlax: no - that is pointless

pootle/apps/pootle_translationproject/models.py
unho: Yes, it seems you are true.

pootle/apps/pootle_translationproject/models.py
unho: pointless? What happened to `Explicit is better than implicit`?

phlax: @unho are you really proposing to go through the code adding `return None` to every function?

unho: @phlax No, I guess I only want that for the ones you want to get landed. Now seriously, lets try to get code landed in the better shape possible.

ta2-1:
lgtm

pootle/apps/pootle_project/models.py
```diff
@@ -184,7 +184,9 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
         ('auto', _('Automatic detection (slower)')),
         ('gnu', _('GNU style: files named by language code')),
         ('nongnu', _('Non-GNU: Each language in its own directory')),
-    )
+        ('none', _('Do not scan or create directories '
+                   '(Select this option if you wish to manage files '
+                   'using pootle_fs)')))
```
julen: Please keep the closing paren on its own line, and add keep the trailing comma too.

unho:
lgtm once nits addressed.

pootle/apps/pootle_project/models.py
```diff
@@ -181,10 +181,12 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
 
     treestyle_choices = (
         # TODO: check that the None is stored and handled correctly
-        ('auto', _('Automatic detection (slower)')),
+        ('auto', _('Automatic detection of gnu/non-gnu file layouts (slower)')),
         ('gnu', _('GNU style: files named by language code')),
         ('nongnu', _('Non-GNU: Each language in its own directory')),
-    )
+        ('none', _('Do not scan or create directories '
+                   '(Select this option if you wish to manage files '
```
julen: This string is probably too technical. Also, what happens when pootle_fs is unused and disabled? Couldn't pootle_fs perhaps provide the value for `get_treestyle()`, maybe making the `auto` option just work?

phlax: i think if you have pootle_fs set up for a project and you want to unconfigure it (unlikely) then you would want to explicitly set the treestyle for the project - i dont think its helpful  for this to work automatically.

unho: This can probably be configured live depending on pootle_fs being enabled, like it is done for checker_choices above.

unho: Ideally `treestyle` should be transparent.

julen: I wasn't mentioning to unconfigure it, but disabling it. The suggestion of making it work automatic was orthogonal; it requires less UI, it's the default and it's probably what everyone will want to use.

phlax: ok, well in the case of configuring pootle_fs, then we needed something to stop project_tree from doing what it does. This seems like the best place to express that.

julen: Well, in case pootle_fs is disabled, the UI will still exhibit a _broken_ option. I don't believe our suggestions above are hare-brained, and IMO it should work better for the reasons outlined in my previous comment.

PR #4760, opened 2016-05-26
Pootle fs files

phlax:
This PR contains the base FSFile class. This wraps a store_fs object, and exposes functions such as retrieving the last_hash, pushing and pulling files.

### DIFF
--- a/pootle/apps/pootle_fs/files.py
+++ b/pootle/apps/pootle_fs/files.py
@@ -1,0 +1,229 @@
+
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import io
+import logging
+import os
+
+from translate.storage.factory import getclass
+
+from django.contrib.auth import get_user_model
+from django.utils.functional import cached_property
+
+from pootle.core.models import Revision
+from pootle_statistics.models import SubmissionTypes
+from pootle_store.models import FILE_WINS, POOTLE_WINS, Store
+
+
+logger = logging.getLogger(__name__)
+
+
+class FSFile(object):
+
+    def __init__(self, store_fs):
+        """
+        :param store_fs: ``StoreFS`` object
+        """
+        from .models import StoreFS
+
+        if not isinstance(store_fs, StoreFS):
+            raise TypeError(
+                "pootle_fs.FSFile expects a StoreFS")
+        self.store_fs = store_fs
+        self.pootle_path = store_fs.pootle_path
+        self.path = store_fs.path
+
+    def __str__(self):
+        return "<%s: %s::%s>" % (
+            self.__class__.__name__, self.pootle_path, self.path)
+
+    def __hash__(self):
+        return hash(
+            "%s::%s::%s::%s"
+            % (self.path,
+               self.pootle_path,
+               self.store_fs.last_sync_hash,
+               self.store_fs.last_sync_revision))
+
+    def __eq__(self, other):
+        return hash(other) == hash(self)
+
+    @property
+    def file_exists(self):
+        return os.path.exists(self.file_path)
+
+    @property
+    def store_exists(self):
+        return self.store is not None
+
+    @property
+    def file_path(self):
+        return os.path.join(
+            self.store_fs.project.local_fs_path,
+            self.path.strip("/"))
+
+    @property
+    def fs_changed(self):
+        latest_hash = self.latest_hash
+        return (
+            latest_hash is not None
+            and (
+                latest_hash
+                != self.store_fs.last_sync_hash))
+
+    @property
+    def latest_hash(self):
+        if self.file_exists:
+            return str(os.stat(self.file_path).st_mtime)
+
+    @property
+    def pootle_changed(self):
+        return (
+            self.store_exists
+            and (
+                self.store.get_max_unit_revision()
+                != self.store_fs.last_sync_revision))
+
+    @cached_property
+    def store(self):
+        return self.store_fs.store
+
+    def add(self):
+        logger.debug("Adding file: %s" % self.path)
+        self.store_fs.resolve_conflict = POOTLE_WINS
+        self.store_fs.save()
+
+    def create_store(self):
+        """
+        Creates a ```Store``` and if necessary the ```TranslationProject```
+        parent ```Directories```
+        """
+        store = Store.objects.create_by_path(
+            self.pootle_path,
+            project=self.store_fs.project)
+        self.store_fs.store = store
+        self.store_fs.save()
+        self.__dict__["store"] = self.store_fs.store
+
+    def delete(self):
+        """
+        Delete the file from FS and Pootle
+
+        This does not commit/push
+        """
+        store = self.store
+        if store and store.pk:
+            store.makeobsolete()
+            del self.__dict__["store"]
+        if self.store_fs.pk:
+            self.store_fs.delete()
+        self.remove_file()
+
+    def fetch(self):
+        """
+        Called when FS file is fetched
+        """
+        logger.debug("Fetching file: %s" % self.path)
+        self.store_fs.resolve_conflict = FILE_WINS
+        self.store_fs.save()
+
+    def on_sync(self):
+        """
+        Called after FS and Pootle have been synced
+        """
+        self.store_fs.resolve_conflict = None
+        self.store_fs.staged_for_merge = False
+        self.store_fs.last_sync_hash = self.latest_hash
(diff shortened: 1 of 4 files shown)


pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class FSFile(object):
+
+    def __init__(self, store_fs):
+        """
+        :param store_fs: ``FSStore`` object
```
ta2-1: StoreFS

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,229 @@
...
+    def fs_changed(self):
+        latest_hash = self.latest_hash
+        return (
+            latest_hash is not None
+            and (
+                latest_hash
+                != self.store_fs.last_sync_hash))
+
+    @property
+    def latest_hash(self):
```
ta2-1: latest_hash but returns st_mtime?

phlax: for the default localfs plugin that is what it uses, as anything else would be too costly, but this is supposed to be generic to any kind of fs, so with git, mercurial etc it will get the repo hash

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        User = get_user_model()
+        resolve_conflict = (
+            pootle_wins
+            and store_models.POOTLE_WINS
+            or store_models.FILE_WINS)
+        if merge:
+            revision = self.store_fs.last_sync_revision
+        else:
+            revision = self.store.get_max_unit_revision() + 1
```
ta2-1: Revision.incr() ?

phlax: that would increase the revision of the system though, whereas we just want to set the revision to something higher than the Store's.

That said, what would be cheaper than get_max_unit_revision is doing Revision.get()

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def deserialize(self):
+        if not self.exists:
+            return
+        deserialized = ""
+        with open(self.file_path) as f:
+            deserialized = f.read()
+        if self.store:
+            return self.store.deserialize(deserialized)
+        serialio = io.BytesIO(deserialized)
```
unho: Nit: use `serial_io` instead.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,224 @@
...
+        resolve_conflict = (
+            self.store_fs.resolve_conflict or FILE_WINS)
+        if merge:
+            revision = self.store_fs.last_sync_revision or 0
+        else:
+            revision = Revision.get() + 1
+        tmp_store = self.deserialize()
+        self.store.update(
+            tmp_store,
+            submission_type=SubmissionTypes.UPLOAD,
```
unho: I hardly doubt this is the best `SubmissionTypes` instance to use. But just checked and it seems there is no submission type for "sync with disk". Perhaps a new type should be added for PootleFS operations.

unho: To clarify, adding a `SubmissionTypes.UPLOAD` submission will reflect in timeline as an upload by some user. But here we want to reflect a synchronization done by `system` user.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,224 @@
...
+            user=user or User.objects.get_system_user(),
```
unho: In which scenarios would be a user different to `system` be used?

phlax: hopefully, in almost all.

Although not part of the logic of this class, the attribution will most likely come from either the FS itself or configuration - there are some implementation details yet to be worked out in this respsect.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def push(self, user=None):
+        """
+        Push Pootle ``Store`` into FS
+        """
+        if not self.store:
+            return
+        current_revision = self.store.get_max_unit_revision()
+        last_revision = self.store_fs.last_sync_revision
+        if self.exists and last_revision == current_revision:
```
unho: Perhaps simpler:
```python
        if self.exists and not self.pootle_changed:
```

phlax: yep, much nicer

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,223 @@
...
+        if not self.store:
+            self.create_store()
+        self._sync_to_pootle(user=user, merge=merge)
+
+    def push(self, user=None):
+        """
+        Push Pootle ``Store`` into FS
+        """
+        dont_push = (
+            not self.store
```
unho: I think this is unneeded because is already included in `pootle_changed`.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,223 @@
...
+        current_hash = self.latest_hash
+        last_hash = self.store_fs.last_sync_hash
+        if self.store and last_hash == current_hash:
+            return
+        logger.debug("Pulling file: %s" % self.path)
+        if not self.store:
+            self.create_store()
+        self._sync_to_pootle(user=user, merge=merge)
+
+    def push(self, user=None):
```
unho: Unused `user`.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,223 @@
...
+        self.store_fs.save()
+        logger.debug("File synced: %s" % self.path)
+
+    def pull(self, user=None, merge=False):
+        """
+        Pull FS file into Pootle
+        """
+        current_hash = self.latest_hash
+        last_hash = self.store_fs.last_sync_hash
+        if self.store and last_hash == current_hash:
```
unho: Perhaps cleaner:

```python
        if self.store and not self.fs_changed:
```

unho: I wonder if `self.store is not None` should be used instead. Even more, perhaps it is better to add a `self.store_exists` property (as opposed to `self.exists`) in order to make the code more meaningful.

pootle/apps/pootle_fs/files.py
unho: `hash` sounds perhaps confusing, but `mtime` is easier to pick.

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,223 @@
...
+            % (self.path,
+               self.pootle_path,
+               self.store_fs.last_sync_hash,
+               self.store_fs.last_sync_revision))
+
+    def __eq__(self, other):
+        return hash(other) == hash(self)
+
+    @property
+    def exists(self):
```
unho: I would rename this to the more meaningful `file_exists`.

pootle/apps/pootle_fs/files.py
phlax: its only mtime in the localfs case.

In the sense that we are getting a unique value for the file at that point, its a `hash` in all cases

pootle/apps/pootle_fs/files.py
phlax: unused in the localfs case, but not in any other

pootle/apps/pootle_fs/files.py
phlax: they are separate conditional clauses

pootle/apps/pootle_fs/files.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Update Pootle ``Store`` with the parsed FS file.
+        """
+        User = get_user_model()
+        resolve_conflict = (
+            self.store_fs.resolve_conflict or FILE_WINS)
+        if merge:
+            revision = self.store_fs.last_sync_revision or 0
+        else:
+            revision = Revision.get() + 1
```
ta2-1: @phlax you told me about commet here

phlax: just added

pootle/apps/pootle_project/models.py
```diff
@@ -301,6 +301,11 @@ def config(self):
         return ObjectConfig(self)
 
     @property
+    def local_fs_path(self):
+        return os.path.join(
+            settings.POOTLE_FS_PATH, self.code)
```
ta2-1: @phlax I've just realized that we don't have any description for `settings.POOTLE_FS_PATH`

phlax: no atm its only in test fixtures, but tbh this class is not yet used, so i would rather add the default/docs in pootle_fs_merge

unho:
lgtm

ta2-1:
lgtm

PR #4763, opened 2016-05-28
Add pootle_fs Response classes

phlax:
Adds the actual Response class used by pootle_fs, and some tests

### DIFF
--- a/pootle/apps/pootle_fs/response.py
+++ b/pootle/apps/pootle_fs/response.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+from collections import OrderedDict
+
+from django.utils.functional import cached_property
+
+from pootle.core.response import ItemResponse, Response
+from pootle_store.models import Store
+
+
+logger = logging.getLogger(__name__)
+
+FS_RESPONSE = OrderedDict()
+FS_RESPONSE["pulled_to_pootle"] = {
+    "title": "Pulled to Pootle",
+    "description":
+        "Stores updated where filesystem version was new or newer"}
+FS_RESPONSE["added_from_pootle"] = {
+    "title": "Added from Pootle",
+    "description":
+        "Files staged from Pootle that were new or newer than their files"}
+FS_RESPONSE["fetched_from_fs"] = {
+    "title": "Fetched from filesystem",
+    "description":
+        ("Files staged from the filesystem that were new or newer than their "
+         "Pootle Stores")}
+FS_RESPONSE["pushed_to_fs"] = {
+    "title": "Pushed to filesystem",
+    "description":
+        "Files updated where Pootle Store version was new or newer"}
+FS_RESPONSE["staged_for_removal"] = {
+    "title": "Staged for removal",
+    "description":
+        ("Files or Stores staged for removal where the corresponding "
+         "file/Store is missing")}
+FS_RESPONSE["removed"] = {
+    "title": "Removed",
+    "description":
+        ("Files or Stores removed that no longer had corresponding files or "
+         "Stores")}
+FS_RESPONSE["staged_for_merge_fs"] = {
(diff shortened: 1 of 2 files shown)


tests/pootle_fs/fs_response.py
```diff
@@ -0,0 +1,207 @@
...
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+
+import pytest
+
+
```
unho: Remove unnecessary extra blank line.

unho:
Nit aside, lgtm.

pootle/apps/pootle_fs/response.py
```diff
@@ -0,0 +1,128 @@
...
+    "title": "Merged from fs",
+    "description":
+        ("Merged - FS won where unit updates conflicted")}
+FS_RESPONSE["merged_from_pootle"] = {
+    "title": "Merged from pootle",
+    "description":
+        ("Merged - Pootle won where unit updates conflicted")}
+
+
+class ProjectFSItemResponse(ItemResponse):
```
ta2-1: Item is a Store, is it?

phlax: no, more like an item of response, which may have a Store/StoreFS or just paths related to it

pootle/apps/pootle_fs/response.py
```diff
@@ -0,0 +1,128 @@
...
+    @cached_property
+    def store(self):
+        if self.fs_state.store:
+            return self.fs_state.store
+        if self.store_fs:
+            return self.store_fs.store
+        try:
+            return Store.objects.get(
+                pootle_path=self.fs_state.pootle_path)
+        except Store.DoesNotExist:
```
ta2-1: What about `MultipleObjectsReturned`?

ta2-1: have I misunderstood this
> just paths related to it

?

phlax: i dont believe its possible to have multiple Stores with the same pootle_path

phlax: in relation to "just paths related"

imagine the following scenario

- file  appears - ie `fs_untracked`
- user diidnt actually want the file (dupes or something)
- user removes 
- we then have an action(/response) that has no `Store`, and no `StoreFS` - the only thing it has is a an `fs_path`, and a corresponding `pootle_path` for where it would have been put had it been kept

pootle/apps/pootle_fs/response.py
```diff
@@ -0,0 +1,128 @@
...
+        return (
+            "<%s(%s)%s: %s %s::%s>"
+            % (self.__class__.__name__,
+               self.response.context,
+               self.failed and " FAILED" or "",
+               self.action_type,
+               self.pootle_path,
+               self.fs_path))
+
+    @property
```
ta2-1: It equals `ItemResponse.failed`.

pootle/apps/pootle_fs/response.py
ta2-1: k, I got it. thanks.

tests/pootle_fs/fs_response.py
```diff
@@ -0,0 +1,206 @@
...
+    action_type = kwargs.pop("action_type")
+    state_type = kwargs.pop("state_type")
+    resp = ProjectFSResponse(DummyContext())
+    state = DummyFSState(DummyContext(), **kwargs)
+    for fs_state in state[state_type]:
+        resp.add(action_type, fs_state=fs_state)
+    assert resp.has_failed is False
+    assert resp.made_changes is True
+    assert resp.response_types == FS_RESPONSE.keys()
+    assert len(list(resp.completed())) == 2
```
ta2-1: can we pass this `2` via function arg here?

ta2-1: as I understand it comes from `range(0, 2)`...

ta2-1:
@phlax thanks.
There is one small nit re using constant in tests still but I think it's not blocker.
And I hope states and actions will match)
lgtm.

tests/pootle_fs/fs_response.py
phlax: ive added it as a param to the test function, but i dont think i need to pass it every time

PR #4766, opened 2016-05-29
Add pootle_fs resources class

phlax:
Adds a resource class for encapsulating fs plugins relationship to Stores/StoreFS

### DIFF
--- a/pootle/apps/pootle_fs/resources.py
+++ b/pootle/apps/pootle_fs/resources.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+from pootle_fs.models import StoreFS
+from pootle_store.models import Store
+
+
+class FSProjectResources(object):
+
+    def __init__(self, project):
+        self.project = project
+
+    def __str__(self):
+        return (
+            "<%s(%s)>"
+            % (self.__class__.__name__,
+               self.project))
+
+    @property
+    def stores(self):
+        return Store.objects.filter(
+            translation_project__project=self.project)
+
+    @property
+    def tracked(self):
+        return StoreFS.objects.filter(
+            project=self.project).select_related("store")
+
+    @property
+    def synced(self):
+        return (
+            self.tracked.exclude(last_sync_revision__isnull=True)
+                        .exclude(last_sync_hash__isnull=True))
+
+    @property
+    def unsynced(self):
+        return (
+            self.tracked.filter(last_sync_revision__isnull=True)
+                        .filter(last_sync_hash__isnull=True))
+
+    @property
+    def trackable_stores(self):
+        return self.stores.exclude(obsolete=True).filter(fs__isnull=True)
--- a/tests/pootle_fs/resources.py
+++ b/tests/pootle_fs/resources.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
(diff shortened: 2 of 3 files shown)


tests/pootle_fs/resources.py
```diff
@@ -0,0 +1,108 @@
...
+    assert str(resources) == "<FSResources(Project 0)>"
+
+
+@pytest.mark.django_db
+def test_project_resources_stores():
+    project = Project.objects.get(code="project0")
+    stores = Store.objects.filter(
+        translation_project__project=project)
+    assert list(FSResources(project).stores) == list(stores)
+    # mark some Stores obsolete - should still show
```
unho: Is this comment about something not yet coded?

unho:
lgtm aside that comment that is not clear to me.

tests/pootle_fs/resources.py
phlax: yep, a todo, ill add

tests/pootle_fs/resources.py
```diff
@@ -0,0 +1,108 @@
...
+    obsoleted = synced.filter(store__obsolete=True).first()
+    obsoleted.last_sync_hash = "FOO"
+    obsoleted.last_sync_revision = 23
+    obsoleted.save()
+    active = synced.exclude(store__obsolete=True).first()
+    active.last_sync_hash = "FOO"
+    active.last_sync_revision = 23
+    active.save()
+    synced = synced.exclude(last_sync_revision__isnull=True)
+    synced = synced.exclude(last_sync_hash__isnull=True)
```
ta2-1: this is a test of course but i thought there are duplicated strings here..

pootle/apps/pootle_fs/resources.py
```diff
@@ -0,0 +1,46 @@
...
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+from pootle_fs.models import StoreFS
+from pootle_store.models import Store
+
+
+class FSResources(object):
```
ta2-1: FSProjectResources

pootle/apps/pootle_fs/resources.py
```diff
@@ -0,0 +1,46 @@
...
+                        .exclude(last_sync_hash__isnull=True))
+
+    @property
+    def unsynced(self):
+        return (
+            self.tracked.filter(last_sync_revision__isnull=True)
+                        .filter(last_sync_hash__isnull=True))
+
+    @property
+    def trackable(self):
```
ta2-1: trackable_stores

ta2-1:
there are some nits. 
lgtm after they are addressed.

tests/pootle_fs/resources.py
phlax: which strings are duplicated?

ta2-1: there are no duplicated strings but I could think so for a second.

```
synced = synced.exclude(
    last_sync_revision__isnull=True,
    last_sync_hash__isnull=True)
```

phlax: i did it this way for a reason, that being that the ORM can do different things with in these 2 situations - altho not always clear to me exactly why.

basically i want an OR exclude, and im not 100% i would get that with the way you show

ta2-1:
```
from django.db.models import Q
synced = synced.exclude(
    Q(last_sync_revision__isnull=True)
    | Q(last_sync_hash__isnull=True)
)
```
I always do `OR` via `Q()`, but I don't know if it works here?

phlax: yep also `Q(~some_file=...)`

tbh i went with what i know works and is relatively simple

ta2-1: I wonder why 
```
qs = qs.exclude(cond1)
qs = qs.exclude(cond2)
``` 
works as `OR`. It has to work as `AND`, doesn't it?

PR #4768, opened 2016-05-30
Add FSPathMatcher util

phlax:
separating out the fs/qs path matcher and adding tests

### DIFF
--- a/pootle/apps/pootle_fs/utils.py
+++ b/pootle/apps/pootle_fs/utils.py
@@ -7,9 +7,63 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from fnmatch import translate
+
+from django.utils.functional import cached_property
+
 from pootle.core.exceptions import MissingPluginError, NotConfiguredError
 
 from .delegate import fs_plugins
+
+
+class PathFilter(object):
+
+    def path_regex(self, path):
+        return translate(path).replace("\Z(?ms)", "$")
+
+
+class StorePathFilter(PathFilter):
+    """Filters Stores (only pootle_path)
+    pootle_path should be file a glob
+    the glob is converted to a regex and used to filter a qs
+    """
+
+    def __init__(self, pootle_path=None):
+        self.pootle_path = pootle_path
+
+    @cached_property
+    def pootle_regex(self):
+        if not self.pootle_path:
+            return
+        return self.path_regex(self.pootle_path)
+
+    def filtered(self, qs):
+        if not self.pootle_regex:
+            return qs
+        return qs.filter(pootle_path__regex=self.pootle_regex)
+
+
+class StoreFSPathFilter(StorePathFilter):
+    """Filters StoreFS
+    pootle_path and fs_path should be file globs
+    these are converted to regexes and used to filter a qs
+    """
+
+    def __init__(self, pootle_path=None, fs_path=None):
+        super(StoreFSPathFilter, self).__init__(pootle_path=pootle_path)
+        self.fs_path = fs_path
+
+    @cached_property
+    def fs_regex(self):
+        if not self.fs_path:
+            return
+        return self.path_regex(self.fs_path)
+
+    def filtered(self, qs):
+        qs = super(StoreFSPathFilter, self).filtered(qs)
+        if not self.fs_regex:
+            return qs
+        return qs.filter(path__regex=self.fs_regex)
 
 
 class FSPlugin(object):
--- a/tests/pootle_fs/utils.py
+++ b/tests/pootle_fs/utils.py
@@ -7,13 +7,18 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from fnmatch import translate
+
 import pytest
 
 from pootle.core.exceptions import MissingPluginError, NotConfiguredError
 from pootle.core.plugin import provider
 from pootle_fs.delegate import fs_plugins
-from pootle_fs.utils import FSPlugin
+from pootle_fs.models import StoreFS
+from pootle_fs.utils import (
+    PathFilter, StoreFSPathFilter, StorePathFilter, FSPlugin)
 from pootle_project.models import Project
+from pootle_store.models import Store
 
 
 class DummyFSPlugin(object):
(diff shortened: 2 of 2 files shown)


tests/pootle_fs/utils.py
```diff
@@ -7,12 +7,15 @@
...
+from fnmatch import translate
+
 import pytest
 
 from pootle.core.exceptions import MissingPluginError, NotConfiguredError
 from pootle.core.plugin import provider
 from pootle_fs.delegate import fs_plugins
-from pootle_fs.utils import FSPlugin
+from pootle_fs.utils import (
+    FSPathFilter, StoreFSPathFilter, StorePathFilter, FSPlugin)
```
unho: `FSPathFilter`? Where is it being defined?

pootle/apps/pootle_fs/utils.py
```diff
@@ -7,11 +7,76 @@
...
+    the glob is converted to a regex and used to filter a qs
+    """
+
+    def __init__(self, pootle_path=None):
+        self.pootle_path = pootle_path
+        self.fs_path = None
+
+    @property
+    def fs_regex(self):
+        return None
```
unho: Is this not the same as the inherited `fs_regex`?

pootle/apps/pootle_fs/utils.py
```diff
@@ -7,11 +7,76 @@
...
+    pass
+
+
+class StorePathFilter(PathFilter):
+    """Filters Stores (only pootle_path)
+    pootle_path should be file a glob
+    the glob is converted to a regex and used to filter a qs
+    """
+
+    def __init__(self, pootle_path=None):
```
unho: Shouldn't this have a second `fs_path` argument?

pootle/apps/pootle_fs/utils.py
```diff
@@ -7,11 +7,76 @@
...
+
+    def __init__(self, pootle_path=None, fs_path=None):
+        self.pootle_path = pootle_path
+        self.fs_path = fs_path
+
+    def filtered_pootle_path(self, qs):
+        if not self.pootle_regex:
+            return qs
+        return qs.filter(
+            pootle_path__regex=self.path_regex(self.pootle_path))
```
ta2-1: `return qs.filter(pootle_path__regex=self.pootle_regex)` ?

pootle/apps/pootle_fs/utils.py
phlax: no, its to make sure that the one in the parent class is  never called

tests/pootle_fs/utils.py
```diff
@@ -35,6 +39,74 @@ def plugin_method(self, foo):
...
+        assert path_filter.filtered(stores) is stores
+    else:
+        assert (
+            list(path_filter.filtered(stores))
+            == list(stores.filter(pootle_path__regex=path_filter.pootle_regex)))
+
+
+@pytest.mark.parametrize(
+    "glob", ["/foo", "/bar/*", "*.baz", "[abc]", "[!xyz]", "language?"])
+def test_fs_path_filter_path_regex(glob):
```
unho: Rename test to get rid of `fs`.

pootle/apps/pootle_fs/utils.py
unho: As I am seeing in this class' `__init__` you are setting `self.fs_path = None`, so in the superclass' `fs_regex` it will just return.

pootle/apps/pootle_fs/utils.py
```diff
@@ -7,11 +7,76 @@
...
+
+    def filtered_fs_path(self, qs):
+        if not self.fs_regex:
+            return qs
+        return qs.filter(path__regex=self.fs_regex)
+
+    def filtered(self, qs):
```
ta2-1: so qs is StoreFS queryset...
do we need `filtered_fs_path` and `filtered_pootle_path` separately from `filtered`?

pootle/apps/pootle_fs/utils.py
phlax: yep. the subclass does not use this attribute, and prevents itself from being instantiated with it

pootle/apps/pootle_fs/utils.py
phlax: good spot, i was updating these

pootle/apps/pootle_fs/utils.py
unho: ok

pootle/apps/pootle_fs/utils.py
ta2-1: reading further... you can ignore my previous comment...

pootle/apps/pootle_fs/utils.py
```diff
@@ -7,11 +7,76 @@
...
+
+
+class StorePathFilter(PathFilter):
+    """Filters Stores (only pootle_path)
+    pootle_path should be file a glob
+    the glob is converted to a regex and used to filter a qs
+    """
+
+    def __init__(self, pootle_path=None):
+        self.pootle_path = pootle_path
```
ta2-1:
```
super(StorePathFilter, self).__init__(pootle_path=pootle_path, fs_path=None)
```
?

tests/pootle_fs/utils.py
phlax: no - its part of the pootle_fs suite and needs to be distinguishable as such

pootle/apps/pootle_fs/utils.py
phlax: well i suppose we dont need them, but both StoreFS and Store need pootle_path filtering, but only StoreFS needs fs_path filtering, so i would rather keep

pootle/apps/pootle_fs/utils.py
phlax: i want to make sure there is no chance of this class being either instantiated with an fs_path, or attempt to filter on it - so i would rather keep as is

pootle/apps/pootle_fs/utils.py
phlax: the property is to prevent this class (or a subclass of it) ever trying to filter Stores on path, again i would rather keep this

pootle/apps/pootle_fs/utils.py
ta2-1: how this class can be instantiated with fs_path if it has the only parameter in __init__?

phlax: it cant, thats the point

phlax: `Store` cannot be filtered on `path` only on `pootle_path`

unho:
lgtm

PR #4769, opened 2016-05-30
Add FSProjectStateResources

phlax:
for caching resources on state object, and to make it easier to review and test

### DIFF
--- a/pootle/apps/pootle_fs/resources.py
+++ b/pootle/apps/pootle_fs/resources.py
@@ -6,9 +6,15 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from fnmatch import fnmatch
 
-from pootle_fs.models import StoreFS
+from django.db.models import F, Max
+from django.utils.functional import cached_property
+
 from pootle_store.models import Store
+
+from .models import StoreFS
+from .utils import StoreFSPathFilter, StorePathFilter
 
 
 class FSProjectResources(object):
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_fs/resources.py
```diff
@@ -47,3 +53,96 @@ def unsynced(self):
     @property
     def trackable_stores(self):
         return self.stores.exclude(obsolete=True).filter(fs__isnull=True)
+
+
+class FSProjectStateResources(object):
```
unho: An explanatory docstring would be appreciated.

pootle/apps/pootle_fs/resources.py
```diff
@@ -47,3 +53,96 @@ def unsynced(self):
...
+
+    def __init__(self, context, pootle_path=None, fs_path=None):
+        self.context = context
```
unho: What is `context`?

phlax: in this case its an FSPlugin, but its kinda agnostic, as long as the context has a resources attribute that (would if we had interfaces) implements ProjectResources (see class above)

phlax: actually it needs `get_fs_path`and a couple of other things so it needs an `FSPlugin` really

unho: As long as there is some comment in code explaining what is needed I am happy.

unho:
lgtm

pootle/apps/pootle_fs/resources.py
```diff
@@ -47,3 +53,96 @@ def unsynced(self):
...
+    @cached_property
+    def fs_translations(self):
+        return sorted(self.context.find_translations(
+            fs_path=self.fs_path, pootle_path=self.pootle_path))
+
+    @cached_property
+    def resources(self):
+        return self.context.resources
+
+    @cached_property
```
ta2-1: why do we need `@cached_property` here?

pootle/apps/pootle_fs/resources.py
```diff
@@ -47,3 +53,96 @@ def unsynced(self):
...
+        if not self.fs_path or fnmatch(path, self.fs_path):
+            return path
+
+    def _exclude_staged(self, qs):
+        return (
+            qs.exclude(staged_for_removal=True)
+              .exclude(staged_for_merge=True))
+
+    @cached_property
+    def fs_translations(self):
```
ta2-1: what are translations? localized files?

phlax: hmm, yep - this defo needs to be renamed

its paths to found translation files

pootle/apps/pootle_fs/resources.py
phlax: this filter is called repeatedly - there is no reason to keep recreating it and the regexes etc

pootle/apps/pootle_fs/resources.py
unho: Addressed.

unho:
lgtm

pootle/apps/pootle_fs/resources.py
```diff
@@ -47,3 +53,111 @@ def unsynced(self):
...
+    If present all resources are filtered accordingly.
+    """
+
+    def __init__(self, context, pootle_path=None, fs_path=None):
+        self.context = context
+        self.pootle_path = pootle_path
+        self.fs_path = fs_path
+
+    def match_fs_path(self, path):
+        """Filter fs_paths using file glob"""
```
ta2-1: It seems "filter" is not a proper word here. Check probs?

phlax: hmm well it does filter in the sense that it only returns if matched (or no match required)

ta2-1:
lgtm

PR #4770, opened 2016-05-31
Fix: use correct calendar month length when calculating dates

phlax:
~~im not 100% sure if this is the correct fix, but it works right now~~

im pretty sure this is correct now


unho:
lgtm

ta2-1:
@phlax I'm ok to fix travis now. So it's lgtm after travis pass. But I don't follow why we should test calendar months here.

phlax:
@ta2-1 there is some functionality in get_units/SearchBackend to search for units by calendar month.

I dont rem off hand what this is used for but i believe something to do with reports

PR #4773, opened 2016-05-31
Avoid bothering translators on failing XHR requests

### LINKED ISSUE #4771: Do not display error messages on failing XHR requests
Current Pootle [handling for failing XHR requests](https://github.com/translate/pootle/blob/master/pootle/static/js/editor/app.js#L1056-L1079) (for example when querying amaGama) is to display some not helpful error messages, see following screenshot:

selection_676

A more user friendly approach was suggested by @jason-p-pickering:

> From a translation perspective, I think if Amagama fails for whatever
reason, Pootle should simply swallow the error (possibly logging the
occurrence in the JavaScript console) but not continually display the error
message to the translator. The TM is there to help the translator, but
if the service does not work, then I don't think we should penalize the
translator to see the error message each and every time it occurs.
>
> I am not sure if this is what happens, but from a usability standpoint this
would seem to make sense.

This issue was originally reported in https://github.com/translate/pootle/issues/4759.

unho:
Fixes #4771.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -2215,7 +2215,7 @@ PTL.editor = {
     fetch({ url: tmUrl, crossDomain: true })
       .then(
         (data) => this.handleTmResults(data, store, unit),
-        this.error
+        (xhr, s) => console.error(`HTTP ${xhr.status} (${s}): ${tmUrl}`)
       );
   },
 


unho:
@translate/core Changed. Please review again.

julen:
No objections once the Travis error has been fixed.

PR #4775, opened 2016-05-31
cleanup for fixtures and adjustment to pootle.core.State class

phlax:
- moves add_store_fs to pytest_pootle.utils
- makes State class work with object attributes
- Adjust StateResources found_file_paths property and tests


unho:
lgtm

ta2-1:
lgtm

PR #4783, opened 2016-06-03
Pootle fs cleanups/tests

phlax:
- adds a path matcher
- FSPlugin delegates equality to wrapped plugin
- cleanups

### DIFF
--- a/pootle/apps/pootle_fs/files.py
+++ b/pootle/apps/pootle_fs/files.py
@@ -133,6 +133,14 @@ class FSFile(object):
         self.store_fs.resolve_conflict = FILE_WINS
         self.store_fs.save()
 
+    def merge(self, pootle_wins):
+        if pootle_wins:
+            self.store_fs.resolve_conflict = POOTLE_WINS
+        else:
+            self.store_fs.resolve_conflict = FILE_WINS
+        self.store_fs.staged_for_merge = True
+        self.store_fs.save()
+
     def on_sync(self):
         """
         Called after FS and Pootle have been synced
--- a/pootle/apps/pootle_fs/files.py
+++ b/pootle/apps/pootle_fs/files.py
@@ -189,6 +189,10 @@ class FSFile(object):
         if self.file_exists:
             os.unlink(self.file_path)
 
+    def rm(self):
+        self.store_fs.staged_for_removal = True
+        self.store_fs.save()
+
     def deserialize(self):
         if not self.file_exists:
             return
--- a/pootle/apps/pootle_fs/matcher.py
+++ b/pootle/apps/pootle_fs/matcher.py
@@ -1,0 +1,122 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+import os
+from fnmatch import fnmatch
+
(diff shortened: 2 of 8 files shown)


tests/pootle_fs/files.py
```diff
@@ -136,6 +136,22 @@ def test_wrap_store_fs_add(store_fs_file_store):
 
 
 @pytest.mark.django_db
+def test_wrap_store_fs_merge_pootle(store_fs_file_store):
+    fs_file = store_fs_file_store
+    fs_file.merge(True)
```
unho: Just for clarity I would use the named argument `pootle_wins`.

unho:
lgtm.

pootle/apps/pootle_fs/matcher.py
```diff
@@ -0,0 +1,101 @@
...
+        return self.context.project
+
+    @property
+    def root_directory(self):
+        return self.context.project.local_fs_path
+
+    @property
+    def translation_path(self):
+        return self.context.project.config[
+            "pootle_fs.translation_paths"]["default"]
```
ta2-1: does "default" translation path limit multiple translation_paths here?

pootle/apps/pootle_fs/matcher.py
```diff
@@ -0,0 +1,101 @@
...
+            ["", language_code,
+             self.project.code]
+            + [m for m in
+               matched.get('dir_path', '').split("/")
+               if m]
+            + ["%s.%s"
+               % (filename, ext)])
+
+    def match_pootle_path(self, pootle_path_match=None, **matched):
+        pootle_path = self.make_pootle_path(**matched)
```
ta2-1: why do we combine matching by `pootle_path_match` and paths codes from `matched`?
when do we use this match?

phlax: its for reverse matching - ie given a set of lang/subdirs/filename would our `translation_path` give us a fs path, given the fs/pootle glob filter

pootle/apps/pootle_fs/matcher.py
phlax: in the original implementation there was an additional method, this iterated through the translation_path matches using a slightly iffy (but maybe ok) algorithm. This was what i removed.

So now we still have a `translation_paths` dictionary, im thinking we might want to just replace that with a `translation_path` string

phlax: ...so answering your q, yep this implementation only uses a single translation path - so in a sense limits it, but it doesnt limit from being implemented in the future.

pootle/apps/pootle_fs/files.py
```diff
@@ -133,6 +133,14 @@ def fetch(self):
         self.store_fs.resolve_conflict = FILE_WINS
         self.store_fs.save()
 
+    def merge(self, pootle_wins):
+        if pootle_wins:
```
ta2-1: I seems you added tests but I can't understand why it is not covered...

phlax: i think it is

unho:
lgtm, nits aside.

unho:
lgtm

pootle/apps/pootle_fs/files.py
```diff
@@ -133,6 +133,14 @@ def fetch(self):
         self.store_fs.resolve_conflict = FILE_WINS
         self.store_fs.save()
 
+    def merge(self, pootle_wins):
```
ta2-1: here

ta2-1:
lgtm

PR #4786, opened 2016-06-07
Cleanup: simplify JS object parsing in templates

julen:
By using the `to_js` filter on any Python object, it'll return all what's needed
to read the value into a native JS object via JSON, e.g.:

  var js_foo = {{ my_python_object|to_js }};

Therefore removing all the error-prone boilerplate needed at the moment.

### DIFF
--- a/pootle/templates/layout.html
+++ b/pootle/templates/layout.html
@@ -1,4 +1,4 @@
-{% load i18n assets locale profile_tags humanize staticfiles %}
+{% load core i18n assets locale profile_tags humanize staticfiles %}
 {% get_current_language as LANGUAGE_CODE %}
 <!DOCTYPE html>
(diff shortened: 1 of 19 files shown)


pootle/templates/layout.html
```diff
@@ -29,7 +29,7 @@
       CONTACT_ENABLED: {{ settings.POOTLE_CONTACT_ENABLED|yesno:'true, false' }},
       SIGNUP_ENABLED: {{ settings.POOTLE_SIGNUP_ENABLED|yesno:'true, false' }},
       MARKUP_FILTER: '{{ settings.POOTLE_MARKUP_FILTER }}',
-      SOCIAL_AUTH_PROVIDERS: {{ SOCIAL_AUTH_PROVIDERS|safe }},
+      SOCIAL_AUTH_PROVIDERS: {{ SOCIAL_AUTH_PROVIDERS|to_js }},
```
unho: No need to load `core`?

julen: It's there (you might have taken a look at the first push).

unho: That is exactly what I did.

Addressed now.

unho:
lgtm

PR #4788, opened 2016-06-08
Error when building PO files

dwaynebailey:
* We now passthrough errors from `msgfmt` when using `build_mo`
* We're only building release languages (non-release langs have some errors)
* We add `--check` to `build_mo` to allow more thorough checks.  Unfortunately some are false positives and this can't be relied on for Travis testing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -63,5 +63,5 @@ commands=
     make travis-assets
     python setup.py sdist
     make docs
-    make mo-all
+    python setup.py build_mo
     make jslint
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -63,5 +63,5 @@ commands=
     make travis-assets
     python setup.py sdist
     make docs
-    make mo-all
+    python setup.py build_mo
```
unho: Why not `make mo`?

dwaynebailey: Because I wanted to see if in future we can actually use `--check` here.  `make mo` is just a wrapper and one I've eliminated from the release docs now anyway.

unho:
lgtm

PR #4794, opened 2016-06-09
Fix: make caighdean MT provider protocol agnostic

### LINKED ISSUE #4600: Caighdean Gaelic Machine Translation provider uses http only and breaks on https sites
The Caighdean MT engine is hardcoded to use http.  This works on http Pootle sites.  But on on https sites.

Solutions:
1. Get Caighdean onto https - that will make it work for Pootle on https or http
2. Disable the checker if Pootle is running on http

phlax:
fixes #4600

### DIFF
--- a/pootle/static/js/editor/mt/providers/CaighdeanTranslate.js
+++ b/pootle/static/js/editor/mt/providers/CaighdeanTranslate.js
@@ -16,7 +16,7 @@ class CaighdeanTranslate extends MTProvider {
       apiKey,
       name: 'caighdean-translate',
       displayName: 'Caighdeán Translate',
-      url: 'http://borel.slu.edu/cgi-bin/seirbhis3.cgi',
+      url: 'https://borel.slu.edu/cgi-bin/seirbhis3.cgi',
       supportedLanguages: ['gd', 'gv', 'ga_IE', 'ga-IE'],
     });
   }


julen:
From Paul Irish's blog, on the topic of [protocol relative URLs](http://www.paulirish.com/2010/the-protocol-relative-url/):
> Now that SSL is encouraged for everyone and doesn’t have performance concerns, this technique is now an anti-pattern. If the asset you need is available on SSL, then always use the https:// asset.
>
> Allowing the snippet to request over HTTP opens the door for attacks like the recent Github Man-on-the-side attack. It’s always safe to request HTTPS assets even if your site is on HTTP, however the reverse is not true.

phlax:
@julen fair point, ill update

pootle/static/js/editor/mt/providers/CaighdeanTranslate.js
```diff
@@ -16,7 +16,7 @@ class CaighdeanTranslate extends MTProvider {
       apiKey,
       name: 'caighdean-translate',
       displayName: 'Caighdeán Translate',
-      url: 'http://borel.slu.edu/cgi-bin/seirbhis3.cgi',
+      url: '//borel.slu.edu/cgi-bin/seirbhis3.cgi',
       supportedLanguages: ['gd', 'gv', 'ga_IE', 'ga-IE'],
```
dwaynebailey: Can we add pure 'ga' here also please

phlax: im not sure if it would need to be recognised by the MT server - @kscanne does the server support plain `ga` ?

unho: It does. But in `foinse` you can now only specify `gd` or `gv` for now.

unho: Also right now this MT doesn't require specifying the target language since it always assumes `ga`, so my guess is that `ga` (or equivalent) is only necessary in `supportedLanguages` to enable this MT button on the translation editor UI.

phlax: cool, ill add

PR #4796, opened 2016-06-09
Editor: Don't update total units count when fetching more units

### LINKED ISSUE #4640: Wrong unit numbering while translating
While translating on MLO (which I assume is running something around current master), I experienced an odd unit numbering behavior in the editor: I could go further beyond the maximum number of units reported for my current filter (_incomplete_).

editor_total_count_bug

unho:
Fixes #4640.

### DIFF
--- a/pootle/static/js/collections.js
+++ b/pootle/static/js/collections.js
@@ -22,6 +22,7 @@ export const UnitSet = Backbone.Collection.extend({
     this.chunkSize = opts.chunkSize;
     this.uIds = [];
     this.total = 0;
+    this.frozenTotal = 0;
   },
 
   comparator(unit) {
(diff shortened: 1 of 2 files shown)


julen:
Thank you, @unho! I haven't tested  the PR (and won't be able to verify it until I'm back), although I quickly checked the code and part of me feels it unnecessarily complicates UI code.

Wouldn't it be better to keep a copy of the value of `total` frozen unless a new `initial` request is made? In case the updated value of `total` is needed to calculate batches/offsets, having the frozen value in a separate property should be enough (something like `this.units.frozenTotal`), and it would help in keeping the UI code simpler, as it would always display the value of `frozenTotal`.

unho:
@julen Updated. I hope this is what you were talking about.

phlax:
lgtm - nice one @unho

unho:
@julen Merged this since we are pretty sure it fixes the issue the way you suggested.

julen:
@unho well done, verified 👍

Issue #4805, opened 2016-06-10
Expanding timeline causes a server error


dwaynebailey:
See https://mozilla.locamotion.org/mni/firefox/translate/browser/chrome/browser/aboutDialog.dtd.po#unit=7407179

`exceptions:KeyError: 'datetime'`

```pytb
Traceback (most recent call last):
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.60.0.46/newrelic/api/web_transaction.py", line 724, in __iter__
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.60.0.46/newrelic/api/web_transaction.py", line 1100, in __call__
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 189, in __call__
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.60.0.46/newrelic/hooks/framework_django.py", line 499, in wrapper
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 57, in _wrapped_view_func
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
File "/var/www/sites/mozilla/src/pootle/apps/pootle_misc/util.py", line 57, in wrapper
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
File "/var/www/sites/mozilla/src/pootle/core/views.py", line 82, in method_wrapper
File "/var/www/sites/mozilla/src/pootle/core/views.py", line 97, in method_wrapper
File "/var/www/sites/mozilla/src/pootle/core/views.py", line 537, in dispatch
File "/var/www/sites/mozilla/env/local/lib/python2.7/site-packages/newrelic-2.60.0.46/newrelic/hooks/framework_django.py", line 872, in wrapper
[...]

PR #4806, opened 2016-06-11
Fix: timeline doesnt try to parse non-existent datetime

phlax:
This is a fix for #4805, altho i think this code needs to be
refactored, and tests obviously need some improvement.

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -380,11 +380,16 @@ class UnitTimelineJSON(PootleUnitJSON):
     def get_entries_group_data(self, context):
         result = []
         for entry_group in context['entries_group']:
+            display_dt = entry_group['datetime']
+            if display_dt is not None:
+                display_dt = dateformat.format(display_dt)
+                iso_dt = entry_group['datetime'].isoformat()
+            else:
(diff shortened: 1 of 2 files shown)


unho:
lgtm

pootle/apps/pootle_store/views.py
```diff
@@ -380,11 +380,16 @@ def render_timeline(self, context):
     def get_entries_group_data(self, context):
         result = []
         for entry_group in context['entries_group']:
+            display_dt = entry_group['datetime']
+            if display_dt is not None:
+                display_dt = dateformat.format(display_dt)
+                iso_dt = entry_group['datetime'].isoformat()
+            else:
+                iso_dt = None
```
ta2-1:
```python
iso_dt = None
if display_dt is not None:
    display_dt = dateformat.format(display_dt)
    iso_dt = entry_group['datetime'].isoformat()
```
won't decrease coverage )

ta2-1:
lgtm

pootle/apps/pootle_store/views.py
phlax: it should decrease coverage - i did it this way intentionally

PR #4811, opened 2016-06-13
Fixes and cleanups for pootle_fs tests and fixtures

phlax:
rationalizing tests and fixtures in preparation for pootle_fs_cli landing


pootle/apps/pootle_fs/display.py
```diff
@@ -130,6 +130,10 @@ def store_exists(self):
             or (self.state_type == "remove"
                 and self.file.store_exists))
 
+    @property
+    def tracked(self):
+        return self.item.store_fs and True or False
```
unho: Perhaps it is simpler:

```python
        return bool(self.item.store_fs)
```

unho:
lgtm

PR #4813, opened 2016-06-14
Add test for pootle_fs.plugin api calls with paths

phlax:
also tidies the main plugin test

### DIFF
--- a/tests/pootle_fs/plugin.py
+++ b/tests/pootle_fs/plugin.py
@@ -10,12 +10,9 @@ import os
 
 import pytest
 
-from pytest_pootle.fs.utils import filtered_fs_stores
-
 from pootle.core.response import Response
 from pootle.core.state import State
 from pootle_fs.matcher import FSPathMatcher
-from pootle_fs.models import StoreFS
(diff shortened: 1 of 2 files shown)


tests/pootle_fs/plugin.py
```diff
@@ -75,51 +72,70 @@ def test_fs_plugin_unstage_staged_response(capsys, localfs_staged_envs):
...
+        changes = ["_fetched"]
+    elif expected[0] in ["pushed_to_fs"]:
+        changes = ["_pushed", "_synced"]
+    elif expected[0] in ["pulled_to_pootle"]:
+        changes = ["_pulled", "_synced"]
+    elif expected[0] in ["staged_for_merge_fs"]:
+        changes = ["_merged", "_merge_fs"]
+    elif expected[0] in ["staged_for_merge_pootle"]:
+        changes = ["_merged", "_merge_pootle"]
+    elif expected[1] and not changes:
```
unho: If you got here through all the `elif` then `changes` will still be `[]`. So this is equivalent to `elif expected[1]`.

unho:
lgtm

PR #4814, opened 2016-06-14
Add --until to contributors command

unho:


phlax:
@unho can i suggest changing until/since to before/after

unho:
> @unho can i suggest changing until/since to before/after

IMHO, until/since are easier to understand.

pytest_pootle/fixtures/models/store.py
```diff
@@ -307,6 +311,17 @@ def en_tutorial_po_member_updated(settings, english_tutorial,
 
 
 @pytest.fixture
+def en_tutorial_po_member_with_email_updated(settings, english_tutorial,
+                                             member_with_email):
```
phlax: please dont use english_tutorial - its extremely inefficient

dwaynebailey:
@unho `python-dateutils` needs to move out of `requirements/test.txt`, it works by accident here and won't work in deployment.

unho:
@dwaynebailey Thanks. Addressed.

phlax:
> and won't work in deployment

i agree that it should be moved to the core requirements - but to be fair its a pretty core dependency already - through django

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -50,16 +53,26 @@ def add_arguments(self, parser):
...
+        "2016-01-24" formats.
+        """
+        if not date_string:
+            return None
 
-        if " " in since:
-            since = "%sT%s%s" % tuple(since.split(" "))
+        try:
+            return make_aware(parse_datetime(date_string))
+        except Exception:
```
phlax: this should be a particular type of exception

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -50,16 +53,26 @@ def add_arguments(self, parser):
                  "are excluded.",
         )
 
-    def _get_revision_from_since(self, since):
-        from pootle.core.dateparse import parse_datetime
-        from pootle.core.utils.timezone import make_aware
-        from pootle_statistics.models import Submission
+    def get_aware_datetime(self, date_string):
```
phlax: now that we are just using dateutil and the existing make_aware function - im not sure that there is any benefit to separating out this method or adding tests for it

tests/commands/contributors.py
```diff
@@ -48,3 +52,83 @@ def test_contributors_get_aware_datetime():
     # Test non-empty string.
     with pytest.raises(CommandError):
         instance.get_aware_datetime("THIS FAILS TOO")
+
+
+@pytest.fixture
+def default_contributors_kwargs():
```
phlax: these fixtures should be moved to pytest_pootle

pootle/apps/pootle_statistics/utils.py
```diff
@@ -83,4 +83,4 @@ def contributors(self):
         return OrderedDict(
             [(user["username"], user)
              for user
-             in qs.values("username", "contributions", "email")])
+             in qs.values("username", "full_name", "contributions", "email")])
```
phlax: can you use `fullname` as i believe that is what is used elsewhere

unho: It is `display_name` what was used before. In order to replicate what `display_name` does we have to retrieve both `username` and `full_name`.

pootle/apps/pootle_app/management/commands/contributors.py
unho: The way I see this, there is less benefit in merging it back and removing tests.

julen:
> to be fair its a pretty core dependency already - through django

Unless I misunderstood, you should know that Django has no dependencies whatsoever. You might be mixing this with pytz, which is exclusively for handling timezones and it's optional (although highly recommended).

phlax:
@julen you are probably right- i wasnt actually able to find django's dependencies, so this probably explains it. I made this assumption because ive never worked in a django env that didnt have dateutil.

Either way - i think we all agree that we need an explicit dep on this.

pootle/apps/pootle_app/management/commands/contributors.py
```diff
@@ -32,107 +31,77 @@ def add_arguments(self, parser):
...
+        return contributors.get()
+
+    def contrib_kwargs(self, **options):
+        kwargs = {
+            k: v
+            for k, v
+            in options.items()
+            if k in ["projects", "languages", "include_anon",
+                     "since", "until", "sort_by"]}
+        kwargs["since"] = self.get_aware_datetime(kwargs["since"])
```
dwaynebailey: You could probably use [type=get_aware_datetime](https://docs.python.org/2/library/argparse.html#type) to achieve the same

unho: Thanks.

phlax:
gtm once the argparse change is added - great work - thanks @unho

docs/server/commands.rst
```diff
@@ -379,28 +379,49 @@ Available options:
...
-  8601 format (``2016-01-24T23:15:22+0000``) or be a string formatted like
-  ``"2016-01-24 23:15:22 +0000"`` (quotes included).
+  8601 format (``2016-01-24T23:15:22+0000`` or ``2016-01-24``) or be a string
+  formatted like ``"2016-01-24 23:15:22 +0000"`` (quotes included).
+
+.. django-admin-option:: --until
+
+  .. versionadded:: 2.8.0
+
+  Only consider contributions until the specified date. Date must be in ISO
```
julen: Now that python-dateutil is in the mix, I believe users are not limited to these two date+time formats only, right?

tests/commands/contributors.py
```diff
@@ -6,56 +6,110 @@
...
+from pootle.core.utils.timezone import make_aware
+from pootle_app.management.commands.contributors import Command
+from pytest_pootle.fixtures.contributors import CONTRIBUTORS_WITH_EMAIL
+
+
+def test_contributors_get_aware_datetime():
+    """Get an aware datetime from a valid string."""
+    instance = Command()
+
+    # Test ISO 8601 datetime.
```
julen: Is this really useful? Parsing is delegated to the underlying library, which is [already tested](https://github.com/dateutil/dateutil/blob/master/dateutil/test/test_parser.py). If something is not (properly) tested there or is missing, improving upstream's test suite sounds like the way to go to me.

pootle/apps/pootle_app/management/commands/contributors.py
unho: Changed added.

docs/server/commands.rst
unho: Will rephrase.

tests/commands/contributors.py
unho: We are here testing that the helper doesn't fail when passing valid strings for the underlying library.

PR #4831, opened 2016-06-16
Fix: prevent no results flashing while editing last items in a list

### LINKED ISSUE #4626: No results intermittently flashes while translating
When translating through units, at some point in time (I assume when performing some async fetch) the _No results_ message flashes for no reason to users. The following recording showcases the issue:

editor_bugs

phlax:
fixes #4626


unho:
I don't have a way to test. Change looks good, so lgtm.

PR #4834, opened 2016-06-16
Fix: remove vfolder listing from dropdowns

phlax:
im not sure this is used anywhere other than in dropdowns where it shouldnt appear


unho:
What about https://github.com/translate/pootle/blob/master/pootle/apps/pootle_project/models.py#L574-L592 ?

phlax:
just checking coveralls now...

unho:
lgtm

PR #4838, opened 2016-06-16
Add Pootle signals

phlax:
Add signals for key events in Pootle


unho:
@phlax Which is the status of this one?

PR #4851, opened 2016-06-22
Proper localize and display

unho:


pootle/templates/admin/xhr_paid_tasks_report.html
```diff
@@ -18,13 +18,13 @@
       <%
         var amount = row.amount;
```
dwaynebailey: This shouldn't be set, its overwritten later, similar for the one later on.

pootle/templates/admin/xhr_summary_report.html
```diff
@@ -12,10 +12,10 @@
     </thead>
     <tbody>
     <% _.each(data.summary, function (row) { %>
-      <tr class="item">
+      <tr class="item is-visible">
```
dwaynebailey: I've no idea why this commit was needed.  The commit message is unclear.

pootle/templates/admin/xhr_paid_tasks_report.html
unho: It sets a default, in case the following if conditions are not matched. The same goes for the latter one.

pootle/templates/admin/xhr_summary_report.html
unho: I will rewrite the commit message, but let me explain here. When I tried to test if the `ngettext` calls worked with my changes I was unable to see the texts rendered because they are in a table with rows that use the `item` class. This class hides the row by default: https://github.com/translate/pootle/blob/be4aa70ca1545a8e3f38e0e1b28a3082b8b71062/pootle/static/css/style.css#L756-L764

pootle/templates/admin/xhr_paid_tasks_report.html
dwaynebailey: OK, but s that really what was intended?  Sounds more like it fits in with the mistake of using concatenation.  I'd rather see an else clause to handle that case more intentionally.

unho: Fair point.

dwaynebailey:
lgtm

PR #4854, opened 2016-06-28
Webpack fix

phlax:
- doesnt add bad paths to the resolver
- fails if there are errors

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ assets:
 	npm --version
 	node --version
 	cd ${JS_DIR} && \
+	npm cache clear && \
 	npm install && \
 	cd ${CWD}
 	python manage.py webpack --extra=--display-error-details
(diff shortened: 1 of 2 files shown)


unho:
lgtm

julen:
Clearing the cache is already there (1d5f3a3143edeb08098b3c920798bc422e153ebb), if you need it for assets you should run `make clean` before.

julen:
_s/exists/exits/_ in the commit message.

phlax:
@julen `make clean` in the unix world is for *after* running a build not before - the problem here is that the npm cache is polluted (afaict) from the system libs

julen:
`make clean` is something you would run after a build and before recompiling to clean build artifacts. Why are system libs polluting the cache in your env?

phlax:
> Why are system libs polluting the cache in your env?

its not so much in my env, its in a clean install on debian (using docker).
It seems as though when you install nodejs and npm packages it pulls in a load of npm packages, and without running `npm cache clear` prior to `make assets` it tries to use the system libs and fails.
Im not entirely clear of the exact cause, but certainly this fixes the problem.

PR #4861, opened 2016-06-30
Set versions dynamically in install docs

phlax:
uses parsed-literal in the affected code blocks which unfortunately means no syntax highlighting


phlax:
fixed in https://github.com/translate/pootle/commit/f97ccb44e93a5d76be55a1cbf15784ef7a60ae50

PR #4864, opened 2016-07-02
Fix: set submission times correctly in pytest_pootle.env

phlax:
this was previously setting the creation times on all submissions repeatedly - rather than just for the submissions for specific units.


unho:
lgtm

PR #4865, opened 2016-07-02
Contributors util

phlax:
Add a contributors util to retrieve per-user contributor counts optionally filtered by time

### DIFF
--- a/pootle/apps/pootle_statistics/utils.py
+++ b/pootle/apps/pootle_statistics/utils.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from collections import OrderedDict
+
+from django.contrib.auth import get_user_model
+from django.db.models import Count, Q
+from django.utils.functional import cached_property
+
+
+User = get_user_model()
+
+
+class Contributors(object):
+
+    def __init__(self, include_anon=False,
+                 project_codes=None, language_codes=None,
+                 since=None, until=None, sort_by="username"):
+        self.include_anon = include_anon
+        self.project_codes = project_codes
+        self.language_codes = language_codes
+        self.since = since
+        self.until = until
+        self.sort_by = sort_by
+
+    @property
+    def user_qs(self):
+        if self.include_anon:
+            return User.objects.exclude(username__in=["system", "default"])
+        return User.objects.hide_meta()
+
+    @property
+    def user_filters(self):
+        return Q(submission__gt=0)
+
+    @property
+    def site_filters(self):
+        q = Q()
+        tp_related = "submission__translation_project__"
+        if self.project_codes:
+            q = q & Q(
+                **{"%sproject__code__in" % tp_related: self.project_codes})
+        if self.language_codes:
+            q = q & Q(
+                **{"%slanguage__code__in" % tp_related: self.language_codes})
+        return q
+
+    @property
+    def time_filters(self):
+        q = Q()
+        if self.since is not None:
+            q = q & Q(submission__creation_time__gte=self.since)
+        if self.until is not None:
+            q = q & Q(submission__creation_time__lte=self.until)
+        return q
+
+    @property
+    def filters(self):
+        return self.user_filters & self.site_filters & self.time_filters
+
+    def __iter__(self):
+        for k in self.contributors:
+            yield k
+
+    def __getitem__(self, k):
+        return self.contributors[k]
+
+    @cached_property
+    def contributors(self):
+        qs = self.user_qs.filter(self.filters).annotate(
+            contributions=Count("submission"))
+        if self.sort_by == "contributions":
+            qs = qs.order_by("-contributions", "username")
+        else:
+            qs = qs.order_by("username")
+        return OrderedDict(
+            [(user["username"], user)
+             for user
+             in qs.values("username", "contributions", "email")])
--- a/tests/statistics/stats_utils.py
+++ b/tests/statistics/stats_utils.py
@@ -1,0 +1,204 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from datetime import timedelta
+
+from collections import OrderedDict
+
+import pytest
+
+from pytest_pootle.factories import UserFactory
+
+from django.contrib.auth import get_user_model
+from django.db.models import Q
+
+from pootle.core.delegate import contributors
+from pootle_statistics.models import Submission
+from pootle_statistics.utils import Contributors
+
+
+User = get_user_model()
+
+
+def _contributors_list(contribs):
+    subs = []
+    for user in contribs.user_qs:
+        q = Q(submitter=user)
+        if contribs.project_codes:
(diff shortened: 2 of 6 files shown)


pootle/apps/pootle_statistics/utils.py
```diff
@@ -0,0 +1,80 @@
...
+from django.contrib.auth import get_user_model
+from django.utils.functional import cached_property
+
+
+User = get_user_model()
+
+
+class Contributors(object):
+
+    def __init__(self, include_anon=False,
```
unho: `include_anonymous` is more explanatory.

phlax: also more typing - i think anon is pretty explicit

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    assert anon not in contribs.user_qs
+    assert (
+        sorted(contribs)
+        == sorted(contribs.contributors)
+        == sorted(
+            set(contribs.user_qs.filter(contribs.user_filters)
+                                .values_list("username", flat=True)))
+        == sorted(
+            set(contribs.user_qs.filter(submission__gt=0)
+                                .values_list("username", flat=True))))
```
unho: Can't this be split into separate assert calls? Also perhaps the long queries can be stored in variables, just for ease reading.

pootle/apps/pootle_statistics/utils.py
unho: I disagree.

pootle/apps/pootle_statistics/utils.py
```diff
@@ -0,0 +1,80 @@
...
+    def time_filters(self):
+        q = Q()
+        if self.since is not None:
+            q = q & Q(submission__creation_time__gte=self.since)
+        if self.until is not None:
+            q = q & Q(submission__creation_time__lte=self.until)
+        return q
+
+    @property
+    def query(self):
```
unho: Perhaps better `all_filters`?

pootle/apps/pootle_statistics/utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, include_anon=False,
+                 project_codes=None, language_codes=None,
+                 since=None, until=None, sort_by=None):
+        self.include_anon = include_anon
+        self.project_codes = project_codes
+        self.language_codes = language_codes
+        self.since = since
+        self.until = until
+        self.sort_by = sort_by
```
unho: `sort_by` should default to `"name"`.

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+from pootle_statistics.utils import Contributors
+from pootle_store.models import Store
+
+
+User = get_user_model()
+
+
+def _contributors_list(contribs,
+                       projects=None, languages=None,
+                       since=None, until=None):
```
unho: No sort_by?

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+    subs = []
+    for user in contribs.user_qs.distinct():
+        q = Q(submitter=user)
+        if projects:
+            q = q & Q(
+                translation_project__project__code__in=projects)
```
unho: Where is `submission__`?

unho: nvm

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        if languages:
+            q = q & Q(
+                translation_project__language__code__in=languages)
+        if since or until:
+            _q = Q()
```
unho: Is it necessary to use `_q` when `q` is already available?

tests/statistics/stats_utils.py
phlax: they all equal the same thing - i think the test is more explicit this way

tests/statistics/stats_utils.py
phlax: its not tested yet

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+            if since:
+                _q = _q & Q(creation_time__gte=since)
+            if until:
+                _q = _q & Q(creation_time__lte=until)
+            q = q & _q
+        submissions = Submission.objects.filter(q)
```
unho: I would call `.count()` here, just to call once.

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        if submissions.count():
+            subs.append((user.username, submissions.count()))
+    return OrderedDict(sorted(subs, key=lambda i: (-i[1], i[0])))
```
unho: We do not necessarily want to sort like this. If sorting by name it would be only `key=lambda i: i[0]`.

phlax: again, sorting is not yet tested

tests/statistics/stats_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+User = get_user_model()
+
+
+def _contributors_list(contribs,
+                       projects=None, languages=None,
+                       since=None, until=None):
+    subs = []
+    for user in contribs.user_qs.distinct():
```
unho: `distinct()`?

tests/statistics/stats_utils.py
phlax: im not 100% sure if it necessary (i think it is - it seemed to be collapsing the query before - not sure) but it structures the query differently

tests/statistics/stats_utils.py
phlax: ??

unho: All users are distinct already, you get the same with `.all()`.

tests/statistics/stats_utils.py
unho: Feels hacky. At least I would like to see some short comment explaining this.

tests/statistics/stats_utils.py
phlax: not always - altho to be fair i have now changed user_qs

try doing this

```
subs = User.objects.filter(submission__gt=0)
print subs.count()
print subs.distinct().count()
```

pootle/apps/pootle_statistics/utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+User = get_user_model()
+
+
+class Contributors(object):
+
+    def __init__(self, include_anon=False,
+                 project_codes=None, language_codes=None,
+                 since=None, until=None, sort_by="username"):
```
unho: I don't feel it is a good idea to put a mutable type here as default. Better stick to `None` and do some checking before assigning.

pootle/apps/pootle_statistics/utils.py
```diff
@@ -0,0 +1,83 @@
...
+        self.project_codes = project_codes
+        self.language_codes = language_codes
+        self.since = since
+        self.until = until
+        self.sort_by = sort_by
+
+    @property
+    def user_qs(self):
+        if self.include_anon:
+            return User.objects.exclude(username__in=["system", "default"])
```
unho: I have the feeling this should be moved to `UserManager`.

pootle/apps/pootle_statistics/utils.py
phlax: ??

you told me to put it here - besides strings **immutable**

pootle/apps/pootle_statistics/utils.py
phlax: no - this is relevant to this class - unless we have a need for this pattern elsewhere im going to keep it here for now

unho:
gtm once tests for no contribs in range are added, and Travis goes green.

PR #4875, opened 2016-07-06
Editor: avoid event propagation upon ctrl+enter

### LINKED ISSUE #4874: Popup confirmation always shows up with keyboard submission
Steps to reproduce:
1. Go to a unit.
2. Alter the target text.
3. Submit via keyboard (<kbd>Ctrl+Enter</kbd>)

Observed results:
* The confirmation popup is displayed, the submission is performed and the editor advances to the next unit.

Expected results:
* Submission is performed without confirmation dialog and editor advances to the next unit.

From what I can tell, this only happens in Chrome and Safari.

julen:
Without this, Safari and Chrome propagate the change to the textarea.

Fixes #4874.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -309,7 +309,8 @@ PTL.editor = {
         this.closeSuggestion();
       }
     });
-    hotkeys.bind('ctrl+return', () => {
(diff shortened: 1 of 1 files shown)


unho:
lgtm

Issue #4872, opened 2016-07-06
Lodash dependency may be wrong in npm requirements


phlax:
Building static resources throws an error without explicitly installing lodash before running `npm install .`

julen:
I'm unable to trigger any error, you'll need to provide more detailed information @phlax and at least share the error. By the way, `npm install` doesn't need any extra parameters such as `.`.

phlax:
the build recipe is here

https://github.com/translate/translate-dockerfiles/blob/master/2.8-build/Dockerfile

The current workaround is these 2 lines:

https://github.com/translate/translate-dockerfiles/blob/master/2.8-build/Dockerfile#L29:L30

Removing triggers this error

```
npm --version
3.10.5
node --version
v0.12.15
cd /home/pootle/pootle_env/src/pootle/pootle/static/js && \
npm cache clear && \
npm install && \
cd /home/pootle/pootle_env/src/pootle
pootle@0.0.0 /home/pootle/pootle_env/src/pootle/pootle/static/js
+-- autolinker@0.11.2 
...snip...
    `-- source-map@0.4.4 

npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.12
python manage.py webpack --extra=--display-error-details
2016-07-08 06:54:30,479 INFO Loading custom settings from '/home/pootle/.pootle/pootle.conf'...
2016-07-08 06:54:31,014 INFO Using Python PO
ModuleNotFoundError: Module not found: Error: Cannot resolve module 'lodash/isPlainObject' in /home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/redux/lib
    at /home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/webpack/lib/Compilation.js:229:38
    at onDoneResolving (/home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/webpack/lib/NormalModuleFactory.js:29:20)
    at /home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/webpack/lib/NormalModuleFactory.js:85:20
    at /home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/async/lib/async.js:726:13
    at /home/pootle/pootle_env/src/pootle/pootle/static/js/node_modules/async/lib/async.js:52:16
[...]

PR #4878, opened 2016-07-07
Webpack: rewrite how manifest entry points are merged

julen:
The new code is simple enough and it doesn't rely on lodash, which allows us to
drop it from the `devDevpendencies` section of npm's package.json.


phlax:
yep, great - ill check it now

phlax:
hi @julen - just to confirm this PR fixes the issue raised in #4872

julen:
Cool I'll merge then.

PR #4882, opened 2016-07-11
Fixes to make it possible to create production Docker image

### LINKED ISSUE #4881: Make `pootle webpack` not require system checks
`pootle webpack` fails if eg the db is not set up/correctly. It would be helpful if it didnt

phlax:
fixes #4881

### DIFF
--- a/pootle/apps/pootle_app/management/commands/webpack.py
+++ b/pootle/apps/pootle_app/management/commands/webpack.py
@@ -19,6 +19,7 @@ from pootle_misc.baseurl import l
 
 class Command(BaseCommand):
     help = 'Builds and bundles static assets using webpack'
+    requires_system_checks = False
 
     def add_arguments(self, parser):
         parser.add_argument(


unho:
lgtm

ta2-1:
@phlax could you clarify which commit fixes your issue with running `webpack`?
And I mentioned on channel I'd prefer to skip checks that cause problem and leave the rest (for example: django version check). Then we could reuse this `SkipChecksMixin` from https://github.com/translate/pootle/pull/4576.
If not then probably can get rid of `SkipChecksMixin` and use `requires_system_checks`.

phlax:
for defo - lets use requires_system_checks - apart from being documented and explicit, anything with the word `Mixin` raises a :red_flag: for me

phlax:
re keeping some checks - im not sure if its possible - not 100% tho.

the reason being that if you only skip some checks - you can only skip warnings not critical errors.

phlax:
so @ta2-1 re which commits are necessary.

i dont think this commit https://github.com/translate/pootle/pull/4882/commits/2a1103a93b71db3d8268fa02e4b4d54c3f600f06 is actually required to fix the issue - but tbh i think its an improvement anyway

julen:
6118c2b will break `make assets` for deployments which run directly from source (this is our case).

GTM for 8b5c00c.

Can you also please put these separately in the future? It'd make life easier for anyone reviewing (I was also confused like @ta2-1 not knowing what the other commits were about with respect to the PR).

phlax:
> 6118c2b will break make assets for deployments which run directly from source (this is our case).

this is our case too, and actually this is situation that this commit is fixing.

@julen can you give some more information about what breaks?

this PR is pretty critical as its impossible to build a production docker server without it.

julen:
>> 6118c2b will break make assets for deployments which run directly from source (this is our case).
> 
> this is our case too, and actually this is situation that this commit is fixing.

If you run Pootle directly from the git sources without any pip installation, the `pootle` command is not available as it's an entry point created by setuptools.

Issue #4817, opened 2016-06-15
Reposition Last updated column in stats table


unho:
Some users have reported that now that the **Need Translation** column has been moved to the right, to the left on RTL layouts (see issue https://github.com/translate/pootle/issues/4630), when using small screens (1024px wide) they have to scroll horizontally some times (when some name in first column is long). Perhaps the best way to deal with this is to move the **Last updated** column further right (on LTR) so they can quickly see links to **Need Translation**.

Another possible option is to move the **Total** column further right, instead of **Last updated** column. The upside is that doing this we can align with the order for the action links, where total count is the last one (green lozenge). Note that on https://github.com/translate/pootle/issues/4630 there was some pushback to this proposal.

milupo:
Hi,
I made a screenshot, it is for Upper Sorbian but Pootle interface is in German here. Top right you see a red rectangle - it contains the beginning of the German column caption "Zu übersetzen" (To be translated) of the Need Translation column. This problem does not occur on every Pootle page but here on "devtools/client" it does occur.

![pootle_to_be_translated_column](https://cloud.githubusercontent.com/assets/1248520/16076974/532f6f70-32f6-11e6-8193-280c3445f726.png)

ta2-1:
@unho probably we can make the reposition but we haven't thought about long file names in general.
And the problem will remain for such long filenames.
![image](https://cloud.githubusercontent.com/assets/682845/16078340/16435b92-3307-11e6-96ba-4714a734e8a0.png)
We can try to limit a visible part of filename.

phlax:
can we limit the length of the filenames and append an elipsis where necessary, and then when you hover it could show the full filename

unho:
And display full filename as popup when hovering? Might be an option. Anyway that requires setting a char limit to trim filename, and also it doesn't actually fix the underlying issue which is that some columns are not visible straight away when using small resolutions, something that we will surely hit badly on mobile https://github.com/translate/pootle/issues/4209.

phlax:
more just a tooltip than a popup

phlax:
@unho dont try to plan for mobile in fixing this - a responsive design will handle this totally differently

dwaynebailey:
I suspect with limited horizontal space we'll likely need to do something a bit more innovative.  We could collapse the three columns into one.  And we display the top priority first (Red, Yellow, Blue).

milupo:
[...]

PR #4883, opened 2016-07-12
Truncate names on browse view

unho:
A tooltip is used to display the full name if truncated.

Part of #4817.

### DIFF
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -1139,6 +1139,21 @@ td.stats-name
 {
     white-space: nowrap;
     line-height: 20px;
+    width: 100%;
+    min-width: 10em;
+}
+
+.truncate-ellipsis {
+    display: table;
+    table-layout: fixed;
+    width: 100%;
+    white-space: nowrap;
+}
+ 
+.truncate-ellipsis > * {
+    display: table-cell;
+    overflow: hidden;
+    text-overflow: ellipsis;
 }
 
 td.stats-number
--- a/pootle/templates/browser/_table.html
+++ b/pootle/templates/browser/_table.html
@@ -20,9 +20,13 @@
 
       {% if 'name' in table.fields %}
       <td class="stats-name {{ item.icon }}"{% if item.description %} title="{{ item.description|striptags }}"{% endif %}>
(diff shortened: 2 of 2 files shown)


julen:
I think I agree with the sentiment that @phlax shared on [yesterday's call](https://github.com/translate/core-notes/blob/master/2016-07/july-13.md): doing this server-side doesn't feel ideal for various reasons, including users resizing their browser window/viewports.

So it's probably worth experimenting with a CSS solution using `text-overflow: ellipsis` ([here's an example](http://collaboradev.com/2015/03/28/responsive-css-truncate-and-ellipsis/)), this way the entirety of the text would still be available if the browser viewport grows.

dwaynebailey:
GTM

Tested on staging server with responsive design tools. Set viewport to 1024 wide. The elispsis works as expected.

iafan:
Hi folks. I don't think this got a proper amount of testing before being merged. This is what we see with this commit:

bad_layout

This is not what we'd expect, so I'd suggest reverting this PR and finding a solution that works properly for higher resolutions, or fixing this ASAP so that it doesn't block the master.

Ellipsis behavior needs to kick in only on smaller resolutions where the first column content would be trimmed. Otherwise, on higher resolution, things need to stay the same — there's no reason to push all columns to the right (which looks bad the higher resolution you have).

dwaynebailey:
@iafan could you propose a CSS solution, I don't think anyone else has such a wide screen.  Depending how you define ASAP I'd suggest you maybe land the solution or revert.

iafan:
Julen has a smaller screen — mine is even bigger. )

We'll talk with Julen on how we can address this. A quick fix would be to only allow the new behavior on screen sizes below the certain threshold.

But to avoid this kind of issues in the future, I and Julen would like to be able to see and formally approve any UI-related changes before they are merged into master. I think we already have a good habit of doing this with Taras, where I can provide my input during multiple stages of the feature development, and generally end up with a bit more polished UI. Would love to have the same kind of iterations with Leandro and whoever else is working with UI-related stuff these days.

dwaynebailey:
@iafan lets keep process discussion where they belong in the channel or in the weekly triage.  I'm sure you'd be welcome to bring it up and discuss it there.

Issue #3367, opened 2014-11-20
Review how DJANGO_SETTINGS_MODULE is set


julen:
The `DJANGO_SETTINGS_MODULE` environment variable indicates what's the entry point to read Django settings. It appears that it's being set like everywhere:
1. `manage.py` already sets it to `pootle.settings`.
2. The `pootle` runner which is used for build installs already sets it to `pootle.settings` too.
3. Every single management command sets it to `pootle.settings`.

The last one looks superfluous to me, so let's look at cleaning it up. If `django-admin.py` was used to run management commands this would probably break, but I don't see `django-admin.py` being suggested for Pootle, plus there's the `pootle` command which already takes care of this too.

PR #4888, opened 2016-07-13
Make pootle buildable for production Docker

phlax:
Fix for making assets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,11 @@ SPRITE_DIR = ${IMAGES_DIR}/sprite
 FORMATS=--formats=bztar
 TEST_ENV_NAME = pootle_test_env
 
+POOTLE_CMD = $(shell sh -c "command -v pootle")
+ifeq ($(POOTLE_CMD),)
+	POOTLE_CMD=python manage.py
+endif
+
 .PHONY: all build clean sprite test pot help docs assets
 
 all: help
@@ -24,10 +29,12 @@ assets:
 	npm cache clear && \
 	npm install && \
 	cd ${CWD}
-	python manage.py webpack --extra=--display-error-details
+	${POOTLE_CMD} webpack --extra=--display-error-details
 	mkdir -p ${ASSETS_DIR}
-	python manage.py collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}
-	python manage.py assets build ${TAIL}
+
+	${POOTLE_CMD} collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}
+	${POOTLE_CMD} assets build ${TAIL}
+
 	chmod 664 ${ASSETS_DIR}.webassets-cache/*
 
 travis-assets:
@@ -39,10 +46,10 @@ travis-assets:
 		cd ${JS_DIR} && \
 		npm install && \
 		cd ${CWD}; \
-		python manage.py webpack --dev --nowatch; \
+		${POOTLE_CMD} webpack --dev --nowatch; \
 		mkdir -p ${ASSETS_DIR}; \
-		python manage.py collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}; \
-		python manage.py assets build ${TAIL}; \
+		${POOTLE_CMD} collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}; \
+		${POOTLE_CMD} assets build ${TAIL}; \
 		chmod 664 ${ASSETS_DIR}.webassets-cache/*; \
 	fi
 


julen:
As said, this won't work in deployments from source without installation. You should check for `which pootle` and fall back to `python manage.py` to ensure nothing is broken.

Also, what does 48823af fix? It'd definitely help to have an explanation in the commit message.

phlax:
> Also, what does 48823af fix? It'd definitely help to have an explanation in the commit message.

not sure that it fixes anything - i can happily remove - but it seems like the changed line might not actually be necessary tho

phlax:
> As said, this won't work in deployments from source without installation.

im struggling to understand why you would deploy without installation. To my mind this would open up the possibility of many other problems, and isnt something we should ordinarily support.

I might be wrong tho - can you explain why you would want to do this

unho:
> > Also, what does 48823af fix? It'd definitely help to have an explanation in the commit message.
> 
> not sure that it fixes anything - i can happily remove - but it seems like the changed line might not actually be necessary tho

I would keep it, but renaming the commit message to "cleanup".

Regarding deploy without installation, we do `pip -e` despite it is something I find disturbing because we can easily forget about upgrading when necessary.

unho:
Regarding replacing `manage.py` I wonder why it is necessary for build? Can't we just copy and adjust the required instructions to the docker script?

phlax:
its necessary because using manage.py doesnt seem to set the correct paths afaict.

unho:
> its necessary because using manage.py doesnt seem to set the correct paths afaict.

No it doesn't, that is why in our build instructions (the ones in the docs) we first set up the paths to find the right settings and so. But this is something we can put in the docker script, even using `pootle` command, isn't it?

julen:
We have had this installation for years, and it was [the only way back then to run commands](https://github.com/translate/pootle/blob/f0aabff8b988189aea20221e9ab6f509bdd5888b/docs/server/commands.rst#running-from-checkout) when deploying from source. Deploying means rebuilding assets, so any command/helper necessary to deal with it must work, including `make assets`. The change proposed here breaks this. There might be instances out there still deployed this way too, but it's hard to tell.

While I agree that having a single installation path would be simpler, I don't want to break lightly the ability to upgrade existing deployments from source. At the same time, our installation works flawlessly and even though we could eventually switch to using `pip install .`, there's no actual need for doing so (_if it ain't broke, don't fix it_).

Having said that, I believe you can support both scenarios easily by implementing the suggestion from my previous message.

julen:
Regarding 48823af I'd suggest to please leave it out, as it's uncertain the change has any clear purpose. Related to this and in terms of setting `pootle.settings`, there's #3367 already open to review how it is set across commands.

Makefile
```diff
@@ -24,10 +24,15 @@ assets:
 	npm cache clear && \
 	npm install && \
 	cd ${CWD}
-	python manage.py webpack --extra=--display-error-details
+	pootle webpack --extra=--display-error-details
```
dwaynebailey: Surely we want something more like:

`POOTLE_CMD = $(shell command -v pootle)`

With some fallback to `python manag.py`

Then it all becomes slightly more readable i.e. `$(POOTLE_CMD) webpack...`

phlax: yep, k

also, i think the logic is the wrong way round

Makefile
```diff
@@ -10,6 +10,11 @@ SPRITE_DIR = ${IMAGES_DIR}/sprite
 FORMATS=--formats=bztar
 TEST_ENV_NAME = pootle_test_env
 
+POOTLE_CMD = $(shell sh -c "command -v pootle")
+ifeq ($(POOTLE_CMD),)
+	POOTLE_CMD="python manage.py"
```
dwaynebailey: This shouldn't have quotes so this works:

```make
POOTLE_CMD=python manage.py
```

Makefile
```diff
@@ -24,10 +29,12 @@ assets:
...
 	mkdir -p ${ASSETS_DIR}
-	python manage.py collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}
-	python manage.py assets build ${TAIL}
+
+	${POOTLE_CMD} collectstatic --noinput --clear -i node_modules -i .tox -i docs ${TAIL}
+	${POOTLE_CMD} assets build ${TAIL}
+
 	chmod 664 ${ASSETS_DIR}.webassets-cache/*
 
 travis-assets:
```
dwaynebailey: I'd suggest we change occurances of `python manage.py` here also, to be consistant and not leave vestiges of an approach we'd like to deprecate.

dwaynebailey:
gtm

Tested setup as follows:

1. Using PR checkout
2. Built tmp venv with `pip install -r requirements/base.txt`
3. `make assets` <- worked fine and is using `python manage.py`
4. `pip install -e .` to make sure `pootle` command is in the path
5. `make assets` <- works fine and is using `pootle` command

PR #4890, opened 2016-07-13
User.top_language: fix condition

### LINKED ISSUE #4889: User contributions for their top language don't work
User contributions for their top language always displays _No contributions in the last 30 days_ in their profile page.

This is likely related to some recent changes on how contributions are retrieved.

julen:
Since ffe4deb5 `User.top_scorers()` doesn't return a list of users anymore, but
a list of annotated dictionaries. Therefore, the actual user is now under the
'user' key of these dictionaries.

Fixes #4889.


unho:
lgtm

PR #4892, opened 2016-07-13
Update commands.rst

multiwebinc:


dwaynebailey:
@multiwebinc the docs on [templates](http://docs.translatehouse.org/projects/pootle/en/latest/features/templates.html) was written to try to better explain the use of templates.  I think it may be a better idea to point to those rather than confuse users with a partial picture of the story.

dwaynebailey:
@multiwebinc thanks for this.

This has been landed as b03cb5ded207db89950bb4c10274f1a04066114b with commit message change. And follow up 81dcefa956dd00c0933fd1cff3105b4ce5514a05 referring people to template docs for the full story.

Issue #4848, opened 2016-06-21
Fix RTL layout for top stats panel and user avatars


ta2-1:
![image](https://cloud.githubusercontent.com/assets/682845/16221728/06d48e66-379d-11e6-950f-6ed2bbcb3def.png)
This screenshot was taken from https://mozilla.locamotion.org/en_ZA/firefox/.

SafaAlfulaij:
I did some fixes but not all of them.
TODO:
- Detailed statistics
- Checks warnings
- Others...

_Done using Stylish_
```
  .breadcrumb li, .path-summary-more .top-scorers .top-scorer, .path-summary-more .top-scorers .user, .path-summary-more .top-scorers .number, .path-summary-more .top-scorers .place, .path-summary-more .path-summary-collapsed label {
    float: right;
  }
  .path-summary-more .path-summary-collapsed {
    margin-left: auto;
    margin-right: -10px;
  }
  .path-summary-more .top-scorers .top-scorer, .path-summary-more .top-scorers .user, .path-summary-more .top-scorers .number, .path-summary-more .top-scorers .place, .path-summary-more .path-summary-collapsed label {
    padding-right: auto;
    padding-left: 3px;
  }
  .top-scorer .user-name {
    margin-left: auto;
    margin-right: 24px;
  }
  .expand-stats {
    float: left;
  }
  #translate-actions li {
    margin-right: 2em;
    margin-left: -0.5em;
  }
  .lightbox-body {
    text-align: right;
  }
  .last-action .user-name {
    margin-left: auto;
    margin-right: 24px;
  }
  
  #detailed {
    text-align: left;
  }
```

unho:
@julen I haven't been able to correctly align the icon besides the project names in the overview stats table. If I am correct the direction is injected through https://github.com/translate/pootle/blob/14b6d628/pootle/static/js/stats.js#L119-L122 Can please you shed some light on this?

Issue #4121, opened 2015-10-01
Placeholders in "Last Activity" appear instead of real values


SafaAlfulaij:
In Mozilla, I found this:
![Last Activity placeholders](http://i.imgur.com/XJsI2I2.png)

unho:
This might be related to 3b74a95b.

julen:
Thanks for the pointer @unho. That commit indeed introduced a regression: it removed template strings (those wrapped around backticks), which do parameter interpolation for free. If using regular strings, the placeholder replacement has to be done via the cumbersome `interpolate()` function.

I'd strongly favor fixing the string extraction methods to be able to seamlessly use template strings.

dwaynebailey:
We're in a pickle here then as I can't find anything to extract ECMAScript template strings.  Not even any custom xgettext extractors.

You don't seem to have any control in xgettext over the quote marks.  But I managed to get the extractor to work if the string is ``gettext(`"Translatable text ${var}"`)`` @julen does that work in code?

julen:
No it doesn't, that'd be the same as `'"foo bar"'` or `'"foo bar"'`, i.e. quotes end up being part of the actual string, which is not desired.

dwaynebailey:
@julen yeah just tested here http://mozilla.locamotion.org/user/khaled/ and I get the stray quotes.

Do you have any pointers to an extraction tool?

dwaynebailey:
Deployed a workaround on mlo but reverting 3b74a95

This needs ``interpolate`` or a proper extraction tool for templates.  Or someone to figure out how xgettext can be made to see backtick surrounded strings as extractable strings.

dwaynebailey:
@julen, nope its still a messup.  I'm afraid you can't escape needing to interpolate.  The problem is that the template is being interpolated before being called by gettext(). So the variables are already filled in and gettext then can't find the translation. 

As a hack I tried `gettext('some ${var}')`

dwaynebailey:
Right done a bit of reading.  I think we'll need to build a template string tag function.  These seem to work as such.

```js
l10n(string, ...values) {
   // Do stuff here to actually use gettext
   // Tehn interpolate
   return $string // Lastly return the now localised and interpolated string
}

l10n`Some ${var}`
```

dwaynebailey:
@julen are you able to help in terms of getting template string working in PO?  This feels a bit beyond my skills in this area.

dwaynebailey:
[...]

PR #4896, opened 2016-07-14
Fix RTL issues

unho:
For issue #4848

### DIFF
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -880,6 +880,12 @@ html[dir="rtl"] .last-action
     margin-left: 24px;
 }
 
+html[dir="rtl"] .last-action .user-name
+{
+    margin-left: auto;
+    margin-right: 24px;
+}
+
 .last-action .extra-item-meta
 {
     color: #999;
@@ -1592,6 +1598,11 @@ html[dir="rtl"] ul.pager li:last-child > *
     height: 0;
 }
 
+html[dir="rtl"] .expand-stats
+{
+    float: left;
+}
+
 .expand-stats .icon-expand-stats,
 .expand-stats .icon-collapse-stats
 {
@@ -1621,6 +1632,12 @@ html[dir="rtl"] ul.pager li:last-child > *
     margin-left: -10px;
     padding: 10px 0 10px 0;
     height: 15px;
+}
+
+html[dir="rtl"] .path-summary-more .path-summary-collapsed
+{
+    margin-left: auto;
+    margin-right: -10px;
 }
 
 .path-summary-more.expand .path-summary-expanded
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -743,6 +743,12 @@ button.close
     padding-right: 3.7em;
 }
 
+html[dir="rtl"] #autorefresh-notice span
+{
+    padding: 0.9em 1.5em;
+    padding-left: 3.7em;
+}
+
 #autorefresh-notice .x
 {
     margin-left: 1em;
@@ -750,6 +756,12 @@ button.close
     position: absolute;
     top: -0.2em;
     color: rgba(0, 0, 0, 0.4);
+}
+
+html[dir="rtl"] #autorefresh-notice .x
+{
+    margin-left: auto;
+    margin-right: 1em;
 }
 
 #autorefresh-notice.dirty span
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -1191,6 +1191,15 @@ table.stats td.last-updated
     text-align: right;
 }
 
+html[dir="rtl"] table.stats .number,
+html[dir="rtl"] th.stats-number,
+html[dir="rtl"] td.stats-number,
+html[dir="rtl"] td.stats-percentage,
+html[dir="rtl"] table.stats td.last-updated
+{
+    text-align: left;
+}
+
 
 td.stats-number.need-translation .stats-data,
 td.stats-number.suggestions .stats-data,
@@ -1676,6 +1685,13 @@ html[dir="rtl"] .path-summary-more .path-summary-collapsed
 .summary-3-col
 {
     float: left;
+}
+
+html[dir="rtl"] .summary-1-col,
+html[dir="rtl"] .summary-2-col,
+html[dir="rtl"] .summary-3-col
+{
+    float: right;
 }
 
 .path-summary-more .summary-3-col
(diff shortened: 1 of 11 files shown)


pootle/static/css/style.css
```diff
@@ -1160,6 +1166,15 @@ table.stats td.last-updated
     text-align: right;
 }
 
+html[dir="rtl"] table.stats .number,
+html[dir="rtl"] th.stats-number,
+html[dir="rtl"] td.stats-number,
+html[dir="rtl"] td.stats-percentage,
+html[dir="rtl"] table.stats td.last-updated
```
SafaAlfulaij: I think it's better that "table.stats td.last-updated" is aligned to right, beacuse that column contains RTL text ..

SafaAlfulaij:
More fixing (sorry :P):
```
html[dir="rtl"] @media screen and (max-width: 1440px) .path-summary-more .summary-1-col
{
    margin: 0 0 0 6.25%;
}

html[dir="rtl"] @media screen and (max-width: 1440px) .path-summary-more .summary-2-col
{
    margin: 0 6.25% 0 0;
}

html[dir="rtl"] .path-summary-more .top-scorers-table tr td.number:last-child
{
    text-align: left;
}

html[dir="rtl"] .check-count
{
    text-align: left;
}

html[dir="rtl"] .last-action
```

Current attribute changes in "html[dir="rtl"] .check-count" should be removed and only apply an alignment change.
"html[dir="rtl"] .last-action" should be removed as it doesn't apply anymore (Old stuff from 2.5?), and it adds some unneeded space.

unho:
@SafaAlfulaij Forgot to say that there is a staging server for this PR http://staging.translatehouse.org:8002/ Thanks for the suggestions.

unho:
@SafaAlfulaij Created and account safa/safa for you in http://staging.translatehouse.org:8002/ and updated with your suggestions.

pootle/static/css/style.css
unho: It is aligned exactly the opposite of non-rtl. Perhaps I am misunderstanding you?

ltr
rtl

dwaynebailey:
Did a quick review.  Some things we might need to look at, but not blockers for this PR:

* Numbers in columns are left aligned, is that right? Or is it just my LTR head
* This alignment causes some issues are they don't make it easy to see what is which column in the browse table
* Expaned navbar view: the icon for hte user appears far left so you actually get the page scrolled left just for the icon
* I'm concerned the icon, name order in the collapsed navbar is not correct
* Score is N+ in LTR its +N.  Does that work in RTL?  Should '+%d' be localised?
* Expanded view.  Last update user icon overlaps with the link to what was translated

SafaAlfulaij:
> Numbers in columns are left aligned, is that right? Or is it just my LTR head

Maybe you are correct. 2.5 was displaying them right-aligned.
> This alignment causes some issues are they don't make it easy to see what is which column in the browse table

If we're going to correct the numbers, then it's safe to make them all right-aligned.
> Expaned navbar view: the icon for hte user appears far left so you actually get the page scrolled left just for the icon

Thanks for the notice! I tried but didn't know how to fix it. :(
> I'm concerned the icon, name order in the collapsed navbar is not correct

Seems like the one above..
> Score is N+ in LTR its +N. Does that work in RTL? Should '+%d' be localised?

I would say adding:
```
html[dir="rtl"] .path-summary-more .top-scorers-table tr td.number:last-child {dir="rtl"}
```
> Expanded view. Last update user icon overlaps with the link to what was translated

That's because the action text isn't translated, this bug: #4121


Now there are these:
screenshot_ _

The switch "button" to the left is really sticked with the "Suggest" button to the right. There should be some small space. Didn't know how to fix it. :(

screenshot_ _

There is some more space than needed to the right of the English text.
For a reason, Chromium is flagging `padding-right=auto` attrib of `"tml[dir="rtl"] .suggestion-original` as invalid. Removing the `padding-right` attrib from the ltr selector makes it right. Strange that it's using the float from the rtl selector not the ltr one.

screenshot_ _

The arrows should be mirrored/flipped. Both of them have the same css and that's beyond my knowledge :smile:
This is for both the simplified and detailed one.

The last thing (for now?) is that the detailed statistics page need to be RTL (main html doesn't have `dir="rtl"` attrib set.

pootle/static/css/style.css
unho: This is now sorted out.

unho:
>> - Numbers in columns are left aligned, is that right? Or is it just my LTR head
>
> Maybe you are correct. 2.5 was displaying them right-aligned.

This is just the opposite of what we have for LTR.


>> - This alignment causes some issues are they don't make it easy to see what is which column in the browse table
>
> If we're going to correct the numbers, then it's safe to make them all right-aligned.
>> - Expanded navbar view: the icon for the user appears far left so you actually get the page scrolled left just for the icon
>
> Thanks for the notice! I tried but didn't know how to fix it. :(

I didn't saw this. Can you please share a screenshot?

Perhaps related, I fixed the separation between top contributors when details are collapsed.

>> - I'm concerned the icon, name order in the collapsed navbar is not correct
>
> Seems like the one above..

Not sure what are you talking about either.


>> - Score is N+ in LTR its +N. Does that work in RTL? Should '+%d' be localised?
>
> I would say adding:
> ```
> html[dir="rtl"] .path-summary-more .top-scorers-table tr td.number:last-child {dir="rtl"}
> ```
>> - Expanded view. Last update user icon overlaps with the link to what was translated
>
> That's because the action text isn't translated, this bug: #4121

Exactly, with correct localization this is gone.


> Now there are these:
> screenshot_ _
>
> The switch "button" to the left is really sticked with the "Suggest" button to the right. There should be some small space. Didn't know how to fix it. :(

Fixed.

> screenshot_ _
>
> There is some more space than needed to the right of the English text.
> For a reason, Chromium is flagging `padding-right=auto` attrib of `"tml[dir="rtl"] .suggestion-original` as invalid. Removing the `padding-right` attrib from the ltr selector makes it right. Strange that it's using the float from the rtl selector not the ltr one.

Fixed.

> screenshot_ _
>
> The arrows should be mirrored/flipped. Both of them have the same css and that's beyond my knowledge :smile:
> This is for both the simplified and detailed one.
>
> The last thing (for now?) is that the detailed statistics page need to be RTL (main html doesn't have `dir="rtl"` attrib set.

unho:
@SafaAlfulaij We are landing the changes in this PR and closing it. I will open a new PR to catch up with the remaining issues, or any new one you might have found. Some of the latest issues you reported are fixed, but other issues were not.

unho:
@SafaAlfulaij Landed the fixes in this PR. These fixes don't cover all the issues you reported, so created #4925 to track them. Feel free to comment or report more RTL problems there.

PR #4898, opened 2016-07-14
virtualfolder: get rid of import warnings

### LINKED ISSUE #4442: Import warning when virtualfolder is not in INSTALLED_APPS
Since recently we started to get some warnings related to virtual folders, even if we explicitly make no use of them:
```python
RemovedInDjango19Warning: Model class virtualfolder.models.VirtualFolder doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class VirtualFolder(models.Model):

RemovedInDjango19Warning: Model class virtualfolder.models.VirtualFolderTreeItem doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class VirtualFolderTreeItem(models.Model, CachedTreeItem):
```

If I'm not mistaken these are reported since `virtualfolder.models` is being imported regardless of `virtualfolder` being in `INSTALLED_APPS`, so Django complains about it in this case.

While adding the suggested `app_label` silences the warning, I'm not 100% sure this is the right way to proceed.

julen:
By delaying the model imports from the virtualfolder app only to where they are
needed, we avoid getting warnings when the virtualfolder app is not in Django's
`INSTALLED_APPS`.

Fixes #4442.

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -30,8 +30,6 @@ from pootle_misc.util import get_date_interval
 from pootle_project.models import Project
 from pootle_statistics.models import (Submission, SubmissionFields,
                                       SubmissionTypes)
-from virtualfolder.helpers import extract_vfolder_from_path
-from virtualfolder.models import VirtualFolderTreeItem
 
 from .fields import to_db
 from .form_fields import (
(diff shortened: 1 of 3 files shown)


unho:
lgtm

julen:
@vakbot yes, totally agree on that one. This is an interim workaround which you should treat as a trade-off though, please read #4442 for further details. Thanks for your feedback!

PR #4905, opened 2016-07-16
[WIP] Localize strings in user statistics

unho:
This was to make localizable all strings in user statistics. The most visible part has been already landed. The remaining part is conflicting (despite it works) because it involves dropping a whole JS package from Pootle codebase. Put on hold until I have time to revisit.


unho:
@julen Not sure if this is the way it is supposed to get the js library from npm, so I will appreciate if you check it out.

julen:
Thank you @unho. I checked the code an I fail to see the relevance of moment.js in order to make user statistics' strings localizable — the last commit is all what's needed (3522993).

unho:
> Thank you @unho. I checked the code an I fail to see the relevance of moment.js in order to make user statistics' strings localizable — the last commit is all what's needed (3522993).

I will try to explain the changes by using the screenshoot from the issue this PR addresses:

Statistics

As you can see in the image above there are two place where strings should be localizable:

- In the filtering for the TPs with contributions table
- Below the contributions graph

The two `moment` commits are related to the first one, while the last commit you pointed to (3522993) addresses the latter.

The reasons to drop `moment` are:

- As the TODO comment says this should be offloaded from Pootle
- I wasn't able to get the `moment` library localization used unless I dropped it, since we weren't picking the localizations.

Regarding the graph, I wasn't able to get any localization for the `flot` library, so we have to inject our own. This is what 3522993 commit does.

unho:
@julen are your concerns solved by my previous comment? If yes it would be nice to get some gtm.

julen:
Thanks for the clarification @unho, I see where it is used now.

Unfortunately, as a side-effect this is generating a big problem because now all locales are being bundled along with the library code, which is doubling the size of the reports bundle (up to 938kB, unminified and ungzip'ed). This is a huge price to pay in order to display a localized month name.

A possibility here is to perform the month formatting ourselves. Another possibility is to lazily load moment.js locales.

Part of me is also wary about lightly upgrading many moment.js versions — we hit some issues with it in the past and I'd rather carefully check what changed before proposing an upgrade.

I suggest you merge 3522993 and address the rest separately.

unho:
> I suggest you merge 3522993 and address the rest separately.

I am doing that right now. Thanks. I also WIPed the PR since the remaining changes are still not in a ready state.

unho:
I will have a look at the other issues in another moment. Right now this PR is in WIP state.

unho:
This view was removed so no longer applies.

PR #4914, opened 2016-07-17
Bump ttk requirement in Pootle 2.8

phlax:
wondering if we should push the newer version, we are likely to break compat v soon anyway with l20n parser im thinking

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -25,6 +25,6 @@ rq>=0.5.0,<=0.5.6
 scandir==1.2
 
 # Translate Toolkit
-translate-toolkit==1.13.0
+translate-toolkit==2.0.0b3
 # If you want to use Translate Toolkit 'master'
 #-e git://github.com/translate/translate.git#egg=translate-toolkit


unho:
If we want to stabilize both TTK and Pootle soon it is better done ASAP. I hope both are stable enough for not delaying final release, but we can still roll back if necessary.

julen:
My feeling is we should not hard-depend on beta software so if there's a need for a newer TTK, TTK should stabilize on a public stable version first.

> we are likely to break compat v soon anyway with l20n parser im thinking

What does this mean?

phlax:
> >    we are likely to break compat v soon anyway with l20n parser im thinking
> What does this mean?

It means we are adding features in ttk, that we want pootle to be able to take advantage of

dwaynebailey:
TTK has gone through a major change to make it Python 3 compatible. So I agree with Julen that we shouldn't make a final Pootle release depend on a pre-release ttk version.  But I think we should aim to release Pootle 2.8 beta's and rc's with the latest ttk to make sure that we've teased out any bugs.  We realy don't have much other choice in terms of stabalisation and testing of such a big change.  If ttk doesn't make stable before Pootle 2.8 release then we can keep things as they are.

phlax:
so related to @dwaynebailey's point that we want people to test ttk2 - whether they are installing the 2.8beta or installing from master. I would therefore propose that we land this just now - and if you are tracking master and really dont want ttk2 yet - you will need to adjust your requirements file

requirements/base.txt
```diff
@@ -25,6 +25,6 @@ rq>=0.5.0,<=0.5.6
 scandir==1.2
 
 # Translate Toolkit
-translate-toolkit==1.13.0
+translate-toolkit==2.0.0b2
```
unho: Beta 3 is out.

julen:
Ok so if I understood correctly this means you guys want to push this to beta users, and depending on the feedback received there, decide to move on with it for the final release or not. If that's the case, fine with me.

unho:
@julen Yes, that is the plan.

PR #4918, opened 2016-07-18
Editor: use wrapper for textarea

julen:
Since in the near-future more elements will need to be combined with the
textarea (live previews etc.), the wrapper provides a convenient way to host all
of these while providing a way to nicely style everything.

Implementation-wise, reacting to focus/blur events is needed because there's no
way to target parent elements in pure CSS (note this is the same solution we use
for the search box).

### DIFF
--- a/pootle/static/css/editor.css
+++ b/pootle/static/css/editor.css
@@ -593,8 +593,7 @@ td.translate-full div.translate-original
 td.translate-full textarea.translation
 {
     width: 100%;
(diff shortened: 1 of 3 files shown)


iafan:
Tested this on staging server and it definitely improves the way we handle complex layouts (see e.g. [units with many placeholders](https://translate.stage.evernote.com/it/web_evernote/translate/resources/StripesResources.properties.po#unit=6145619)). 

+1 from me.

Issue #4916, opened 2016-07-18
Editor leaks memory (again)


phlax:
Profiling the editor it seems there are again issues with mem leaks. In fact even without clicking (just wiggling the mouse) its possible to farm memory.

I havent as yet got to the root of the problem

phlax:
I believe the problem is to do with leaking nodes, perhaps in a react component.

I undertook the following test

- switched off tipsy and odometer
- reloaded
- garbage collected
- started profiling
- click back and forth between the same 2 units
- garbage collected
- click back and forth between the same 2 units
- garbage collected

This is a screenshot of the results

![mem-profile-same-two](https://cloud.githubusercontent.com/assets/454682/16915979/6a41bad6-4cf2-11e6-8ab1-f8f7e574aa34.png)

basically the nodecount increases by around 10 every time a unit is focused, and this seems to be what is leaking the memory

julen:
Depending on the version of autosize in use, jackmoore/autosize#270 can be related (the fix was released in version 3.0.14).

phlax:
apologies - i disabled autosize too in the above test.

I have also now disabled updateRelativeDates, so not much left running, but i still get a small increase in uncollected nodes on unit focus - and a ~significant increase in mem

iafan:
Just thinking out loud: if there was no such a memory use pattern before, probably `git bisect` could help? Also, it's worth checking this pattern against different browsers — this might be related to how GC works on a particular browser (and changes between browser versions) rather than to the changes in JavaScript code itself. I tried to profile JS heap on Chrome, did some navigation between units for ~50 times and didn't see any indication that the memory is leaking.

phlax:
I believe this if fixed - although the performance of the editor is still a bit of an issue - ive opened a related ticket https://github.com/translate/pootle/issues/4941

julen:
I re-ran a profile on current master (switching between 2 consecutive units and forcing GC) and this is what I get:
![screen shot 2016-07-22 at 10 49 40](https://cloud.githubusercontent.com/assets/16768/17051738/6dd5bffc-4ffa-11e6-83a4-ecc165e8e717.png)

Is it in line with other people's?

phlax:
if i gc - i return almost to baseline (in master) - with pr - i return to baseline

julen:
Okay here are a couple of updated screenshots because I was using a debug build and the screenshot was missing the important part that was auto-cropped by Chrome.

The same test in current master:
[...]

PR #4920, opened 2016-07-19
JS cleanups

phlax:
- Remove ajax objects from queue on completion

### DIFF
--- a/pootle/static/js/shared/utils/fetch.js
+++ b/pootle/static/js/shared/utils/fetch.js
@@ -29,7 +29,7 @@ function fetch({ url, body, method = 'GET', dataType = 'json', queue = null,
       data: body,
     })
   );
-
+  requests[queueName].done(() => {requests[queueName] = null;});
   return requests[queueName];
 }
 


pootle/static/js/shared/utils/fetch.js
```diff
@@ -29,7 +29,9 @@ function fetch({ url, body, method = 'GET', dataType = 'json', queue = null,
       data: body,
     })
   );
-
+  requests[queueName].done(function(data) {
```
julen: To conform with ESLint:
```js
requests[queueName].done(() => delete requests[queueName]);
```

phlax: thanks, ill update

pootle/static/js/shared/utils/fetch.js
```diff
@@ -29,7 +29,7 @@ function fetch({ url, body, method = 'GET', dataType = 'json', queue = null,
       data: body,
     })
   );
-
+  requests[queueName].done(() => delete requests[queueName]);
```
julen: Actually, [after a read on `delete`](http://bertanguven.com/preventing-memory-leaks-in-javascript-null-vs-delete), it seems like we'd prefer to nullify the value `requests[queueName] = null`.

phlax:
contributed to resolving #4916

PR #4922, opened 2016-07-19
Static jsi18n

phlax:
Add static jsi18n app and configure

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,6 +11,7 @@ django-overextends==0.4.1
 django-redis==4.4.3
 django-rq>=0.8.0,<=0.9.1
 django-transaction-hooks==0.2
+django-statici18n
 
 # Libraries
 cssmin==0.2.0
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ assets:
 	npm cache clear && \
 	npm install && \
 	cd ${CWD}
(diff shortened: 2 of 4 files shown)


requirements/base.txt
```diff
@@ -11,6 +11,7 @@ django-overextends==0.4.1
 django-redis==4.4.3
 django-rq>=0.8.0,<=0.9.1
 django-transaction-hooks==0.2
+django-statici18n
```
unho: No pinned version? Or at least a minimum version?

unho:
Does this PR finally allows to drop https://github.com/translate/pootle/blob/master/pootle/urls.py#L16-L19 ?

phlax:
@unho good catch - yes it will - when i get it working 8/

dwaynebailey:
Landed in df94bac0ce41875b618f2af10bb6b30394d90365..a60c627be285ca7f6cab545ff89188ff0e92cea3

julen:
Is there a reason why after running `compilejsi18n` the files are located at `STATIC_ROOT` and not at `STATICFILES_DIRS` as expected? Also, any ideas why running this with `USE_I18N = False` creates files for all languages?

julen:
As a heads-up, I have opened #4990 to fix the problems I got with this PR being merged to master.

PR #4927, opened 2016-07-19
Fix width of browsing table's name column

### LINKED ISSUE #4913: Name column is too wide
Since we added the truncating name column, the name column is now too wide.

name-too-wide

This problem is even worse if you have the sidebar expanded is it makes most of the information invisible

name-too-wide-expanded

julen:
Fixes #4913.


phlax:
tested, gtm, thanks for fixing this

PR #4930, opened 2016-07-20
[RFC] Add CSSLint

dwaynebailey:
@julen maybe more a question for you.

Would it make sense to add csslint? This PR just adds it to installs and makes it possible to run the command.  Ideally I'd want to enable in Travis but that would require deciding what rules we want to enforce and doing any needed cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,10 @@ jslint:
 	cd ${JS_DIR} \
 	&& npm run lint
 
+csslint:
+	cd ${JS_DIR} \
+	&& npm run csslint -- ../css/
+
 publish-pypi:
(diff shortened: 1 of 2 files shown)


julen:
Thanks for bringing this up Dwayne.

To be honest I'd prefer to rather go with [stylelint](http://stylelint.io/), which is parsed using [PostCSS](http://postcss.org/), a widely used tool for CSS-related transformations, and integrates nicely with npm/webpack. Another reason is that I can also see ourselves benefit from more of that machinery in the future (autoprefixer, CSS modules etc.)

dwaynebailey:
Closed in favour of #5141

PR #4931, opened 2016-07-20
Remove language sorting on select nodes

phlax:
Im not sure if removing this altogether breaks any expectations, but based on comment from @julen 

>  I'd prefer to only sort the data when rendering (and not some DOM structure)

which i agree with - ive PRed to remove this

### DIFF
--- a/pootle/static/js/common.js
+++ b/pootle/static/js/common.js
@@ -159,36 +159,5 @@ PTL.common = {
       const target = $(this).attr('href') || $(this).data('target');
       $(target).toggle();
     });
-
-    /* Sorts language names within select elements */
-    const ids = ['id_languages', 'id_alt_src_langs', '-language',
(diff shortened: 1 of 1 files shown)


unho:
Despite it is not a critical breakage since Select2 provides search, I expect the list of languages to be sorted. To illustrate, this is the result with this PR:

pr

Note that Spanish is not listed where it belongs since we are sorting by language code (apparently).

And this is what we get in master:

master

phlax:
> I expect the list of languages to be sorted. 

I would agree - but i think this is something that should happen on the server. Thanks for the example - ill take a look @unho

phlax:
this was done and resolved i believe

phlax:
reopening to check this is the case

phlax:
this code no longer exists, closing

PR #4937, opened 2016-07-21
Fix: prevent node leaking 

phlax:
This is a significant contributor to #4916

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -1341,15 +1341,13 @@ PTL.editor = {
 
   /* reDraws the translate table rows */
   reDraw(newTbody) {
-    const $oldRows = this.$editorBody.find('tr');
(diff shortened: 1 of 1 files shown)


pootle/static/js/editor/app.js
```diff
@@ -1341,15 +1341,13 @@ PTL.editor = {
 
   /* reDraws the translate table rows */
   reDraw(newTbody) {
-    const $oldRows = this.$editorBody.find('tr');
-
     // Remove autosize event listeners for textarea before removing
     autosize.destroy(document.querySelectorAll('textarea.expanding'));
 
-    $oldRows.remove();
+    const $oldRows = this.$editorBody.find('tr').remove();
```
unho: There is no need to save a `const` since there is no other reference to `$oldRows`.

PR #4940, opened 2016-07-21
Fix: dont break editor on gotoIndex <= 0

### LINKED ISSUE #4625: Can't go past 18 incomplete units
As already reported in #4560, when going through 18+ incomplete units, the editor gets stuck in the 18th unit, and it doesn't go further. Check the following recording as a reference:

editor_bugs

### LINKED ISSUE #4870: "Go to string number" fails to set URL properly
This bug seems to occur in master and to have existed for at least the past few months, I have hopefully nailed down the specific conditions to reproduce it, it requires a view across multiple po files (it wont be observed when translating the strings from only a single po file).

Enter a string number in the string number box and hit enter.

Pootle will navigate to that number, however it *if this number is in a different po file from where you started, it will fail to set the unit_id*
For example, I enter "25" and it sets the hash component as simply `#filter=all&offset=18`

Because it has failed to set the unit_id, if you refresh the page or attempt to use the URL pootle does not know what unit_id to go to and as such you end up at a different string number, for example 19.

The unit_id is set correctly when going from one string number within a po to another string number within the same po, for example if I click the string number box and hit enter again, because it is "navigating" within the same po file it correctly sets the url, for example `#filter=all&offset=18&unit=22`, browser refresh and using the URL then works as expected.

phlax:
fixes #4625

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -1742,10 +1742,12 @@ PTL.editor = {
     }
 
     e.preventDefault();
-    const index = parseInt(this.unitIndexEl.textContent, 10) - 1;
+
+    let index = parseInt(this.unitIndexEl.textContent, 10);
     if (isNaN(index)) {
       return;
     }
+    index = Math.max(0, index - 1);
 
     // if index is outside of current uids clear units first
     if (index < this.initialOffset || index >= this.getOffsetOfLastUnit()) {


unho:
Tested and works. GTM

dwaynebailey:
Tested on a staging instance passes by test.  GTM

PR #4942, opened 2016-07-21
Fix: set unitid when jumping to different offset

### LINKED ISSUE #4870: "Go to string number" fails to set URL properly

phlax:
fixes #4870

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -551,8 +551,8 @@ PTL.editor = {
             uId = this.units.uIds[this.offsetRequested - this.initialOffset - 1];
           } else {
             uId = this.units.uIds[0];
-            $.history.load(utils.updateHashPart('unit', uId));
           }
+          $.history.load(utils.updateHashPart('unit', uId));
         }
         this.offsetRequested = 0;
         this.setUnit(uId);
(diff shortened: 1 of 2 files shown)


unho:
Reviewed and tested. GTM

ta2-1:
gtm

pootle/static/js/editor/app.js
```diff
@@ -1747,19 +1747,25 @@ PTL.editor = {
...
-      this.offsetRequested = index + 1;
-      const newHash = utils.updateHashPart('offset',
-                                           this.getStartOfChunk(index),
-                                           ['unit']);
-      $.history.load(newHash);
-    } else if (index >= 0 && index <= this.units.total) {
-      const uId = this.units.uIds[(index - this.initialOffset)];
-      const newHash = utils.updateHashPart('unit', uId);
-      $.history.load(newHash);
+    index = Math.max(0, index - 1);
```
unho: Is this not duplicated?

phlax: yep, ill remove...

ta2-1:
gtm again

phlax:
landed in #d3b44e6cac

PR #4944, opened 2016-07-21
Use a positive margin so footer is not displayed on top of statistics…

### LINKED ISSUE #4943: Footer sometimes overlaps with statistics table
In https://mozilla.locamotion.org/projects/firefox/ I always see the footer overlaps the bottom of the statistics table, thus hiding some items:

overlap

unho:
… table

Fixes #4943.

### DIFF
--- a/pootle/static/css/style.css
+++ b/pootle/static/css/style.css
@@ -382,7 +382,7 @@ td.stats-name a:hover > i + span,
     min-height: 100%;
     height: auto !important;
     height: 100%;
-    margin-bottom: -62px; /* footer margin-top + height */
+    margin-bottom: 62px; /* footer margin-top + height */
 }
 
 #content


unho:
@iafan Any review appreciated.

unho:
@iafan Pinging you again for you to review this simple fix.

iafan:
@unho thanks for pinging again — the first notification somehow slipped through the cracks. Unfortunately, this doesn't look like a proper fix. There's something going on with your specific setup (OS / browser / larger fonts?) that results in incorrect rendering. Did anyone else report the same rendering issues?

This is how it is expected to look like (footer never overlaps, and there's the same nice margin between the table and the footer and between left/right edges of the table and the edge of the screen):

a

And this is how it looks with your change (too much blank space at the bottom):

b

The negative margin in CSS wasn't a mere typo — it was there for a reason. So before we start experimenting with the styles, let's understand what is there specific to your environment that causes incorrect rendering. Based on that we can see if there's a better solution that might work for you without breaking the rendering for everyone else.

Your older problem with the 'globe' icon (the screenshots that you provided) suggest that you have pretty large fonts, even though the monitor resolution is not that high. Mind sharing some more information about your monitor/UI/accessibility settings?

unho:
> Mind sharing some more information about your monitor/UI/accessibility settings?

I am using a laptop with with Ubuntu 14.04 using default accessibility settings. The browser is Chromium 51 with default zoom, despite using font size large in browser settings.

The overlapping not always shows up. It also happened with Firefox using font size 16 px (not sure if this was the default).

I hope it helps.

iafan:
Thanks! And just to reiterate: do we have any other users reporting the same issue?

I'll experiment with larger fonts on my side, and I'd also suggest experimenting with regular-sized fonts on your side, to make sure this is definitely related to larger fonts and not something else.

unho:
> Thanks! And just to reiterate: do we have any other users reporting the same issue?

No, nobody else reported this.

> I'll experiment with larger fonts on my side, and I'd also suggest experimenting with regular-sized fonts on your side, to make sure this is definitely related to larger fonts and not something else.

This is some error that comes and goes and I was never able to find the reason why it happens nor find a fixed list of steps to reproduce. If I hit this again I will do as you suggested.

unho:
I am closing this for now.

PR #4951, opened 2016-07-22
Editor: avoid extra loading indicator on page load

### LINKED ISSUE #4950: Loading spinner breaks its position on page load
I don't know if this problem happens cause i have 800K strings in db... but i can assure you this doesn't happen if you have 50K strings or less (tested)... i don't know at what specific amout of lines the gif breaks its positioning... compared with the two tests i made, i noticed that when i had 50K strings, the loading was quickly than this... so i wanted to report it. Maybe i'm the first one to test pootle with so many strings in it. :D

This is the test video: https://dl.dropboxusercontent.com/u/89718989/Screenshot/pootle/pootle-loading-translation.mp4

julen:
When the initial loading time goes past a few seconds, the generic indicator
(via jQuery's `ajaxStart()`) will overlap the page load indicator, making the
spinner move its position. This commit avoids the generic indicator from kicking
in on the initial page load, hence avoiding the extra indicator.

Fixes #4950.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -354,6 +354,10 @@ PTL.editor = {
     /* XHR activity indicator */
     $(document).ajaxStart(() => {
       clearTimeout(this.delayedActivityTimer);
+      if (this.isLoading) {
+        return;
+      }
+
       this.delayedActivityTimer = setTimeout(() => {
         this.showActivity();
       }, 3000);


phlax:
@julen do we need the page load indicator?

otherwise lgtm

julen:
Yeah we do; initially displaying a blank page won't help much in terms of feedback.

PR #4956, opened 2016-07-22
Remove external link icon in about page

unho:
This is to fix the icon for the external link in the About page. Maybe it is better to just remove the icon since it is only used there.

@iafan Please review.

### DIFF
--- a/pootle/templates/about.html
+++ b/pootle/templates/about.html
@@ -18,8 +18,7 @@
     teams to do translation and translation management. This community
     tool is Free Software, and any team can use it and contribute to the
     Pootle community. Read more on the
-    <a href="http://pootle.translatehouse.org/"><i class="icon-external-link"></i>
-    project website</a>.
+    <a href="http://pootle.translatehouse.org/">project website</a>.
     {% endblocktrans %}
   </p>
   <p>


unho:
@iafan what do you think about instead dropping the icon besides external links?

iafan:
@unho do you happen to have screenshots that compare the rendering in LTR/RTL mode? And can you describe what the problem with this link was? Because I don't see any problems with the icon in LTR mode.

We don't use this icon much, but apart from About page we also use it in static pages (and others can use it too — I believe it can be helpful).

unho:
Related #4960. As I suggested, perhaps it is just simpler to drop the icon.

This is in master LTR:

master_ltr


Note the icon is rendered too close to the external link, and that there is plenty of blank space before the icon.

And this is master RTL:

master_rtl



This is this PR in LTR:

pr_ltr

Perhaps now there is too much space between the icon and the external link.

This is this PR in RTL:

pr_rtl

iafan:
Leandro, what is your display resolution / page zoom level? This is how the links look for me under Chrome and Firefox at 100% zoom. They are properly aligned vertically and definitely don't need any changes in margins:

screenclip

So before changing CSS it would be good to understand why these are rendered incorrectly for you.

If you don't want to use this icon, I'd rather suggest removing its reference from the About page template for the time being (which will also 'fix' #4960). The icon will still be there in resources, though, for those who still use it.

unho:
> Leandro, what is your display resolution / page zoom level? This is how the links look for me under Chrome and Firefox at 100% zoom.

My resolution is 1360 x 768 and I am using 100% zoom.

> If you don't want to use this icon, I'd rather suggest removing its reference from the About page template for the time being (which will also 'fix' #4960). The icon will still be there in resources, though, for those who still use it.

If you don't mind I will be updating this PR to do that.

iafan:
> If you don't mind I will be updating this PR to do that.

Sure. Will you undo the margins, then?

unho:
@iafan Updated, any review is appreciated.

iafan:
GTM

Issue #4957, opened 2016-07-22
Add support for monolingual formats


phlax:
The lack of `.properties` support in particular is an issue for Pootle

I would propose that we support monolingual formats in the following way.

- Stored in Pootle as bilingual format (ie source/target unit)
- Stores are only created/exist if they exist in source_language/templates
- Stores are automatically created/updated on change in source_language/templates
- De/serializers are added to parse/dump monolingual files

iafan:
How will you store/determine the list of target languages for each monolingual file or for the project?

ta2-1:
@iafan why doesn't translation_project model suit for that?

iafan:
@ta2-1 I'm trying to understand the concept, nothing more. Will this simply reuse the old "Update from templates" mechanism and UI to manually add new languages to the project, or will there be something else? Does this involve the pootle_fs project somehow? I thought pootle_fs was meant to simplify Pootle and make it format-agnostic, and the description of the issue suggests that we want to add support for more file formats directly into Pootle. So I just want to understand this whole thing better.

dwaynebailey:
@iafan at a general level.  We need to get back monolingual support into Pootle. A number of users are blocked in their migration because the previous monolingual support was dropped.

In terms of the actual feature.  Yes this will involve Pootle FS.  But remember that while Pootle FS is a few things centrally it is VCS synchornisation and nothing more.  Its primary aim is not format support, Pootle does that relatively well itself already.   The proposal though is to use the Pootel FS serialisers to manage the output of the monolingual files, so in some ways yes Pootel FS is handling the file format.

Pootle's internal model is heavily influenced by PO and yes the long term aim is to help that internal model to become more format agnostic (Call this a step along the way).  But that is not the primary aim of bringing back properties support.  A spin off of tackling properties is that we allow monolingual support.  This is something that we need for the various other monolingual formats including l20n.

By using the existing bilingual model in Pootle we're able to implement monolingual files without dramatic schema changes.  We create a stable representation of the data and can manage template upgrades much more safely as we're dealing with it as a bilingual representation.

The update mechanism we're looking at is automatic.  As in, you change templates, you change the target files.  We're aware that there are various valid approaches to templates but for now we're looking simply at these being authoratative and automatic.  We won't be bringing back in-UI updating against anytime templates.  And we're not looking at automatic anything with templates in bilingual projects.

iafan:
[...]

Issue #4929, opened 2016-07-20
Allow multiple filetypes in a single project


phlax:
Its not unreasonable for a project to have several types of localisation files to deal with eg js, python, etc

Atm there is a `localfiletype` setting - I would propose to migrate this to `localfiletypes` and adjust related code and tests as necessary

phlax:
closed in #4963

PR #4963, opened 2016-07-24
Pootle formats

phlax:
Adds models and plugin setup for format registry, which can be used by `Project` and `Store` in the case of having mutiple filetype formats - see #4929

PR #4966, opened 2016-07-25
Pootle formats base delegation, models and classes

phlax:
Adds a pootle_formats app to register, retrieve and override format handling in Pootle

This is useful for implement multi-format projects #4929 and is also useful for implementing monolingual formats #4957

### DIFF
--- a/pootle/apps/pootle_format/default.py
+++ b/pootle/apps/pootle_format/default.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+POOTLE_FORMATS = [
+    ("po",
+     dict(title='Gettext PO',
+          extension="po",
+          template_extension="pot")),
+    ("xliff",
+     dict(title='XLIFF',
+          extension="xliff",
+          template_extension="xliff")),
+    ("xlf",
+     dict(title='XLIFF',
+          extension="xlf",
+          template_extension="xlf")),
+    ("ts",
+     dict(title='TS',
+          extension="ts",
+          template_extension="ts")),
+    ("lang",
+     dict(title='Mozilla Lang',
+          extension="lang",
+          template_extension="lang"))]
--- a/pootle/apps/pootle_format/migrations/0001_initial.py
+++ b/pootle/apps/pootle_format/migrations/0001_initial.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='FileExtension',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('name', models.CharField(unique=True, max_length=15, verbose_name=b'Format filetype extension', db_index=True)),
+            ],
+            options={
+                'abstract': False,
+                'db_table': 'pootle_fileextension',
+            },
+        ),
+        migrations.CreateModel(
+            name='Format',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('name', models.CharField(unique=True, max_length=30, verbose_name='Format name', db_index=True)),
+                ('title', models.CharField(max_length=255, verbose_name='Format title', db_index=True)),
+                ('enabled', models.BooleanField(default=True, verbose_name='Enabled')),
+                ('monolingual', models.BooleanField(default=False, verbose_name='Monolingual format')),
+                ('extension', models.ForeignKey(related_name='formats', to='pootle_format.FileExtension')),
+                ('template_extension', models.ForeignKey(related_name='template_formats', to='pootle_format.FileExtension')),
+            ],
+            options={
+                'abstract': False,
+                'db_table': 'pootle_format',
+            },
+        ),
+        migrations.AlterUniqueTogether(
+            name='format',
+            unique_together=set([('title', 'extension')]),
+        ),
+    ]
--- a/pootle/apps/pootle_format/registry.py
+++ b/pootle/apps/pootle_format/registry.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from collections import OrderedDict
+
+from django.utils.functional import cached_property
+
+from pootle.core.delegate import format_registration
+from pootle_format.models import FileExtension, Format
+
+
+class FormatRegistry(object):
+
+    def initialize(self):
+        for filetype, info in format_registration.gather().items():
+            self.register(filetype, **info)
+
+    def register(self, name, extension, title=None, template_extension=None):
+        update = False
+        template_extension = template_extension or extension
+        ext, created = FileExtension.objects.get_or_create(
+            name=extension)
+        if template_extension != extension:
+            template_ext, created = FileExtension.objects.get_or_create(
+                name=template_extension)
+        else:
+            template_ext = ext
+        try:
+            filetype = Format.objects.get(name=name)
+        except:
+            filetype = None
+        if not filetype:
+            filetype = Format.objects.create(
+                name=name, extension=ext, template_extension=template_ext)
+        if filetype.extension != ext:
+            filetype.extension = ext
+            update = True
+        if not filetype.title:
+            filetype.title = title or name.capitalize()
+            update = True
+        elif title and filetype.title != title:
+            filetype.title = title
+            update = True
+        extension = filetype.template_extension.name
+        if template_extension and extension != template_extension:
+            template_ext, created = FileExtension.objects.get_or_create(
+                name=template_extension)
+            filetype.extension = template_ext
+            update = True
+        if update:
+            filetype.save()
+        if update or created:
+            self.clear()
+        return filetype
+
+    def clear(self):
+        if "formats" in self.__dict__:
+            del self.__dict__["formats"]
+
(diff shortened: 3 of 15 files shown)


pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,26 @@
...
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+POOTLE_FORMATS = [
+    ("po",
+     dict(title='Gettext PO file',
```
dwaynebailey: Does this not need lazy localisation?

pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,26 @@
...
+          extension="po",
+          template_extensions=["po", "pot"])),
+    ("xliff",
+     dict(title='XLIFF Translation File',
+          extension="xliff",
+          template_extensions=["xliff"])),
+    ("xlf",
```
dwaynebailey: Is there no way to combine xlf and xliff, they're the same thing the xlf more correct than the xliff.

pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,26 @@
...
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+POOTLE_FORMATS = [
+    ("po",
+     dict(title='Gettext PO file',
+          extension="po",
+          template_extensions=["po", "pot"])),
```
dwaynebailey: Templates are always `.pot`

pootle/apps/pootle_format/default.py
phlax: maybe it does - but i had added a unique_together with extension (to prevent menu duplication) so i should probs remove it in that case - maybe not

pootle/apps/pootle_format/default.py
phlax: even in a source_language ?

phlax: tbh it would greatly simplify things @dwaynebailey if we didnt need to support multiple template_extensions for a particular filetype

pootle/apps/pootle_format/default.py
phlax: not in the current schema

pootle/apps/pootle_format/default.py
dwaynebailey: In Gettext land there isn't be a concept of `source_language`, just templates.  Though the POT files could be in a language other than English I've only rarely seen it done like that.  The msgid is simply a key so from a programming perspective its not important.  I'm almost sure the Gettext tools wouldn't care (haven't checked though).

There is a model of using PO as a monolingual file, though Ive never seen that done in the wild.

pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+
+POOTLE_FORMATS = [
```
ta2-1: @phlax why did you put too few formats? Should we link them with ttk formats?

pootle/apps/pootle_format/migrations/0001_initial.py
```diff
@@ -0,0 +1,44 @@
...
+                'db_table': 'pootle_fileextension',
+            },
+        ),
+        migrations.CreateModel(
+            name='Format',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('name', models.CharField(unique=True, max_length=30, verbose_name='Format name', db_index=True)),
+                ('title', models.CharField(max_length=255, verbose_name='Format title', db_index=True)),
+                ('enabled', models.BooleanField(default=True, verbose_name='Enabled')),
```
ta2-1: what does this mean?

pootle/apps/pootle_format/migrations/0001_initial.py
```diff
@@ -0,0 +1,44 @@
...
+                ('monolingual', models.BooleanField(default=False, verbose_name='Monolingual format')),
+                ('extension', models.ForeignKey(related_name='formats', to='pootle_format.FileExtension')),
+                ('template_extensions', models.ManyToManyField(related_name='template_formats', to='pootle_format.FileExtension')),
```
ta2-1: did we decide that it's enough to use a single extension here?

pootle/apps/pootle_format/default.py
phlax: any formats we add we will need proper support for. Im thinking initially it should just be po and xliff actually.

pootle/apps/pootle_format/migrations/0001_initial.py
phlax: it allows you to dis/enable a Filetype for a site.

If you delete it, it will be recreated the next time format_registry.initialize() is called

pootle/apps/pootle_format/migrations/0001_initial.py
phlax: it seems not - it seems we are going to support both `po` and `pot` - although tbh if that is the only use case we probs dont need a many field for it

pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,30 @@
...
+    ("po",
+     dict(title='Gettext PO file',
+          extension="po",
+          template_extension="pot")),
+    ("xliff",
+     dict(title='XLIFF Translation File',
+          extension="xliff",
+          template_extension="xliff")),
+    ("xlf",
+     dict(title='XLIFF Translation File',
```
unho: How are you going to differentiate both XLIFF entries if they have the same title?

unho: In filetypes code we used different strings: https://github.com/translate/pootle/blob/master/pootle/apps/pootle_store/filetypes.py#L23-L24

pytest_pootle/fixtures/formats/registry.py
```diff
@@ -0,0 +1,12 @@
+
```
unho: No copyright notice?

pootle/apps/pootle_format/default.py
phlax: Because the title is always displayed with the extension - and there is a unique_together on them for that reason.

pytest_pootle/fixtures/formats/registry.py
phlax: ill add

pootle/apps/pootle_format/default.py
unho: If we are displaying this title to the users then it must be localizable. If you want to keep a `unique_together`, something that makes sense because of all the formats that share the .properties extension, then I suggest adding an internal name that can be used for the unique check, like:

```python
POOTLE_FORMATS = [
    ("po",
     dict(title=_('Gettext PO'),
          name='Gettext PO'
          extension="po",
          template_extension="pot")),
```

pootle/apps/pootle_format/default.py
```diff
@@ -0,0 +1,30 @@
...
+          extension="xlf",
+          template_extension="xlf")),
+    ("ts",
+     dict(title='TS File',
```
unho: I would use "Qt TS File".

pootle/apps/pootle_format/default.py
phlax: @unho its unnecssary duplication - if you see in the code it has a name `po`

pootle/apps/pootle_format/default.py
phlax: k - will do

pootle/apps/pootle_format/default.py
unho: Displayed together in the UI like "Gettext PO file (.po)"?

phlax: yep

pootle/apps/pootle_format/default.py
phlax: also re localizing the title - as this is stored in the db my understanding is that this is not the place to localize (im happy to be wrong tho)

pootle/apps/pootle_format/default.py
unho: I am not seeing anything like that in https://github.com/translate/pootle/pull/4966/commits/59d58aa7d11eafa6a4dcf84994ea562def85b738#diff-99489b5e96e0f37f2daaa11c328f4d0dR31

pootle/apps/pootle_format/registry.py
```diff
@@ -0,0 +1,86 @@
+
```
unho: No copyright notice here?

pootle/apps/pootle_format/registry.py
```diff
@@ -0,0 +1,86 @@
...
+class FormatRegistry(object):
+
+    def initialize(self):
+        for filetype, info in format_registration.gather().items():
+            self.register(filetype, **info)
+
+    def register(self, name, extension, title=None, template_extension=None):
+        update = False
+        template_extension = template_extension or extension
+        try:
```
unho: Nit: move this `try` clause to line 29 so it is before the `if` that checks for `filetype`. The code in between is unrelated.

pootle/apps/pootle_format/default.py
phlax: i can add it further up the stack - probs on `__str__` is a good idea actually - but for ref - here is where its actually used - https://github.com/translate/pootle/pull/4963/files#diff-83014e45acc5cd6424075fd6a8861be1R37

pootle/apps/pootle_format/registry.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.formats.__getitem__(k)
+
+    def keys(self):
+        return self.formats.keys()
+
+    def values(self):
+        return self.formats.values()
+
+    def items(self):
+        return self.formats.items()
```
unho: I think this is not the first time I see a class that wraps an OrderedDict and then exposes these methods again. I do wonder if it is worth creating a mixin to handle this repeated code.

phlax: `UserDict` would be the answer, but im reluctant to use a full `UserDict`

unho:
lgtm

PR #4967, opened 2016-07-25
Checks: add placeholders check for Plurr-formatted strings

julen:
This new check will ensure that placeholders used in translations for
Plurr-formatted strings actually exist in source strings, which helps catch
early on potential chances to break apps.


iafan:
GTM

unho:
lgtm since it doesn't directly affect us.

PR #4970, opened 2016-07-26
FormSelectInput: accommodate for multi-value data

julen:
Removes extra boilerplate by handling everything in a single component. For
multi-select inputs, simply pass the `multiple` prop to it.


pootle/static/js/shared/components/FormElement.js
```diff
@@ -11,19 +11,20 @@ import React from 'react';
...
 import FormSelectInput from './FormSelectInput';
-import FormMultiSelectInput from './FormMultiSelectInput';
 
 
 const FormElement = React.createClass({
 
   propTypes: {
     type: React.PropTypes.string,
     label: React.PropTypes.string.isRequired,
+    multiple: React.PropTypes.bool,
```
phlax: this and all the other references to `multiple` need to be changed to `multi`

pootle/static/js/shared/components/FormSelectInput.js
```diff
@@ -41,19 +44,23 @@ const FormSelectInput = React.createClass({
...
+    const { value } = this.props;
+    /* FIXME: react-select#25 prevents using non-string values */
+    const selectValue = this.props.multiple ? value : value.toString();
     return (
       <Select
         onChange={this.handleChange}
         {...this.props}
-        /* FIXME: react-select#25 prevents using non-string values */
-        value={this.props.value.toString()}
+        multi={this.props.multiple}
```
phlax: why make it more difficult for others to understand the code?

https://github.com/JedWatson/react-select#multiselect-options

"""You can enable multi-value selection by setting multi={true}."""

julen: Being a mirror of react-select's API is definitely not a goal of`FormElement` + `FormSelectInput` components, but rather provide a thin abstraction with sane defaults (I know this for a fact because I wrote them). In that context, [`multiple` is a natural and well-known choice](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attr-multiple) to denote that a select element accepts multiple values.

The only code which cares and has to deal directly with react-select is `FormSelectInput`, it uses `multi={<bool>}` and if that's hard to understand, I'm definitely missing the point.

Having said that, I have other simplifications for `FormElement` in mind which will make it easier to reason about the rendering of form controls — after all this is one of the initial React components I wrote around 2 years ago, and best practices, common patterns and experience have evolved since then.

phlax:
also - you have made `multiple` `{false}` by default - and this throws jsx warnings afaict

phlax:
> Being a mirror of react-select's API is definitely not a goal ofFormElement

Like i said, why make it more difficult for others to understand by explicitly breaking the docs?

iafan:
> Like i said, why make it more difficult for others to understand by explicitly breaking the docs?

Though I have no personal attachment to either `multi` or `multiple`, I think it's the author of the component who ultimately decides how the property should be named (if it doesn't break other naming conventions). I think Julen provided a compelling reason: he wants the properties of the component to be more understandable for himself an other users of _his_ component, and influenced by widely adopted HTML standards, not the unconventional choice of the single author. Encapsulation is all about providing  clean API to work with, and isolating you from the quirks of internal implementation of sub-components.

phlax:
> I think Julen provided a compelling reason: he wants the properties of the component to be more understandable for himself an other users of his component

this is in the **pootle** codebase

iafan:
Yet he's the author and primary maintainer of that code. And he's historically the primary developer, maintainer and consumer of the entire frontend code; he has been known to sticking to pretty strict standards, making others stick to the same standards, and has a pretty good record of making good decisions. So I'd leave the final judgement to him.

julen:
> also - you have made multiple {false} by default - and this throws jsx warnings afaict

As [clarified on IRC](https://gitter.im/translate/dev?at=5798c7d23383eb622147bab6), such warnings are unrelated to this PR and are caused by the misuse of JSX attributes in other code.

PR #4971, opened 2016-07-26
pylint cleanups

dwaynebailey:
Fixes:

- W1201 - let logging do interpolation of strings
- W0404 - don't reimport imports


phlax:
never understood why logging had this interpolation thing - but now i understand http://stackoverflow.com/questions/11615053/why-use-multiple-arguments-to-log-instead-of-interpolation

lgtm

PR #4973, opened 2016-07-26
Cleanup: use constant name convention for imported constants

dwaynebailey:


unho:
gtm

julen:
Nice Dwayne 👍 GTM.

PR #4980, opened 2016-07-27
Mousetrap: always prevent default browser behavior

julen:
Mousetrap doesn't offer a default behavior or global hook to disallow events
from propagating, therefore this is left for callbacks to do it explicitly.
That's the case for Pootle: we want to hijack shortcuts entirely, and not bubble
up the event, hence avoiding potential side-effects.

As an example, on Windows machines, pressing Ctrl+Up/Down in the editor would
trigger Pootle's logic AND the native behavior, resulting in potential scrolls
as the event could be handled after the editor changed its layout.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -313,7 +313,8 @@ PTL.editor = {
     const hotkeys = mousetrap(document.body);
 
     // FIXME: move binding to `SuggestionFeedbackForm` component
-    hotkeys.bind('esc', () => {
+    hotkeys.bind('esc', (e) => {
+      e.preventDefault();
       if (this.selectedSuggestionId !== undefined) {
         this.closeSuggestion();
       }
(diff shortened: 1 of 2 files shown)


phlax:
related to this @julen i was wondering if we can attach the `mousetrap` listeners lower in the `DOM` ?

iafan:
Tested on https://translate.stage.evernote.com/ — Ctrl+Up/Down works correctly now.

unho:
lgtm

pootle/static/js/editor/app.js
```diff
@@ -313,7 +313,8 @@ PTL.editor = {
     const hotkeys = mousetrap(document.body);
```
phlax: im aware that its not directly the purpose of this PR - but should this be `moustrap("#editor")`

iafan: @phlax we want to catch hotkeys even if the editor (textarea or any other control inside `#editor`) doesn't have focus. And without such focus `#editor` won't get the keyboard event.

PR #4986, opened 2016-07-28
Fixture fix

phlax:
Atm some Stores are being created with a `/` in their name, and the wrong parent in tests.

This PR addresses that.

### DIFF
--- a/pytest_pootle/fixtures/import_export_fixtures.py
+++ b/pytest_pootle/fixtures/import_export_fixtures.py
@@ -65,6 +65,9 @@ def ts_directory(po_directory, request, tmpdir, settings):
 @pytest.fixture
 def en_tutorial_ts(settings, english_tutorial, ts_directory):
     """Require the en/tutorial/tutorial.ts store."""
+
+    english_tutorial.project.localfiletype = "ts"
+    english_tutorial.project.save()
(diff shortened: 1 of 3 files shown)


pytest_pootle/fixtures/import_export_fixtures.py
```diff
@@ -65,6 +65,9 @@ def _cleanup():
 @pytest.fixture
 def en_tutorial_ts(settings, english_tutorial, ts_directory):
     """Require the en/tutorial/tutorial.ts store."""
+
+    english_tutorial.project.localfiletype = "ts"
+    english_tutorial.project.save()
```
unho: I think it is better to instead use a different project instead of hijacking tutorial.

phlax: you are probs right - but that is how its done currently - and im reluctant to change as tbh it needs a bit of a rewrite.

This change only allows ts in that Project - which should have been happening anyway. Im going to raise a related but that Store.create should check for localfiletype (soon to be filetypes)

phlax: opened https://github.com/translate/pootle/issues/4988

unho:
lgtm.

PR #4990, opened 2016-07-28
Make statici18n work in development mode

julen:
In this PR the output path for static files generated by `statici18n` is adjusted so that it matches the expected behavior of having source files under `static/` and let `collectstatic` manage the rest.

It also removes the need for extra commands to handle the case for `en-us`.


Makefile
```diff
@@ -30,8 +30,6 @@ assets:
 	npm install && \
 	cd ${CWD}
 	${POOTLE_CMD} compilejsi18n
-	# The site language en-us is not compiled by default, so handle it specially
-	${POOTLE_CMD} compilejsi18n -l en-us
```
phlax: this is a required step in the build process

julen: This is taken care of. Did you perhaps miss to check some commit in the PR?

pootle/settings/40-apps.conf
```diff
@@ -131,6 +131,10 @@ POOTLE_SIGNUP_ENABLED = True
 # http://django-statici18n.readthedocs.io/en/latest/faq.html#using-a-placeholder-directory
 STATICI18N_ROOT = working_path('static')
 
+# Directory where generated files will be written to. Customized to avoid name
+# clashing with `static/js`.
+STATICI18N_OUTPUT_DIR = 'translations'
```
phlax: there is a real risk of this being confused with the Pootle directory of the same name

Makefile
phlax: where is the symbolic link?

pootle/settings/40-apps.conf
julen: If you read the comment, you'll realize the main point of adding this is to avoid a name clash with `static/js`. If you have a better name which doesn't start with `js`, I'm in.

Makefile
julen: > where is the symbolic link?

[It's here.](https://github.com/julen/pootle/blob/32160c5dd1b5145abbdb2301fd45374c46731756/pootle/locale/en_US)

phlax: thanks - aplogies - ill run a build now

phlax:
tested - and works find - gtm - thanks @julen 

(apologies for stalling)

unho:
lgtm. Please land asap.

PR #4993, opened 2016-07-28
Pootle formats Project/Store db fields

phlax:
adds

- a sorted m2m key Project -> Filetype (allowing for prioritised relationship)
- an fk Store -> Filetype

### DIFF
--- a/pootle/apps/pootle_format/utils.py
+++ b/pootle/apps/pootle_format/utils.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+from collections import OrderedDict
+
+from .exceptions import UnrecognizedFiletype
+
+
+class ProjectFiletypes(object):
+
+    def __init__(self, project):
+        self.project = project
+
+    @property
+    def filetypes(self):
+        return self.project.filetypes.all()
+
+    def choose_filetype(self, filename):
+        ext = os.path.splitext(filename)[1][1:]
+        formats = self.project.filetypes.filter(
+            extension__name=ext)
+        if formats.exists():
+            return formats.first()
+        formats = self.project.filetypes.filter(
+            template_extension__name=ext)
+        if formats.count():
+            return formats.first()
+        filetypes = (
+            ", ".join(
+                self.project.filetypes.values_list(
+                    "extension__name", flat=True)))
+
+        # The filename's extension is not recognised in this Project
+        raise UnrecognizedFiletype(
+            "File '%s' is not recognized for Project "
+            "'%s', available extensions are %s"
+            % (filename,
+               self.project.fullname,
+               filetypes))
+
+    @property
+    def filetype_extensions(self):
+        return list(
+            self.filetypes.values_list(
+                "extension__name", flat=True))
+
+    @property
+    def template_extensions(self):
+        return list(
+            self.filetypes.values_list(
+                "template_extension__name", flat=True))
+
+    @property
+    def valid_extensions(self):
+        """this is the equiv of combining 2 sets"""
+        return list(
+            OrderedDict.fromkeys(
+                self.filetype_extensions
+                + self.template_extensions))
--- a/pootle/apps/pootle_project/migrations/0007_migrate_localfiletype.py
+++ b/pootle/apps/pootle_project/migrations/0007_migrate_localfiletype.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.delegate import formats
+
+
+def migrate_localfiletype(apps, schema_editor):
+    format_registry = formats.get()
+
+    projects = apps.get_model("pootle_project.Project").objects.all()
+    filetypes = apps.get_model("pootle_format.Format").objects.all()
+
+    for project in projects:
+        filetype = filetypes.get(
+            pk=format_registry[project.localfiletype]['pk'])
+        if filetype not in project.filetypes.all():
+            project.filetypes.add(filetype)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_format', '0002_default_formats'),
+        ('pootle_project', '0006_project_filetypes'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_localfiletype),
+    ]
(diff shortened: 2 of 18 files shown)


pootle/core/initdb.py
```diff
@@ -244,15 +253,16 @@ def create_default_projects(self):
         """
         en = self.require_english()
 
+        po = Format.objects.get(name="po")
```
unho: Nit: move this down to before `tutorial, created = self._create_object(Project, **criteria)`. This way changes are closer and this method resembles like previous one.

phlax: in this case i think its better to set the local var explicitly at the top - as we do for `en` as its reused in the method

phlax: i think i should remove line above to make that a bit clearer

pootle/apps/pootle_format/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    def filetypes(self):
+        return self.project.filetypes.all()
+
+    def choose_filetype(self, filename):
+        ext = os.path.splitext(filename)[1][1:]
+        formats = self.project.filetypes.filter(
+            extension__name=ext)
+        if formats.count():
```
unho: `count`? Why not `exists`?

phlax: yep - ill change - i never rem which is the cheapest `models` `models.exists` or `models.count` - not the last tho i think

pootle/apps/pootle_format/utils.py
```diff
@@ -0,0 +1,63 @@
...
+            return formats.first()
+        formats = self.project.filetypes.filter(
+            template_extension__name=ext)
+        if formats.count():
+            return formats.first()
+        filetypes = (
```
ta2-1: can we add empty line before this? I've thought that this function always raises exception.

pootle/apps/pootle_project/migrations/0007_migrate_localfiletype.py
```diff
@@ -0,0 +1,31 @@
...
+
+def migrate_localfiletype(apps, schema_editor):
+    format_registry = formats.get()
+
+    projects = apps.get_model("pootle_project.Project").objects.all()
+    filetypes = apps.get_model("pootle_format.Format").objects.all()
+
+    for project in projects:
+        filetype = filetypes.get(
+            pk=format_registry[project.localfiletype]['pk'])
```
ta2-1: what if there is no localfiletype in format_registry?

pootle/apps/pootle_project/models.py
```diff
@@ -56,6 +59,20 @@
 
 class ProjectManager(models.Manager):
 
+    def create(self, *args, **kwargs):
```
ta2-1: I raised a question re porting filetypes logic to ProjectManager.create(). And I understood @phlax was going to check this.

pootle/apps/pootle_store/migrations/0010_set_store_filetypes.py
```diff
@@ -0,0 +1,41 @@
...
+    stores = apps.get_model("pootle_store.Store").objects.all()
+    filetypes = apps.get_model("pootle_format.Format").objects.all()
+
+    for project in projects:
+        for filetype in project.filetypes.all():
+            stores = (
+                stores.filter(translation_project__project=project)
+                      .filter(filetype__isnull=True)
+                      .filter(name__endswith=".%s" % filetype.extension))
+            stores.update(filetype=filetype)
```
ta2-1: should we check if all stores were updated?

pootle/apps/pootle_format/utils.py
phlax: yep - ill add a comment too

pootle/apps/pootle_store/models.py
```diff
@@ -1274,6 +1276,23 @@ def live(self):
         """Filters non-obsolete stores."""
         return self.filter(obsolete=False)
 
+    def create(self, *args, **kwargs):
```
ta2-1: The same concern as for ProjectManager re using StoreManager.create() for filetypes logic.

pootle/apps/pootle_project/migrations/0007_migrate_localfiletype.py
phlax: this was discussed - basically we only allow `po` and `xliff` atm so its a safe assumption that it will be

pootle/apps/pootle_project/models.py
phlax: my understanding was `Project.save()` - yep - i will

pootle/apps/pootle_store/models.py
```diff
@@ -1342,6 +1358,13 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
                                             related_name='stores',
                                             db_index=True, editable=False)
 
+    filetype = models.ForeignKey(
+        Format,
+        related_name='stores',
+        null=True, blank=True,
+        db_index=True)
+    is_template = models.BooleanField(default=False)
```
ta2-1: It's definitely not a part of this PR.

pootle/apps/pootle_store/migrations/0010_set_store_filetypes.py
phlax: im thinking more about adding a warning check to say "You have Stores without filetype in your system, you should fix in some way 8/"

pootle/apps/pootle_project/models.py
phlax: k - i experimented with this. My feeling that in terms of the CRUD cycle the manager is the right place, and has the advantage of making it easier to create a project passing a list of string names for `filetypes`

pootle/apps/pootle_store/models.py
phlax: i disagree.

This PR creates the db fields necessary to remove `localfiletype` this is one of them

pootle/apps/pootle_format/utils.py
unho: Kinda addressed, since the blank line and the comment are after this line, not before it.

pootle/apps/pootle_store/models.py
```diff
@@ -1342,6 +1358,12 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
                                             related_name='stores',
                                             db_index=True, editable=False)
 
+    filetype = models.ForeignKey(
+        Format,
+        related_name='stores',
+        null=True, blank=True,
```
unho: If we are using more than a line, can then put each item on its own line?

pootle/apps/pootle_store/migrations/0010_set_store_filetypes.py
unho: Where is that check?

phlax: not added for now - as there is some clarity required around handling Stores with no filetype (or for example changing the filetype of a Project). This will be addressed one way or another as the logic for this lands

unho:
lgtm

PR #4996, opened 2016-07-29
i18n: add `nt()` helper function

julen:
For completeness, `t()`'s counterpart with support for plurals.


unho:
Do we have some use for this one?

unho:
lgtm

PR #4998, opened 2016-07-29
Pootle formats pootle_fs hookup

phlax:
add `filetypes` support to pootle_fs

### DIFF
--- a/pootle/apps/pootle_fs/management/commands/init_fs_project.py
+++ b/pootle/apps/pootle_fs/management/commands/init_fs_project.py
@@ -11,6 +11,7 @@ import logging
 from django.core.management import BaseCommand, CommandError
 from django.db import IntegrityError
 
+from pootle_format.models import Format
 from pootle_fs.utils import FSPlugin, parse_fs_url
 from pootle_language.models import Language
 from pootle_project.models import Project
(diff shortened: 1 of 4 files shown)


pootle/apps/pootle_fs/management/commands/init_fs_project.py
```diff
@@ -94,11 +93,15 @@ def handle(self, **options):
             project = Project.objects.create(
                 code=code,
                 fullname=name,
-                localfiletype=options['file_type'],
                 treestyle='none',
                 checkstyle=options['checkstyle'],
-                source_language=source_language
-            )
+                source_language=source_language)
+            for filetype in options["filetypes"] or ["po"]:
```
dwaynebailey: Since you already set 'po' as a default, you shoudn't need to `or` here at all.

phlax: it was my lack of argparse-fu - it was failing every way i tried it - apart from this

unho: Didn't `default=["po"]` in `parser.add_argument()` above work?

phlax: no, with append it was making [["po"]]

im sure there is a better way tho

unho:
lgtm since argparse seems not to be able to handle specifying a default when using `append`.

pootle/apps/pootle_fs/management/commands/init_fs_project.py
unho: Just for the record this is not what I got, but I hit another different issue (https://bugs.python.org/issue16399). My stance is to land the PR as it is.

PR #5001, opened 2016-07-30
Add migration to correctly set Store.filetypes

phlax:
The previous migration to do this did not work correctly, and im figuring its better to add the correct one, in case anyone is following master, and have already migrated.


unho:
lgtm

PR #5005, opened 2016-08-02
Docs: Flush cached snippets on upgrade

unho:
This will make sure cached snippets pointing to outdated static
assets are not served.

### DIFF
--- a/docs/server/upgrading.rst
+++ b/docs/server/upgrading.rst
@@ -12,9 +12,8 @@ current release.
 Stop your running Pootle
 ------------------------
 
-You may want to stop your running Pootle while you upgrade to prevent updates
-to your data during the migration process. If you have RQ workers running you
-may want to stop those also.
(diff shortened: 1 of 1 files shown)


dwaynebailey:
lgtm

ta2-1:
@unho after discussion with @phlax I've added this commit https://github.com/translate/pootle/pull/4659/commits/c099ab4c28756fc5173d1bc4c2db99085a1d8677 where html snippets are removed within migration. 
I remember that @phlax suggested to add similar migrations when it is necessary.

TBH I don't think that migrations is the best place for flushing cache snippets. Though it allows us to force people to do it.

lgtm re this PR.

phlax:
@ta2-1 you are probably right, i was keen not to increase the number of commands required to upgrade, and in that situation we knew we needed to clear the cache - but its not a great solution

unho:
Yep, the least steps in upgrade the better. But for now this should work and also educates people. I am merging.

PR #5008, opened 2016-08-02
Use templates for robots.txt

dwaynebailey:
This drops the autogenerated robots.txt and replaces it with a default
and customisable templated file.

The rules ensure:
* Allow the whole site by default
* Disallow any of the editor and export views
* Allows access to the browse views
* Allows access to static pages
* Allows access to user pages but not to user stats


iafan:
Thanks, Dwayne. I think Pootle would benefit from a note added somewhere to the documentation (feel free to adjust it — this is just a proposal):

>By default, Pootle serves its own robots.txt file that prevents indexing certain Pootle pages. Since it is a static template, you can use standard Django template overriding mechanism if you want to tweak the file to your needs. Keep in mid, however, that robots.txt must be served from the root of the web site, so it won't work if Pootle is served from user non-root path. In this case we imply that you have Pootle installed side-by-side with some other website or application, so you need to manage robots.txt completely outside Pootle. In this case we recommend to copy the default contents of provided templates/robots.txt file into your own robots.txt and adjust the paths to allowed/disallowed resources accordingly.

pootle/templates/robots.txt
```diff
@@ -0,0 +1,13 @@
+User-agent: *
+Allow: /
+Allow: /projects/
+Allow: /projects/*/
+Allow: /pages/
```
unho: If we allow `/` is it necessary to restate these?

dwaynebailey: I added it for clarity of intention, i.e. explicit over implicit

phlax:
my only concern here is that like with the stats pages, there is a risk of unleashing the bots on bad links etc.

My feeling is that this should be checked out a bit before unleashing them on these pages.

phlax:
lgtm, concerns aside

unho:
Despite I would perhaps prefer to blacklist everything and then open specific URLs, lgtm from my side.

dwaynebailey:
@iafan added some docs to the optimisation and customisation sections. This doesn't deal with non-root installs though, didn't want to overburden the docs already with that.

PR #5011, opened 2016-08-03
Editor: ensure checks can be muted/unmuted

### LINKED ISSUE #5010: Unmuting checks doesn't work
1. Open some unit that has muted checks
2. Unmute the check (the background will turn pink)
3. Reload the page

Expected: the unit will have the check unmuted

Observed: the unit has the check muted, and there's no record about unmuting the check in the history. Basically, it's a data loss issue.

julen:
Fixes #5010.


unho:
Reviewed and tested. GTM

PR #5015, opened 2016-08-03
Editor: fix un/muting non-critical checks

### LINKED ISSUE #5013: Muting non-critical checks doesn't work
image

- Style of non-critical check is not changed after it is muted;
- Unmuting seems not to work;

We can go with tow options:
- block muting/unmuting of non-critical checks;
- fix it.

julen:
The JS selector for the DOM element has been missing (dbc038b), so this couldn't
work at all.

Fixes #5013.


unho:
Reviewed and tested. GTM.

PR #5019, opened 2016-08-04
Set Store.is_template on Store creation

phlax:
Adds code/tests to ensure that Store.is_template is correctly set on Store creation

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1280,6 +1280,8 @@ class StoreManager(models.Manager):
         if "filetype" not in kwargs:
             filetypes = ProjectFiletypes(kwargs["translation_project"].project)
             kwargs['filetype'] = filetypes.choose_filetype(kwargs["name"])
+        if kwargs["translation_project"].is_template_project:
+            kwargs["is_template"] = True
         kwargs["pootle_path"] = (
             "%s%s"
             % (kwargs["parent"].pootle_path, kwargs["name"]))
@@ -1287,9 +1289,17 @@ class StoreManager(models.Manager):
 
     def get_or_create(self, *args, **kwargs):
         store, created = super(StoreManager, self).get_or_create(*args, **kwargs)
-        if created and "filetype" not in kwargs:
+        if not created:
+            return store, created
+        update = False
+        if store.translation_project.is_template_project:
+            store.is_template = True
+            update = True
+        if "filetype" not in kwargs:
             filetypes = ProjectFiletypes(store.translation_project.project)
             store.filetype = filetypes.choose_filetype(store.name)
+            update = True
+        if update:
             store.save()
         return store, created
 
--- a/tests/models/store.py
+++ b/tests/models/store.py
@@ -15,7 +15,8 @@ import six
 
 import pytest
 
-from pytest_pootle.factories import LanguageDBFactory
+from pytest_pootle.factories import (
+    LanguageDBFactory, ProjectDBFactory, TranslationProjectFactory)
 from pytest_pootle.utils import update_store
 
 from translate.storage.factory import getclass
@@ -848,10 +849,13 @@ def test_store_create():
         parent=tp.directory,
         translation_project=tp)
     assert store.filetype == po
+    assert not store.is_template
     store = Store.objects.create(
         name="store.pot",
         parent=tp.directory,
         translation_project=tp)
+    # not in source_language folder
+    assert not store.is_template
     assert store.filetype == po
     store = Store.objects.create(
         name="store.xliff",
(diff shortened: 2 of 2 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -1280,16 +1280,27 @@ def create(self, *args, **kwargs):
         if "filetype" not in kwargs:
             filetypes = ProjectFiletypes(kwargs["translation_project"].project)
             kwargs['filetype'] = filetypes.choose_filetype(kwargs["name"])
+        tp = kwargs["translation_project"]
+        if tp.language == tp.project.source_language:
```
ta2-1: `tp.is_template_project`?

phlax: it would be probs be better, but im not entirely clear if this is the best logic

```
    def get_template_translationproject(self):
        """Returns the translation project that will be used as a template                                                                                                                          
        for this project.                                                                                                                                                                           
                                                                                                                                                                                                    
        First it tries to retrieve the translation project that has the                                                                                                                             
        special 'templates' language within this project, otherwise it                                                                                                                              
        falls back to the source language set for current project.                                                                                                                                  
        """
        try:
            return self.translationproject_set.get(language__code='templates')
        except ObjectDoesNotExist:
            try:
                return self.translationproject_set \
                           .get(language=self.source_language_id)
            except ObjectDoesNotExist:
                pass
```

phlax: in particular, im not clear what happens throughout the system if you were to set a source_language (ie not templates) and then also add a templates tp

pootle/apps/pootle_store/models.py
```diff
@@ -1280,16 +1280,27 @@ def create(self, *args, **kwargs):
...
+        if created:
+            tp = store.translation_project
+            if tp.language == tp.project.source_language:
+                store.is_template = True
+                update = True
         if created and "filetype" not in kwargs:
             filetypes = ProjectFiletypes(store.translation_project.project)
             store.filetype = filetypes.choose_filetype(store.name)
+            update = True
+        if update:
```
ta2-1: Hmm... `get_or_create()` will create and update every time if store doesn't exist? Two queries?

phlax: it only updates `if created` - not sure why this would add another query

phlax: ill tidy the conditionals here...

tests/models/store.py
```diff
@@ -882,3 +886,29 @@ def test_store_create():
             name="another_store.foo",
             parent=tp.directory,
             translation_project=tp)
+
+
+@pytest.mark.django_db
+def test_store_create_templates(templates):
+    project = ProjectDBFactory(source_language=templates)
```
ta2-1: I like that we are trying to avoid mess if we have `Templates` and `English` both and project.source_language is `English`. Though `Templates` can be in any language? I'd prefer to have language set explicitly for `Templates`.

pootle/apps/pootle_store/models.py
ta2-1: the first query - `store, created = super(StoreManager, self).get_or_create(*args, **kwargs)`
the second query - `store.save()`

tests/models/store.py
phlax: yep, not sure what is best - to make it *always* defer to source_language would make this more straightforward, but it would be a change from what is there now

pootle/apps/pootle_store/models.py
```diff
@@ -1280,16 +1280,28 @@ def create(self, *args, **kwargs):
...
             % (kwargs["parent"].pootle_path, kwargs["name"]))
         return super(StoreManager, self).create(*args, **kwargs)
 
     def get_or_create(self, *args, **kwargs):
         store, created = super(StoreManager, self).get_or_create(*args, **kwargs)
-        if created and "filetype" not in kwargs:
+        if not created:
+            return store, created
+        update = False
+        tp = store.translation_project
```
unho: Nit: Don't create unnecessary variables.

pootle/apps/pootle_store/models.py
phlax: im not sure this is a big issue at Store creation time to set a foreign key, and really no different to being activated in Store.save

It might be possible to set the kwargs similar to how it is done in create, but tbh i think this makes it more complex

unho:
Other than raised nit, lgtm.

tests/models/store.py
phlax: ive added a separate issue to deal with issues around template authority #5020

ta2-1:
lgtm

Issue #5017, opened 2016-08-03
Stats: empty stats when zero number of top scorers


ta2-1:
Browse page stats table is not filled if there are no top contributors available.

PR #5021, opened 2016-08-04
Followup fixes on #5017

julen:
This PR removes the dependency on the DOM to determine whether there are top contributors or not. It also moves the rendering logic to the component where it belongs.

Refs. https://github.com/translate/pootle/commit/52abd64c99070d75e53d8ca5ff3989c7f14e5c33#commitcomment-18515994

### DIFF
--- a/pootle/static/js/browser/components/Stats.js
+++ b/pootle/static/js/browser/components/Stats.js
@@ -11,7 +11,7 @@ import React from 'react';
 import StatsAPI from 'api/StatsAPI';
 import { t } from 'utils/i18n';
 
-import TopContributors from './TopContributors';
+import TopContributorsTable from './TopContributorsTable';
 
 
 const Stats = React.createClass({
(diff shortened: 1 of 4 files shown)


pootle/static/js/browser/components/Stats.js
```diff
@@ -45,12 +47,19 @@ const Stats = React.createClass({
   },
 
   render() {
+    if (!this.state.topContributors.length) {
```
unho: This way we are not rendering the "Top Contributors for the Last 30 Days" header. Not sure if that is the best idea.

unho:
Reviewed and tested. Other than the raised concern, lgtm.

pootle/static/js/browser/components/Stats.js
ta2-1: @unho see https://github.com/translate/pootle/commit/52abd64c99070d75e53d8ca5ff3989c7f14e5c33#commitcomment-18518144

julen: The behavior is in no way different from the current code, which checks for `top_scorers` in the template.

unho: Yes, my bad.

PR #5026, opened 2016-08-05
Allow translators to auto-accept exact suggestions

### LINKED ISSUE #5025: PermissionDenied in pootle_store.views:manage_suggestion
We have seen several instances of this in MLO server. We are trying to figure out if it is either:

- The link is exposed to users without the required right, or
- The permissions checking in the view is flawed.

unho:
Instead of sending a new translation, suggestions that
match 100% are accepted instead. This helps reducing the
number of suggestions and also credit the involved people.

This was failing because user could have "translate" right
but not "review" right as previously required by the view.

Fixes #5025.

### DIFF
--- a/pootle/apps/pootle_store/decorators.py
+++ b/pootle/apps/pootle_store/decorators.py
@@ -31,7 +31,15 @@ def get_permission_message(permission_code):
     }.get(permission_code, default_message)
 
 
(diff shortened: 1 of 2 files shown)


unho:
Closing since this has been addressed by #5029.

PR #5031, opened 2016-08-08
[RFC] Cleanup/pylint

dwaynebailey:
I did a number of pylint cleanups and added these to permanent checks.

* New checks: uneeded-not, unused-import, unused-variable
* Enabled checking in migrations to get some added cleanups
* With `unused-variables` I took a naming decision that such vars are either `__` (double underscore) or `sensible_variable_name_` (where you still want to convey the variable name but indicate that it is unused with a trailing underscore). 

Would love feedback on what is proposed here.

### DIFF
--- a/pootle/apps/pootle_app/management/commands/update_tmserver.py
+++ b/pootle/apps/pootle_app/management/commands/update_tmserver.py
@@ -122,7 +122,7 @@ class FileParser(object):
                 continue
 
             if os.path.isdir(filename):
-                for dirpath, dirs, fnames in os.walk(filename):
+                for dirpath, dirs_, fnames in os.walk(filename):
                     if (os.path.basename(dirpath) in
                         ["CVS", ".svn", "_darcs", ".git", ".hg", ".bzr"]):
 
--- a/pootle/apps/pootle_app/management/commands/update_tmserver.py
+++ b/pootle/apps/pootle_app/management/commands/update_tmserver.py
@@ -352,5 +352,4 @@ class Command(BaseCommand):
         if self.is_local_tm:
             self._set_latest_indexed_revision(**options)
 
-        success, _ = helpers.bulk(self.es,
-                                  self._parse_translations(**options))
+        helpers.bulk(self.es, self._parse_translations(**options))
--- a/pootle/apps/pootle_fs/resources.py
+++ b/pootle/apps/pootle_fs/resources.py
@@ -163,7 +163,7 @@ class FSProjectStateResources(object):
 
     def reload(self):
         """Uncache cached_properties"""
-        for k, v in self.__dict__.items():
+        for k, v_ in self.__dict__.items():
             if k in ["context", "pootle_path", "fs_path"]:
                 continue
             del self.__dict__[k]
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -142,7 +142,7 @@ class ProjectURLMixin(object):
     """
 
     def get_absolute_url(self):
-        lang, proj, dir, fn = split_pootle_path(self.pootle_path)
+        proj = split_pootle_path(self.pootle_path)[1]
 
         if proj is not None:
             pattern_name = 'pootle-project-browse'
@@ -154,7 +154,7 @@ class ProjectURLMixin(object):
         return reverse(pattern_name, args=pattern_args)
 
     def get_translate_url(self, **kwargs):
-        lang, proj, dir, fn = split_pootle_path(self.pootle_path)
+        proj, dir, fn = split_pootle_path(self.pootle_path)[1:]
 
         if proj is not None:
             pattern_name = 'pootle-project-translate'
@@ -577,7 +577,7 @@ def invalidate_resources_cache(sender, instance, **kwargs):
         (not kwargs['created'] or kwargs['raw'])):
         return
 
-    lang, proj, dir, fn = split_pootle_path(instance.pootle_path)
+    proj = split_pootle_path(instance.pootle_path)[1]
     if proj is not None:
         cache.delete(make_method_key(Project, 'resources', proj))
 
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -464,7 +464,7 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
     def _detect_treestyle(self):
         try:
             dirlisting = os.walk(self.get_real_path())
-            dirpath, dirnames, filenames = dirlisting.next()
+            dirpath_, dirnames, filenames = dirlisting.next()
 
             if not dirnames:
                 # No subdirectories
@@ -479,7 +479,7 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
                 return "nongnu"
 
             # No language subdirs found, look for any translation file
-            for dirpath, dirnames, filenames in os.walk(self.get_real_path()):
+            for dirpath_, dirnames, filenames in os.walk(self.get_real_path()):
                 if filter(self.file_belongs_to_project, filenames):
                     return "gnu"
         except:
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -250,7 +250,7 @@ def check_settings(app_configs=None, **kwargs):
         ))
 
     try:
-        markup_filter, markup_kwargs = settings.POOTLE_MARKUP_FILTER
+        markup_filter = settings.POOTLE_MARKUP_FILTER[0]
     except AttributeError:
         errors.append(checks.Warning(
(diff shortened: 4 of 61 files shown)


phlax:
>  unused-import, unused-variable

These would already be picked up by pyflakes i figure, but no harm to add

> Enabled checking in migrations to get some added cleanups

Not really sure if there is any point to this, as most migrations are created and run automatically. The other migrations are `do once and forget` so there may be some benefit to linting, but it would be pretty limited

> With unused-variables I took a naming decision that such vars are either __

Looks like a good plan

pootle/apps/pootle_app/management/commands/update_tmserver.py
```diff
@@ -352,5 +352,4 @@ def handle(self, **options):
         if self.is_local_tm:
             self._set_latest_indexed_revision(**options)
 
-        success, _ = helpers.bulk(self.es,
-                                  self._parse_translations(**options))
+        __, __ = helpers.bulk(self.es, self._parse_translations(**options))
```
unho: In this case it might make more sense to just remove the assignment.

pootle/apps/pootle_fs/resources.py
```diff
@@ -163,7 +163,7 @@ def pootle_changed(self):
 
     def reload(self):
         """Uncache cached_properties"""
-        for k, v in self.__dict__.items():
+        for k, __ in self.__dict__.items():
```
unho: Why not `v_` following what you did in other places?

pootle/apps/pootle_project/models.py
```diff
@@ -142,7 +142,7 @@ class ProjectURLMixin(object):
     """
 
     def get_absolute_url(self):
-        lang, proj, dir, fn = split_pootle_path(self.pootle_path)
+        __, proj, __, __ = split_pootle_path(self.pootle_path)
```
unho: I would use here `whatever_` syntax.

pootle/apps/pootle_store/migrations/0010_set_store_filetypes.py
```diff
@@ -16,7 +16,6 @@ def migrate_store_filetypes(apps, schema_editor):
 
     projects = apps.get_model("pootle_project.Project").objects.all()
     stores = apps.get_model("pootle_store.Store").objects.all()
-    filetypes = apps.get_model("pootle_format.Format").objects.all()
```
unho: This belongs to a separate cleanup commit.

pootle/apps/pootle_fs/resources.py
dwaynebailey: I've only used `$var_` if I thought it added value to understanding.  I think it's pretty clear here that this is `key`, `value`. And I don't think it helps here.

pootle/core/views.py
```diff
@@ -778,7 +778,7 @@ def get_context_data(self, *args, **kwargs):
             raise Http404(
                 ValidationError(search_form.errors).messages)
 
-        total, start, end, units_qs = search_backend.get(Unit)(
+        total, __, __, units_qs = search_backend.get(Unit)(
```
unho: In cases where several `__` are assigned at once I would keep the `whatever_` syntax.

pootle/checks.py
```diff
@@ -267,11 +267,11 @@ def check_settings(app_configs=None, **kwargs):
...
             try:
                 if markup_filter == 'textile':
-                    import textile  # noqa
+                    import textile  # noqa pylint:disable=unused-variable
                 elif markup_filter == 'markdown':
-                    import markdown  # noqa
+                    import markdown  # noqa pylint:disable=unused-variable
                 elif markup_filter == 'restructuredtext':
-                    import docutils  # noqa
+                    import docutils  # noqa pylint:disable=unused-variable
```
unho: This unused variables cleanups belong to another commit.

pootle/core/markup/filters.py
```diff
@@ -72,11 +72,11 @@ def get_markup_filter():
...
             return (None, "unset")
         elif markup_filter == 'textile':
-            import textile  # noqa
+            import textile  # noqa pylint:disable=unused-variable
         elif markup_filter == 'markdown':
-            import markdown  # noqa
+            import markdown  # noqa pylint:disable=unused-variable
         elif markup_filter == 'restructuredtext':
-            import docutils  # noqa
+            import docutils  # noqa pylint:disable=unused-variable
```
unho: And these unused variables cleanups belong to another commit too.

pootle/apps/pootle_project/models.py
dwaynebailey: I took the decision that `split_pootle_path` is pretty well understood and thus `lang_` clouds the issue and is not useful.  What this does show though is that we probably want helper functions that actually return `proj` from `self.pootle_path`

unho:
lgtm overall, only some nits. Just wondering about Travis, but just restarted the job to see if it is actually picking something.

pootle/core/views.py
dwaynebailey: My rationale was simple.  `__` makes it much more readable then `something_` so I'm aiming for readability.  The only reason to keep vars is if the function creating them is not that well known and you want to keep that in the code.

pootle/apps/pootle_fs/resources.py
phlax: in this case we dont need the placeholder at all, just remove the `.items()`

pootle/apps/pootle_project/models.py
phlax: how about:

```
proj  = split_pootle_path(self.pootle_path)[1]
```

pootle/checks.py
phlax: no!

i defo dont agree to *any* linting that requires disabling comments

phlax: also, this is highlighting bad coding - please dont disable it

dwaynebailey: We already do.  I'm also anti, except where the places we need it are minor and the overall benefit is larger enough.  I.e. where we have no false positives other that these ones which we can tightly control and where we're operating slightly out of normal.

Of course suggestions to tighten the code are the other option.

phlax: these are not false positives

phlax: in fact if anything we should remove the `#noqa` and fix, not hide again

dwaynebailey: Open to suggestions on how to fix the errors that will present here.

dwaynebailey:
Had a quick chat with @phlax to discuss use of `pylint` comments.  Our general conclusion:

1. These are usually bad bits of code and they shouldn't be hidden
2. Land the fixes that the check picked up
3. Flag a cleanup bug and highlight that the check can be enabled if fixed
4. Don't enable check by default till its properly fixed, especially if we have alternatives in other linters that are doing similar things without reporting an issue.

julen:
>> Enabled checking in migrations to get some added cleanups
> 
> Not really sure if there is any point to this, as most migrations are created and run automatically. 

I agree, I don't believe there's much point on this either.

pootle/apps/pootle_fs/resources.py
unho: :+1:

pootle/checks.py
phlax: @dwaynebailey i reckon we could fix this relatively easily with importlib

heres a SO discussion about similar http://stackoverflow.com/questions/14050281/how-to-check-if-a-python-module-exists-without-importing-it

pytest_pootle/utils.py
```diff
@@ -132,7 +132,7 @@ def get_translated_storefile(store, pootle_path=None):
     """Returns file store with added translations for untranslated units."""
     storeclass = store.get_file_class()
     filestore = store.convert(storeclass)
-    for i, unit in enumerate(filestore.units):
+    for i_, unit in enumerate(filestore.units):
```
phlax: if the `i` is not used we can safely remove the enumeration

tests/misc/checks.py
```diff
@@ -402,7 +402,7 @@ def test_get_qualitycheck_schema():
             'url': ''
         })
 
-    result = sorted([item for code, item in d.items()],
+    result = sorted([item for code_, item in d.items()],
                     key=lambda x: x['code'],
```
phlax: we can use `d.values()` and remove `code_`

tests/views/admin.py
```diff
@@ -249,7 +249,7 @@ def test_admin_view_projects(client, request_users, english):
     languages = Language.objects.exclude(code='templates')
     language_choices = [(lang.id, unicode(lang)) for lang in languages]
     filetypes = []
-    for name, info in formats.get().items():
+    for name_, info in formats.get().items():
```
phlax: we can use `formats.get().values()` here to get rid of `name_`

PR #5034, opened 2016-08-09
[WIP] Template authority

### LINKED ISSUE #5020: Make source_language authoritative
atm, if you have a tp with a language called `templates` it overrides the source_language setting for the project

The problem with this is that its implicit, and its not clear what happens when both templates and a a source_language are present

I would propose to

- Only use source_language (and check the code for anywhere that uses templates)
- Add a migration to make the templates tp the source_language anywhere that it is present

This will keep the current behaviour (i believe) and simplify the confusion around template authority

phlax:
- Update ProjectAdminForm with Project.template_tp
- Add Project.template_tp OneToOneField

fixes #5020


pootle/apps/pootle_project/models.py
```diff
@@ -210,6 +210,13 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
     source_language = models.ForeignKey(
         'pootle_language.Language', db_index=True,
         verbose_name=_('Source Language'))
+    template_tp = models.OneToOneField(
+        'pootle_translationproject.TranslationProject',
+        db_index=True,
+        null=True,
+        blank=True,
+        related_name="template_project",
+        verbose_name=_('Templates transation project'))
```
unho: Typo: transation

pootle/apps/pootle_app/views/admin/projects.py
```diff
@@ -41,6 +41,7 @@ def get_context_data(self, **kwargs):
                 'checkstyle': Project.checker_choices,
                 'filetypes': filetypes,
                 'source_language': language_choices,
+                'template_tp': language_choices,
```
unho: But `language_choices` is a `Language` queryset, not a TP queryset. But even if we want to use languages, `templates` language is explicitly being excluded.

pootle/apps/pootle_app/forms.py
```diff
@@ -100,6 +98,51 @@ def clean_code(self):
         return value
 
 
+class ProjectEditForm(ProjectForm):
+    template_tp = forms.ModelChoiceField(
+        label=_('Template tp'),
```
unho: I would use TP in uppercase.

pootle/apps/pootle_app/forms.py
```diff
@@ -100,6 +98,51 @@ def clean_code(self):
...
+        self.fields["template_tp"].choices = Language.objects.values_list(
+            "pk", "fullname")
+
+    def save(self, *args, **kwargs):
+        template_language = self.cleaned_data.pop("template_tp", None)
+        if template_language:
+            # create disabled initially
+            disabled = self.cleaned_data.get("disabled")
+            self.cleaned_data["disabled"] = True
+            template_lang = Language.objects.get(pk=template_language)
```
unho: This can be moved down to where it is actually used.

pootle/apps/pootle_app/forms.py
dwaynebailey: Please be aware the TP, Store, Unit are things we talk about in the Pootle internals.  They are not necessarily things we want to expose to users.

phlax:
rather than making templates tp configurable, we are going to make it always `templates`.

this makes things a lot simpler and negates the need for this PR

PR #5043, opened 2016-08-10
l10n: Use incomplete for untranslated+needs work

unho:
This way it is less likely to be confused with "needs work".
Also aligns with the naming already used in the editor, and
the filters in the editor URLs.


dwaynebailey:
lgtm

PR #5044, opened 2016-08-10
Remove monkey patching i18n from lazy functions

dwaynebailey:
This uses the safe Pootle gettext functions directly instead of monkey patching the Django versions.

### DIFF
--- a/pootle/i18n/gettext.py
+++ b/pootle/i18n/gettext.py
@@ -10,6 +10,7 @@ from translate.lang import data as langdata
 
 from django.conf import settings
 from django.utils import translation
+from django.utils.functional import lazy
 from django.utils.translation import _trans
 
 
@@ -44,6 +45,12 @@ def ungettext(singular, plural, number, variables=None):
 
 def ngettext(singular, plural, number, variables=None):
     return _format_translation(_trans.ngettext(singular, plural, number), variables)
+
+
+gettext_lazy = lazy(gettext, str)
+ugettext_lazy = lazy(ugettext, unicode)
+ngettext_lazy = lazy(ngettext, str)
+ungettext_lazy = lazy(ungettext, unicode)
 
 
 def tr_lang(language_name):
--- a/pootle/apps/accounts/adapter.py
+++ b/pootle/apps/accounts/adapter.py
@@ -10,9 +10,10 @@ import logging
 
 from django.conf import settings
 from django.http import JsonResponse
-from django.utils.translation import ugettext_lazy as _
 
 from allauth.account.adapter import DefaultAccountAdapter
+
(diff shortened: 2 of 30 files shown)


phlax:
lgtm, but sadly travis disagrees

pootle/i18n/gettext.py
```diff
@@ -29,7 +29,9 @@ def _format_translation(message, vars=None):
 
 
 def ugettext(message, vars=None):
-    return _format_translation(translation.ugettext(message), vars)
+    django_message = translation.ugettext(message)
+    safe_message = _format_translation(safe_message, vars)
```
phlax: you dont need to create the extra `safe_message` var

julen:
in 43c32d3 the commit message mentions _remove monkeypatch_, but I don't see monkeypatching being done there — it switches from using a _private_ member to a public one.

Also, when adapting from using Django's `gettext()` to Pootle's `i18n.gettext`, what's the motivation for doing so? I'm assuming the point is safeguarding against variable errors more gracefully as stated by `_format_translation()`, although it'd help to confirm and have a clarifying comment in the commit message.

And speaking of `_format_translation()`, its docstring says:
> Overrides the gettext function, handling variable errors more gracefully.
> This is needed to avoid tracebacks on translation errors with live translation.

Live translation is long gone, therefore I wonder why do we need our own wrappers in the first place?

dwaynebailey:
Nope the commit doesn't remove the monkeypatch part, bad message, later commits are extracting that correctly.

@julen thanks for raising the do we need own gettext functions.  Here are my thoughts

* There is still more that is patched that may or may not be relevant to us in newer Django versions.  Its not just the Gettext functions that have been monkeypatched
* While initially designed for live translation.  Where we didn't want broken variables to cause broken interpolation to cause broken site.  I think that is still useful in that it allows us to release translations that are partially complete without breaking our site.  But the burden of maintenance might be more than the benefit.

julen:
> * There is still more that is patched that may or may not be relevant to us in newer Django versions. Its not just the Gettext functions that have been monkeypatched

I don't believe this relates to the doubt of needing wrappers for gettext or not, which was the doubt I was raising. 
The other monkeypatching is for improved language negotiation and BiDi support. Without checking it deeper, I'm not sure about the former, but the latter could probably be ported upstream to Django.

> While initially designed for live translation. Where we didn't want broken variables to cause broken interpolation to cause broken site. I think that is still useful in that it allows us to release translations that are partially complete without breaking our site. But the burden of maintenance might be more than the benefit.

I don't have a quick way to test this but it'd be good to check if bad interpolation from the localizers' side really breaks sites.

dwaynebailey:
Closing this for now.  Will PR bits as we unpick i18n

PR #5046, opened 2016-08-10
Clear resources dropdown cache when directory has no stores

### LINKED ISSUE #4932: Files dropdown keeps references to empty directories
In https://mozilla.locamotion.org/templates/firefox/toolkit/chrome/mozapps/plugins/ there were files that have been removed a while ago, but you can still select this empty directory in the dropdown. This **plugins** directory is not displayed in the statistics table at https://mozilla.locamotion.org/templates/firefox/toolkit/chrome/mozapps/

unho:
Fixes #4932.

### DIFF
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -583,8 +583,13 @@ def invalidate_resources_cache(**kwargs):
         return
 
     # Don't invalidate if the save didn't create new objects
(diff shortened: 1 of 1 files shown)


unho:
It hasn't been tested yet.

unho:
Merged without tests given that it is not so easy to code some test that verifies cache, and because the fix is needed.

PR #5051, opened 2016-08-11
Editor: allow selecting similar translations' text

### LINKED ISSUE #5009: Allow user to copy "Similar translations" text instead of automatically replacing on click.
1. Go to the edit page for a unit. Eg. https://pootle.com/pootle/projects/translations/translate/#filter=suggestions&unit=3169216
2. Scroll down to the "Similar translations" section at the bottom.
3. Attempt to right click and highlight part of a translation
4. On mouse up, the entire similar translation is copied over the contents of the text box

Our translators have provided feedback that they'd like to be able to copy and paste segments into the text box. They often forget that they can't until it's too late and they'll lose their in-progress translation.

julen:
Potentially this could have been integrated into `copyText()`, although with the
will to avoid near-term conflicts with the raw font branch, a new handler via a
different explicit selector has been added to address this particular case.

Fixes #5009.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -220,6 +220,7 @@ PTL.editor = {
 
     /* Write TM results, special chars... into the currently focused element */
     $('#editor').on('click', '.js-editor-copytext', (e) => this.copyText(e));
+    $('#editor').on('click', '.js-editor-copy-tm-text', (e) => this.copyTMText(e));
 
     /* Copy translator comment */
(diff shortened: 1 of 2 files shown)


phlax:
it would defo be good to limit the number of listeners and handle these together - but as this addresses an immediate  issue i would say lgtm

unho:
@julen How can we test this?

ta2-1:
Tested. GTM.

unho:
I wasn't able to verify on testing. But given it worked for @ta2-1 I am lgtm.

PR #5054, opened 2016-08-11
Cleanups from Django 1.9

dwaynebailey:
These are cleanups picked up in the Django 1.9 branch that are not strictly Django 1.9 related and can be backported to Dango 1.8


unho:
lgtm

PR #5056, opened 2016-08-12
Email suggestion accept/reject comment

unho:


pootle/apps/pootle_store/views.py
```diff
@@ -629,6 +630,26 @@ def handle_suggestion_comment(request, suggestion, comment):
...
+
+        ctx = {
+            'suggestion_id': suggestion.id,
+            'unit_url': request.build_absolute_uri(unit.get_translate_url()),
+            'comment': comment,
+        }
+        if action == "rejected":
+            message = loader.render_to_string(
+                'editor/email/suggestion_rejected_with_comment.txt', ctx)
+            subject = _(u"Suggestion %s rejected with comment" % suggestion.id)
```
ta2-1: I'm not sure re using suggestion id here. Will user be able to use that `id`? He won't afaict.
Probs: "Your suggestion was rejected"?

unho: I think that in case the suggester receives several emails like this it helps differentiate threads, one for each suggestion.

unho: Also using "Yours" with negative feedback can demoralize suggester. Or maybe I am overthinking this.

phlax: if you are going to ref something it should be the unit/unit url - the sugg.id is a leaky abstraction to a user

phlax: just saw it was a subject so not url - but perhaps the unit source?

unho: This is about suggestion, not unit. And the email already includes link to unit in case suggester is curious. The email is just to communicate acceptance/rejection feedback comment.

ta2-1: > I think that in case the suggester receives several emails like this it helps differentiate threads, one for each suggestion.

I'd prefer to get them grouped instead of having separate threads containing the only message.

Probs we can put date into subject?

pootle/apps/pootle_store/views.py
```diff
@@ -629,6 +630,28 @@ def handle_suggestion_comment(request, suggestion, comment):
...
+        else:
```
phlax: how about

```
+            message = loader.render_to_string(
+                'editor/email/suggestion_%s_with_comment.txt' % action, ctx)
+            subject = _(u"Suggestion %s %s with comment" % (suggestion.id, action))
```

unho: That is breaking l10n. So no.

unho: Well, regarding the l10n message. To pick the template it is possible, but I still feel it is better to keep the full string in order to ease grep.

phlax: im not really seeing why it would break l10n

phlax: also - the suggestion id is of **no** use to the user - so there is no reason to include in the subject

unho: The aim of that was to keep email threads separate, but since you are pestering about it, it is just quicker to please you all. Removing then.

PR #5058, opened 2016-08-12
Editor: do not append history entry on filter initialization

### LINKED ISSUE #4793: Sometimes "Back" button in browser doesn't work
- go to browse view page https://mozilla.locamotion.org/projects/ios/
- click on any of `Need works` links to open translate view page
- click on "Back" button in browser

Sometimes it works but sometimes you are redirected on the same translate view page.

julen:
While having the unit id in the URL for initial loads is desirable, the existing
implementation prevents the browser back button from working as expected.

The ideal solution would be using the history API instead of relying on the
`onhashchange` event, although while that is not in place, not updating the hash
on the first load is an acceptable compromise — in fact, this is how it worked
until this issue was introduced.

Fixes #4793.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -571,7 +571,6 @@ PTL.editor = {
           } else {
             uId = this.units.uIds[0];
           }
-          $.history.load(utils.updateHashPart('unit', uId));
         }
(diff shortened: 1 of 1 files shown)


phlax:
lgtm

PR #5062, opened 2016-08-13
Add tp_tool for manipulating TranslationProjects

phlax:
allows us to reassign/clone/update a tp to a different language (and does the required path-fu)

this is required to be able to create a testable migration for template tps, and for monolingual support

it will also be useful in the future for copying/cloning/moving tps more generally

### DIFF
--- a/pootle/apps/pootle_translationproject/utils.py
+++ b/pootle/apps/pootle_translationproject/utils.py
@@ -1,0 +1,163 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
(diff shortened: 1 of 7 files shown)


pootle/apps/pootle_translationproject/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    def update_store(self, source, target):
+        differ = StoreDiff(target, source, target.get_max_unit_revision() + 1)
+        diff = differ.diff()
+        if diff is None:
+            return
+        system = User.objects.get_system_user()
+        update_revision = Revision.incr()
+        target.update_from_diff(
+            source,
+            target.get_max_unit_revision() + 1,
```
ta2-1: can we reuse `target.get_max_unit_revision()` from the first line in this method?

phlax: yep, updated

ta2-1:
@phlax Tested and GTM 👍 .

PR #5065, opened 2016-08-13
Improvements to StoreDiff

phlax:
- make it pluggable
- make it work for diffing 2 db stores

### DIFF
--- a/pootle/apps/pootle_store/diff.py
+++ b/pootle/apps/pootle_store/diff.py
@@ -9,7 +9,10 @@
 import difflib
 from collections import OrderedDict
 
+from django.db import models
 from django.utils.functional import cached_property
+
+from pootle.core.delegate import format_diffs
 
 from .fields import to_python as multistring_to_python
 from .unit import UnitProxy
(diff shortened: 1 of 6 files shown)


pootle/apps/pootle_store/diff.py
```diff
@@ -51,32 +51,37 @@ def target(self):
 
 class StoreDiff(object):
 
-    def __init__(self, db_store, file_store, file_revision):
-        self.db_store = db_store
-        self.file_store = file_store
-        self.db_revision = db_store.get_max_unit_revision()
-        self.file_revision = file_revision
+    def __init__(self, target_store, source_store, source_revision):
```
ta2-1: @phlax target before source?

pootle/apps/pootle_store/diff.py
```diff
@@ -51,32 +51,37 @@ def target(self):
...
+        self.target_store = target_store
+        self.source_store = source_store
+        self.source_revision = source_revision
+
+    @cached_property
+    def target_revision(self):
+        return self.target_store.get_max_unit_revision()
 
     @cached_property
     def active_units(self):
```
ta2-1: active_target_units?

phlax:
@ta2-1 in this case i think yes - because the target is the thing to be updated - and the update code does something similar

not sure tho - thought about this quite a bit and didnt come to a firm conclusion.

also worth mentioning - this is how it is now - just that it uses more confusing vars

pootle/apps/pootle_store/diff.py
phlax: yep, good idea

pootle/apps/pootle_store/diff.py
phlax: @ta2-1 in this case i think yes - because the target is the thing to be updated - and the update code does something similar

not sure tho - thought about this quite a bit and didnt come to a firm conclusion.

also worth mentioning - this is how it is now - just that it uses more confusing vars

pootle/apps/pootle_store/diff.py
```diff
@@ -51,32 +51,37 @@ def target(self):
...
-        self.db_revision = db_store.get_max_unit_revision()
-        self.file_revision = file_revision
+    def __init__(self, target_store, source_store, source_revision):
+        self.target_store = target_store
+        self.source_store = source_store
+        self.source_revision = source_revision
+
+    @cached_property
+    def target_revision(self):
+        return self.target_store.get_max_unit_revision()
```
ta2-1: IIRC we should get db_revision before we start update process to avoid any of race conditions.
I'm not clear when we call this method first time in the current implementation.

phlax: k - ill put back or make sure its initialized in __init__

tests/models/store.py
```diff
@@ -984,3 +986,207 @@ def test_store_get_or_create_templates(templates):
...
+    assert result["update"][0] == set([update_unit.pk])
+    assert result["update"][1] == {}
+    assert len(result["add"]) == 0
+    assert len(result["index"]) == 0
+
+    differ = StoreDiff(
+        target_store,
+        source_store,
+        0)
+    assert result["update"][0] == set([update_unit.pk])
```
ta2-1: `result = differ.diff()` is missing.

phlax: good spot!

tests/models/store.py
```diff
@@ -984,3 +986,207 @@ def test_store_get_or_create_templates(templates):
...
+    result = differ.diff()
+    assert result["update"][0] == set([target_unit.pk])
+    assert result["update"][1] == {}
+    assert len(result["add"]) == 0
+    assert len(result["index"]) == 0
+    differ = StoreDiff(
+        target_store,
+        source_store,
+        0)
+    assert result["update"][0] == set([target_unit.pk])
```
ta2-1: `result = differ.diff()` is missing.

phlax: at least im consistent with my mistakes 8/

tests/models/store.py
```diff
@@ -984,3 +986,207 @@ def test_store_get_or_create_templates(templates):
...
+    assert result["update"][1] == {}
+    assert len(result["add"]) == 0
+    assert len(result["index"]) == 0
+
+
+@pytest.mark.django_db
+def test_store_diff_custom(diffable_stores):
+    target_store, source_store = diffable_stores
+    start_receivers = format_diffs.receivers
```
phlax: move to fixture...

pootle/apps/pootle_store/diff.py
```diff
@@ -49,53 +52,117 @@ def target(self):
...
+        return diffs["default"]
 
-    def __init__(self, db_store, file_store, file_revision):
-        self.db_store = db_store
-        self.file_store = file_store
-        self.db_revision = db_store.get_max_unit_revision()
-        self.file_revision = file_revision
+    @cached_property
+    def target_revision(self):
+        return self.target_store.get_max_unit_revision()
```
ta2-1: IIRC we should get db_revision before we start update process to avoid any of race conditions.
I'm not clear when we call this method first time in the current implementation.

phlax: as said above - ill move or initializie in `__init__`

ta2-1:
@phlax lgtm after comments are addressed. thanks for your great job.

PR #5066, opened 2016-08-13
Template authority 2

### LINKED ISSUE #5020: Make source_language authoritative

phlax:
fixes #5020 

makes templates tp always be called `templates`


pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
```diff
@@ -0,0 +1,50 @@
...
+    Language = apps.get_model("pootle_language.Language")
+    to_update = None
+    for project in projects:
+        source_lang = project.source_language
+        try:
+            templates = project.translationproject_set.get(
+                language__code="templates")
+            # there is a tp called "templates"
+            # - dont do anything
+            continue
```
unho: Nit: move the `continue` to an `else` clause.

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
```diff
@@ -0,0 +1,50 @@
...
+        except TP.DoesNotExist:
+            pass
```
unho: Instead of using `try` clauses, why not use `exists` on the queryset? IMHO it would be cleaner code.

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
unho: Even better, instead of using `try` clauses, why not use `exists` on the queryset? IMHO it would be cleaner code.

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
```diff
@@ -0,0 +1,50 @@
...
+        except TP.DoesNotExist:
+            # there is nothing that looks like a templates tp
+            continue
+        to_update.append((project, templates))
+    if not to_update:
+        return
+    try:
+        template_lang = Language.objects.get(code="templates")
+    except:
+        template_lang = Language.objects.create(code="templates")
```
unho: Use `get_or_create` https://docs.djangoproject.com/en/1.8/ref/models/querysets/#get-or-create

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
phlax: im not clear wheter exists would create an additional query

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
phlax: i had a problem with get_or_create - but now i think it was probs unrelated

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
```diff
@@ -0,0 +1,50 @@
...
+from pootle_translationproject.utils import TPTool
+
+
+def rename_templates(apps, schema_editor):
+    projects = apps.get_model("pootle_project.Project").objects.all()
+    TP = apps.get_model("pootle_translationproject.TranslationProject")
+    Language = apps.get_model("pootle_language.Language")
+    to_update = None
+    for project in projects:
+        source_lang = project.source_language
```
unho: I would postpone this until we are sure no `templates` TP exists for the project.

pootle/apps/pootle_translationproject/migrations/0004_rename_templates.py
unho: Who cares. This is a migration running once. And you are not going to deal with a trillion projects scenario.

unho:
Some blank lines in the function's code won't kill anyone.

PR #5073, opened 2016-08-14
Cleanup/pylint fixes (more)

phlax:
more cleanups


unho:
This is a nice work of cleanup, but I am just wondering if it would be easier to change the regexp to match `kwargs` as well as any variable name ending with `_` or being `__` in order to skip when checking for unused variables.

phlax:
no - i think its better to be explicit that we want to ignore in these cases - rather than blanket ignoring kwargs

PR #5074, opened 2016-08-14
i18n: adapt Django ugettext instances to Pootle gettext

phlax:
use pootles `_` explicitly

### DIFF
--- a/pootle/apps/pootle_app/views/admin/dashboard.py
+++ b/pootle/apps/pootle_app/views/admin/dashboard.py
@@ -16,13 +16,13 @@ from django.contrib.auth import get_user_model
 from django.core.cache import cache
 from django.http import HttpResponse
 from django.shortcuts import render
-from django.utils.translation import ugettext as _, ungettext
 
 from django_rq.queues import get_failed_queue, get_queue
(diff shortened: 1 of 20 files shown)


julen:
It would be great to have at least some clarity on this one. I spent time reviewing and providing my impressions on #5044 (which the commit in this PR originates from) along with raising some doubts, and now that I see this PR is being merged without further ado it only makes me feel like it has been miserably ignored 😞.

phlax:
@julen what was there was wrong in every way - it was using monkeypatched functions - this was causing **lots** of pylint **errors** - and wasted a load of my time.

we have not done away with the pootle.i18n functions, nor (for now) have we removed the monkey patch. What was done here was to make these `_` calls explicitaly use the functions that they are using.

julen:
I know there's monkeypatching and I know about some of the motivations behind it. No one said it's cool and no one said it needs to stay.

From my side I took some time to review the suggestion proposed in the PR and I provided my input to improve it and raised some doubts whether even the wrappers were needed. If you prefer to ignore that and listen only to pylint to increase a score value, that's your call.

PR #5076, opened 2016-08-14
Mozilla lang

phlax:
Add native support for the mozilla_lang format

### DIFF
--- a/pootle/apps/pootle_format/formats/mozilla_lang.py
+++ b/pootle/apps/pootle_format/formats/mozilla_lang.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from pootle_store.constants import FUZZY, UNTRANSLATED
+from pootle_store.diff import DiffableStore
+from pootle_store.syncer import StoreSyncer, UnitSyncer
+
+
+class LangUnitSyncer(UnitSyncer):
+
+    @property
+    def target(self):
+        if self.isfuzzy:
+            return ""
+        return self.unit.target
+
+
+class LangStoreSyncer(StoreSyncer):
+    unit_sync_class = LangUnitSyncer
+
+
+class DiffableLangStore(DiffableStore):
+
+    def get_unit_state(self, file_unit):
+        return (
+            FUZZY
+            if (file_unit["unitid"] in self.target_units
+                and self.target_units[file_unit["unitid"]]["state"] == FUZZY
+                and file_unit["state"] == UNTRANSLATED)
+            else file_unit["state"])
+
+    def get_unit_target(self, file_unit):
+        return (
+            self.target_units[file_unit["unitid"]]["target_f"]
+            if file_unit["state"] == FUZZY
+            else file_unit["target"])
+
+    def get_file_unit(self, unit):
+        file_unit = super(DiffableLangStore, self).get_file_unit(unit)
+        file_unit["state"] = self.get_unit_state(file_unit)
+        file_unit["target"] = self.get_unit_target(file_unit)
+        return file_unit
--- a/tests/formats/mozilla_lang.py
+++ b/tests/formats/mozilla_lang.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from pytest_pootle.factories import StoreDBFactory
+
+from pootle_format.models import Format
+from pootle_store.constants import FUZZY, TRANSLATED
+from pootle_store.models import Unit
+
+
+@pytest.mark.django_db
+def test_mozlang_update(tp0):
+    mozlang = Format.objects.get(name="lang")
+    tp0.project.filetypes.add(mozlang)
+
+    foo_lang = StoreDBFactory(
+        name="foo.lang",
+        filetype=mozlang,
+        parent=tp0.directory,
+        translation_project=tp0)
+
+    store0 = tp0.stores.get(name="store0.po")
+
+    # deserialize as source
+    foo_lang.update(store0.deserialize(store0.serialize()))
+
+    # get serialized lang store
+    serialized = foo_lang.serialize()
+
+    # mark a translated unit as fuzzy
+    translated = foo_lang.units.filter(state=TRANSLATED).first()
+    translated.state = FUZZY
+    translated.save()
+
+    # source is translated
+    foo_lang.update(foo_lang.deserialize(serialized))
+    # unit is set back to TRANSLATED
+    translated = Unit.objects.get(pk=translated.pk)
+    assert translated.state == TRANSLATED
+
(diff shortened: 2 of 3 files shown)


unho:
lgtm

tests/formats/mozilla_lang.py
```diff
@@ -0,0 +1,68 @@
...
+    ttkunit.target = ""
+    foo_lang.update(ttk)
+    # unit stays FUZZY
+    translated = Unit.objects.get(pk=translated.pk)
+    assert translated.state == FUZZY
+
+    # NOT WORKING - bug in ttk it seems
+
+    # set source == target > untranslated
+    # ttkunit.target = ttkunit.source
```
ta2-1: this line is wrong. ttk api doesn't know anything about `{ok}`.
if you set target for a ttkunit then this means the unit is translated.

phlax: not in this spec

dwaynebailey: I think @ta2-1 is correct, you're dealing with internal ttk unit not a moz_lang unit here.  It is true one serialised that this would man untranslated.

phlax: (Pdb) unit
<translate.storage.mozilla_lang.LangUnit object at 0x7fce72e5b2d0>
(Pdb) unit.source
'Open'
(Pdb) unit.target
(Pdb) unit.istranslated()
False
(Pdb) unit.target = "Something else"
(Pdb) unit.istranslated()
True
(Pdb) unit.target = "Open"
(Pdb) unit.istranslated()

phlax: whatever way i look at this - this is broken.

its not a po unit its `translate.storage.mozilla_lang.LangUnit` and i would expect it to follow the spec

phlax: this is what i would expect

```
(Pdb) unit.source
'Open'
(Pdb) unit
<translate.storage.mozilla_lang.LangUnit object at 0x7f3b503da2d0>
(Pdb) unit.target
''
(Pdb) unit.istranslated()
False
(Pdb) unit.target = unit.source
(Pdb) unit.target
'Open'
(Pdb) unit.istranslated()
False
(Pdb) unit.target = "something else"
(Pdb) unit.istranslated()
True
(Pdb) unit.target = "something else {ok}"
(Pdb) unit.istranslated()
True
(Pdb) unit.target
'something else'
 
```

phlax: ref https://github.com/translate/translate/pull/3484

dwaynebailey: The idea of ttk is that you don't need to know how the format works.  Setting target to `Open` and knowing that in .lang that would mean it is untranslated means you need to knw how the format is operating.  The job of ttk is to hide that and its doing it quit well.

As the user of .lang you shouldn't be setting {ok} at all or even knowing that you have to do that.  It's a completely stupid idea built because someone realised after the fact that their format is in fact broken.  So as a user of the format you shouldn't need to care.  And ttk handling means that you don't.

phlax: the point is this - with the current code *you cant manipulate the units without reparsing the store*

this is a completely arbitrary limitation - and it means `LangUnit` meets neither po nor lang's spec

pootle/apps/pootle_format/formats/mozilla_lang.py
```diff
@@ -0,0 +1,47 @@
...
+from pootle_store.syncer import StoreSyncer, UnitSyncer
+
+
+class LangUnitSyncer(UnitSyncer):
+
+    @property
+    def target(self):
+        if self.isfuzzy:
+            return ""
+        return self.unit.target
```
phlax: i think this should be `self.unit.source`

pootle/apps/pootle_format/formats/mozilla_lang.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.unit.source
+
+
+class LangStoreSyncer(StoreSyncer):
+    unit_sync_class = LangUnitSyncer
+
+
+class DiffableLangStore(DiffableStore):
+
+    def get_unit_state(self, file_unit):
```
unho: I am comparing this class to `DiffableStore` and the latter doesn't have this method despite it makes sense to have it there.

unho: I won't block on it now that I realize this will require extensive refactoring.

pootle/apps/pootle_format/formats/mozilla_lang.py
```diff
@@ -0,0 +1,47 @@
...
+from pootle_store.syncer import StoreSyncer, UnitSyncer
+
+
+class LangUnitSyncer(UnitSyncer):
+
+    @property
+    def target(self):
+        if self.isfuzzy:
+            return ""
+        return self.unit.source
```
unho: Why returning source? I was under the impression this is a bilingual format...

tests/formats/mozilla_lang.py
```diff
@@ -0,0 +1,85 @@
...
+    foo_lang.update(store0.deserialize(store0.serialize()))
+
+    # mark the unit as fuzzy
+    unit = foo_lang.units.filter(state=TRANSLATED).first()
+    unit.markfuzzy()
+    unit.save()
+
+    ttk = foo_lang.deserialize(foo_lang.serialize())
+    ttk_unit = ttk.findid(unit.getid())
+    assert not ttk_unit.istranslated()
```
unho: `assert ttk_unit.isfuzzy()` ?

pootle/apps/pootle_format/formats/mozilla_lang.py
```diff
@@ -0,0 +1,47 @@
...
+            if (file_unit["unitid"] in self.target_units
+                and self.target_units[file_unit["unitid"]]["state"] == FUZZY
+                and file_unit["state"] == UNTRANSLATED)
+            else file_unit["state"])
+
+    def get_unit_target(self, file_unit):
+        return (
+            self.target_units[file_unit["unitid"]]["target_f"]
+            if file_unit["state"] == FUZZY
+            else file_unit["target"])
```
unho: I am not following why this is needed.

pootle/apps/pootle_format/formats/mozilla_lang.py
phlax: the format expects `source == target` for untranslated

pootle/apps/pootle_format/formats/mozilla_lang.py
phlax: if the unit is fuzzy in pootle, but untranslated in file - its treated as fuzzy

pootle/apps/pootle_format/formats/mozilla_lang.py
unho: This is returning the source for translated units.

phlax: ahh - good point - ill fix

tests/formats/mozilla_lang.py
phlax: no - its not fuzzy at this point its untranslated - its only fuzzy in pootle

pootle/apps/pootle_format/formats/mozilla_lang.py
phlax: turns out not

pootle/apps/pootle_format/formats/mozilla_lang.py
phlax: Its not really a DiffableStore method - more of a DiffableUnit one - but in this case it needs to happen here as we have access to both stores

unho:
lgtm

PR #5079, opened 2016-08-15
Cleanup: renames of 'obj' and 'split_pootle_path

dwaynebailey:
* Since `object` is a builtin in Python, prefer `obj` instead of `ob`
* `split_pootle_path` has various naming conventions for the returned values.  Renamed `lang`, `proj`, `dir`, `fn` to more explicit `lang_code`, `proj_code`, `dir_path` (avoids a builtin also), `filename`.  Where `language_code` or `project_code` i.e. in fullform exists I've changed some but left others where the code felt too complicated to intervene.

### DIFF
--- a/pootle/apps/pootle_comment/forms.py
+++ b/pootle/apps/pootle_comment/forms.py
@@ -52,13 +52,13 @@ class CommentForm(DjCommentForm):
             raise CommentNotSaved(dict(comment=should_not_save))
 
     def save(self):
-        ob = self.comment
-        ob.user = self.cleaned_data["user"]
-        ob.submit_date = datetime.now()
-        ob.save()
+        comment = self.comment
+        comment.user = self.cleaned_data["user"]
+        comment.submit_date = datetime.now()
+        comment.save()
         comment_was_saved.send(
-            sender=ob.__class__,
-            comment=ob)
+            sender=comment.__class__,
+            comment=comment)
 
 
 class UnsecuredCommentForm(CommentForm):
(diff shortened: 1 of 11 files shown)


pootle/apps/pootle_comment/forms.py
```diff
@@ -52,13 +52,13 @@ def clean(self):
             raise CommentNotSaved(dict(comment=should_not_save))
 
     def save(self):
-        ob = self.comment
-        ob.user = self.cleaned_data["user"]
-        ob.submit_date = datetime.now()
-        ob.save()
+        obj = self.comment
```
phlax: how about we make this one `comment` -  we dont really need a local var at all tbh tho

tests/views/tp.py
```diff
@@ -53,15 +53,15 @@ def _test_browse_view(tp, request, response, kwargs):
     pootle_path = "%s%s" % (tp.pootle_path, resource_path)
 
     if not (kwargs["dir_path"] or kwargs.get("filename")):
-        ob = tp.directory
```
phlax: im wondering if this would be better as `directory`

phlax:
nits aside - lgtm

tests/views/tp.py
dwaynebailey: In the `if` clause we have `obj = Store.objects.get(pootle_path=pootle_path)` so they're not always directories.  So I felt I couldn't use that.

phlax: perhaps `context` then? i usually think `obj` is a var of last resort

phlax: ...or where it can be objects of totally different sorts

feel free to ignore  - its a nit on a test

PR #5081, opened 2016-08-15
Fix: only use live units in StoreDiff.source_db_store

phlax:
when doing a diff from a db Store with obsolete units it was trying to readd them to target


ta2-1:
lgtm

PR #5085, opened 2016-08-16
Fix broken admin URLs

unho:
Now we can use /admin/languages/ instead of /adminlanguages/

This was exposed by 3bef6f52.


ta2-1:
lgtm

dwaynebailey:
lgtm tested on staging.

PR #5091, opened 2016-08-17
Pluggable sync

phlax:
adds pluggable sync from Stores.

this allows us to handle different formats or templates - and refactors the code in Store.sync

### DIFF
--- a/pootle/apps/pootle_format/formats/po.py
+++ b/pootle/apps/pootle_format/formats/po.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from translate.storage import poheader
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.utils import timezone
+
+from pootle.core.mixins import CachedMethods
+from pootle.core.utils import dateformat
+from pootle.core.utils.timezone import datetime_min
+from pootle.core.utils.version import get_major_minor_version
+from pootle_statistics.models import Submission
+from pootle_store.syncer import StoreSyncer
+
+
+class PoStoreSyncer(StoreSyncer):
+
+    def get_latest_submission(self, mtime):
+        user_displayname = None
+        user_email = None
+        fields = (
+            "submitter__username",
+            "submitter__full_name",
+            "submitter__email")
+        submissions = self.translation_project.submission_set.exclude(
+            submitter__username="nobody")
+        try:
+            username, fullname, user_email = (
+                submissions.filter(creation_time=mtime)
+                           .values_list(*fields).latest())
+        except Submission.DoesNotExist:
+            try:
+                _mtime, username, fullname, user_email = (
+                    submissions.values_list("creation_time", *fields)
+                               .latest())
+                mtime = min(_mtime, mtime)
+            except ObjectDoesNotExist:
+                pass
+        if user_email:
+            user_displayname = (
+                fullname.strip()
+                if fullname.strip()
+                else username)
+        return mtime, user_displayname, user_email
+
+    def get_po_revision_date(self, mtime):
+        return (
+            "%s%s"
+            % (mtime.strftime('%Y-%m-%d %H:%M'),
+               poheader.tzstring()))
+
+    def get_po_mtime(self, mtime):
+        return (
+            '%s.%06d'
+            % (int(dateformat.format(mtime, 'U')),
+               mtime.microsecond))
+
+    def get_po_headers(self, mtime, user_displayname, user_email):
+        headerupdates = {
+            'PO_Revision_Date': self.get_po_revision_date(mtime),
+            'X_Generator': "Pootle %s" % get_major_minor_version(),
+            'X_POOTLE_MTIME': self.get_po_mtime(mtime)}
+        headerupdates['Last_Translator'] = (
+            user_displayname and user_email
+            and ('%s <%s>'
+                 % (user_displayname,
+                    user_email))
+            or 'Anonymous Pootle User')
+        return headerupdates
+
+    def update_po_headers(self, mtime, user_displayname, user_email):
+        self.disk_store.updateheader(
+            add=True,
+            **self.get_po_headers(mtime, user_displayname, user_email))
+        if self.language.nplurals and self.language.pluralequation:
+            self.disk_store.updateheaderplural(
+                self.language.nplurals,
+                self.language.pluralequation)
+
+    def update_store_header(self, **kwargs):
+        super(PoStoreSyncer, self).update_store_header(**kwargs)
+        user = kwargs.get("user")
+        mtime = self.store.get_cached_value(CachedMethods.MTIME)
+        if mtime is None or mtime == datetime_min:
+            mtime = timezone.now()
+        user_displayname = None
+        user_email = None
+        if user is None:
+            (mtime,
+             user_displayname,
+             user_email) = self.get_latest_submission(mtime)
+        elif user.is_authenticated:
+            user_displayname = user.display_name
+            user_email = user.email
+        self.update_po_headers(mtime, user_displayname, user_email)
--- a/pootle/apps/pootle_store/syncer.py
+++ b/pootle/apps/pootle_store/syncer.py
@@ -1,0 +1,354 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+import os
+from collections import namedtuple
+
+from translate.storage.factory import getclass
+
+from django.utils.functional import cached_property
+
+from pootle.core.delegate import format_classes
+from pootle.core.log import log
+
+from .models import Unit
+from .util import get_change_str
+
+
+class UnitSyncer(object):
+
+    def __init__(self, unit):
+        self.unit = unit
+
+    @property
+    def context(self):
+        return self.unit.getcontext()
+
+    @property
+    def developer_notes(self):
+        return self.unit.getnotes(origin="developer")
+
+    @property
+    def isfuzzy(self):
+        return self.unit.isfuzzy()
+
+    @property
+    def isobsolete(self):
+        return self.unit.isobsolete()
+
+    @property
+    def locations(self):
+        return self.unit.getlocations()
+
+    @property
+    def source(self):
+        return self.unit.source
+
+    @property
+    def target(self):
+        return self.unit.target
+
+    @property
+    def translator_notes(self):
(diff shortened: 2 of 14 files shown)


tests/models/unit.py
```diff
@@ -317,3 +318,108 @@ def test_unit_ts_plurals(store_po, test_fs):
...
+
+
+@pytest.mark.django_db
+def test_unit_syncer_fuzzy(unit_syncer):
+    unit, unit_class = unit_syncer
+    syncer = UnitSyncer(unit)
+    newunit = syncer.convert(unit_class)
+    unit.state = FUZZY
+    unit.save()
+    newunit = syncer.convert(unit_class)
```
unho: Why `newunit = syncer.convert(unit_class)` twice?

pootle/apps/pootle_format/formats/po.py
```diff
@@ -0,0 +1,102 @@
...
+from pootle.core.utils import dateformat
+from pootle.core.utils.timezone import datetime_min
+from pootle.core.utils.version import get_major_minor_version
+from pootle_statistics.models import Submission
+from pootle_store.syncer import StoreSyncer
+
+
+class PoStoreSyncer(StoreSyncer):
+
+    def get_latest_submission(self, mtime):
```
ta2-1: why is get_latest_submission in PoStoreSyncer?

tests/models/unit.py
phlax: the first is not required ill remove

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,365 @@
...
+    @property
+    def source(self):
+        return self.unit.source
+
+    @property
+    def target(self):
+        return self.unit.target
+
+    @property
+    def translator_notes(self):
```
unho: Nit: Put all the notes properties together.

pootle/apps/pootle_format/formats/po.py
phlax: it wants to get a user for the po header

pootle/apps/pootle_store/syncer.py
phlax: ive put the properties in alphabetical order - i think that makes more sense

pootle/apps/pootle_format/formats/po.py
```diff
@@ -0,0 +1,102 @@
...
+            except ObjectDoesNotExist:
+                pass
+        if user_email:
+            user_displayname = (
+                fullname.strip()
+                if fullname.strip()
+                else username)
+        return mtime, user_displayname, user_email
+
+    def get_po_revision_date(self, mtime):
```
ta2-1: do we need to use `po` in method names of PoStoreSyncer?

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,365 @@
...
+    def set_locations(self, newunit):
+        locations = self.locations
+        if locations:
+            newunit.addlocations(locations)
+
+    def set_obsolete(self, newunit):
+        if self.isobsolete:
+            newunit.makeobsolete()
+
+    def set_notes(self, newunit):
```
unho: I would drop this one.

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,365 @@
...
+        self.set_developer_notes(newunit)
+        self.set_translator_notes(newunit)
+
+    def set_target(self, newunit):
+        newunit.target = self.target
+
+    def set_translator_notes(self, newunit):
```
unho: Nit: Put close to `set_developer_notes`.

pootle/apps/pootle_format/formats/po.py
phlax: tbh - i think they should probs get factored out at some stage - but yep i would rather keep it clear that this is not part of the default format sync

pootle/apps/pootle_store/syncer.py
phlax: im not sure i follow

pootle/apps/pootle_store/syncer.py
phlax: they are in alphabetical order

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,365 @@
...
+            not self.store.file.exists()
+            or (last_revision >= self.store.last_sync_revision))
+
+    def get_dbid_index(self, update=False, obsolete=False):
+        """build a quick mapping index between unit ids and database ids"""
+        if update or not hasattr(self.store, "dbid_index"):
+            units = self.store.unit_set.all()
+            if not obsolete:
+                units = units.filter(state__gt=OBSOLETE)
+        return dict(units.values_list('unitid', 'id'))
```
phlax: good spot

pootle/apps/pootle_format/formats/po.py
```diff
@@ -0,0 +1,102 @@
...
+            '%s.%06d'
+            % (int(dateformat.format(mtime, 'U')),
+               mtime.microsecond))
+
+    def get_po_headers(self, mtime, user_displayname, user_email):
+        headerupdates = {
+            'PO_Revision_Date': self.get_po_revision_date(mtime),
+            'X_Generator': "Pootle %s" % get_major_minor_version(),
+            'X_POOTLE_MTIME': self.get_po_mtime(mtime)}
+        # FIXME: maybe insert settings.POOTLE_TITLE or domain here?
```
unho: Drop this fixme.

tests/models/unit.py
```diff
@@ -317,3 +318,108 @@ def test_unit_ts_plurals(store_po, test_fs):
...
+    assert not newunit.isobsolete()
+    assert not newunit.istranslated()
+    _test_unit_syncer(unit, newunit)
+
+
+@pytest.mark.django_db
+def test_unit_syncer_untranslated(unit_syncer):
+    unit, unit_class = unit_syncer
+    syncer = UnitSyncer(unit)
+    newunit = syncer.convert(unit_class)
```
ta2-1: remove this too

pootle/apps/pootle_store/syncer.py
unho: Instead of adding this method that just calls two other methods, just get rid of this method and replace any call to this method by the two calls to the other two methods.

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,365 @@
...
+    def convert(self, fileclass=None):
+        """export to fileclass"""
+        fileclass = fileclass or self.file_class
+        logging.debug(
+            u"Converting %s to %s",
+            self.store.pootle_path,
+            fileclass)
+        output = fileclass()
+        output.settargetlanguage(self.language.code)
+        # FIXME: we should add some headers
```
ta2-1: should we?

phlax: this was a leftover - im not clear where the headers should be inserted - i think here is probs better - but i would like to look at this after

unho: Probably here `update_store_header` should be called. Then leave to target format to do their thing.

pootle/apps/pootle_store/syncer.py
unho: Yes, but luckily it doesn't happen in current code.

pootle/apps/pootle_store/syncer.py
phlax: lets deal with this after - changing this could significantly change the behaviour of po being im/exported

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,363 @@
...
+        self.store.save()
+
+    def update_newer(self, last_revision):
+        return (
+            not self.store.file.exists()
+            or (last_revision >= self.store.last_sync_revision))
+
+    def get_dbid_index(self, update=False, obsolete=False):
+        """build a quick mapping index between unit ids and database ids"""
+        if update or not hasattr(self.store, "dbid_index"):
```
ta2-1: afaict `store.dbid_index` was defined by `require_dbid_index`. this PR drops `require_dbid_index` method. I wonder if   `hasattr(self.store, "dbid_index")` condition is still necessary here.

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,363 @@
...
+            del_unit = True
+        elif not conservative:
+            obsolete = 1
+            unit.makeobsolete()
+            if not unit.isobsolete():
+                del_unit = True
+        if del_unit:
+            deleted = 1
+            del unit
+        return obsolete, deleted
```
unho: I would simplify it like:

```python
        obsoleted = unit.istranslated() and not conservative
        deleted = (not unit.istranslated() or
                   (not conservative and not unit.isobsolete()))

        if obsoleted:
            unit.makeobsolete()

        if deleted:
            del unit

        return obsolete, deleted
```

phlax: it doesnt catch the logic correctly

in the original it only deletes an obsolete unit if obsoleting didnt work

ta2-1: I was trying to get this logic during some time. So this is about that if format supports obsoleting, yes?

ta2-1: and if a file format supports obsoleting then clashes are possible. because obsoleting and deleting can be different operations for the file format but pootle supports obsoleting only.

unho: > it doesnt catch the logic correctly

You are right. I am not going to check how is it possible that makeobsolete can fail, but it still can be simplified:

```python
        deleted = not unit.istranslated()
        obsoleted = not deleted and not conservative

        if obsoleted:
            unit.makeobsolete()
            deleted = not unit.isobsolete()

        if deleted:
            del unit

        return obsolete, deleted
```

phlax: some formats dont have obsolete

tests/models/store.py
```diff
@@ -1213,3 +1215,283 @@ def test_store_diff_delete_obsoleted_source_unit(diffable_stores):
...
+    assert not obsolete
+    assert not deleted
+
+    # not conservative and the unit is deleted
+    obsolete, deleted = store.syncer.obsolete_unit(newunit, False)
+    assert obsolete
+    assert not deleted
+
+
+@pytest.fixture
```
ta2-1: should it go to fixtures?

ta2-1:
lgtm when tests pass. it's really great PR. thanks @phlax 👍

pootle/apps/pootle_store/syncer.py
```diff
@@ -0,0 +1,354 @@
...
+    def update_newer(self, last_revision):
+        return (
+            not self.store.file.exists()
+            or (last_revision >= self.store.last_sync_revision))
+
+    @cached_property
+    def dbid_index(self):
+        """build a quick mapping index between unit ids and database ids"""
+        return dict(
+            self.store.unit_set.live().values_list('unitid', 'id'))
```
unho: This should be filtering out obsolete units.

phlax: `.live()` does that

unho:
lgtm once tests pass

PR #5095, opened 2016-08-18
Added support for configuring currencies in config files

KillerX:
Should be fully backwards compatible since the default is the same as the values are now.

### DIFF
--- a/docs/server/settings.rst
+++ b/docs/server/settings.rst
@@ -286,6 +286,20 @@ Configuration settings for applications used by Pootle.
   - ``ANALYZE`` - coefficient to calculate an user score change for
     rejecting suggestions and penalty for the rejected suggestion.
 
+.. setting:: POOTLE_CURRENCIES
+
+``POOTLE_CURRENCIES``
+  Default::
+
+  (
+    ('USD', 'USD'),
+    ('EUR', 'EUR'),
+    ('CNY', 'CNY'),
+    ('JPY', 'JPY')
+  )
+
+  A collection currencies that can be selected for reporting of user activities
+
 
 60-translation.conf
 ^^^^^^^^^^^^^^^^^^^
--- a/docs/server/settings.rst
+++ b/docs/server/settings.rst
@@ -291,12 +291,12 @@ Configuration settings for applications used by Pootle.
 ``POOTLE_CURRENCIES``
   Default::
 
-  (
-    ('USD', 'USD'),
-    ('EUR', 'EUR'),
-    ('CNY', 'CNY'),
-    ('JPY', 'JPY')
-  )
+    (
+       ('USD', 'USD'),
+       ('EUR', 'EUR'),
+       ('CNY', 'CNY'),
+       ('JPY', 'JPY')
+    )
 
   A collection currencies that can be selected for reporting of user activities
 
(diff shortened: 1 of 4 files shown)


pootle/apps/accounts/migrations/0001_initial.py
```diff
@@ -5,7 +5,7 @@
 import django.utils.timezone
 import re
 import django.core.validators
-
```
unho: Please add back this blank line after imports.

pootle/settings/40-apps.conf
```diff
@@ -155,3 +155,6 @@ POOTLE_SCORE_COEFFICIENTS = {
     'SUGGEST': 0.2,
     'ANALYZE': 0.1,
 }
+
+POOTLE_CURRENCIES = (('USD', 'USD'), ('EUR', 'EUR'), ('CNY', 'CNY'), ('JPY', 'JPY'))
+
```
unho: Remove this blank line at the end.

pootle/apps/accounts/models.py
```diff
@@ -41,7 +41,7 @@
 __all__ = ('User', )
 
 
-CURRENCIES = (('USD', 'USD'), ('EUR', 'EUR'), ('CNY', 'CNY'), ('JPY', 'JPY'))
+CURRENCIES = settings.POOTLE_CURRENCIES
```
unho: Replace `CURRENCIES` everywhere with the setting.

docs/server/settings.rst
```diff
@@ -286,6 +286,20 @@ Configuration settings for applications used by Pootle.
...
+  Default::
+
+    (
+       ('USD', 'USD'),
+       ('EUR', 'EUR'),
+       ('CNY', 'CNY'),
+       ('JPY', 'JPY')
+    )
+
+  A collection currencies that can be selected for reporting of user activities
```
unho: Trailing period.

dwaynebailey:
@KillerX thanks for the contribution.

I'm afraid we can't edit existing migrations.  Any migration needs to happen after the sequence of migrations.  The code as you have it here will work only in a situation where you set the currencies before running `pootle migrate`, if you change them later I'm pretty sure you'll have a situation where your db models don't match your DB.  This certainly won't work for anyone who has already deployed Pootle and they're upgrading.

For this to work, it will need to be coded such that the currency choices can be set and are dynamic outside of the models.  I'm sure the rest of the @translate/core will correct me if I'm wrong.

iafan:
@KillerX, thanks for contributing! Out of curiosity: are you currently using currencies (paid translator functionality) in production? You may want to also look at PR #4648 that makes this functionality feature-complete.

KillerX:
@dwaynebailey very true, I wasn't really thinking about that. How about adding all/many currencies to the model but not actually show them in the dropdown and have configuration what to show in the drop down? This way the model stays the same and can be extended at any time, but the drop down contents still remain relevant for the use case.
Alternatively a new DB table with its associated UI and linking would need to be created but I'm personally not too eager to write that, since I don't see it worth the effort in comparison to how often I would expect it used.

@iafan Yes, that is exactly what we are doing. We just started a month or so ago and so far it has been really nice. I saw the PR you mentioned. I think at least one person here would be interested in that so I'll take a closer look a bit later.

dwaynebailey:
@KillerX could work.  I think best place to ask and get help getting the PR into shape is https://gitter.im/translate/dev where I think we can quickly help arrive at a workable solution.

KillerX:
As discussed I will try to implement this as a big list of currencies, limited for display by a config variable. Closing this in favor of a clean PR later.

PR #5096, opened 2016-08-18
DOM: use wrappers for querySelector[All]

### LINKED ISSUE #5089: Array spread breaks IE11
Babel transpiles Array spreads using `Array.prototype.from`, which is not available in IE11, and breaks it. The object spread is mostly used to [convert a `NodeList` into an `Array`](https://developer.mozilla.org/de/docs/Web/API/NodeList#Converting_a_NodeList_to_an_Array).

An option would be to use the `loose` mode of the ES2015 spread transform, and then feed an array to spreads (i.e. convert the `NodeList` first to an array via `Array.prototype.slice.call`). Another option would be to combine `Object.keys` and `forEach`.

In any of the cases creating a wrapper for iterating over `NodeList`s will be more than convenient.

julen:
Most notably this is handy to be able to use `Array.prototype` methods over the collection of `NodeList`s returned by `querySelectorAll`.

Update: forgot to mention this basically fixes #5089.

### DIFF
--- a/pootle/static/js/admin/app.js
+++ b/pootle/static/js/admin/app.js
@@ -11,6 +11,8 @@ import 'imports?Backbone=>require("backbone")!backbone-move';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
+import { q } from 'utils/dom';
+
 import AdminController from './components/AdminController';
 import User from './components/User';
 import Language from './components/Language';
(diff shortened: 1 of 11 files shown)


unho:
Reviewed and tested. GTM.

ta2-1:
lgtm

PR #5107, opened 2016-08-19
Cleanups from adding StoreSyncer

phlax:
- StoreSyncer.convert allows fileclass to be empty and uses a default
- removes unnecessary or unused methods from Store


unho:
lgtm

PR #5112, opened 2016-08-19
fetchUnits: report no results for initial requests

### LINKED ISSUE #5111: If editor filter retrieves no units, TypeError is thrown
When filtering in the editor by some filter that doesn't return any unit, some JavaScript error is shown on the JavaScript console:

> Uncaught TypeError: Cannot read property 'id' of undefined

Observed in https://mozilla.locamotion.org/ach/firefox/translate/ when changing the filter to "Suggestions" (this TP has no suggestions).

The same happens if the filter is changed to "My suggestions" somewhere you never contributed.

julen:
`this.units` is not guaranteed to be empty when firing up initial requests,
hence it could provide false positives if previously the editor was populated
with units and users were changing filters.

Fixes #5111.

### DIFF
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -1465,7 +1465,7 @@ PTL.editor = {
       }
       return UnitAPI.fetchUnits(reqData)
         .then(
-          (data) => this.storeUnitData(data),
+          (data) => this.storeUnitData(data, { isInitial: initial }),
           this.error
         ).always(() => this.markAsFetched(offsetToFetch));
     }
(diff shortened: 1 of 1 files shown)


julen:
Since this checks for initial requests now, I'm almost confident #4626 doesn't regress, but we'll want to double-check and keep an eye on it.

unho:
Commit message is not referencing the issue.

unho:
Did some testing and the reported issue is fixed (issue #4626). Also the flashing did not came back (issue #5111). What I did hit was the error when you are sent back to overview with "You have completed all units" message a few units before reaching the last unit in the editor (I hit this twice), but I think this error is unrelated. GTM

julen:
> Commit message is not referencing the issue.

This is intended, not to spam issues. I almost always have to rebase anyway...

julen:
> What I did hit was the error when you are sent back to overview with "You have completed all units" message a few units before reaching the last unit in the editor (I hit this twice), but I think this error is unrelated.

Thanks for the pointer. I tried in master and could reproduce what you describe; filed #5116.

PR #5113, opened 2016-08-19
Cleanup: customize login form via allauth mechanisms

julen:
This customizes the login form using allauth's own suggested way while helping
reduce unnecessary boilerplate.


unho:
Created some user with active flag enabled, but user not yet verified and this fails. This fails too in master so given that the rest of the scenarios do work I would say GTM.

PR #5118, opened 2016-08-20
Prevent cycling updates for obsolete units in update/sync stores

### LINKED ISSUE #5117: Update/sync update obsolete units every time with --force --overwrite
While testing recent changes to the Store.sync code i noticed that on every sync units that were obsolete were being shown as `deleted`.

On further testing it appears that Store.update does something similar with the same units, effectively generating noop submission records.

In the case of Store.update it appears that this has been the case for some time.

phlax:
fixes #5117

### DIFF
--- a/pootle/apps/pootle_store/syncer.py
+++ b/pootle/apps/pootle_store/syncer.py
@@ -205,10 +205,10 @@ class StoreSyncer(object):
              in new_ids - old_ids])
 
     def get_units_to_obsolete(self, old_ids, new_ids):
-        return (
-            self.disk_store.findid(uid)
-            for uid
-            in old_ids - new_ids)
+        for uid in old_ids - new_ids:
+            unit = self.disk_store.findid(uid)
+            if unit and not unit.isobsolete():
+                yield unit
 
     def obsolete_unit(self, unit, conservative):
         deleted = not unit.istranslated()
--- a/pootle/apps/pootle_store/updater.py
+++ b/pootle/apps/pootle_store/updater.py
@@ -123,7 +123,8 @@ class UnitUpdater(object):
     def should_update_index(self):
         return (
             self.update.change_indices
-            and self.uid in self.update.indices)
+            and self.uid in self.update.indices
+            and self.unit.index != self.update.get_index(self.uid))
 
     @property
     def should_update_target(self):
--- a/tests/models/store.py
+++ b/tests/models/store.py
@@ -1369,7 +1369,9 @@ def _test_get_obsolete(results, syncer, old_ids, new_ids):
     assert list(results) == list(
         syncer.disk_store.findid(uid)
         for uid
-        in old_ids - new_ids)
+        in old_ids - new_ids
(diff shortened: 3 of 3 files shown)


pootle/apps/pootle_store/updater.py
```diff
@@ -123,7 +123,8 @@ def should_create_suggestion(self):
     def should_update_index(self):
         return (
             self.update.change_indices
-            and self.uid in self.update.indices)
+            and self.uid in self.update.indices
+            and self.unit.index != self.update.get_index(self.uid))
```
ta2-1: @phlax I'm recalling that we have allowed to have gaps in index ordering if they don't break a real sort order. Though I'm not sure how this change is related.

phlax: with this update - the index would still get updated if required (not sure how gaps are related, but it would do the same as now in that respect.

pootle/apps/pootle_store/syncer.py
```diff
@@ -205,10 +205,10 @@ def get_new_units(self, old_ids, new_ids):
              in new_ids - old_ids])
 
     def get_obsolete_units(self, old_ids, new_ids):
-        return (
-            self.disk_store.findid(uid)
-            for uid
-            in old_ids - new_ids)
+        for uid in old_ids - new_ids:
+            unit = self.disk_store.findid(uid)
+            if unit and not unit.isobsolete():
```
unho: `not isobsolete`? The method is `get_obsolete_units`. Maybe I am missing something.

pootle/apps/pootle_store/updater.py
unho: Are there tests to ensure this doesn't change behavior?

pootle/apps/pootle_store/syncer.py
phlax: i think its not very well named - it should be `get_units_to_obsolete`

pootle/apps/pootle_store/syncer.py
```diff
@@ -205,10 +205,10 @@ def get_new_units(self, old_ids, new_ids):
              in new_ids - old_ids])
 
     def get_obsolete_units(self, old_ids, new_ids):
-        return (
-            self.disk_store.findid(uid)
-            for uid
-            in old_ids - new_ids)
+        for uid in old_ids - new_ids:
```
unho: I would create a variable `old_untouched_uids = old_ids - new_ids`.

phlax: that would make it more confusing for me tbh

pootle/apps/pootle_store/updater.py
phlax: yes - the index/order/update tests

pootle/apps/pootle_store/syncer.py
unho: Ok, then just forget about this.

ta2-1:
lgtm

PR #5120, opened 2016-08-20
Cleanup: remove method only used in tests

phlax:
remove getorig method that is only used in tests

### DIFF
--- a/tests/models/unit.py
+++ b/tests/models/unit.py
@@ -49,11 +49,11 @@ def _update_translation(store, item, new_values, sync=True):
 
 @pytest.mark.django_db
 def test_getorig(af_tutorial_po):
-    """Tests that the in-DB Store and on-disk Store match by checking that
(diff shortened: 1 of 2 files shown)


tests/models/unit.py
```diff
@@ -49,11 +49,11 @@ def _update_translation(store, item, new_values, sync=True):
 
 @pytest.mark.django_db
 def test_getorig(af_tutorial_po):
-    """Tests that the in-DB Store and on-disk Store match by checking that
+    """Tests that the inDB Store and ondisk Store match by checking that
```
julen: If there's something wrong with using an hyphen in compound constructions, I'd be interested to know why.

phlax: no - it was an inadvertent replace

julen: Cool, that's reassuring. I'll bring it back.

PR #5122, opened 2016-08-20
Test/fixture cleanups

phlax:
remove unreachable/unnecessary code in fixtures and tests

### DIFF
--- a/pytest_pootle/utils.py
+++ b/pytest_pootle/utils.py
@@ -90,10 +90,7 @@ def items_equal(left, right):
     """Returns `True` if items in `left` list are equal to items in
     `right` list.
     """
-    try:
-        return sorted(left) == sorted(right)
-    except TypeError:  # non-iterable
-        return False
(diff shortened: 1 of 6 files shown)


pytest_pootle/utils.py
```diff
@@ -114,15 +111,12 @@ def create_api_request(rf, method='get', url='/', data='', user=None,
     return request
 
 
-def update_store(store, units=None, store_revision=None, add_headers=False,
+def update_store(store, units=None, store_revision=None,
                  user=None, submission_type=None, resolve_conflict=None):
     from pootle_store.models import POOTLE_WINS
     if resolve_conflict is None:
         resolve_conflict = POOTLE_WINS
     store_headers = {}
```
unho: AFAICT there is no need to keep this variable and pass it below.

phlax: thanks cleanup added https://github.com/translate/pootle/pull/5126

PR #5124, opened 2016-08-21
Update django-overextends to 0.4.2

rene0:
This version adds support for Django 1.10 and drops support for Django 1.7


unho:
Merged in 320e3c32. Thanks for your contribution!

PR #5125, opened 2016-08-21
Test/fixture cleanups (/more)

phlax:
Cleanup: remove old fixtures from changed_langs test

### DIFF
--- a/pytest_pootle/fixtures/models/store.py
+++ b/pytest_pootle/fixtures/models/store.py
@@ -370,14 +370,6 @@ def ru_tutorial_po(settings, russian_tutorial):
 
 
 @pytest.fixture
-def ru_update_save_changed_units_po(settings, russian_tutorial):
-    """Require the /ru/tutorial/tutorial.po store."""
-    return _require_store(russian_tutorial,
-                          settings.POOTLE_TRANSLATION_DIRECTORY,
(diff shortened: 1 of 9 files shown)


pytest_pootle/fixtures/models/store.py
```diff
@@ -370,14 +370,6 @@ def ru_tutorial_po(settings, russian_tutorial):
 
 
 @pytest.fixture
-def ru_update_save_changed_units_po(settings, russian_tutorial):
-    """Require the /ru/tutorial/tutorial.po store."""
-    return _require_store(russian_tutorial,
-                          settings.POOTLE_TRANSLATION_DIRECTORY,
-                          'update_save_changed_units.po')
```
unho: Just wondering if the files in https://github.com/phlax/pootle/tree/master/pytest_pootle/data/po/tutorial/ru should be dropped too.

phlax: there are some other ru fixtures - so not sure tbh

unho: If this specific file is not used anywhere else then I would say drop it.

phlax: currently it is used - but if there are significant benefit to removing if that wasnt the case - the fiels in the tp folders are copied on every test

PR #5131, opened 2016-08-22
Cleanup fix

phlax:
Currently, when creating a Store, StoreSyncer doesnt respect whether the project is gnu/nongnu.

Also, it sets the wrong path for the `store.file`

This PR fixes these 2 problems - and sets the default test projects to "nongnu" explicitly. Currently they were being treated as both gnu and non-gnu according to the state of the the filesystem when it was being checked

### DIFF
--- a/pootle/apps/pootle_store/syncer.py
+++ b/pootle/apps/pootle_store/syncer.py
@@ -148,6 +148,14 @@ class StoreSyncer(object):
             *split_pootle_path(self.store.pootle_path)[2:])
 
     @property
+    def relative_file_path(self):
+        path_parts = split_pootle_path(self.store.pootle_path)
+        path_prefix = [path_parts[1]]
+        if self.project.get_treestyle() != "gnu":
+            path_prefix.append(path_parts[0])
+        return os.path.join(*(path_prefix + list(path_parts[2:])))
+
+    @property
     def unit_class(self):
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/syncer.py
```diff
@@ -235,10 +243,10 @@ def create_store_file(self, last_revision, user):
         store = self.convert()
         if not os.path.exists(os.path.dirname(self.store_file_path)):
             os.makedirs(os.path.dirname(self.store_file_path))
+        self.store.file = self.relative_file_path
```
julen: Previously this set an absolute file path, whereas now it is a relative one. Correct me if I'm wrong, but is this now possible because the `FieldFile` descriptor in `store.file` is using `PootleFileSystemStorage`, which sets its base location to `POOTLE_TRANSLATION_DIRECTORY`? If so, please let's add a comment in the commit message to make it clear.

phlax: im not sure how it worked correctly previously - but what i know is that sometimes test where using gnu conventions and sometimes nongnu - and sometimes a mixture in tests.

afaict tho you are right that it needs a relative path for `PootleFileSystemStorage`

ta2-1:
@phlax this PR looks good. I'm happy to land it.

Issue #5016, opened 2016-08-03
Show more stats view in admin dashboard is unperformant


unho:
https://github.com/translate/pootle/blob/master/pootle/apps/pootle_app/views/admin/dashboard.py#L58-L96

ta2-1:
refs https://github.com/translate/pootle/issues/4846

julen:
Without further details, it's a dupe, I'd say.

phlax:
@julen the performance on one particular view is a critical performance issue - it completely hoses our database

julen:
Ah yeah, already reported but in the fork repository (evernote/pootle#290).

unho:
It is not a duplicate, since the other issue is about exposing more data on that view, while this issue is about increasing performance.

unho:
Closing this since #4846 can work as reimplement this feature now removed by #5132.

PR #5132, opened 2016-08-22
Remove unperformant dashboard show more stats

unho:
This is rarely used, but when used it has a big impact on server
performance.

References #5016.


pootle/apps/pootle_app/views/admin/urls.py
```diff
@@ -50,9 +50,9 @@
         name='pootle-admin-permissions'),
 
     # XHR
-    url(r'^more-stats/?$',
-        dashboard.server_stats_more,
-        name='pootle-admin-more-stats'),
+#    url(r'^more-stats/?$',
+#        dashboard.server_stats_more,
+#        name='pootle-admin-more-stats'),
```
phlax: dont comment out - we can add it back if we need it - but no point distributing commented code

phlax:
@unho would this stop the link from showing?

unho:
Removed the code instead of commenting it, and also removed the link.

julen:
If disabled, I'd be happy to see the code go away—any proper implementation will very likely be different anyway.

Nit: this is not a speedup on its own, but prevents an accidental resource hog.

unho:
@julen Updated to comply with your requests.

phlax:
lgtm, nice one @unho

julen:
You can also remove the [`a.slide` style declaration](https://github.com/translate/pootle/blob/f7c601d96271e82ef2182e2149945ea8466cae12/pootle/static/css/style.css#L2054). GTM.

PR #5136, opened 2016-08-22
Move autosize to an standalone component

julen:
This will ease reuse in many places. It also bumps the package version to include many interesting bug fixes affecting memory leaks and height calculation.


unho:
lgtm

PR #5138, opened 2016-08-22
Update for tp_tool

phlax:
- Add get/[] methods to tp_tool to access tps by lang_code
- add param to prevent cache-fu in tp_tool methods


unho:
lgtm

PR #5141, opened 2016-08-23
Cleanup add stylelint for CSS linting

dwaynebailey:
This adds stylelint to do CSS linting (and replaced #4930 which tried using csslint)

This also uses the stylelint-standard-config and implements the linting suggested. Minor deviations from this guide are added to follow our suggested style and comment anomalies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ put-translations:
 linguas:
 	@${SRC_DIR}/tools/make-LINGUAS.sh 80 > ${SRC_DIR}/locale/LINGUAS
 
-jslint:
+lint-js:
 	cd ${JS_DIR} \
 	&& npm run lint
 
--- a/Makefile
+++ b/Makefile
@@ -94,9 +94,15 @@ put-translations:
 linguas:
 	@${SRC_DIR}/tools/make-LINGUAS.sh 80 > ${SRC_DIR}/locale/LINGUAS
 
+lint: lint-js lint-css
+
 lint-js:
 	cd ${JS_DIR} \
 	&& npm run lint
+
+lint-css:
+	cd ${JS_DIR} \
+	&& npm run stylelint
 
 publish-pypi:
 	python setup.py sdist ${FORMATS} upload
--- a/pootle/static/css/auth.css
+++ b/pootle/static/css/auth.css
@@ -133,6 +133,7 @@
 {
     margin: 1em auto;
 }
+
 .auth-window .actions.password-forgotten
 {
     margin-top: -1.2em;
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ install:
       -r requirements/base.txt
       -r requirements/travis.txt
       -e .
+  # http://entulho.fiatjaf.alhur.es/guias/how-to-use-node-along-with-other-language-on-travis-ci/
+  - source $HOME/.nvm/nvm.sh
+  - nvm install stable
+  - nvm use stable
 script:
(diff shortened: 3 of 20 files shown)


unho:
I am not sure if I like all the rules enabled, but overall looks really good.

Makefile
```diff
@@ -94,10 +94,16 @@ put-translations:
...
 
+lint: lint-js lint-css
+
 lint-js:
 	cd ${JS_DIR} \
 	&& npm run lint
 
+lint-css:
+	cd ${JS_DIR} \
+	&& npm run stylelint -- --config=stylelint.json "../css/**/*.css"
```
julen: As I understand this will only check for CSS files under the _static/css/_ folder, but not CSS files which live next to components and we'll want to check too.

julen: Also, to my eyes these flags belong to the `"scripts"` section of _package.json_.

pootle/static/css/auth.css
```diff
@@ -63,9 +63,7 @@
     box-sizing: border-box;
     padding: 1em;
     width: 50%;
-
```
julen: hmmm such empty lines have been intentional to split blocks of similar declarations. Probably worth checking the [`declaration-block-properties-order`](http://stylelint.io/user-guide/rules/declaration-block-properties-order/) rule and set groups there, although I understand that might become a bit noisy.

julen:
Thanks for taking the lead on this one @dwaynebailey. Overall the PR looks good; let's try to address the couple of raised points though.

dwaynebailey:
@unho this might help https://github.com/stylelint/stylelint-config-standard#stylelint-config-standard - it explains how its designed.  By default stylelint enforces no rules and I wanted to start from something instead of nothing.  And stylint-config-standard seems to use generally accepted styles and avoids ones that are divisive.

So unless you pick up anything specific which we can turn off and deal with elsewhere I'm assuming that all are good enough.

Makefile
dwaynebailey: @julen I wanted it to be more generally useful, but I'll look at what to move into scripts

unho:
@dwaynebailey My worries were about several of the enabled rules, but after checking my only concern is `declaration-block-no-ignored-properties`. While I agree that final CSS assets shouldn't contain these redundancies I am not sure about if we should remove in the source CSS in case any of those is removed from the places where it is inherited from.

pootle/static/css/auth.css
dwaynebailey: I think this is the way to go.  But it's quite a bit of work to setup groups and then cleanup and I'm not sure it works well without a lot of maintenance.

I suggest I drop that and switch off that check for now.  I'll play a bit with the properties order and try PR that apart from this.

dwaynebailey:
@julen 

* moved config into "scripts" section
* search for css also in the scripts section, and made some fixes there
* Experimented with declaration-block-properties-order, nice future fix, not now.  Instead disabled the whitespace in deceleration check

@unho 

* Not sure I can do much about `declaration-block-no-ignored-properties`.  if it's broke, it's broke, so lets fix it is my view.  You could probably take the view of any inherited param that it might get removed from the underlying object so we should duplicate it here.

That's all the feedback integrated, any GTMs?

unho:
lgtm

pootle/static/css/auth.css
julen: Sounds like a plan!

julen:
GTM.

Issue #4359, opened 2016-01-11
Some test fixtures lack docstrings


julen:
Running `py.test --fixtures` allows one to see all available fixtures for testing along with a (short) explanation of what each of them does.

Currently there are a few which don't contain any docstrings, and it'd be great to have them added. The full list is:

* `file_import_failure`
* `store_diff_tests`
* `param_update_store_test`
* `site_matrix`
* `site_permissions`
* `site_matrix_with_announcements`
* `site_matrix_with_subdirs`
* `site_matrix_with_vfolders`
* `site_root`
* `project_views`
* `tp_views`
* `language_views`
* `bad_views`

PR #5144, opened 2016-08-24
Migrations in tests

phlax:
- remove the `--nomigrations` flag
- add a fixture to run migrations if db is used in tests

this makes it easier to test migrations, and makes the test env closer to the env that users actually have


ta2-1:
lgtm

unho:
lgtm

julen:
Please don't forget to add valuable docstrings to fixtures or `py.test --fixtures` becomes useless (cf. #4359).

PR #5146, opened 2016-08-24
[WIP] Denormalize Store.get_max_unit_revision

phlax:
It would speed a lot of operations up if we didnt need to keep doing Store.get_max_unit_revision.

This PR denorms the max revision onto store, and provides signal handlers for updating

In cases where you dont want to update the revision on the store until you have completed updating units - it will ensure that this is possible

### DIFF
--- a/pootle/apps/pootle_store/receivers.py
+++ b/pootle/apps/pootle_store/receivers.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import Unit
+
+
+@receiver(post_save, sender=Unit)
+def unit_postsave_handler(**kwargs):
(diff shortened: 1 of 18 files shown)


pootle/apps/pootle_store/receivers.py
```diff
@@ -0,0 +1,22 @@
...
+# AUTHORS file for copyright and authorship information.
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import Unit
+
+
+@receiver(post_save, sender=Unit)
+def unit_postsave_handler(**kwargs):
```
ta2-1: @phlax I'm worried if it leads to performance down for store.update(), where you will need to call store.save() after every unit update.

phlax: @ta2-1 - please see the description of the PR - its not finished yet - but also - this in particular makes it (almost) possible to do that

ta2-1: @phlax sorry missed the last part of PR description.
In this case store.update() should be wrapped within transaction, as I can understand.

unho: I have read the last part of the PR description and I don't get how it will wait until all units for a store are updated.

phlax: it doent yet - that is still under discussion @unho

phlax:
no longer relevant

PR #5147, opened 2016-08-24
Set unsynced units to the same revision in store.updater

phlax:
This will massively improve performance on store.update

### DIFF
--- a/pootle/apps/pootle_store/updater.py
+++ b/pootle/apps/pootle_store/updater.py
@@ -210,11 +210,9 @@ class StoreUpdater(object):
             'revision__gt': self.target_store.last_sync_revision,
             'revision__lt': update_revision,
             'state__gt': OBSOLETE}
-        count = 0
-        for unit in self.target_store.unit_set.filter(**filter_by):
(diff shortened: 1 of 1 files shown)


unho:
I was wondering myself about this particular cleanup. lgtm

pootle/apps/pootle_store/updater.py
```diff
@@ -210,11 +210,9 @@ def increment_unsynced_unit_revision(self, update_revision):
...
             'revision__lt': update_revision,
             'state__gt': OBSOLETE}
-        count = 0
-        for unit in self.target_store.unit_set.filter(**filter_by):
-            unit.save(revision=Revision.incr())
-            count += 1
-        return count
+        units = self.target_store.unit_set.filter(**filter_by)
+        units.update(revision=Revision.incr())
+        return units.count()
```
ta2-1: @phlax we've just updated unit revisions in the previous line and we query for `count()` here. 
It looks like updated revisions already affect the result. And we will get `0` in any case.

phlax: good spot @ta2-1

PR #5151, opened 2016-08-24
Add cached template loader in tests (and make tests DEBUG=False)

phlax:
in the words of the django manual - this should give a "drastic" improvement in performance

EDIT:

seems like it was only not being used in tests (or if DEBUG=True) - ive changed the PR to instead use DEBUG=False in tests/settings

### DIFF
--- a/tests/misc/state.py
+++ b/tests/misc/state.py
@@ -12,6 +12,13 @@ import pytest
 from pootle.core.state import State, ItemState
 
 
+def test_settings(settings):
+    assert not settings.DEBUG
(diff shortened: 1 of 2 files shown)


unho:
lgtm

phlax:
turns out this is not necessary - i tested on TRAVIS and its using both DEBUG=False and the correct TEMPLATE loader.

phlax:
reopening

it seems that even tho DEBUG = True, tests are not using the cached template loader

PR #5154, opened 2016-08-25
Make sure projects are created with a code and a fullname

### LINKED ISSUE #5114: Creating project without a title causes broken select2 dropdown
If you:

1. Create a project
2. With no human readable title
3. The project dropdown will appear with a blank line
4. When in the Project, there will be no dropdown to change Project

Either we could:
1. Prevent you creating a project without a human readable name
2. Use the project code if no human readable name is defined

I'd lean toward 2.

unho:
Fixes #5114.

### DIFF
--- a/pootle/apps/pootle_app/forms.py
+++ b/pootle/apps/pootle_app/forms.py
@@ -14,7 +14,7 @@ from django.contrib.auth import get_user_model
 from django.utils.translation import ugettext_lazy as _
 
 from pootle_language.models import Language
-from pootle_project.models import RESERVED_PROJECT_CODES, Project
+from pootle_project.models import Project
 from pootle_store.models import Store
 
 
@@ -80,6 +80,9 @@ class ProjectForm(forms.ModelForm):
                           % filetype))
         return value
 
+    def clean_fullname(self):
+        return self.cleaned_data['fullname'].strip()
+
     def clean_treestyle(self):
         value = self.cleaned_data.get('treestyle', None)
         if not value:
@@ -87,12 +90,7 @@ class ProjectForm(forms.ModelForm):
         return value
 
     def clean_code(self):
-        value = self.cleaned_data['code']
-        if value in RESERVED_PROJECT_CODES:
-            raise forms.ValidationError(
-                _('"%s" cannot be used as a project code', value)
-            )
-        return value
+        return self.cleaned_data['code'].strip()
 
 
 class UserForm(forms.ModelForm):
--- a/pootle/apps/pootle_fs/management/commands/init_fs_project.py
+++ b/pootle/apps/pootle_fs/management/commands/init_fs_project.py
@@ -8,8 +8,8 @@
 
 import logging
 
+from django.core.exceptions import ValidationError
 from django.core.management import BaseCommand, CommandError
-from django.db import IntegrityError
 
 from pootle_format.models import Format
 from pootle_fs.utils import FSPlugin, parse_fs_url
@@ -96,9 +96,7 @@ class Command(BaseCommand):
                 treestyle='none',
                 checkstyle=options['checkstyle'],
                 source_language=source_language)
-        except IntegrityError as e:
-            self.stdout.write('Project code "%s" already exists.'
-                              % options['code'])
+        except ValidationError as e:
             raise CommandError(e)
 
         for filetype in options["filetypes"] or ["po"]:
--- a/pootle/apps/pootle_project/models.py
+++ b/pootle/apps/pootle_project/models.py
@@ -17,7 +17,7 @@ from translate.lang.data import langcode_re
 
 from django.conf import settings
 from django.core.cache import cache
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.urlresolvers import reverse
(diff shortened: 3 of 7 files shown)


pootle/apps/pootle_project/models.py
```diff
@@ -420,6 +423,13 @@ def delete(self, *args, **kwargs):
 
         directory.delete()
 
+    def clean(self):
+        if not self.code:
+            raise ValidationError('A code must be provided.')
```
phlax: i dont think this is necessary

julen: Without this, one can create projects without code. Such a project would be useless and accidental, but it's technically possible.

phlax: in which case im thinking that `code` needs to be made `blank=False`

phlax:
i think this PR needs some tests

pootle/apps/pootle_project/models.py
unho: `blank=False` only applies to forms. So this is still required.

julen: [As already commented](https://github.com/translate/pootle/issues/5114#issuecomment-241642828), `blank=False` is the default and applies to forms.

julen:
Leandro, in the `ProjectForm` I would also add:
```python
    def clean_fullname(self):
        return self.cleaned_data['fullname'].strip()

File: pootle/apps/pootle_project/models.py (hunk 1 above)
Comment (unho) [inline pootle/apps/pootle_project/models.py]: It turns out that `blank=False` is enforced if in `save()` the `full_clean()` method is called.

File: pootle/apps/pootle_fs/management/commands/init_fs_project.py [hunk 2]
```diff
@@ -96,18 +96,22 @@ def handle(self, **options):
...
-                    project.filetypes.add(filetype)
-                except Format.DoesNotExist as e:
-                    raise CommandError(e)
-            project.update_all_cache()
-        except IntegrityError as e:
-            self.stdout.write('Project code "%s" already exists.'
-                              % options['code'])
+        except ValidationError as e:
+            if 'code' in e and 'already exists' in e['code']:
+                self.stdout.write('Project code "%s" already exists.'
```
phlax: raise a CommandError with this message

tests/models/project.py
```diff
@@ -10,12 +10,67 @@
...
 from pytest_pootle.fixtures.models.permission_set import _require_permission_set
 from pytest_pootle.utils import items_equal
 
 from pootle_format.models import Format
-from pootle_project.models import Project
+from pootle_project.models import Project, RESERVED_PROJECT_CODES
+
+
+@pytest.mark.django_db
+def test_create_project_with_valid_arguments(english):
```
phlax: this should be test_create_project_bad - and separate out the valid_args test

unho: Thanks. Will do. And probably will parametrize too.

unho:
@julen @phlax All concerns addressed. GTM?

pootle/apps/pootle_project/models.py
```diff
@@ -388,7 +388,21 @@ def resources(self):
...
         return self.fullname
 
+    def clean(self):
+        self.code = self.code.strip()
+
+        if not self.code:
+            raise ValidationError("Blank strings cannot be used as a project "
+                                  "code")
+
+        if self.code in RESERVED_PROJECT_CODES:
```
julen: This is also present in the project form. Given that such form will use this model, doesn't this validation already kick in? In other words, isn't the duplicated validation in `ProjectForm` unnecessary?

pootle/apps/pootle_project/models.py
```diff
@@ -181,7 +181,7 @@ class Project(models.Model, CachedTreeItem, ProjectURLMixin):
                             db_index=True, verbose_name=_('Code'), blank=False,
                             help_text=code_help_text)
 
-    fullname = models.CharField(max_length=255, null=False,
+    fullname = models.CharField(max_length=255, null=False, blank=True,
```
julen: I don't think making `fullname` optional was needed or agreed upon, but rather ensuring it was non-empty.

pootle/apps/pootle_app/forms.py
```diff
@@ -80,14 +80,20 @@ def clean_filetypes(self):
...
         value = self.cleaned_data.get('treestyle', None)
         if not value:
             value = self.instance.treestyle
         return value
 
     def clean_code(self):
-        value = self.cleaned_data['code']
+        value = self.cleaned_data['code'].strip()
+        if not value:
+            raise forms.ValidationError(_("Blank strings cannot be used as a "
```
julen: Note that after cleaning if `value` evaluates to `False`, Django will already provide an error message (provided the model has `blank=False`, the default). Given the low chances for this happening, personally I wouldn't add an extra string for localizers as we are already covered.

pootle/apps/pootle_project/models.py
unho: Yes. Will do the required adjustments. Thanks for spotting.

unho:
Updated to address all @julen input. Any GTM?

phlax:
really nice work @unho - lgtm

PR #5155, opened 2016-08-25
[WIP] Clear cache signals

phlax:
Puts clear_cache calls into signal receivers

This allows the signal to be overridden in the event that you dont want to clear cache - and a contextmanager is provided for this purpose


phlax:
no longer relevant

PR #5159, opened 2016-08-25
Pootle data 0

phlax:
Adds a pootle_data app, initially just with a tool for getting store data efficiently. It will also hold aggregated data tables after

### DIFF
--- a/pootle/apps/pootle_data/utils.py
+++ b/pootle/apps/pootle_data/utils.py
@@ -1,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from collections import Counter
+
+from translate.filters.decorators import Category
+
+from django.db import models
+from django.utils.functional import cached_property
+
+from pootle_statistics.models import SubmissionTypes
+from pootle_store.constants import FUZZY, TRANSLATED, UNTRANSLATED
+from pootle_store.util import SuggestionStates
+
+
+class StoreDataTool(object):
+    """Sets the stats for a store"""
+
+    def __init__(self, store):
+        self.store = store
+
+    @property
+    def units(self):
+        """Non-obsolete units in this Store"""
+        return self.store.unit_set.live()
+
+    @cached_property
+    def last_created_unit(self):
+        """Last unit created in this Store"""
+        return (
+            self.units.exclude(creation_time__isnull=True)
+                      .order_by('-creation_time', '-revision').first())
+
+    @cached_property
+    def last_updated_unit(self):
+        """Last unit updated in this Store"""
+        return self.units.order_by('-revision', '-mtime').first()
+
+    @cached_property
+    def last_submission(self):
+        """Last non-UNIT_CREATE submission for this store"""
+        return (
+            self.store.submission_set.exclude(type=SubmissionTypes.UNIT_CREATE)
+                                     .latest())
+
+    @cached_property
+    def wordcount(self):
+        """Returns a dictionary of translated,fuzzy,total words for store"""
+        # XXX: `order_by()` here is important as it removes the default
+        # ordering for units. See #3897 for reference.
+        wordcount = (
+            self.units.order_by().values_list('state')
+                      .annotate(wordcount=models.Sum('source_wordcount')))
+        result = dict(total=0, translated=0, fuzzy=0)
+        for state, wordcount in wordcount.iterator():
+            result['total'] += wordcount
+            if state == TRANSLATED:
+                result['translated'] = wordcount
+            elif state == FUZZY:
+                result['fuzzy'] = wordcount
+        return result
+
+    @property
+    def suggestion_count(self):
+        """Return the count of pending suggetions for the store"""
+        return (
+            self.units.filter(suggestion__state=SuggestionStates.PENDING)
+                      .values_list("suggestion").count())
+
+    def get_max_unit_revision(self, unit=None):
+        """Returns the max revision of the store"""
+        return (
+            unit.revision
+            if unit is not None
+            else self.last_updated_unit.revision)
+
+    @cached_property
+    def checks(self):
+        """Dictionary of checks and critical check count for the store"""
+        qc_fields = (
+            "qualitycheck__category",
+            "qualitycheck__name",
+            "qualitycheck__false_positive")
+        qc_qs = (
+            self.store.unit_set.filter(state__gt=UNTRANSLATED)
+                               .filter(qualitycheck__isnull=False)
+                               .values_list(*qc_fields)
+                               .iterator())
+        result = dict(critical_count=0, checks={})
+        for (category, name, muted), count in Counter(qc_qs).items():
+            if muted:
+                continue
+            if category == Category.CRITICAL:
+                result["critical_count"] += count
+            result["checks"][name] = count
+        return result
--- a/pootle/apps/pootle_data/__init__.py
+++ b/pootle/apps/pootle_data/__init__.py
@@ -1,0 +1,7 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
--- a/tests/pootle_data/data_utils.py
+++ b/tests/pootle_data/data_utils.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from translate.filters.decorators import Category
+
+from pootle_data.utils import StoreDataTool
+from pootle_store.constants import (
+    FUZZY, OBSOLETE, TRANSLATED, UNTRANSLATED)
+from pootle_store.models import Suggestion
+from pootle_store.util import SuggestionStates
+from pootle_statistics.models import SubmissionTypes
+from pootle_store.models import QualityCheck
+
+
+@pytest.mark.django_db
+def test_data_store_util(store0):
+    data_tool = StoreDataTool(store0)
+    assert data_tool.store == store0
+
+
+@pytest.mark.django_db
+def test_data_store_util_last_created(store0):
+    last_created_unit = (
+        store0.unit_set.live()
+                       .exclude(creation_time__isnull=True)
+                       .order_by('-creation_time', '-revision').first())
+    assert StoreDataTool(store0).last_created_unit == last_created_unit
+    last_created_unit.creation_time = None
+    last_created_unit.save()
+    assert not StoreDataTool(store0).last_created_unit == last_created_unit
+
+
+@pytest.mark.django_db
+def test_data_store_util_last_updated(store0):
+    last_updated_unit = store0.unit_set.live().order_by(
+        '-revision', '-mtime').first()
+    assert StoreDataTool(store0).last_updated_unit == last_updated_unit
+    other_unit = store0.unit_set.exclude(pk=last_updated_unit.pk).first()
+    other_unit.target = "SOMETHING ELSE"
+    other_unit.save()
+    assert StoreDataTool(store0).last_updated_unit == other_unit
+
+
+@pytest.mark.django_db
+def test_data_store_util_last_submission(store0):
+    last_submission = store0.submission_set.exclude(
+        type=SubmissionTypes.UNIT_CREATE).latest()
+    assert StoreDataTool(store0).last_submission == last_submission
+    last_submission.type = SubmissionTypes.UNIT_CREATE
+    last_submission.save()
+    assert not StoreDataTool(store0).last_submission == last_submission
+
+
+@pytest.mark.django_db
+def test_data_store_util_wordcount(store0):
+    expected = dict(total=0, translated=0, fuzzy=0)
+
+    # ensure a fuzzy unit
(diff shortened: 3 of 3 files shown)


pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    @property
+    def units(self):
+        return self.store.unit_set.live()
+
+    @cached_property
+    def last_created_unit(self):
+        """Last unit created in this Store"""
+        return (
+            self.units.exclude(creation_time__isnull=True)
+                      .order_by('-creation_time', '-revision').first())
```
unho: Why `revision`?

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @cached_property
+    def last_updated_unit(self):
+        """Last unit updated in this Store"""
+        return self.units.order_by('-mtime', '-revision').first()
+
+    @cached_property
+    def last_submission(self):
+        """Last non-UNIT_CREATE submisison for this store"""
```
unho: Nit: typo.

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_max_unit_revision(self, unit):
+        return unit.revision or self.last_updated_unit.revision
+
+    # WIP...
+    @cached_property
+    def checks(self):
+        qs = self.store.unit_set.filter(unit__state__gt=UNTRANSLATED)
+        qs_values = qs.values_list("qualitycheck__category", "qualitycheck__name")
+        qs_values = qs_values.exclude(qualitycheck__false_positive=True)
```
unho: Shouldn't this happen on the `qs`?

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+        return qs_values.order_by("pk", "-qualitycheck__category")
```
unho: I find it more natural to first filter the queryset and then use `values_list` on it.

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+            elif state == FUZZY:
+                result['fuzzy'] = wordcount
+        return result
+
+    @property
+    def suggestion_count(self):
+        return Suggestion.objects.filter(
+            unit__store=self,
+            unit__state__gt=OBSOLETE,
+            state=SuggestionStates.PENDING).count()
```
unho: When I see stuff like that I always wonder why not using `self.store.suggestion_set` and thus drop the `Suggestion` import?

pootle/apps/pootle_data/utils.py
phlax: on mysql there is not microsecond precision, so we want the actually most recently created/updated

pootle/apps/pootle_data/utils.py
phlax: i havent tested this code yet

pootle/apps/pootle_data/utils.py
phlax: for sure, i will

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_max_unit_revision(self, unit):
+        return unit.revision or self.last_updated_unit.revision
+
+    # WIP...
+    @cached_property
+    def checks(self):
+        qs = self.store.unit_set.filter(unit__state__gt=UNTRANSLATED)
```
unho: `self.units`?

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+from django.db import models
+from django.utils.functional import cached_property
+
+from pootle_statistics.models import SubmissionTypes
+from pootle_store.constants import FUZZY, OBSOLETE, TRANSLATED, UNTRANSLATED
+from pootle_store.models import Suggestion
+from pootle_store.util import SuggestionStates
+
+
+User = get_user_model()
```
ta2-1: is it used anywhere?

pootle/apps/pootle_data/utils.py
unho: Forget about the above comment.

pootle/apps/pootle_data/utils.py
phlax: no - ill remove

pootle/apps/pootle_data/utils.py
ta2-1: revision value is not unique when we do store.update()

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+    def last_submission(self):
+        """Last non-UNIT_CREATE submisison for this store"""
+        return (
+            self.store.submission_set.exclude(type=SubmissionTypes.UNIT_CREATE)
+                                     .latest())
+
+    @cached_property
+    def wordcount(self):
+        wordcount = (
+            self.units.order_by().values_list('state')
```
unho: Why `order_by`?

pootle/apps/pootle_data/utils.py
unho: I would add that valuable bit of info as a comment there.

phlax: this is secondary ordering - current code uses only creation time

phlax: also - if there are genuinely 2 units with same creation/revision then its moot which one is teh latest

pootle/apps/pootle_data/utils.py
phlax: ill copy the comment from the original - for efficiency - to remove the ordering

unho: Sure, forgot about that. Having a comment helps.

pootle/apps/pootle_data/utils.py
ta2-1: can we live with this moot thing?

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @property
+    def suggestion_count(self):
+        return Suggestion.objects.filter(
+            unit__store=self,
+            unit__state__gt=OBSOLETE,
+            state=SuggestionStates.PENDING).count()
+
+    def get_max_unit_revision(self, unit):
+        return unit.revision or self.last_updated_unit.revision
```
ta2-1: will mtime be updated when you increment revision for unsynced units in store.update() directly via qs.update()?
it won't afaict. so I'm not sure if it's reliable to use self.last_updated_unit here. 
also store.get_max_revision() query may be faster than self.last_updated_unit, though you are probably relying that it's already cached here.

pootle/apps/pootle_data/utils.py
phlax: moot means its of no consequence

pootle/apps/pootle_data/utils.py
ta2-1: just checked, it works on values_list. so I'm surprised.

pootle/apps/pootle_data/utils.py
phlax: > will mtime be updated when you increment revision 

it shouldnt be - those units didnt change - more that we want to mark that they are part of a revision

> also store.get_max_revision() query may be faster 

cool, not sure what is faster - defo worth testing

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,81 @@
...
+            if state == TRANSLATED:
+                result['translated'] = wordcount
+            elif state == FUZZY:
+                result['fuzzy'] = wordcount
+        return result
+
+    @property
+    def suggestion_count(self):
+        """Return the count of pending suggetions for the store"""
+        return self.units.filter(suggestion__state=SuggestionStates.PENDING).count()
```
unho: You are filtering in units, not in suggestions.

pootle/apps/pootle_data/utils.py
ta2-1: but we update in consequence, that's why I'm asking. probs revision is unnecessary if there can be the cases which are of no consequence.

pootle/apps/pootle_data/utils.py
phlax: good point - i think 

```
        return self.units.filter(suggestion__state=SuggestionStates.PENDING).values_list("suggestion_id")count() 
```

pootle/apps/pootle_data/utils.py
phlax: perhaps - you want the last unit that was updated?

pootle/apps/pootle_data/utils.py
unho: No!

```python
        return self.store.suggestion_set.filter(
            suggestion__state=SuggestionStates.PENDING).count() 
```

phlax: there is no store.suggestion_set afaict its a unit thin hence this code

unho: Shit. It seems there is no `suggestion_set`.

phlax: im thinking i shoud change the order of the filter/values_list

unho: The thing is that units don't have suggestion states because they are units, not suggestions. So I think you have to bring back the code I thought it was wrong. Or add another FK in Suggestion pointing to Store.

pootle/apps/pootle_data/utils.py
unho: Anyway it is confusing to me.

phlax: its the way you can get data from an m2m field - im updating with slightly changed code - although it uses same pattern

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @cached_property
+    def checks(self):
+        """Dictionary of checks and critical check count for the store"""
+        qs = (
+            self.store.unit_set.filter(state__gt=UNTRANSLATED)
+                               .filter(qualitycheck__isnull=False))
+        checks = Counter(
+            qs.values_list("qualitycheck__category", "qualitycheck__name")
+              .order_by("qualitycheck__category")
```
phlax: this is pointless

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,102 @@
...
+                      .order_by('-creation_time', '-revision').first())
+
+    @cached_property
+    def last_updated_unit(self):
+        """Last unit updated in this Store"""
+        return self.units.order_by('-revision', '-mtime').first()
+
+    @cached_property
+    def last_submission(self):
+        """Last non-UNIT_CREATE submisison for this store"""
```
unho: Typo

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,102 @@
...
+            "qualitycheck__false_positive")
+        qc_qs = (
+            self.store.unit_set.filter(state__gt=UNTRANSLATED)
+                               .filter(qualitycheck__isnull=False)
+                               .order_by("qualitycheck__category")
+                               .values_list(*qc_fields)
+                               .iterator())
+        result = dict(critical_count=0, checks={})
+        for (category, name, muted), count in Counter(qc_qs).items():
+            if muted:
```
unho: Is it not better to exclude these on the queryset?

phlax: its not possible afaict - which is why ive done it this way - after testing this

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    def checks(self):
+        """Dictionary of checks and critical check count for the store"""
+        qc_fields = (
+            "qualitycheck__category",
+            "qualitycheck__name",
+            "qualitycheck__false_positive")
+        qc_qs = (
+            self.store.unit_set.filter(state__gt=UNTRANSLATED)
+                               .filter(qualitycheck__isnull=False)
+                               .order_by("qualitycheck__category")
```
unho: I don't see any need to order by category.

phlax: good point

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,102 @@
...
+            elif state == FUZZY:
+                result['fuzzy'] = wordcount
+        return result
+
+    @property
+    def suggestion_count(self):
+        """Return the count of pending suggetions for the store"""
+        return (
+            self.units.filter(suggestion__state=SuggestionStates.PENDING)
+                      .values_list("suggestion").count())
```
unho: I do wonder if this is more performant than previous:

```python
        return Suggestion.objects.filter(
            unit__store=self,
            unit__state__gt=OBSOLETE,
            state=SuggestionStates.PENDING).count()
```

unho: Also I am worried that this might be only counting units, and no suggestions. If a unit has 3 suggestions does it return 3 or 1?

pootle/apps/pootle_data/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get_max_unit_revision(self, unit=None):
+        """Returns the max revision of the store"""
+        return (
+            unit.revision
+            if unit is not None
+            else self.last_updated_unit.revision)
```
unho: I found shorter and clearer this:

```python
        if unit is None:
            return self.last_updated_unit.revision
        return unit.revision
```

pootle/apps/pootle_data/utils.py
phlax: > I do wonder if this is more performant than previous

i would imagine very similar if no the same

> Also I am worried that this might be only counting units,

suggestions - see the tests

pootle/apps/pootle_data/utils.py
phlax: i think this is a matter of dev style tbh

unho:
lgtm

PR #5161, opened 2016-08-26
[WIP] Store stats data

phlax:
This PR will
- use pootle_data tool to generate stats for store
- trigger stats refresh on unit/store save
- add migration to set stats for existing stores
- allow for suppression of refresh - eg during store.update
- add tools for efficiently retreiving stats for tps/projects/diretories


iafan:
Since there's no spec, let me ask some questions:

1. Am I right to assume that this ultimately moves stats from Redis to the DB?
2. What are the expected changes in the DB size (e.g. on MLO instance)?
3. Will the data be stored for every node (store/folder/tp/project/language) as it was before in Redis? If not, how it will be aggregated?
4. Am I right to assume that stats will be recalculated right on unit save?
5. Will calculation be done in a transaction? How do we deal with multiple people (and `sync_stores`) updating stats at the same time? Will there be a guarantee that the stats will be correct on all levels? (to give some context, this was a **huge** problem in the past when we stored stats in DB and was one of the reasons we pursued the route of having one process calculating the stats which fixed it once and for all)

These are the performance metrics I would like to see before/after breakdown for:

1. What are the changes in unit save time? Does the position of the unit (store) in the project folder structure matters?
2. What are the changes in browse page rendering for different levels:

    2.1. Language page
    2.2. Project page
    2.3. TP page
    2.4.  some folder with deep structure within a TP (one language)
    2.5.  some folder with deep structure across all languages

3. How does time depend on the folder structure? What is the impact of having very deep folder structure?

phlax:
in summary @iafan the stats will be stored in the db per-store and per-tp.

they will be updated in a unit save - except during updates when they will be updated immediately after the last unit was saved

these changes will make **updates** no more/less transactional than they are now - stats obv will be immediately correct - so in a sense a part of the transaction

not only does this remove a lot of unnecessary code, calls to the db and a very sub-optimal caching regime - it removes the need for a lot of very indeterminate and totally untested async code

iafan:
> the stats will be stored in the db per-store and per-tp

@phlax so no per-folder stats? Does this mean we're expecting to have different timing for large top-level folders within a TP with deep structure, especially if we're about to see stats across all languages? Let's make sure to test this on the most deep folder project structure we have on MLO and Evernote databases (@julen, for us it will likely be https://translate.evernote.com/projects/website2_evernote/business/, but you can probably query the DB to find the top level folder with most files in it).

Also, I would like us to understand how the performance of the translation (unit saving) will be affected when e.g. `sync_stores` is in the progress for the same file, and how will we are going to deal with transaction failures (and especially how we are going to monitor/log such failures).

phlax:
> Does this mean we're expecting to have different timing for large top-level folders within a TP

yes, but in my initial testing the views are still faster - even without caching

phlax:
...and we have added various optimizations since - so i will see when i test the views again

julen:
In addition to what @iafan already commented and as I was mentioning yesterday on IRC, besides having before/after numbers, I believe it is crucial to know what specific problems are being solved here: which measured bottlenecks exist, where, and under which circumstances.

These might be obvious for someone hitting them, especially in larger installations, but I don't see any context being referenced anywhere. As far as our production environment is concerned, we are not aware of measured bottlenecks making it difficult for the site to be responsive, but we don't ignore the fact there is always room for improvement.

All in all I hold to the belief that it would definitely help everyone to first understand the problems before diving into a storm of PRs, because that's undoubtedly the most important first step to solve any problem (unless the will is to address all unilaterally).

phlax:
> not only does this remove a lot of unnecessary code, calls to the db and a very sub-optimal caching regime - it removes the need for a lot of very indeterminate and totally untested async code

phlax:
closed in favour of https://github.com/translate/pootle/pull/5332

PR #5164, opened 2016-08-28
Use revision get

phlax:
atm increment_unsynced_units increments the unit revision rather than making it the same as the last unit to be edited - which we set it to from incrementing per unit. Im thinking this makes revisions unhelpful for a store as these units appear to have been revised more recently


phlax:
after discussing with @ta2-1 i now understand that we need to increment the units revision above the last_sync_revision to ensure that the units are synced to file

PR #5166, opened 2016-08-28
Store data model

phlax:
- Add data model for Store
- Add tools for efficiently retrieving data

### DIFF
--- a/tests/models/datatables.py
+++ b/tests/models/datatables.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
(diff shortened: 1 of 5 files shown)


tests/models/datatables.py
```diff
@@ -0,0 +1,51 @@
...
+        == '<StoreData: %s>' % store0.pootle_path)
+
+
+@pytest.mark.django_db
+def test_data_tp_bad():
+    """Test that you cant add a duplicate file extension
+    """
+    # needs a TP
+    with pytest.raises(IntegrityError):
+        StoreData.objects.create()
```
unho: Which is the difference between this test and `test_data_store_bad` above?

phlax: good spot - it should test tp not store

unho:
Nit on tests aside, looks gtm. But I don't see it being used anywhere.

phlax:
@unho its not used in this PR - i separated out because i felt that the data model was worth reviewing by itself - the next PR adds some tools for getting/setting this data and event handlers to update as necessary

PR #5167, opened 2016-08-29
Releasing 2.8.0b4

unho:


julen:
In case you find it helpful, I'd like to make an observation on release notes: I get the feeling these beta/RC changelogs are unnecessarily verbose and a bit difficult to digest, as they list the changes for the entire 2.8 cycle.

I believe it would be clearer for the reader only to highlight the changes across beta/RC versions, and leave the full details for the final changelog (Django does this well, check their [unreleased notes on version 1.11](https://docs.djangoproject.com/en/dev/releases/1.11/) for instance).

unho:
The idea is to keep collecting all the changes and do the final filtering for the final release notes. Despite it might be too verbose at least the data is there, as apposed to old Pootle releases.

julen:
Hmmm maybe you misunderstood; as far as I know I haven't proposed to do whatever any old Pootle releases were doing. I believe that it might be saner to keep track of entire release changes in one place instead of duplicating information across minor betas altogether (check the Django example, which they link to from every release announcement). Data is there.

unho:
Somehow this PR was on a branch that I wrongly pushed to origin repo instead of mine. New PR is https://github.com/translate/pootle/issues/5177

PR #5172, opened 2016-08-29
Do not create stores with (back)slashes on their names

### LINKED ISSUE #4989: Store.create should not allow /s in the Store.name
atm its possible to create a Store that has `/` in the name - which foos with the directory system

unho:
Fixes #4989.

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1169,7 +1169,8 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
                                 db_index=True)
     creation_time = models.DateTimeField(auto_now_add=True, db_index=True,
                                          editable=False, null=True)
-    last_sync_revision = models.IntegerField(db_index=True, null=True)
+    last_sync_revision = models.IntegerField(db_index=True, null=True,
+                                             blank=True)
     obsolete = models.BooleanField(default=False)
 
     objects = StoreManager()
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -18,6 +18,7 @@ from translate.storage import base
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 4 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -1228,6 +1238,11 @@ def save(self, *args, **kwargs):
 
         self.pootle_path = self.parent.pootle_path + self.name
 
+        # Force validation of "name" field.
+        exclude_fields = list(set([f.name for f in self._meta.get_fields()]) -
+                              set(["name"]))
+        self.clean_fields(exclude=exclude_fields)
```
unho: In case someone is asking why only "name" field is being validated, it is because if we use `full_clean` basically all tests fail.

pootle/apps/pootle_store/models.py
```diff
@@ -1228,6 +1238,11 @@ def save(self, *args, **kwargs):
 
         self.pootle_path = self.parent.pootle_path + self.name
 
+        # Force validation of "name" field.
+        exclude_fields = list(set([f.name for f in self._meta.get_fields()]) -
```
phlax: can you put the `-` at the start of next line - it tooik a while to see it

pootle/apps/pootle_store/models.py
phlax: why ae the tests failing?

pootle/apps/pootle_store/models.py
unho: Done

pootle/apps/pootle_store/models.py
unho: I didn't had a deep look on all the failures there, but the first ones, the ones I checked, where because of some validation error on `last_sync_revision`.

unho: For the record: https://travis-ci.org/translate/pootle/jobs/156187908

phlax:
lgtm - thanks for picking this up

PR #5174, opened 2016-08-30
Adjust StoreDataTool.get_max_unit_revision

phlax:
As obsolete units can have a higher revision than the the last `updated` unit, this PR adds a last `changed` unit to include obsolete units, and retrieve the correct max_unit_revision


unho:
lgtm

PR #5176, opened 2016-08-30
Add index_together on unit__store with revision and mtime

phlax:
this seems to give significant performance boosts to queries involivng unit revision in a store - eg get_max_unit_revision - as units are also sometimes sorted my mtime this is added tooo


ta2-1:
we have tested adding of these indices and it speeded up timing on get_max_revision query to `0.0 sec`. And it doesn't look like it affects update performance. so GTM.

PR #5179, opened 2016-08-31
StoreSyncer: fix bogus logic

### LINKED ISSUE #5178: Severe performance degradation of sync_stores
Earlier today we updated production translation server (merged 'Raw font' PR branch with master and switched to master). So this release includes recent changes related to optimizations. Immediately after that our sync cycle time increased from typical 18-20 minutes to 1.5 hours. I had to write a tool to analyze and compare our logs from past sync cycles and extract timing information. Here's the output:

screenclip

Some comments on the screenshot. It compares our main steps of sync cycle between four logs (vertical columns). The first log is from a morning run, some time before the release. The second one is soon after the release, and the last two are most recent ones. As one might see, the `pull-ts` step is the culprit. It went from mere 3 minutes up to more than an hour. During this step all we do is run `manage.py sync_stores --skip-missing --project=<project_id>` and wait for it to finish.

julen:
Optimizations were not kicking in because the logic would lead to process all
stores regardless of the last sync revision, therefore hitting the disk for
every single store and badly degrading performance.

Fixes #5178.

### DIFF
--- a/pootle/apps/pootle_store/syncer.py
+++ b/pootle/apps/pootle_store/syncer.py
@@ -256,7 +256,8 @@ class StoreSyncer(object):
     def update_newer(self, last_revision):
         return (
             not self.store.file.exists()
-            or (last_revision >= self.store.last_sync_revision))
+            or last_revision > self.store.last_sync_revision
+        )
 
     @cached_property
     def dbid_index(self):


unho:
lgtm

ta2-1:
gtm

Issue #4968, opened 2016-07-26
Pootle refuses to run on TTK 1.13


julen:
I updated our staging server moments ago and I was surprised that after restarting the RQ workers, they were not showing up on the list of processes being run. After digging up a bit I realize `manage.py rqworker` refuses to start because TTK's minimum hard dependency is set to version 2.

My understanding of #4914 was that there was a will to push TTK2 to beta users, however [not to make it a minimum hard dependency right away](https://github.com/translate/pootle/commit/c090e51070c2a06a74305bd0974dd7da72d9fdca). From a business perspective, and given the major overhauling that it went through, we have no motivation to bump up our TTK dependency and would like to remain in 1.13 for the time being because it's known to work well.

So we'd appreciate if you would consider reverting the change to the minimum dependency.

phlax:
@julen please see this comment https://github.com/translate/pootle/pull/4914#issuecomment-233708697

afaia you dont install - so this really shouldnt be a problem for you

unho:
> afaia you dont install - so this really shouldnt be a problem for you

But the [checks for runtime](https://github.com/translate/pootle/blob/master/pootle/checks.py#L15) force this minimum TTK to be in place. Reverting that while keeping the requirement should be enough for Evernote.

phlax:
wfm

julen:
Yes, my suggestion was to change the [minimum dependency check](https://github.com/translate/pootle/blob/b635565a93bfc085b70dc080ebbd06c8c43308bb/pootle/checks.py#L15) so anyone can still run Pootle along with a stable TTK version.

PR #5180, opened 2016-08-31
Bump TTK requirement to 2.0.0b5

unho:


dwaynebailey:
lgtm

Did we want to bump `TTK_MINIMUM_REQUIRED_VERSION` at all?  IIRC we didn't just wondering out loud.

julen:
> Did we want to bump TTK_MINIMUM_REQUIRED_VERSION at all? IIRC we didn't just wondering out loud.

No, we don't. It's worth recalling [TTK2 is in a pre-release state](https://github.com/translate/pootle/pull/4914), and raising the minimum requirement [would make it impossible to run Pootle with a stable TTK](https://github.com/translate/pootle/issues/4968).

PR #5181, opened 2016-08-31
assets: pass `--production` flag to `npm install`

julen:
This prevents `devDependencies` from being installed, which are unnecessary in
production builds.


julen:
It seems we are out of luck as `webpack` is within the `devDependencies` so it won't be installed and building won't work...

Issue #3704, opened 2015-05-02
Logging: add a Pootle logging reference


dwaynebailey:
Pootle should also need its own [logging] reference (see [comment in #3701](https://github.com/translate/pootle/pull/3701#discussion_r29498214)), but that requires reviewing existing loggers and naming them consistently.

julen:
I had a quick overview of all the loggers Pootle is defining and actively using. Here's a summary:

Logger name | Notes
----------------- | ---------
`stats` | Used in `calculate_checks`, `refresh_stats`, `refresh_stats_rq` management commands, as well as through the `TreeItem` methods
`action` | In `pootle.core.log`, log()`, `action_log()`, `cmd_log()` and `store_log()` leverage this logger. They are tightly coupled to the output and their logging settings are basically hard-coded
`pootle.markup` | Used in `pootle.core.markup` for a couple of caching-related messages and configuration-related errors
`pootle.settings` | Used in *95-outro.conf* just for a `POOTLE_SETTINGS`-related message
`''` (catch-all) | Used all over the place, i.e. when no loggers are created explicitly (`import logging; logging.<method>(...)`)

This only displays what's being currently used, and it's far from any sort of classification because most of the calls fall into the last catch-all bucket.

It's also worth noting that many management commands are using logging calls as a pure mechanism to report feedback to user. In these cases `self.stdout.write()`/`self.stderr.write()` needs to be used instead.

On the other hand, naming consistently here means using explicit loggers all the time, with a naming scheme such as `pootle.<foo>.<bar>`. This way it's possible to configure loggers for an entire namespace, while optionally allowing users to have more fine-grained logging for specific parts.

*As an example: using `pootle.commands.<command-name>` allows to capture logging output from any command by configuring a logger for `pootle.commands`. At the same time, an specific command's logger can be configured in `pootle.commands.<command-name>`.*

PR #5183, opened 2016-09-01
Debug tests

phlax:
This adds debugging fixtures and a cli option to specify a file to log to - also a context manager for debugging sql calls.

It logs query counts and timings for each tests - and does some very basic slow query analysis

test can be debugged as:

```
py.test --debug-tests my.log
```

you can use the context manager so:

```python
from pootle.core.debug import debug_sql
with debug_sql():
   do_something_with_sql_calls()
```
there is also a timings contextmanager - which can be used as:

```python
from pootle.core.debug import timings

with timings("name of thing being timed"):
    do_something_you_want_timings_from()

```

the contextmanagers can be used outside of tests - but in combination with `--debug-tests` will log to the same file

### DIFF
--- a/pootle/core/debug.py
+++ b/pootle/core/debug.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+import time
+from contextlib import contextmanager
+
+
+logger = logging.getLogger("POOTLE_DEBUG")
+
+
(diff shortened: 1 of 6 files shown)


unho:
lgtm

julen:
Very nice stuff @phlax!

pootle/core/debug.py
```diff
@@ -0,0 +1,63 @@
...
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import logging
+import time
+from contextlib import contextmanager
+
+
+logger = logging.getLogger("POOTLE_DEBUG")
```
julen: We probably want to consider using `pootle.*` names here to work towards consistency (see #3704 for further background on the matter).

PR #5185, opened 2016-09-01
Relative times: minor refactoring

julen:
The implementation for relative times has been split into pieces that can be
tested on their own and likewise tests have been added for these.

Now time deltas are calculated independently, and the display messages are only
a reflectiong of this data, hence allowing to implement any type of messaging
which involves relative times.

### DIFF
--- a/pootle/static/js/shared/utils/time.test.js
+++ b/pootle/static/js/shared/utils/time.test.js
@@ -1,0 +1,477 @@
+/*
+ * Copyright (C) Pootle contributors.
+ *
+ * This file is a part of the Pootle project. It is distributed under the GPL3
+ * or later license. See the LICENSE file for a copy of the license and the
+ * AUTHORS file for copyright and authorship information.
+ */
+
+import expect from 'expect';
+import lolex from 'lolex';
+import { describe, it } from 'mocha';
+
+import { relativeTime, relativeTimeMessage, timeDelta } from './time';
+
+
+describe('timeDelta', () => {
+  it('returns an empty structure for invalid datetimes', () => {
+    const undefinedDelta = {
+      isFuture: null,
+      seconds: null,
+      minutes: null,
+      hours: null,
+      days: null,
+      weeks: null,
+      months: null,
+      years: null,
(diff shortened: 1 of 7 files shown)


pootle/static/js/shared/utils/time.test.js
```diff
@@ -0,0 +1,507 @@
...
+          minutes: 59,
+          hours: 0,
+          days: 0,
+          weeks: 0,
+          months: 0,
+          years: 0,
+        },
+      },
+      {
+        unit: 'minutes',
```
unho: Does this, and the next test provide anything else that the two previous tests are not?

unho:
Very nice work. Really good to see JavaScript being tested too. Just some questions, how can I run those tests? Is Travis running those tests?

Reviewed the code and tested, raised nit aside GTM.

pootle/static/js/shared/utils/time.test.js
julen: The previous tests are for seconds. Are you perhaps referring to the other two tests for minutes? Two of them can potentially go, I feel like they are leftovers from some experimentation.

julen:
Leandro, you can run tests via `npm test` (`cd` to the _js_ folder first). Alternatively extra flags can be passed to the [Mocha runner](http://mochajs.org/) by passing arguments after `--`, e.g. `npm test -- --watch`. Note this is common to any [_package.json_ script](https://docs.npmjs.com/misc/scripts).

pootle/static/js/shared/utils/time.test.js
unho: Yes, I meant there are 4 tests for minutes while I think just 2 is enough. Somehow I posted in the wrong place (I thought I was in the third minutes test when commenting). My apologies.

unho:
@julen Do you think it is possible to trigger those JavaScript tests on Travis?

julen:
Technically possible, although in our current configuration that is not natively supported (see https://github.com/travis-ci/travis-ci/issues/4090 for details).

PR #5186, opened 2016-09-02
Validate directory name

unho:
Proper directory validation will prevent creating potentially broken directories. Checking for (back)slashes in directory name and that name is empty and parent is unset only for root directory. Adjusting the schema so the special case for root directory has no name and no parent is accepted.

### DIFF
--- a/pootle/apps/pootle_app/models/directory.py
+++ b/pootle/apps/pootle_app/models/directory.py
@@ -53,7 +53,7 @@ class Directory(models.Model, CachedTreeItem):
     # any changes to the `pootle_path` field may require updating the schema
     # see migration 0005_case_sensitive_schema.py
     pootle_path = models.CharField(max_length=255, null=False, db_index=True,
-                                   unique=True)
+                                   unique=True, default='/')
     obsolete = models.BooleanField(default=False)
 
     is_dir = True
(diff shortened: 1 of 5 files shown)


pootle/apps/pootle_app/models/directory.py
```diff
@@ -33,13 +34,22 @@ def projects(self):
...
 class Directory(models.Model, CachedTreeItem):
 
     # any changes to the `name` field may require updating the schema
     # see migration 0005_case_sensitive_schema.py
-    name = models.CharField(max_length=255, null=False)
+    name = models.CharField(max_length=255, null=False, blank=True,
+                            validators=[validate_no_slashes])
     parent = models.ForeignKey('Directory', related_name='child_dirs',
-                               null=True, db_index=True)
+                               null=True, blank=True, db_index=True)
```
julen: Is there a remote chance to create a root directory (i.e. a directory with a blank parent) from a form? I don't believe so... (Note how [`blank` only affects form validation](https://docs.djangoproject.com/en/1.8/ref/models/fields/#blank)).

phlax: @julen blank also affects model.full_clean afaiaa - good info here on model validation http://schneck.cc/node/41

julen: Ah right, I missed the `full_clean()` below.

phlax:
lgtm

PR #5187, opened 2016-09-03
Add virtualfolder data

phlax:
Adds 
- a data table for storing site-wide virtualfolder stats
- an m2m table for matched vfolders <> stores
- project and lang fk fields

### DIFF
--- a/pootle/apps/virtualfolder/migrations/0015_vfdata.py
+++ b/pootle/apps/virtualfolder/migrations/0015_vfdata.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pootle_statistics', '0004_fill_translated_wordcount'),
+        ('pootle_store', '0019_remove_unit_priority'),
+        ('virtualfolder', '0014_remove_location'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='VFData',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('max_unit_mtime', models.DateTimeField(db_index=True, null=True, blank=True)),
+                ('max_unit_revision', models.IntegerField(default=0, null=True, db_index=True, blank=True)),
+                ('critical_checks', models.IntegerField(default=0, db_index=True)),
+                ('pending_suggestions', models.IntegerField(default=0, db_index=True)),
+                ('total_words', models.IntegerField(default=0, db_index=True)),
+                ('translated_words', models.IntegerField(default=0, db_index=True)),
+                ('fuzzy_words', models.IntegerField(default=0, db_index=True)),
+                ('last_created_unit', models.OneToOneField(related_name='last_created_for_vfdata', null=True, blank=True, to='pootle_store.Unit')),
+                ('last_submission', models.OneToOneField(related_name='vfdata_stats_data', null=True, blank=True, to='pootle_statistics.Submission')),
+                ('last_updated_unit', models.OneToOneField(related_name='last_updated_for_vfdata', null=True, blank=True, to='pootle_store.Unit')),
+                ('vf', models.OneToOneField(related_name='data', to='virtualfolder.VirtualFolder')),
+            ],
+            options={
+                'db_table': 'pootle_vf_data',
+            },
+        ),
+    ]
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -18,6 +18,7 @@ from pootle.core.url_helpers import (get_all_pootle_paths, get_editor_filter,
                                      split_pootle_path)
 from pootle.i18n.gettext import ugettext_lazy as _
 from pootle_app.models import Directory
+from pootle_data.abstracts import AbstractPootleData
 from pootle_language.models import Language
 from pootle_project.models import Project
 from pootle_store.models import Store
@@ -272,3 +273,17 @@ class VirtualFolderTreeItem(models.Model, CachedTreeItem):
                 if p.count('/') > 3]
 
     # # # /TreeItem
+
+
+class VFData(AbstractPootleData):
+
+    class Meta(object):
+        db_table = "pootle_vf_data"
+
+    vf = models.OneToOneField(
+        "virtualfolder.VirtualFolder",
+        db_index=True,
+        related_name="data")
+
+    def __unicode__(self):
+        return unicode(self.vf)
--- a/tests/models/virtualfolder.py
+++ b/tests/models/virtualfolder.py
@@ -16,7 +16,7 @@ from pytest_pootle.factories import VirtualFolderDBFactory
 
 from pootle_language.models import Language
 from pootle_store.models import Store
-from virtualfolder.models import VirtualFolder, VirtualFolderTreeItem
+from virtualfolder.models import VirtualFolder, VirtualFolderTreeItem, VFData
 
 
 @pytest.mark.django_db
@@ -251,3 +251,11 @@ def test_vfolder_membership_new_store(tp0):
     assert Store.objects.get(pk=normal_store.pk).priority == 1.0
     vf0.delete()
     assert Store.objects.get(pk=wierd_store.pk).priority == 1.0
+
+
+@pytest.mark.pootle_vfolders
+@pytest.mark.django_db
+def test_vfdata_repr(vfolder0):
+    assert (
+        repr(VFData(vf=vfolder0))
+        == "<VFData: %s>" % vfolder0.name)


pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+                else qs.filter(pootle_path__startswith=rule_path))
+
+    def filter_from_rules(self, qs):
+        filtered = None
+        for rule in self.filter_rules:
+            filtered_qs = self.filter_by_rule(qs, rule)
+            filtered = (
+                filtered_qs
+                if filtered is None
+                else (filtered | filtered_qs))
```
unho: This if might not be necessary by setting `filtered = qs.none()` before the loop.

phlax: hmm - yep - not sure 8/

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            self.path_regex
+            if self.use_regex
+            else self.vf.location)
+
+    def get_rule_path(self, rule):
+        return "%s%s" % (self.path, rule)
+
+    def filter_by_rule(self, qs, rule):
+        rule_path = self.get_rule_path(rule)
+        is_file = "." in os.path.basename(rule)
```
unho: Not sure if we want to exclude directories with dots in the name.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.use_regex:
+            return (
+                qs.filter(pootle_path__regex="^%s$" % rule_path)
+                if is_file
+                else qs.filter(pootle_path__regex="^%s" % rule_path))
+        else:
+            return (
+                qs.filter(pootle_path=rule_path)
+                if is_file
+                else qs.filter(pootle_path__startswith=rule_path))
```
unho: Nit: no need to nest `else`

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        proj_code = None
+    return lang_code, proj_code, dir_path, filename
+
+
+class VirtualFolderStoreMatcher(object):
+
+    def __init__(self, vf):
+        self.vf = vf
+
+    def add_related_stores(self):
```
phlax: just thinking we might need to change this method to `update_related_stores` and include removing no longer related stores

pootle/apps/virtualfolder/utils.py
phlax: yep, wasnt sure what the best algo for `is_fileish` was

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @property
+    def use_regex(self):
+        return (
+            not self.lang_code
+            or not self.proj_code)
+
+    @property
+    def path_regex(self):
+        regex = "/[^/]*"
```
unho: If `{LANG}` or `{PROJ}` are in the vfolder location then they must appear once yes, or yes. So I am wondering if `*` should be replaced by `{1}`.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        return "%s%s" % (self.path, rule)
+
+    def filter_by_rule(self, qs, rule):
+        rule_path = self.get_rule_path(rule)
+        is_file = "." in os.path.basename(rule)
+        if self.use_regex:
+            return (
+                qs.filter(pootle_path__regex="^%s$" % rule_path)
+                if is_file
+                else qs.filter(pootle_path__regex="^%s" % rule_path))
```
unho: Just wondering if taking in account `is_file` provides us any performance boost.

phlax: i would imagine so - its the equiv of difference between endswith and matches - but on a more complex query

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            self.vf.stores.add(store)
+
+    @property
+    def filter_rules(self):
+        return (
+            r.strip()
+            for r
+            in self.vf.filter_rules.split(","))
+
+    @property
```
unho: Is it not better to cache this?

pootle/apps/virtualfolder/utils.py
phlax: no - this will match the actual lang/proj `code`

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            filtered = (
+                filtered_qs
+                if filtered is None
+                else (filtered | filtered_qs))
+        return (
+            qs
+            if filtered is None
+            else filtered)
+
+    def find_related_stores(self):
```
unho: Nit: I would use "matching" instead of "related".

pootle/apps/virtualfolder/utils.py
phlax: i was gonna say...

> for this type of tool - im reluctant to cache anything - because its cached on the model - and so if it was reused elsewhere in the code it might get unexpected results

but in this case i think you are right - we might want to think about the api for changing a vf location in that case tho, so that this is cleared

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def add_related_stores(self):
+        existing_stores = self.vf.stores.all()
+        related_stores = self.find_related_stores()
+        for store in related_stores:
+            if store not in existing_stores:
+                self.vf.stores.add(store)
+
+    def add_store_if_related(self, store):
+        if store in self.find_related_stores():
```
unho: I find this ineficient. First I would check if it is already in `self.vf.stores.all()` and if it is not in there I wouldn't user `find_related_stores` to find out if store matches any rule since it hits the DB. There should be a faster way of matching the store's pootle_path against all the location+filter_rules.

phlax: @unho im not sure i agree

if you look 2 lines above there is an optimized version, for the use case you are kinda describing.

this is for the situation where we have a single store and we need to check authoritatively whether the store should be a member

phlax: ...having said that we probs  want to make it more like `update_store_relations` - and check to see if it requires removal also

unho:
One missing feature is removing relationships to stores that no longer match. Note that vfolder locations or rules can change.

phlax:
yep, for sure, lets update and test

pootle/apps/virtualfolder/utils.py
unho: k, it has been some time since last time I did regexes. Had to grab the book.

pootle/apps/virtualfolder/utils.py
unho: For a single store we only need to iterate over location+filter_rules and match store pootle_path against each. It is faster since we don't hit the DB at all.

phlax: good point

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @property
+    def store_qs(self):
+        qs = Store.objects
+        if self.lang_code:
+            qs = qs.filter(translation_project__language__code=self.lang_code)
+        if self.proj_code:
+            qs = qs.filter(translation_project__project__code=self.proj_code)
+        if self.use_regex:
+            if self.lang_code or self.proj_code:
```
unho: Is it not better to merge this `if` with its parent `if`?

phlax: i think not because of the else

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+            in self.filter_rules)
+
+    @property
+    def path(self):
+        return (
+            self.path_regex
+            if self.use_regex
+            else self.vf.location)
+
+    def get_rule_path(self, rule):
```
unho: Is this necessary now that we have `regexes`?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        return qs
+
+    @property
+    def use_regex(self):
+        return (
+            not self.lang_code
+            or not self.proj_code)
+
+    @property
+    def path_regex(self):
```
unho: Please rename to `location_regex`. Lets not introduce more misleading names.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def filter_from_rules(self, qs):
+        if not self.filter_rules:
+            return qs
+        filtered_qs = qs.none()
+        for rule in self.filter_rules:
+            filtered_qs = filtered_qs | self.filter_by_rule(qs, rule)
+        return filtered_qs
+
+    def find_matching_stores(self):
```
unho: This can be merged with `filter_from_rules`.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        return self.filter_from_rules(self.store_qs)
+
+    def path_matches(self, path):
+        """Returns bool of whether path is valid for this VF.
+        """
+        (lang_code, proj_code,
+         dir_path, filename) = split_pootle_path(path)
+        dir_path = "/".join(dir_path.split("/")[1:])
+        no_match = (
+            self.lang_code and lang_code != self.lang_code
```
unho: I would enclose this line in parentheses to make easier to understanding the logic.

pootle/apps/virtualfolder/utils.py
phlax: not sure which is necessary  - `regexes` doesnt seem to be used afaict

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+            or (self.proj_code and proj_code != self.proj_code)
+            or (self.dir_path and not self.dir_path.startswith(dir_path)))
+        if no_match:
+            return False
+        # the path matches the vfolder - but we need to check the filter rules
+        # to ensure any remaining dir_path also matches
+        remaining_dir_path = dir_path[len(self.dir_path):]
+        for rule in self.filter_rules:
+            if rule.startswith(remaining_dir_path):
```
unho: You are assuming that vfolder location can only be `/lang/proj/` but it can actually can exists a vfolder with location `/af/firefox/browser/chrome/browser/preferences/` and rules `advanced.dtd.po,colors.dtd.po`. I suggest you to expand location+rules to get all possible combinations and then check if passed path matches or starts with any of those.

pootle/apps/virtualfolder/utils.py
phlax: i would rather keep these separate tbh

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def path_matches(self, path):
+        """Returns bool of whether path is valid for this VF.
+        """
+        (lang_code, proj_code,
+         dir_path, filename) = split_pootle_path(path)
+        dir_path = "/".join(dir_path.split("/")[1:])
+        no_match = (
+            self.lang_code and lang_code != self.lang_code
+            or (self.proj_code and proj_code != self.proj_code)
```
unho: Note that vfolder location can just be `/{LANG}/` or `/af/`, so the project code won't be here but in the rules (if any). In that case this line won't match and also the next one won't match either.

pootle/apps/virtualfolder/utils.py
unho: Thinking it better, this inner `if` is unnecessary. I mean, if either `self.lang_code` or `self.proj_code` are `None` then `self.use_regex` is `True`.

pootle/apps/virtualfolder/utils.py
unho: But it can be. Instead of building the regex here use instead `regexes`.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,155 @@
...
+            self.vf.stores.add(*to_add)
+        if to_delete:
+            self.vf.stores.delete(*to_delete)
+
+    def add_store_if_matching(self, store):
+        pootle_dir_path = "".join(
+            ("/", ) + split_pootle_path(store.pootle_path)[2:])
+        no_add = (
+            not self.path_matches(pootle_dir_path)
+            or self.vf.stores.through.objects.filter(store_id=store.id).exists())
```
phlax: `store in self.vf.stores`

phlax: not sure about this - using the `through` we are saving a table hop

phlax: maybe

phlax: ill `debug_sql` it

unho:
lgtm

PR #5190, opened 2016-09-03
Pootle data tools

phlax:
- adds data_tools for pootle types
- adds tests to ensure that they get/set correct data

### DIFF
--- a/pootle/apps/pootle_data/language_data.py
+++ b/pootle/apps/pootle_data/language_data.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from .utils import RelatedTPsDataTool
+
+
+class LanguageDataTool(RelatedTPsDataTool):
+    """Retrieves aggregate stats for a Language"""
+
+    def filter_data(self, qs):
+        return qs.filter(tp__language=self.context)
--- a/pootle/apps/pootle_data/project_data.py
+++ b/pootle/apps/pootle_data/project_data.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+from .utils import RelatedStoresDataTool, RelatedTPsDataTool
+
+
+class ProjectDataTool(RelatedTPsDataTool):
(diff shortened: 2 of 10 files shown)


pootle/apps/pootle_project/models.py
```diff
@@ -560,6 +564,10 @@ def __eq__(self, other):
             self.pootle_path == other.pootle_path
             and list(self.get_children()) == list(other.get_children()))
 
+    @cached_property
+    def data_tool(self):
+        return data_tool.get(self.__class__)(self)
```
unho: Is this not repeated in the same model?

pootle/apps/pootle_data/language_data.py
```diff
@@ -0,0 +1,30 @@
...
+    def data_model(self):
+        return TPData.objects
+
+    @property
+    def stat_data(self):
+        return self.data_model.filter(
+            tp__language=self.context)
+
+    def get_root_child_path(self, child):
+        return "-".join(child[field] for field in self.group_by)
```
unho: This is very misleading. It doesn't seem to match the method name at all.

pootle/apps/pootle_project/models.py
unho: Forget about that. Somehow missed the class when looking for it.

pootle/apps/pootle_data/project_data.py
```diff
@@ -0,0 +1,71 @@
...
+        return split_pootle_path(self.context.pootle_path)[1]
+
+    @property
+    def dir_path(self):
+        return split_pootle_path(self.context.pootle_path)[2]
+
+    @property
+    def filename(self):
+        path_parts = split_pootle_path(self.context.pootle_path)
+        return path_parts[3]
```
unho: No need for `path_parts`.

pootle/apps/pootle_data/language_data.py
phlax: k - we can revisit this - i changed it to this to try and make it clearer, but obv didnt work. Any suggestions?

unho: No suggestions, since I am not 100% sure what is this expected to return. But I would expect the path to be joined with `/` instead.

phlax: its ouputting the expectation of the js on the page - in the case of langs it matches the children (which are tps) with `lang`-`proj`

pootle/apps/pootle_data/utils.py
```diff
@@ -424,3 +499,42 @@ def get_lastupdated(self, **kwargs):
...
+            for k, v
+            in stats.items()}
+        stats["lastaction"] = None
+        stats["lastupdated"] = None
+        stats["suggestions"] = None
+        return stats
+
+    def show_all_to(self, user):
+        if user and user.is_superuser:
+            return True
```
unho: Nit: `return user and user.is_superuser`

unho:
Nit aside lgtm.

PR #5192, opened 2016-09-05
underscore: use package from npm

julen:
The reason why the vendor'ed package was kept in the repository is that at the
time of switching the codebase to use modules, underscore wasn't available on
npm.

Now it is and we pin to the specific version we have been using for a while.
This will allow for `underscore` imports to work in non-webpack contexts — right
now it worked only because webpack internally aliases underscore to the
vendor'ed package.


unho:
lgtm

dwaynebailey:
`js/README` also needs cleaning.

PR #5193, opened 2016-09-05
JS README: just refer to vendor'ed packages

julen:
The rest of the references were helpful in a world where only `<script>` tags
existed, but it's not the case anymore, and having now multiple small modules
makes keeping this list non-scalable and a bit redundant.


unho:
Not sure about loosing own modules description, those were useful.

julen:
I don't believe you do the same for Python modules, do you? That would be insane 😄. One can still decide to add descriptions in modules themselves when deemed necessary. Please read the PR's description for the general reasoning.

unho:
Sure, lgtm.

PR #5195, opened 2016-09-05
Relative times: keep private members private

julen:
Besides moving the code specific to relative times' messaging to its own module (this will grow easily), the functions that were meant to be private are now actually private. The only reason preventing us from doing so was the inability to import the private functions for testing, but with the help of the rewire Babel plugin that is now possible.


pootle/static/js/shared/utils/relativeTime.js
```diff
@@ -0,0 +1,153 @@
...
+import { nt, t } from './i18n';
+
+
+/**
+ * A version of `Math.trunc` which returns absolute values.
+ *
+ * @param {Number} number - the number to truncate.
+ * @return {Number} - positive integer
+ */
+function absTrunc(number) {
```
unho: This is still on `pootle/static/js/shared/utils/time.js`. Do we keep copies?

pootle/static/js/shared/utils/relativeTime.js
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param {String} isoDateTime - a date and time combination in ISO 8601 format
+ * as understood by browsers' `Date.parse()`.
+ * @return {Object} - An object with the delta difference specified in seconds,
+ * minutes, hours, days, weeks and months. The `isFuture` boolean
+ * property indicates whether the difference refers to a future time.
+ * When the value passed is invalid or cannot be parsed, every property in the
+ * object will be `null`.
+ */
+export function timeDelta(isoDateTime) {
```
unho: This one is too on `pootle/static/js/shared/utils/time.js`.

unho: The same goes for the constants below.

pootle/static/js/shared/utils/relativeTime.js
julen: Oh, good catch, this is a copying & pasting remnant. I removed it, and also adjusted the relevant import statements.

julen: To clarify: with _this_ I meant the copies in the _time.js_ module.

PR #5196, opened 2016-09-05
In memory testing on travis

phlax:
Pytest-django3 broke support for in-memory testing with transaction_hooks

This PR uses my branch of pytest-django and updates the travis db name as required.

Ill PR to pytest-django - but in the meantime this speeds up testing massively

### DIFF
--- a/pootle/settings/91-travis.conf
+++ b/pootle/settings/91-travis.conf
@@ -24,7 +24,7 @@ if os.environ.get("TRAVIS"):
             'PORT': '',
             'ATOMIC_REQUESTS': True,
             'TEST': {
-                'NAME': 'test_pootle',
+                'NAME': '',
             }
         }
     }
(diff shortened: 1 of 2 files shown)


julen:
Good catch! So long as a DB name is specified, pytest-django will never use an in-memory DB. This seems to be the case regardless of using transaction hooks or not in combination with SQLite3.

Another possible workaround might be explicitly specifying `:memory:` in the test DB name.

By the way, [pytest-django 2.9 exhibited the same behavior](https://github.com/pytest-dev/pytest-django/blob/78258372191e6f95169cf05aafb35b06dafa6404/pytest_django/db_reuse.py#L62-L64), so pytest-django didn't break any support explicitly.

phlax:
> By the way, pytest-django 2.9 exhibited the same behavior, so pytest-django didn't break any support explicitly.

maybe it was an older version - i noticed that my sqlite testing slowed right down when i upgraded to 3

julen:
> Ill PR to pytest-django - but in the meantime this speeds up testing massively

Please don't forget to do so as soon as possible. Note pytest-django is a project run by volunteers and the change might take a while to be considered, merged and released.

PR #5197, opened 2016-09-06
Move priority from unit to store

phlax:
in the current implementation (of vfolders) all units in a store have the same priority.

moving priority to store makes this a lot less expensive, and will allow us to more easily shift vfolder assocs from unit to store

### DIFF
--- a/pootle/apps/pootle_store/constants.py
+++ b/pootle/apps/pootle_store/constants.py
@@ -73,3 +73,7 @@ STATES_NAMES = {
     UNTRANSLATED: "untranslated",
     FUZZY: "fuzzy",
     TRANSLATED: "translated"}
+
+# Default store priority - used by vfolders atm
+#  - valid range = 0 < n <= 999.99
+DEFAULT_PRIORITY = 1.0
--- a/pootle/apps/pootle_store/constants.py
+++ b/pootle/apps/pootle_store/constants.py
@@ -14,7 +14,7 @@ from django.utils.translation import ugettext_lazy as _
 #: will be used against the DB.
 ALLOWED_SORTS = {
     'units': {
-        'priority': '-priority',
+        'priority': '-store__priority',
         'oldest': 'submitted_on',
         'newest': '-submitted_on',
     },
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -52,7 +52,7 @@ from pootle_statistics.models import (Submission, SubmissionFields,
                                       SubmissionTypes)
 
 from .constants import (
-    FUZZY, NEW, OBSOLETE, PARSED, POOTLE_WINS,
+    DEFAULT_PRIORITY, FUZZY, NEW, OBSOLETE, PARSED, POOTLE_WINS,
     TRANSLATED, UNTRANSLATED)
 from .fields import (PLURAL_PLACEHOLDER, SEPARATOR, MultiStringField,
                      TranslationStoreField)
@@ -490,24 +490,6 @@ class Unit(models.Model, base.TranslationUnit):
         information from the database as the target format can support.
         """
         return self.unit_syncer.convert(unitclass)
-
-    def calculate_priority(self):
-        if not vfolders_installed():
-            return 1.0
-
-        priority = (
-            self.vfolders.order_by("-priority")
-                         .values_list("priority", flat=True)
-                         .first())
-        if priority is None:
-            return 1.0
-        return priority
-
-    def set_priority(self):
-        priority = self.calculate_priority()
-
-        if priority != self.priority:
-            Unit.objects.filter(pk=self.pk).update(priority=priority)
 
     def sync(self, unit):
         """Sync in file unit with translations from the DB."""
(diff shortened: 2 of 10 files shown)


iafan:
Note that current vfolder implemenation is incomplete. The original specification on vfolders implies a per-unit granularity.

phlax:
@iafan atm it pointlessly triggers updates on every unit change, for data taht is not used

iafan:
...because the implementation is incomplete. I'm just saying that this PR moves us one step back further from implementing vfolders the way they were supposed to work.

phlax:
@iafan afaiaa you have never used the vfolder functionality

atm its **very** broken

this causes actual users a lot of pain - i am addressing this in the context of the current functionality - not a spec from several years ago

iafan:
> afaiaa you have never used the vfolder functionality

We tried, but it was pretty useless for us, because — as you might have already guessed — _the implementation is incomplete_. I was hoping, however, that we will at some point start moving towards making it really usable, but I guess with this PR we're accepting the current state of implementation (store-level granularity only) as a de-facto specification amendment.

phlax:
youll have something you can actually build on if you ever do

dwaynebailey:
@iafan by addressing vfolders we've taken test query count down from 120k to 80k.  This step will reduce that further I believe.  We're simply matching the model with current possible use cases.  I see real value in per unit vfolders.  But honestly as *nobody* has worked further to implement the remainder of the spec it makes much more sense to address pain points that are on live servers.  If anyone starts to address unit level use cases then that can be implemented on top of a more robust implementation.

pootle/apps/pootle_store/models.py
```diff
@@ -1178,6 +1178,11 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
                                              blank=True)
     obsolete = models.BooleanField(default=False)
 
+    # this is calculated from virtualfolders if installed and linked
+    priority = models.FloatField(
+        db_index=True, default=1,
+        validators=[MinValueValidator(0)])
```
unho: This accepts priority equal to zero. The specification and the docs do require greater than zero.

phlax: its just copied from `Unit` - i would prefer if we could we deal with data validation issues separately if poss

pootle/apps/virtualfolder/receivers.py
```diff
@@ -92,20 +92,22 @@ def vfolder_unit_priority_presave_handler(**kwargs):
                                 vfolders=original))
         for unit in removed_units.iterator():
             unit.vfolders.remove(original)
-            unit.set_priority()
             stores_affected.add(unit.store)
 
     for store in stores_affected:
         update_vfolder_tree(original, store)
+        store.set_priority()
```
unho: This is unnecessary. Previously obsoleted units were recalculated priority. But for stores doesn't apply. Even if all units on the store is obsoleted.

pootle/apps/pootle_store/models.py
```diff
@@ -1282,6 +1264,26 @@ def delete(self, *args, **kwargs):
         for p in parents:
             p.update_all_cache()
 
+    def calculate_priority(self):
+        if not vfolders_installed():
+            return 1.0
```
unho: Just wondering if this should be moved to a constant.

pootle/apps/pootle_store/models.py
unho: Sure, just noticed this was already there. 100% agree on dealing with this afterwards.

pootle/apps/pootle_store/models.py
phlax: wfm

pootle/apps/virtualfolder/receivers.py
phlax: great!

pootle/apps/virtualfolder/migrations/0002_set_unit_priorities.py
```diff
@@ -36,6 +36,17 @@ def forwards(apps, schema_editor):
                 Unit.objects.filter(pk=pk).update(priority=new_priority)
 
 
+def forwards(apps, schema_editor):
+    # as we have no real way of controlling whether this will be
+    # run before or after priority was moved to store, we need to
+    # test the field exists
+    try:
+        apps.get_model("pootle_store.Unit").objects.filter(priority=1)
+        return forwards_units(apps, schema_editor)
```
unho: I would move this to the `else` clause of `try`.

phlax: yep k

pootle/apps/virtualfolder/receivers.py
phlax: turns out this is necessary in this implementation - the signals will change anyway in the next related PR

pootle/apps/pootle_store/constants.py
```diff
@@ -73,3 +73,7 @@
     UNTRANSLATED: "untranslated",
     FUZZY: "fuzzy",
     TRANSLATED: "translated"}
+
+# Default store priority - used by vfolders atm
+#  - valid range = 0 < n <= 5.0
```
unho: Why `5.0`? I agree on setting some upper limit, but at least on MLO we have a vfolder with priority `6.0`, and in the [example in docs](http://docs.translatehouse.org/projects/pootle/en/latest/features/virtual_folders.html#format-for-the-json-file) we have `999.99`.

pootle/apps/virtualfolder/receivers.py
unho: ok

pootle/apps/pootle_store/constants.py
phlax: my bad - i thought that was the range - ill change to 999.99

pootle/apps/pootle_store/constants.py
```diff
@@ -73,3 +73,7 @@
     UNTRANSLATED: "untranslated",
     FUZZY: "fuzzy",
     TRANSLATED: "translated"}
+
+# Default store priority - used by vfolders atm
+#  - valid range = 0 < n <= 999.99
```
unho: It would be great to enforce it as we do for lower limit. But lets leave that for further PR.

unho:
lgtm

PR #5198, opened 2016-09-06
Tests: fix encoding issue

julen:
Reading from a store with non-ASCII characters would fail with:
```
pytest_pootle/fixtures/views.py:411: in tp_uploads
    updated_store = create_store(store.pootle_path, 0, updated_units)
pytest_pootle/utils.py:75: in create_store
    io_store = io.BytesIO(string_store.encode())
E   UnicodeEncodeError: 'ascii' codec can't encode character u'\xb4' in position
1867: ordinal not in range(128)
```


unho:
lgtm

PR #5200, opened 2016-09-06
Cleanup: use `JsonResponse` in `APIView`

julen:
We already have this logic implemented, so let's reuse it and benefit from
consistent behavior.


phlax:
lgtm

PR #5204, opened 2016-09-07
Optimization: use bulk_create when recalculating checks

phlax:
quick cleanup to give a small boost when recalculating checks

### DIFF
--- a/pootle/core/checks/checker.py
+++ b/pootle/core/checks/checker.py
@@ -127,6 +127,7 @@ class UnitQualityCheck(object):
         Removes members of self.original_checks as they have been compared.
         """
         updated = False
+        new_checks = []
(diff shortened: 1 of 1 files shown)


pootle/core/checks/checker.py
```diff
@@ -140,13 +141,15 @@ def update_checks(self):
                 continue
 
             # the check didnt exist previously - so create it
-            self.checks_qs.create(
-                unit_id=self.unit.id,
-                name=name,
-                message=self.check_failures[name]['message'],
-                category=self.check_failures[name]['category'])
+            new_checks.append(
+                self.checks_qs.model(
```
unho: Why not just `QualityCheck`?

phlax: more extensible/reusable

PR #5206, opened 2016-09-07
Tests: serialize using jsonify in create_api_request

julen:
This ensures the same behavior as in application code. Note we don't use a plain
`json.dumps()`, but we provide our own encoding class which supports encoding
more data types than the default implementation.

The conditional import statement is necessary to avoid circular imports and
to allow tests to be run without breaking pytest-django.

### DIFF
--- a/pytest_pootle/utils.py
+++ b/pytest_pootle/utils.py
@@ -9,7 +9,6 @@
 """Random utilities for tests."""
 
 import io
-import json
 import sys
 import time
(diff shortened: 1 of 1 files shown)


phlax:
> The conditional import statement is necessary to avoid circular imports

just wondering if there is a cleaner way to fix the circular import

phlax:
apologies - just realised that this is in a fixture - you cant import pootle code there

Issue #3192, opened 2014-03-28
Language dropdown not showing template languages


chhitz:
*Version: 2.5.1*

The language dropdown menu is only displaying languages that are _not_ specified as source language in any project.

Deactivating this filter shows all possible languages and does not seem to have any undesirable side effects:

```diff
diff --git a/pootle/apps/pootle_language/models.py b/pootle/apps/pootle_language/models.py
index 03dbd96..9611602 100644
--- a/pootle/apps/pootle_language/models.py
+++ b/pootle/apps/pootle_language/models.py
@@ -56,7 +56,6 @@ class LiveLanguageManager(models.Manager):
         return super(LiveLanguageManager, self).get_query_set().filter(
                 ~models.Q(code='templates'),
                 translationproject__isnull=False,
-                project__isnull=True,
             ).distinct()
```

jleclanche:
@julen is that what you guys just fixed?

julen:
We removed the explicit check for `templates`, this talks about source languages in general. Feel free to check it out and close if it doesn't apply anymore.

julen:
`project__isnull=True` was there to avoid routing users to the English source language, which leads to a misleading empty page in our case and the vast majority of deployments, but it's not always the desired behavior as shown by the OP.

In any case the actual fix is not *just* taking that line out, but taking out languages with no live TPs. As described in the manager's docstring: *A live language is any language containing at least a project with translatable files.* The pushed commit is a regression in this aspect, hence reopening.

unho:
Just for the record, commit 21b59033 has been reverted in 9387c920. So this issue is still not fixed as told in the previous commit.

unho:
There is agreement on including any template language on the dropdown, but instead hide any language without visible TPs.

PR #5208, opened 2016-09-07
Display in dropdown only languages with visible TPs

unho:
Addresses issue #3192.


pootle/apps/pootle_language/models.py
```diff
@@ -33,7 +33,9 @@ class LiveLanguageManager(models.Manager):
     def get_queryset(self):
         return super(LiveLanguageManager, self).get_queryset().filter(
             translationproject__isnull=False,
-            project__isnull=True,
```
phlax: do we not need the project__isnull ?

unho: No, we do no need that anymore. That bit is to exclude languages that are used as `source_language` in any project. With this PR we are trying to expose those.

unho:
This PR introduces plenty of new queries (the first test is already on master):

```
================================================================================
TESTS START: 2016-09-16 11:16:59.834071
TESTS: py.test -k tests/models/language.py -vv --debug-tests ../mytest.log
================================================================================
test_language_repr                                                          0.0042     1         
test_unused_language_is_not_live                                            0.0074     15        
test_language_only_in_disabled_proj_is_not_live                             0.0262     51        
test_language_only_in_obsolete_tp_is_not_live                               0.0267     53        
test_language_only_in_disabled_project_and_obsolete_tp_is_not_live          0.0524     89        
test_language_both_in_disabled_project_and_regular_tp_is_live               0.0421     81        
test_language_both_in_obsolete_and_regular_tp_is_live                       0.0496     96        
test_language_in_regular_tp_and_disabled_project_and_obsolete_tp_is_live    0.0605     119       
test_templates_language_is_live                                             0.0209     44        
================================================================================
TESTS END: 2016-09-16 11:17:00.140990
TESTS AVERAGE query time: 0.000559018175024
TESTS TOTAL test time: 0.306900978088
TESTS TOTAL queries: 549
================================================================================
```

tests/models/language.py
```diff
@@ -17,3 +20,143 @@ def test_language_repr():
...
+    assert language not in Language.live.all()
+
+
+@pytest.mark.django_db
+def test_language_both_in_disabled_project_and_regular_tp_is_live():
+    language = LanguageDBFactory()
+
+    # Create TP for disabled project.
+    project = ProjectDBFactory(source_language=language)
+    project.disabled = True
```
phlax: use the disabled_project fixture

tests/models/language.py
```diff
@@ -17,3 +20,143 @@ def test_language_repr():
...
+    assert language in Language.live.all()
+
+
+@pytest.mark.django_db
+def test_language_both_in_obsolete_and_regular_tp_is_live():
+    language = LanguageDBFactory()
+
+    # Create obsolete TP.
+    language = LanguageDBFactory()
+    tp = TranslationProjectFactory(
```
phlax: use tp0

tests/models/language.py
```diff
@@ -17,3 +20,143 @@ def test_language_repr():
...
+def test_unused_language_is_not_live():
+    unused_language = LanguageDBFactory()
+    assert unused_language.translationproject_set.count() == 0
+    assert unused_language not in Language.live.all()
+
+
+@pytest.mark.django_db
+def test_language_only_in_disabled_proj_is_not_live():
+    language = LanguageDBFactory()
+    project = ProjectDBFactory(source_language=language)
```
unho: It fails that way. But found another way to get it working with less queries and that is considerably faster.

pootle/apps/pootle_language/models.py
julen: > hiding unused languages and languages used only in disabled projects or in obsolete TPs.

Correct me if my understanding is wrong, but since admins can still access disabled projects/TPs, shouldn't we display for them languages used only in disabled projects too? Otherwise accessing such a TP would have a broken drop-down.

julen:
Looks good to me and works fine, although we should clarify the matter with disabled TPs.

tests/models/language.py
unho: It doesn't exist.

tests/models/language.py
unho: It fails too. Using `tp0` forces the test to use `language0`, but that is being used in several TPs and therefore we cannot ensure if it is fit for the test.

unho:
> Looks good to me and works fine, although we should clarify the matter with disabled TPs.

@julen I don't think any clarification is needed at all. Live languages are those that have at least a non-obsolete TP for a non-disabled project. If you mean that admin users should be able to see languages only used in disabled projects, then I guess that can be discussed afterwards.

unho:
Thinking better I will be addressing the admins-should-see-disabled scenario in this PR.

unho:
Updated with new code to handle different language lists for admin users and regular users, and also improved the tests to both increase coverage and make them more performant.

unho:
The new times are:

```
================================================================================
TESTS START: 2016-09-20 19:35:58.360612
TESTS: py.test -k tests/models/language.py -vv --debug-tests ../mytesdsfdsds.log
================================================================================
test_language_repr                                              0.0033     1         
test_language_only_in_disabled_proj_liveness                    0.0181     30        
test_language_liveness                                          0.1054     188       
================================================================================
TESTS END: 2016-09-20 19:35:58.492306
TESTS AVERAGE query time: 0.000601018400497
TESTS TOTAL test time: 0.131623029709
TESTS TOTAL queries: 219
================================================================================
```

pootle/apps/pootle_app/views/index/index.py
```diff
@@ -28,7 +28,9 @@ def view(request):
     lang = request.COOKIES.get(COOKIE_NAME, None)
 
     if lang is None:
-        supported = Language.live.cached_dict()
+        supported = Language.live.cached_dict(
```
phlax: this seems to be untested

unho: Yes, it is untested. But it was untested before too. This code is just touched to pass the param. I am not adding tests for this in this PR.

phlax: k

pootle/apps/pootle_language/models.py
```diff
@@ -31,24 +31,41 @@ class LiveLanguageManager(models.Manager):
...
+            disabled and enabled projects) or only live languages for enabled
+            projects.
         :return: an `OrderedDict`
         """
-        key = make_method_key(self, 'cached_dict', locale_code)
+        key_prefix = 'all_cached_dict' if show_all else 'cached_dict'
+        key = make_method_key(self, key_prefix, locale_code)
         languages = cache.get(key, None)
         if languages is None:
+            qs = self.get_all_queryset() if show_all else self.all()
```
ta2-1: @unho what is difference if we use `self.get_queryset()` instead of `self.all()`?
triple `all` looks not ideally.

unho: There is no difference. The reason why I used `self.all()` is because it is what was previously being used.

PR #5211, opened 2016-09-07
Add requires.io directive markers

dwaynebailey:
Using [directives](https://requires.io/features/#rq-directives) allows requires.io to limit its checking to a specific range.

We're using this to:

* Limit our Django to LTS 1.8
* Keep Elasticsearch on 1.x which supports the version we actually deploy against

Now requires.io is fully useful in that green means everything is fine.

https://requires.io/github/dwaynebailey/pootle/requirements/?branch=cleanup%2Frequireio-directives

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,6 +17,10 @@ Pootle
    :target: https://coveralls.io/github/translate/pootle?branch=master
    :alt: Test Coverage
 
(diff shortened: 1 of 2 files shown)


phlax:
can we add a brag badge also (it would help to see if it changes apart from anything else)

phlax:
lgtm

PR #5212, opened 2016-09-07
Have the checker retrieval in a single place

### LINKED ISSUE #5209: Custom POOTLE_QUALITY_CHECKER created without any configuration
The issue is that when I write a custom quality checker and set the class into `POOTLE_QUALITY_CHECKER` my checks get called with no language set. This makes writing language dependent checks impossible.

I have tracked this issue down a bit, but I am not a django coder so this is a bit hack and slash.

It appears to be that the quality checker is being created via
  `apps/pootle_misc/checks.py:get_checker`

`get_checker` has this code:
```python
    if settings.POOTLE_QUALITY_CHECKER:
        return import_func(settings.POOTLE_QUALITY_CHECKER)()
```

So you can see that the quality checker has no parameters specified, resulting in an empty configuration with no language code. 

`get_checker` comes from the `apps/pootle_store/models.py` path.  There is another nicer code path that results in a `TeeChecker` and a much better set of configuration happening in `apps/pootle_translationproject/models.py` for for some reason my install is not using that code path at all.

unho:
Now all checker retrieval is in TranslationProject.checker. This
also allows to pass the target language to the checkers.

Fixes #5209.

### DIFF
--- a/pootle/apps/pootle_misc/checks.py
+++ b/pootle/apps/pootle_misc/checks.py
@@ -250,13 +250,6 @@ def clean_plurr_placeholder(string):
     return plurr_plural_suffix_regex.sub('', string)
 
 
-def get_checker(unit):
-    if settings.POOTLE_QUALITY_CHECKER:
-        return import_func(settings.POOTLE_QUALITY_CHECKER)()
-
-    return unit.store.translation_project.checker
-
-
 def get_category_id(code):
     return CATEGORY_IDS.get(code)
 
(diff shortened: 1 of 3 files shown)


phlax:
as discussed in channel im lgtm on this for now - altho would like to move the checkers to a signal/plugin

julen:
It looks good to me and removes some duplication, although I don't see how this addresses the mentioned issue or does anything special to allow passing target languages to checkers.

unho:
@julen In https://github.com/translate/pootle/blob/master/pootle/apps/pootle_translationproject/models.py#L226 the language is passed. In some of the duplication removed in this PR it is not.

PR #5213, opened 2016-09-07
Add tests for timezone utilities

julen:
This PR mostly adds tests and docstrings for timezone utilities. When doing so I needed to bring in the pytz library for sanity, which I'm adding in the base requirements also because Django strongly recommends it.

At the same there is a minor enhancement to the existing functions to allow passing timezones explicitly and hence override the default of using `django.utils.timezone.get_default_timezone` (I'm happy to move this to another PR if necessary).

### DIFF
--- a/pootle/core/utils/timezone.py
+++ b/pootle/core/utils/timezone.py
@@ -19,6 +19,10 @@ if settings.USE_TZ:
 
 
 def make_aware(value):
+    """Makes a `datetime` timezone-aware.
+
+    :param value: `datetime` object to make timezone-aware.
+    """
     if getattr(settings, 'USE_TZ', False) and timezone.is_naive(value):
         tz = timezone.get_default_timezone()
         value = timezone.make_aware(value, tz)
@@ -27,6 +31,10 @@ def make_aware(value):
 
 
 def make_naive(value):
+    """Makes a `datetime` naive, i.e. not aware of timezones.
+
+    :param value: `datetime` object to make timezone-aware.
+    """
     if getattr(settings, 'USE_TZ', False) and timezone.is_aware(value):
         tz = timezone.get_default_timezone()
         value = timezone.make_naive(value, tz)
@@ -35,4 +43,8 @@ def make_naive(value):
 
 
 def aware_datetime(*args, **kwargs):
+    """Creates a `datetime` object and makes it timezone-aware.
+
+    :param args: arguments passed to `datetime` constructor.
+    """
     return make_aware(datetime.datetime(*args, **kwargs))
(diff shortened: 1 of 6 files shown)


pootle/core/utils/timezone.py
```diff
@@ -46,5 +52,8 @@ def aware_datetime(*args, **kwargs):
     """Creates a `datetime` object and makes it timezone-aware.
 
     :param args: arguments passed to `datetime` constructor.
+    :param tz: timezone in which the `datetime` should be constructed. Note that
+        this bypasses passing `tzinfo` to the `datetime` constructor, as it is
+        known not to play well with DST (unless only UTC is used).
     """
-    return make_aware(datetime.datetime(*args, **kwargs))
+    return make_aware(datetime.datetime(*args), tz=kwargs.pop('tz', None))
```
unho: We are loosing the ability to pass the `kwargs`. Please pop out the tz first, and keep passing the kwargs to datetime.

tests/core/utils/timezone.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_make_naive_use_tz_false(settings):
+    """Tests datetimes are left intact if `USE_TZ` is not in effect."""
+    settings.USE_TZ = False
+
+    datetime_object = datetime(2016, 1, 2, 21, 52, 25, tzinfo=pytz.utc)
+    assert timezone.is_aware(datetime_object)
+    naive_datetime = make_naive(datetime_object)
+    assert timezone.is_aware(naive_datetime)
```
unho: How do you check datetime is left intact?

pootle/core/utils/timezone.py
```diff
@@ -18,21 +18,42 @@
...
-        tz = timezone.get_default_timezone()
-        value = timezone.make_naive(value, tz)
+        use_tz = tz if tz is not None else timezone.get_default_timezone()
+        value = timezone.make_naive(value, timezone=use_tz)
 
     return value
 
 
 def aware_datetime(*args, **kwargs):
-    return make_aware(datetime.datetime(*args, **kwargs))
```
unho: We are loosing the ability to pass the `kwargs`. Please pop out the tz first, and keep passing the kwargs to datetime.

tests/core/utils/timezone.py
julen: `make_naive()` only deals with `tzinfo`: it removes it in case there was any. It does no timezone localization, there is no need to check for that.

unho: Well, the tests are for preventing this from changing in the future. I would add some check here even if we are sure about it.

pootle/core/utils/timezone.py
unho: This is now addressed.

tests/core/utils/timezone.py
julen: The underlying `make_naive` is Django's own, it has a very specific function and it's well tested. We don't need to test for that and note this function won't change lightly: it would be a backwards incompatible change affecting not a few Django sites.

unho:
lgtm

PR #5223, opened 2016-09-08
linting tools go bikeshedding

dwaynebailey:
pep8 -> pycodestyle
pep257 -> pydocstyle

They have their reasons, but I feel a red for the bikeshed would have been a better choice


unho:
becamse? Please rename commit.

PR #5227, opened 2016-09-09
Add PTL.unitApiRoot var

phlax:
this allows plugins - eg virtualfolders - to set the /xhr/units url dynamically and therefore use their own views in the editor


pootle/templates/layout.html
```diff
@@ -31,6 +31,7 @@
       MARKUP_FILTER: '{{ settings.POOTLE_MARKUP_FILTER }}',
       SOCIAL_AUTH_PROVIDERS: {{ SOCIAL_AUTH_PROVIDERS|to_js }},
     };
+    PTL.unitApiRoot = '{% if unit_api_root %}{{ unit_api_root }}{% else %}/xhr/units/{% endif %}'
```
unho: This should only be initialized for the editor. So wrong template. Also I think it is better to use some initialization code to pass it.

PR #5233, opened 2016-09-10
Move vfolder location and filter_rules to projects/languages/glob matching

phlax:
this PR will

- add projects/languages m2m tables with vfolder
- glob matching for filter_rules
- migrate existing location/filter_rules fields to projects/locations and update the filter_rules if required

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1624,13 +1624,12 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
         """
         pootle_paths = super(Store, self).all_pootle_paths()
         if 'virtualfolder' in settings.INSTALLED_APPS:
-            vftis = self.parent_vf_treeitems.values_list(
-                "vfolder__location", "pootle_path")
-            for location, pootle_path in vftis:
+            vftis = self.parent_vf_treeitems
+            for pootle_path in vftis.values_list("pootle_path", flat=True):
                 pootle_paths.extend(
                     [p for p
                      in get_all_pootle_paths(pootle_path)
-                     if p.count('/') > location.count('/')])
+                     if p.count('/') > 3])
         return pootle_paths
 
     # # # /TreeItem
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1268,8 +1268,11 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
             return DEFAULT_PRIORITY
         return priority
 
-    def set_priority(self):
-        priority = self.calculate_priority()
+    def set_priority(self, priority=None):
+        priority = (
+            self.calculate_priority()
+            if priority is None
+            else priority)
 
         if priority != self.priority:
             Store.objects.filter(pk=self.pk).update(priority=priority)
--- a/pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
+++ b/pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.url_helpers import split_pootle_path
+
+
+def parse_vfolder_rules(vf):
+    languages = set()
+    projects = set()
+    new_rules = set()
+
+    full_rules = [vf.location.strip() + rule.strip()
+                  for rule in vf.filter_rules.split(",")]
+
+    for full_rule in full_rules:
+        lang_code, proj_code, dir_path, filename = split_pootle_path(full_rule)
+        if filename:
+            new_rules.add(dir_path + filename)
+        else:
+            new_rules.add(dir_path + "*")
+        languages.add(lang_code)
+        projects.add(proj_code)
+
+    if "{LANG}" in languages:
+        languages = set()
+
+    if "{PROJ}" in projects:
+        projects = set()
+
+    new_rules=",".join(new_rules)
+
+    return languages, projects, new_rules
+
+
+def set_projects_and_languages(app, schema):
+    VirtualFolder = app.get_model("virtualfolder.VirtualFolder")
+    Project = app.get_model("pootle_project.Project")
+    Language = app.get_model("pootle_language.Language")
+    for vf in VirtualFolder.objects.all():
+        languages, projects, new_rules = parse_vfolder_rules(vf)
+        if projects:
+            vf.projects.add(*Project.objects.filter(code__in=projects))
+        if languages:
+            vf.languages.add(*Language.objects.filter(code__in=languages))
+        vf.filter_rules = new_rules
+        vf.all_projects = not projects
+        vf.all_languages = not languages
+        vf.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('virtualfolder', '0012_add_all_proj_lang_flags'),
+    ]
+
+    operations = [
+        migrations.RunPython(set_projects_and_languages)
+    ]
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -81,44 +81,6 @@ class VirtualFolder(models.Model):
     def path_matcher(self):
         return path_matcher.get(self.__class__)(self)
 
-    @property
-    def all_locations(self):
-        """Return a list with all the locations this virtual folder applies.
-
-        If the virtual folder location has no {LANG} nor {PROJ} placeholders
-        then the list only contains its location. If any of the placeholders is
-        present, then they get expanded to match all the existing languages and
-        projects.
-        """
-        if "{LANG}/{PROJ}" in self.location:
-            locations = []
-            for lang in Language.objects.all():
-                temp = self.location.replace("{LANG}", lang.code)
-                for proj in Project.objects.all():
-                    locations.append(temp.replace("{PROJ}", proj.code))
-            return locations
-        elif "{LANG}" in self.location:
-            try:
-                project = Project.objects.get(code=self.location.split("/")[2])
-                languages = project.languages.iterator()
-            except Exception:
-                languages = Language.objects.iterator()
-
-            return [self.location.replace("{LANG}", lang.code)
-                    for lang in languages]
-        elif "{PROJ}" in self.location:
-            try:
-                projects = Project.objects.filter(
-                    translationproject__language__code=self.location.split("/")[1]
-                ).iterator()
-            except Exception:
-                projects = Project.objects.iterator()
-
-            return [self.location.replace("{PROJ}", proj.code)
-                    for proj in projects]
-
-        return [self.location]
-
     def __unicode__(self):
         return self.name
 
@@ -140,70 +102,6 @@ class VirtualFolder(models.Model):
             raise ValidationError(u'Priority must be greater than zero.')
         if not self.filter_rules:
             raise ValidationError(u'Some filtering rule must be specified.')
-
-    def get_adjusted_location(self, pootle_path):
-        """Return the virtual folder location adjusted to the given path.
-
-        The virtual folder location might have placeholders, which affect the
-        actual filenames since those have to be concatenated to the virtual
-        folder location.
-        """
-        count = self.location.count("/")
-
-        if pootle_path.count("/") < count:
-            raise ValueError("%s is not applicable in %s" % (self,
-                                                             pootle_path))
-
-        pootle_path_parts = pootle_path.strip("/").split("/")
-        location_parts = self.location.strip("/").split("/")
-
-        try:
-            if (location_parts[0] != pootle_path_parts[0] and
-                location_parts[0] != "{LANG}"):
-                raise ValueError("%s is not applicable in %s" % (self,
-                                                                 pootle_path))
-
-            if (location_parts[1] != pootle_path_parts[1] and
-                location_parts[1] != "{PROJ}"):
-                raise ValueError("%s is not applicable in %s" % (self,
-                                                                 pootle_path))
-        except IndexError:
-            pass
-
-        return "/".join(pootle_path.split("/")[:count])
-
-    def get_adjusted_pootle_path(self, pootle_path):
-        """Adjust the given pootle path to this virtual folder.
-
-        The provided pootle path is converted to a path that includes the
-        virtual folder name in the right place.
-
-        For example a virtual folder named vfolder8, with a location
-        /{LANG}/firefox/browser/ in a path
-        /af/firefox/browser/chrome/overrides/ gets converted to
-        /af/firefox/browser/vfolder8/chrome/overrides/
-        """
-        count = self.location.count('/')
-
-        if pootle_path.count('/') < count:
-            # The provided pootle path is above the virtual folder location.
-            path_parts = pootle_path.rstrip('/').split('/')
-            pootle_path = '/'.join(path_parts +
-                                   self.location.split('/')[len(path_parts):])
-
-        if count < 3:
-            # If the virtual folder location is not long as a translation
-            # project pootle path then the returned adjusted location is too
-            # short, meaning that the returned translate URL will have the
-            # virtual folder name as the project or language code.
-            path_parts = pootle_path.split('/')
-            return '/'.join(path_parts[:3] + [self.name] + path_parts[3:])
-
-        # If the virtual folder location is as long as a TP pootle path and
-        # the provided pootle path isn't above the virtual folder location.
-        lead = self.get_adjusted_location(pootle_path)
-        trail = pootle_path.replace(lead, '').lstrip('/')
-        return '/'.join([lead, self.name, trail])
 
 
 class VirtualFolderTreeItemManager(models.Manager):
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -76,6 +76,14 @@ class VirtualFolder(models.Model):
         Store,
         db_index=True,
         related_name='vfolders')
+    projects = models.ManyToManyField(
+        Project,
+        db_index=True,
+        related_name='vfolders')
+    languages = models.ManyToManyField(
+        Language,
+        db_index=True,
+        related_name='vfolders')
 
     class Meta(object):
         unique_together = ('name', 'location')
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -68,10 +68,12 @@ class VirtualFolder(models.Model):
         Store,
         db_index=True,
         related_name='vfolders')
+    all_projects = models.BooleanField(default=False)
     projects = models.ManyToManyField(
         Project,
         db_index=True,
         related_name='vfolders')
+    all_languages = models.BooleanField(default=False)
     languages = models.ManyToManyField(
         Language,
         db_index=True,
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -194,19 +194,19 @@ class VirtualFolderTreeItem(models.Model, CachedTreeItem):
         return self.pootle_path
 
     def save(self, *args, **kwargs):
-        self.pootle_path = self.vfolder.get_adjusted_pootle_path(
-            self.directory.pootle_path
-        )
+        parts = split_pootle_path(self.directory.pootle_path)
+        path_parts = ["", parts[0], parts[1], self.vfolder.name]
+        if parts[2]:
+            path_parts.append(parts[2].strip("/"))
+        path_parts.append(parts[3])
+        self.pootle_path = "/".join(path_parts)
 
         # Trigger the creation of the whole parent tree up to the vfolder
-        # adjusted location.
-        if (self.directory.pootle_path.count('/') >
-                self.vfolder.location.count('/')):
-            parent = VirtualFolderTreeItem.objects.get_or_create(
+        # tp
+        if self.directory.pootle_path.count('/') > 3:
+            self.parent = VirtualFolderTreeItem.objects.get_or_create(
                 directory=self.directory.parent,
-                vfolder=self.vfolder,
-            )[0]
-            self.parent = parent
+                vfolder=self.vfolder)[0]
 
         super(VirtualFolderTreeItem, self).save(*args, **kwargs)
 
@@ -269,6 +269,6 @@ class VirtualFolderTreeItem(models.Model, CachedTreeItem):
         don't want to mess with regular CachedTreeItem stats.
         """
         return [p for p in get_all_pootle_paths(self.get_cachekey())
-                if p.count('/') > self.vfolder.location.count('/')]
+                if p.count('/') > 3]
 
     # # # /TreeItem
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -40,14 +40,6 @@ class VirtualFolder(models.Model):
         null=True,
         max_length=255)
 
-    # any changes to the `location` field may require updating the schema
-    # see migration 0003_case_sensitive_schema.py
-    location = models.CharField(
-        _('Location'),
-        blank=False,
-        max_length=255,
-        help_text=_('Root path where this virtual folder is applied.'),
-    )
     filter_rules = models.TextField(
         # Translators: This is a noun.
         _('Filter'),
@@ -84,9 +76,6 @@ class VirtualFolder(models.Model):
         Language,
         db_index=True,
         related_name='vfolders')
-
-    class Meta(object):
-        unique_together = ('name', 'location')
 
     @cached_property
     def path_matcher(self):
@@ -131,7 +120,7 @@ class VirtualFolder(models.Model):
         return [self.location]
 
     def __unicode__(self):
-        return ": ".join([self.name, self.location])
+        return self.name
 
     def save(self, *args, **kwargs):
         # Force validation of fields.
@@ -149,14 +138,6 @@ class VirtualFolder(models.Model):
         """Validate virtual folder fields."""
         if self.priority <= 0:
             raise ValidationError(u'Priority must be greater than zero.')
-
-        elif self.location == "/":
-            raise ValidationError(u'The "/" location is not allowed. Use '
-                                  u'"/{LANG}/{PROJ}/" instead.')
-        elif self.location.startswith("/projects/"):
-            raise ValidationError(u'Locations starting with "/projects/" are '
-                                  u'not allowed. Use "/{LANG}/" instead.')
-
         if not self.filter_rules:
             raise ValidationError(u'Some filtering rule must be specified.')
 
--- a/pootle/apps/virtualfolder/models.py
+++ b/pootle/apps/virtualfolder/models.py
@@ -9,6 +9,7 @@
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
 from django.db import models
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 
 from pootle.core.markup import MarkupField, get_markup_filter_display_name
@@ -20,6 +21,8 @@ from pootle_app.models import Directory
 from pootle_language.models import Language
 from pootle_project.models import Project
 from pootle_store.models import Store
+
+from .delegate import path_matcher
 
 
 class VirtualFolder(models.Model):
@@ -76,6 +79,10 @@ class VirtualFolder(models.Model):
 
     class Meta(object):
         unique_together = ('name', 'location')
+
+    @cached_property
+    def path_matcher(self):
+        return path_matcher.get(self.__class__)(self)
 
     @property
     def all_locations(self):
--- a/pootle/apps/virtualfolder/receivers.py
+++ b/pootle/apps/virtualfolder/receivers.py
@@ -6,98 +6,31 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-from django.db.models.signals import post_save
+from django.db.models.signals import post_save, pre_delete
 from django.dispatch import receiver
 
 from pootle_store.models import Store
 
-from .models import VirtualFolder, VirtualFolderTreeItem
-
-
-def update_vfolder_tree(vf, store):
-    """For a given VirtualFolder and Store update the VirtualFolderTreeItem
-    """
-    # Create missing VirtualFolderTreeItem tree structure for affected Stores
-    # after adding or unobsoleting Units.
-    vfolder_treeitem, created = (
-        VirtualFolderTreeItem.objects.get_or_create(
-            directory=store.parent, vfolder=vf))
-
-    if not created:
-        # The VirtualFolderTreeItem already existed, so
-        # calculate again the stats up to the root.
-        vfolder_treeitem.update_all_cache()
-
-
-def add_store_to_vfolders(store):
-    """For a given Unit check for membership of any VirtualFolders
-    """
-    pootle_path = store.pootle_path
-
-    for vf in VirtualFolder.objects.iterator():
-        store_added = False
-        for location in vf.all_locations:
-            if not pootle_path.startswith(location):
-                continue
-
-            for filename in vf.filter_rules.split(","):
-                if pootle_path == "".join([location, filename]):
-                    vf.stores.add(store)
-                    store_added = True
-                    break
-
-            if store_added:
-                break
-
-        if store_added:
-            update_vfolder_tree(vf, store)
-
-
-@receiver(post_save, sender=VirtualFolder)
-def vfolder_save_handler(sender, instance, created, **kwargs):
-    """Remove Units from VirtualFolder when vfolder changes
-
-    - Check the original VirtualFolder object's locations
-    - Check the new VirtualFolders object's locations
-    - Remove any units from the VirtualFolder that are only in original's
-      locations and reset Unit priority
-    - Update VirtualFolderTree for any Stores that may have been affected
-    """
-    locations = set()
-    for location in instance.all_locations:
-        for filename in instance.filter_rules.split(","):
-            locations.add("".join([location, filename]))
-
-    stores_we_want = Store.objects.none()
-    for location in locations:
-        stores_we_want = stores_we_want | Store.objects.filter(
-            pootle_path__startswith=location)
-    to_remove = list(
-        instance.stores.exclude(
-            pk__in=stores_we_want.values_list("pk", flat=True)))
-    to_add = list(
-        stores_we_want.exclude(
-            pk__in=instance.stores.values_list("pk", flat=True)))
-    instance.stores.remove(*to_remove)
-    instance.stores.add(*to_add)
-    for store in to_add:
-        update_vfolder_tree(instance, store)
-    for store in to_remove:
-        store.set_priority()
-    for store in stores_we_want:
-        store.set_priority()
+from .delegate import vfolder_finder
+from .models import VirtualFolder
 
 
 @receiver(post_save, sender=Store)
-def vfolder_store_postsave_handler(**kwargs):
-    """Match VirtualFolders to Unit and update Unit.priority
+def handle_store_save(sender, instance, created, **kwargs):
+    if not created:
+        return
+    vfolder_finder.get(
+        instance.__class__)(instance).add_to_vfolders()
 
-    - If unit was newly created, then check vfolders for membership
-    - Update Unit priority from vfolder membership on Unit.save or create
 
-    """
-    instance = kwargs["instance"]
-    created = kwargs.get("created", False)
-    if created:
-        add_store_to_vfolders(instance)
-    instance.set_priority()
+@receiver(post_save, sender=VirtualFolder)
+def handle_vfolder_save(sender, instance, created, **kwargs):
+    instance.path_matcher.update_stores()
+
+
+@receiver(pre_delete, sender=VirtualFolder)
+def handle_vfolder_delete(sender, instance, **kwargs):
+    for store in instance.stores.all():
+        instance.stores.remove(store)
+        if store.priority == instance.priority:
+            store.set_priority()
--- a/pootle/apps/virtualfolder/utils.py
+++ b/pootle/apps/virtualfolder/utils.py
@@ -1,0 +1,219 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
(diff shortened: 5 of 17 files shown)


dwaynebailey:
@phlax from what I read here {ANY} means any at this point in time, as opposed to any current and future languages. With the former you would need to update languages when each new Lang is added to Pootle instead of it just being covered.  Or have I read the code wrong?

phlax:
@dwaynebailey - thats correct yep

we could make it so that not setting any projects/languages == all 

i havent initally at least because i reckon its going to be more efficient getting related objects through the m2m tables

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
```diff
@@ -0,0 +1,81 @@
...
+            if rule_proj:
+                projects.add(rule_proj)
+            vf_rules.add(new_rule)
+        if any_project and any_lang:
+            langs_to_add = Language.objects.all()
+            projs_to_add = Project.objects.all()
+        elif any_project:
+            projs_to_add = Project.objects.filter(
+                translation_project__language__code__in=languages)
+            langs_to_add = Language.objects.filter(code__in=languages)
```
unho: In this case `languages` only has a single language, so not need to use `__in`.

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
```diff
@@ -0,0 +1,81 @@
...
+        elif any_lang:
+            langs_to_add = Language.objects.filter(
+                translation_project__project__code__in=projects)
+            projs_to_add = Project.objects.filter(code__in=projects)
```
unho: The same as the previous comment, but for `projects`.

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            langs_to_add = Project.objects.filter(code__in=projects)
+            projs_to_add = Language.objects.filter(code__in=languages)
```
unho: In this case both `projects` and `languages` have a single element each.

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
phlax: not necessarily - langs can be restricted by the filter_rules - no?

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
phlax: ditto

pootle/apps/virtualfolder/migrations/0012_set_projects_languages.py
phlax: and again 8/

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
-            directory=store.parent, vfolder=vf))
-
+    created = False
+    try:
+        vfolder_treeitem = (
+            VirtualFolderTreeItem.objects.get(
+                directory=store.parent, vfolder=vf))
+    except VirtualFolderTreeItem.DoesNotExist:
+        parts = split_pootle_path(store.parent.pootle_path)
+        path_parts = ["", parts[0], parts[1], vf.name]
```
unho: You are assuming that vfolder name is always just after language/project, but the location could have been deeper.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
-    for store in stores_we_want:
+        if store.priority != instance.priority:
+            store.set_priority()
+    stores = Store.objects.filter(
+        pk__in=deleted).exclude(priority=instance.priority)
+    for store in stores:
         store.set_priority()
 
 
-@receiver(post_save, sender=Store)
```
unho: Why is this removed?

unho: Just saw it was moved up. This just increased the diff and mades more difficult to review.

tests/models/virtualfolder.py
```diff
@@ -152,17 +104,18 @@ def test_vfolder_membership(tp0, store0):
     vf1.save()
     assert store not in vf1.stores.all()
     vf1.filter_rules = ",".join([store_name, vf1.filter_rules])
+    vf1._FOO = 1
```
unho: What is this?

tests/models/virtualfolder.py
```diff
@@ -152,17 +104,18 @@ def test_vfolder_membership(tp0, store0):
...
 @pytest.mark.django_db
-def test_vfolder_store_priorities():
+def test_vfolder_store_priorities(project0):
 
     # remove the default vfolders and update units to reset priorities
-    VirtualFolder.objects.all().delete()
-    [store.save() for store in Store.objects.all()]
-
+    for vf in VirtualFolder.objects.all():
+        vf.delete()
```
unho: Does not `VirtualFolder.objects.all().delete()` trigger the right signals?

pootle/apps/virtualfolder/models.py
```diff
@@ -194,19 +194,19 @@ def __unicode__(self):
...
+        path_parts.append(parts[3])
+        self.pootle_path = "/".join(path_parts)
 
         # Trigger the creation of the whole parent tree up to the vfolder
-        # adjusted location.
-        if (self.directory.pootle_path.count('/') >
-                self.vfolder.location.count('/')):
-            parent = VirtualFolderTreeItem.objects.get_or_create(
+        # tp
+        if self.directory.pootle_path.count('/') > 3:
```
unho: Are we assuming that a vfolder directory cannot be upper than a TP directory? What if the vfolder is for all languages and all projects?

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
```diff
@@ -0,0 +1,86 @@
...
+    any_project = proj_code == "{PROJ}"
+    any_lang = lang_code == "{LANG}"
+    languages = set()
+    projects = set()
+    new_rules = set()
+    old_rules = [
+        r.strip()
+        for r
+        in vf.filter_rules.split(",")
+        if r.strip()]
```
unho: I don't think there is any need for `strip()`.

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from pootle.core.url_helpers import split_pootle_path
+
+
+def parse_filter_rule(dir_path, filter_rule, lang_from_filters, proj_from_filters):
+    filter_rule = "/%s/" % filter_rule.strip("/")
```
unho: Are we appending `/` to filepaths?

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
```diff
@@ -0,0 +1,86 @@
...
+    rule_lang = rule_proj = None
+    new_rule = filter_rule
+    if lang_from_filters:
+        rule_lang, rule_proj = split_pootle_path(filter_rule)[:2]
+        new_rule = filter_rule[len(rule_proj) + len(rule_lang) + 2:]
+    elif proj_from_filters:
+        rule_proj = split_pootle_path("/X/%s/" % filter_rule)[1]
```
unho: If the original rule is for example `firefox/dir/file.po`, then  `filter_rule` would be `/firefox/dir/file.po/`, and the interpolation in here will yield `/X//firefox/dir/file.po/`, therefore `rule_proj` will be `""` (empty string).

tests/models/virtualfolder.py
phlax: why, surely thats obvious - its a `_FOO` !

tests/models/virtualfolder.py
phlax: i have a feeling not - lets check

pootle/apps/virtualfolder/models.py
phlax: not sure - but this wont matter once there are no vftis - ill check this bit tho

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
phlax: hmm -  think its a good idea tbh - i think we should switch this field to a `JSONField` but this just makes `foo, bar`  work the same as `foo,bar`

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
phlax: it normalizes them so that filter rules can start with/out a slash

pootle/apps/virtualfolder/migrations/0013_set_projects_languages.py
```diff
@@ -0,0 +1,86 @@
...
+        new_rule = filter_rule[len(rule_proj) + 1:]
+    elif dir_path:
+        new_rule = filter_rule[len(dir_path):]
```
unho: Assuming original location was `/af/firefox/dir/` then `dir_path` will be `dir`. If we assume some original rule `subdir/file.po` then `filter_rule` will be `/subdir/file.po/` thus the new rule will be `bdir/file.po/`.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
+    # limit to vfolders that could be possibly related
+    vfolders = (
+        project.vfolders.filter(all_languages=True)
+        | language.vfolders.filter(all_projects=True)
+        | project.vfolders.filter(languages=language)
+        | VirtualFolder.objects.filter(
+            all_languages=True, all_projects=True))
+    for vf in vfolders:
+        added = vf.path_matcher.add_store_if_matching(instance)
+        if added:
```
unho: This is misleading. Please rename to `store_matches` or something that truly reflects the meaning.

phlax: it doesnt cjust match - it adds

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
+            update_vfolder_tree(vf, instance)
+            instance.set_priority()
```
unho: I think it is better to call this once after traversing all vfolders.

phlax: good spot!

phlax: on that note @unho we could probs do with some code to set priority for multiple stores at once

unho: In fact we can keep the highest priority in a variable for all the vfolders it was added to, and then directly set the priority and save the store.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
-                    break
-
-            if store_added:
-                break
-
-        if store_added:
-            update_vfolder_tree(vf, store)
+@receiver(post_save, sender=Store)
+def handle_store_save(sender, instance, created, **kwargs):
+    if not created:
```
unho: This should be skipped if `virtualfolder` app is not enabled.

pootle/apps/virtualfolder/receivers.py
```diff
@@ -19,85 +20,67 @@ def update_vfolder_tree(vf, store):
...
+def handle_vfolder_save(sender, instance, created, **kwargs):
+    added, deleted = instance.path_matcher.update_matching_stores()
+    for store in added:
         update_vfolder_tree(instance, store)
-    for store in to_remove:
-        store.set_priority()
-    for store in stores_we_want:
+        if store.priority != instance.priority:
+            store.set_priority()
+    stores = Store.objects.filter(
```
unho: Nit: `deleted_stores`

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        existing_stores = self.vf.stores.all()
+        matching_stores = self.find_matching_stores()
+        to_add = []
+        to_delete = []
+        for store in matching_stores:
+            if store not in existing_stores:
+                to_add.append(store)
+        for store in existing_stores:
+            if store not in matching_stores:
+                to_delete.append(store.id)
```
unho: Why the `id` and not the Store object?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+                translation_project__project__in=self.vf.projects.all())
+        qs = qs.filter(pootle_path__regex="^%s" % self.path)
+        return qs
+
+    @property
+    def path(self):
+        # TODO: filter to projects/languages if not all_*
+        regex = "[^/]*"
+        return (
+            "/%s/%s"
```
unho: No trailing slash?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+                translation_project__language__in=self.vf.languages.all())
+        if not self.vf.all_projects:
+            qs = qs.filter(
+                translation_project__project__in=self.vf.projects.all())
+        qs = qs.filter(pootle_path__regex="^%s" % self.path)
+        return qs
+
+    @property
+    def path(self):
+        # TODO: filter to projects/languages if not all_*
```
unho: This needs to be implemented.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        regex = "[^/]*"
+        return (
+            "/%s/%s"
+            % (regex, regex))
+
+    def get_rule_path(self, rule):
+        return (
+            "%s/%s"
+            % (self.path, rule.lstrip("/")))
```
unho: Rules should not start with `/`. Also this fits in a single line and using `join` makes it cleaner.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        if to_add:
+            self.vf.stores.add(*to_add)
+        if to_delete:
+            self.vf.stores.remove(
+                *self.vf.stores.filter(id__in=to_delete))
+        return to_add, to_delete
+
+    def add_store_if_matching(self, store):
+        pootle_dir_path = "".join(
+            ("/", ) + split_pootle_path(store.pootle_path)[2:])
```
unho: Is not this what `to_tp_relative_path` is supposed to return?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+# AUTHORS file for copyright and authorship information.
+
+import os
+from fnmatch import fnmatch
+
+from pootle.core.url_helpers import split_pootle_path
+from pootle_store.models import Store
+
+
+class VirtualFolderPathMatcher(object):
```
unho: I don't recall seeing any test for this, despite it is a critical component of vfolder support.

pootle/apps/virtualfolder/utils.py
phlax: im not sure if it does - what i meant here is a regex like `(af)(es)(ru)` which could get a bit unwieldy if its all projects - not sure what is more efficient and in what situation - for now tho this has been pretty performant in the limited testing ive done

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        if to_delete:
+            self.vf.stores.remove(
+                *self.vf.stores.filter(id__in=to_delete))
+        return to_add, to_delete
+
+    def add_store_if_matching(self, store):
+        pootle_dir_path = "".join(
+            ("/", ) + split_pootle_path(store.pootle_path)[2:])
+        no_add = (
+            not self.path_matches(pootle_dir_path)
```
unho: If all rules are directory paths, and this store is within one of those directories, then the store won't be added, am I wrong?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+            "%s/%s"
+            % (self.path, rule.lstrip("/")))
+
+    def filter_by_rule(self, qs, rule):
+        rule_path = self.get_rule_path(rule)
+        is_file = "." in os.path.basename(rule)
+        return (
+            qs.filter(pootle_path__regex="^%s$" % rule_path)
+            if is_file
+            else qs.filter(pootle_path__regex="^%s" % rule_path))
```
unho: IMHO this is more difficult to read than just:

```python

        if is_file:
            return qs.filter(pootle_path__regex="^%s$" % rule_path)
        return qs.filter(pootle_path__regex="^%s" % rule_path))
```

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def filter_from_rules(self, qs):
+        if not self.filter_rules:
```
unho: This can't happen.

pootle/apps/virtualfolder/utils.py
unho: Or
```python
        regex = "^%s" % self.get_rule_path(rule)
        if "." in os.path.basename(rule):  # Is file.
            regex += "$"

        return qs.filter(pootle_path__regex=regex)
```

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+            if fnmatch(rule, path):
+                return True
+        return False
+
+    def dir_path_matches(self, path):
+        """Returns bool of whether path is valid for this VF.
+        or as a parent of any valid paths
+        """
+        dir_path = "%s%s" % (self.path, path)
+        if self.use_regex:
```
unho: Where is `use_regex` defined?

pootle/apps/virtualfolder/utils.py
phlax: yep, i was thining more like the interpolated version of this

phlax: also im wondering if this should be in  the path property

pootle/apps/virtualfolder/utils.py
phlax: no - it shouldnt - ill remove

pootle/apps/virtualfolder/utils.py
phlax: hmm - i removed it - because it seemed unused - im gonna check this method too

pootle/apps/virtualfolder/receivers.py
phlax: it wouldnt be triggered 8)

pootle/apps/virtualfolder/utils.py
unho: If we have a vfolder that has been set for a specific project, then this returns a regex to match all projects, effectively matching stores in any project.

pootle/apps/virtualfolder/utils.py
phlax: hmm - we might need to convert the path rules to `/path/*` rules

pootle/apps/virtualfolder/utils.py
phlax: dev style?

pootle/apps/virtualfolder/utils.py
phlax: i think you are right - i have some very minimal tests somewhere - but i think it needs unit tests - ill add

pootle/apps/virtualfolder/utils.py
unho: Lets focus on the "Rules should not start with `/`". This means that `lstrip` is unnecessary.

pootle/apps/virtualfolder/utils.py
phlax: good point

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @property
+    def path(self):
+        # TODO: filter to projects/languages if not all_*
+        regex = "[^/]*"
+        return (
+            "/%s/%s"
+            % (regex, regex))
+
+    def get_rule_path(self, rule):
```
unho: It is not a path, it is a regex.

pootle/apps/virtualfolder/utils.py
phlax: actually no - if its set for specific projects - they are already filtered

pootle/apps/virtualfolder/utils.py
unho: Actually this method is unused AFAICT.

pootle/apps/virtualfolder/utils.py
phlax: somewhere in between - a glob

pootle/apps/virtualfolder/utils.py
phlax: the lstrip was kinda defensive coding - i wasnt clear what the rules rules were

pootle/apps/virtualfolder/utils.py
unho: Then it is useless to keep this property? I wonder why you didn't chose to put full paths in the filters. I thought that was the plan, even if it was regex.

pootle/apps/virtualfolder/utils.py
unho: It is concatenating a regex and a string, so the result is a regex to me.

phlax: hmm - im wondering about that - i think i implemented this slightly differently originally - ill check this further

phlax: ...but the string is still a glob 8/

pootle/apps/virtualfolder/receivers.py
unho: Right, sorry.

pootle/apps/virtualfolder/utils.py
phlax: no - the project/lang part are dealt with by assoc - and all filter_rules are tp-relative

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,233 @@
...
+    def remove_stores(self, stores):
+        self.vf.stores.remove(*stores)
+
+    def should_add_store(self, store):
+        if self.store_matches(store) and not self.store_associated(store):
+            return True
+        return False
+
+    def store_associated(self, store):
+        return self.vf.stores.through.objects.filter(store_id=store.id).exists()
```
phlax: i checked this and this is the difference in sql

with this:
```
'SELECT (1) AS "a" FROM "virtualfolder_virtualfolder_stores" WHERE ("virtualfolder_virtualfolder_stores"."store_id" = %s AND "virtualfolder_virtualfolder_stores"."virtualfolder_id" = %s) LIMIT 1' - PARAMS = (31, 6)
```
and using `self.vf.stores.filter(store_id=store_id).exists()`

```
u'SELECT (1) AS "a" FROM "pootle_store_store" INNER JOIN "virtualfolder_virtualfolder_stores" ON ( "pootle_store_store"."id" = "virtualfolder_virtualfolder_stores"."store_id" ) WHERE ("virtualfolder_virtualfolder_stores"."virtualfolder_id" = %s AND "pootle_store_store"."id" = %s) LIMIT 1' - PARAMS = (6, 31)

```

phlax: ...although actually the algo is a little wrong there - correcting now...

pootle/apps/virtualfolder/models.py
phlax: @unho i debugged this - afaict vftis are created all the way to the root - if i set to any less it tries to create parents for root nodes and fails

tests/models/virtualfolder.py
phlax: seems to work - updating

pootle/apps/pootle_store/models.py
```diff
@@ -1624,13 +1624,12 @@ def all_pootle_paths(self):
...
-            vftis = self.parent_vf_treeitems.values_list(
-                "vfolder__location", "pootle_path")
-            for location, pootle_path in vftis:
+            vftis = self.parent_vf_treeitems
+            for pootle_path in vftis.values_list("pootle_path", flat=True):
                 pootle_paths.extend(
                     [p for p
                      in get_all_pootle_paths(pootle_path)
-                     if p.count('/') > location.count('/')])
+                     if p.count('/') > 3])
```
unho: This method probably should rely on vfti's `all_pootle_paths` method:

```python
        pootle_paths = super(Store, self).all_pootle_paths()

        if 'virtualfolder' in settings.INSTALLED_APPS:
            for vfti in self.parent_vf_treeitems:
                pootle_paths.extend(vfti.all_pootle_paths())

        return pootle_paths
```

phlax: @unho is going to be removed in a couple of PR anyway

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+from pootle_store.models import Store
+
+from .models import VirtualFolder, VirtualFolderTreeItem
+
+
+def update_vfolder_tree(vf, store):
+    """For a given VirtualFolder and Store update the VirtualFolderTreeItem
+    """
+    # Create missing VirtualFolderTreeItem tree structure for affected Stores
+    # after adding or unobsoleting Units.
```
unho: Typo: resurrecting.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+    created = False
```
unho: What about using `try..except..else` instead?

pootle/apps/virtualfolder/utils.py
phlax: this is just copied from the original https://github.com/translate/pootle/pull/5233/files#diff-362cf55e6d6252afb9292eeae52046d5L21

and again will be removed very soon

pootle/apps/virtualfolder/utils.py
phlax: its going to be removed

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+        vfolder_treeitem = (
+            VirtualFolderTreeItem.objects.select_related("directory").get(
+                directory=store.parent, vfolder=vf))
+    except VirtualFolderTreeItem.DoesNotExist:
+        parts = split_pootle_path(store.parent.pootle_path)
+        path_parts = ["", parts[0], parts[1], vf.name]
+        if parts[2]:
+            path_parts.append(parts[2].strip("/"))
+        path_parts.append(parts[3])
+        pootle_path = "/".join(path_parts)
```
unho: It is not the first time I see this code. If it is going to stay after all vfolder-related PRs I would say refactor out.

phlax: its going to be removed

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+        """Store qs containing all stores that match
+        project, language, and rules for this vfolder
+        """
+        return self.filter_from_rules(self.store_qs)
+
+    @property
+    def rules(self):
+        """Glob matching rules"""
+        return (
+            "%s" % r.strip()
```
unho: This `strip()` thing should happen on vfolder validation, not here every time we try to match anything.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+            for r
+            in self.vf.filter_rules.split(","))
+
+    @property
+    def store_manager(self):
```
unho: I see this property as completely unnecessary.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+        return False
+
+    def store_associated(self, store):
+        return self.vf.stores.through.objects.filter(
+            store_id=store.id,
+            virtualfolder_id=self.vf.id).exists()
+
+    def store_matches(self, store):
+        return self.path_matches(
+            "".join(split_pootle_path(store.pootle_path)[2:]))
```
unho: Use `to_tp_relative_path`?

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+                return True
+        return False
+
+    def remove_stores(self, stores):
+        self.vf.stores.remove(*stores)
+
+    def should_add_store(self, store):
+        if self.store_matches(store) and not self.store_associated(store):
+            return True
+        return False
```
unho:
```python
        return self.store_matches(store) and not self.store_associated(store)
```

pootle/apps/virtualfolder/utils.py
phlax: im going to move these to json and we can add validation then - for now i think a comma-separated string that is not ws-tolerant in this way is a bad idea

pootle/apps/virtualfolder/utils.py
phlax: unless you wanted to subclass it

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def store_associated(self, store):
+        return self.vf.stores.through.objects.filter(
+            store_id=store.id,
+            virtualfolder_id=self.vf.id).exists()
```
unho: If `self.vf.stores.through` is just the same as `Store.objects` I would use the latter instead.

pootle/apps/virtualfolder/utils.py
```diff
@@ -0,0 +1,220 @@
...
+            "".join(split_pootle_path(store.pootle_path)[2:]))
+
+    def update_stores(self):
+        """Add and delete Store associations as necessary, and set the
+        priority for any affected Stores
+        """
+        added, removed = self.add_and_remove_stores()
+        for store in added:
+            update_vfolder_tree(self.vf, store)
+            if store.priority != self.vf.priority:
```
unho: Only do if store priority is lower than vf priority!

pootle/apps/virtualfolder/utils.py
phlax: its not - see the debugging info i posted you

pootle/apps/virtualfolder/utils.py
phlax: yep - will do

pootle/apps/virtualfolder/utils.py
unho: Ok

pootle/apps/virtualfolder/utils.py
phlax: yep - ill update

pootle/apps/virtualfolder/utils.py
phlax: this is updated with store.path

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+
+@pytest.mark.pootle_vfolders
+@pytest.mark.django_db
+def test_vfolder_path_matcher_languages(language0, language1):
+    vfolder = VirtualFolder.objects.create(
+        name="avfolder",
+        filter_rules="FOO,BAR")
+    vfolder.languages.add(language0)
+    vfolder.languages.add(language1)
+    # no projects set so no stores
```
unho: This is not what I would expect. It should try matching stores for all TPs in the passed languages

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+    assert list(vfolder.path_matcher.existing_stores) == []
+    assert (
+        sorted(vfolder.path_matcher.languages)
+        == [language0.pk, language1.pk])
+    vfolder.all_languages = True
+    vfolder.save()
+    assert vfolder.path_matcher.languages is None
+    assert list(vfolder.path_matcher.projects) == []
```
unho: This should match stores in all TPs. But without TPs and stores it can't retrieve anything.

tests/vfolders/path_matcher.py
unho: Discussed in chat. It is ok if not `all_projects` set.

tests/vfolders/path_matcher.py
phlax: i think the answer is the same as above

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+    assert (
+        list(vfolder.path_matcher.existing_stores.order_by("pk"))
+        == list(Store.objects.order_by("pk")))
+    vfolder.all_languages = False
+    vfolder.save()
+    assert (
+        list(vfolder.path_matcher.existing_stores.order_by("pk"))
+        == [])
+    vfolder.all_languages = True
+    vfolder.all_languages = False
```
unho: `all_projects`?

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+    vfolder.save()
+    assert vfolder.path_matcher.projects is None
+    assert list(vfolder.path_matcher.languages) == []
+    # no languages set so no stores
+    assert list(vfolder.path_matcher.existing_stores) == []
+    vfolder.all_projects = False
+    vfolder.save()
+    # didnt forget lang preferences
+    vfolder.all_projects = False
+    vfolder.save()
```
unho: Repeated code.

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+    assert vfolder.path_matcher.languages is None
+    assert list(vfolder.path_matcher.projects) == []
+    # no projects set so no stores
+    assert list(vfolder.path_matcher.existing_stores) == []
+    vfolder.all_languages = False
+    vfolder.save()
+    # didnt forget lang preferences
+    vfolder.all_languages = False
+    vfolder.save()
```
unho: Repeated code.

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+@pytest.fixture(params=VF_RULE_TESTS)
+def vf_rules(request):
+    return request.param
+
+
+@pytest.mark.pootle_vfolders
+@pytest.mark.django_db
+def test_vfolder_path_matcher_get_rule_regex(vfolder0, vf_rules):
```
unho: No parametrize?

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+    vfolder.languages.add(tp0.language)
+    vfolder.projects.add(tp0.project)
+    # m2m handler?
+    vfolder.save()
+    assert (
+        list(vfolder.path_matcher.existing_stores.order_by("pk"))
+        == list(tp0.stores.order_by("pk")))
+    vfolder.languages.add(language1)
+    vfolder.save()
+    assert (
```
unho: I think it is worth testing also a vfolder with single language and multiple projects.

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+        list(vfolder0.path_matcher.rules)
+        == ["foo", "bar"])
+
+
+@pytest.mark.pootle_vfolders
+@pytest.mark.django_db
+def test_vfolder_path_matcher_paths_regex(vfolder0):
+    vfolder0.filter_rules = "foo,bar"
+    assert (
+        list(vfolder0.path_matcher.rules)
```
phlax: no. its not supposed to be

tests/vfolders/path_matcher.py
unho: Also add testing for single language and all projects, and for all projects and single language.

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,194 @@
...
+        == ["foo", "bar"])
+
+    # convert to json field? for now be ws tolerant
+    vfolder0.filter_rules = "foo, bar"
+    assert (
```
unho: Worth adding another test that actually ensures the stores are actually matched provided some rules.

tests/vfolders/path_matcher.py
phlax: yep - its is parametrized

tests/vfolders/path_matcher.py
phlax: @unho we can add more comprehensive testing after - i have added a test to cover your first suggestion - but can i suggest that might be something that you might want to follow up

tests/vfolders/path_matcher.py
phlax: added - about to push

tests/vfolders/path_matcher.py
phlax: apologies - this was a copied out test - removing

tests/vfolders/path_matcher.py
phlax: removed

tests/vfolders/path_matcher.py
phlax: removed

tests/vfolders/path_matcher.py
phlax: fixture moved

unho: Then why it get a parameter `vf_rules`? Can't it be named `rule`?

tests/vfolders/path_matcher.py
```diff
@@ -0,0 +1,171 @@
...
+        sorted(vfolder.path_matcher.projects.values_list("code", flat=True))
+        == ["project0", "project1"])
+    vfolder.all_projects = True
+    vfolder.save()
+    assert vfolder.path_matcher.projects is None
+    assert list(vfolder.path_matcher.languages) == []
+    # no languages set so no stores
+    assert list(vfolder.path_matcher.existing_stores) == []
+    # didnt forget lang preferences
```
unho: Looks like repeated comment from previous test.

tests/vfolders/path_matcher.py
phlax: as a rule parametrized fixturs are always plural - we might want to think about namespacing them in some wayin the future

unho:
lgtm

PR #5235, opened 2016-09-11
Remove _units from Store model as its no longer used

phlax:
afaict this is no longer used anywhere, and certainly isnt triggered by any tests.

I wanted to check tho, before removing

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -1191,9 +1191,6 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     @property
     def units(self):
-        if hasattr(self, '_units'):
-            return self._units
-
         return self.unit_set.filter(state__gt=OBSOLETE).order_by('index')
 
     @units.setter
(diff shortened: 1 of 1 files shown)


julen:
I _believe_ this was used for the auto-terminology feature that created virtual stores from the web UI (this feature was already removed). It _might_ as well have been used for monolingual file format support in the past, but again, I'm not entirely sure.

julen:
It would be good if someone else could have a look and confirm.

phlax:
this is what grepping pootles codebase gives
```
(pootle)phlax@aio:~/.virtualenvs/pootle/pootle$ grep -r "\._units" pootle/ | grep -v log | grep -v Binary | grep -v [#~]
pootle/apps/pootle_store/models.py:            return self._units
pootle/apps/pootle_store/models.py:            for i, unit in enumerate(self._units):
pootle/apps/pootle_store/models.py:                self._units = FakeQuerySet()
pootle/apps/pootle_store/models.py:            self._units.append(newunit)
```

PR #5238, opened 2016-09-12
Optimization: use _ids and bulk_create in scorelog

phlax:
by using ids in scorelog the related models are not pulled from the db.

this PR also uses bulk_create for creating the scorelogs

### DIFF
--- a/pootle/apps/pootle_statistics/models.py
+++ b/pootle/apps/pootle_statistics/models.py
@@ -320,10 +320,18 @@ class Submission(models.Model):
     def save(self, *args, **kwargs):
         super(Submission, self).save(*args, **kwargs)
 
-        if self.needs_scorelog():
-            for score in ScoreLog.get_scorelogs(submission=self):
-                if 'action_code' in score and score['user'] is not None:
-                    ScoreLog.objects.create(**score)
+        if not self.needs_scorelog():
+            return
+
+        scorelogs_created = []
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_statistics/models.py
```diff
@@ -320,10 +320,18 @@ def get_submission_info(self):
...
-                    ScoreLog.objects.create(**score)
+        if not self.needs_scorelog():
+            return
+
+        scorelogs_created = []
+        for score in ScoreLog.get_scorelogs(submission=self):
+            if 'action_code' in score and score['user'] is not None:
+                if not isinstance(score["user"], models.Model):
+                    score["user_id"] = score["user"]
+                    del score["user"]
```
unho: This looks like handling a corrupt data situation. Shouldn't it be done in a data migration?

ta2-1:
@phlax I tested scorelog.save() is triggered.
GTM after merge with further cleanup https://github.com/phlax/pootle/pull/1.

pootle/apps/pootle_statistics/models.py
phlax: @unho no its not that - sometimes user is a real user and sometime just a pk - this works out which and adjusts the dictionary used to create

phlax:
closed in favour of https://github.com/translate/pootle/issues/5243

PR #5239, opened 2016-09-12
Use proxy unit for source_unit when updating from diff

phlax:
When StoreDiff creates its diff of changes the code then retrieves the source units individually (either from the file or the db) - although it already has the information required to update.

This PR uses a proxy class to update the unit without going back to db/fs

*this needs tests to ensure that a plural from a source db unit is updated correctly*

UPDATE: **added some tests to ensure that plural form units are added correctly when copied from db**

### DIFF
--- a/pootle/apps/pootle_store/diff.py
+++ b/pootle/apps/pootle_store/diff.py
@@ -32,6 +32,27 @@ class UnitDiffProxy(UnitProxy):
     def __ne__(self, other):
         return not self == other
 
+    def hasplural(self):
+        return (
+            self.source is not None
+            and (len(self.source.strings) > 1
+                 or getattr(self.source, "plural", None)))
+
+    def getnotes(self, origin=None):
+        return self.unit.get("%s_comment" % origin, "")
+
+    def getcontext(self):
+        return self.unit["context"]
+
+    def isfuzzy(self):
+        return self.unit["state"] == FUZZY
+
+    def isobsolete(self):
+        return self.unit["state"] == OBSOLETE
+
+    def getid(self):
+        return self.unit["unitid"]
+
 
 class DBUnit(UnitDiffProxy):
     pass
@@ -50,6 +71,9 @@ class FileUnit(UnitDiffProxy):
     @property
     def target(self):
         return multistring_to_python(self.unit["target"])
+
+    def hasplural(self):
+        return self.unit["hasplural"]
 
 
 class DiffableStore(object):
@@ -92,6 +116,7 @@ class DiffableStore(object):
             "source": unit.source,
             "target": unit.target,
             "state": state,
+            "hasplural": unit.hasplural(),
             "developer_comment": unit.getnotes(origin="developer"),
             "translator_comment": unit.getnotes(origin="translator")}
 
@@ -271,12 +296,16 @@ class StoreDiff(object):
     def get_units_to_add(self):
         offset = 0
         to_add = []
+        proxy = (
+            isinstance(self.source_store, models.Model)
+            and DBUnit or FileUnit)
+
         for (insert_at, uids_add, next_index_, delta) in self.insert_points:
             for index, uid in enumerate(uids_add):
-                source_unit = self.source_store.findid(uid)
-                if source_unit and source_unit.getid() not in self.target_units:
+                source_unit = self.source_units.get(uid)
+                if source_unit and uid not in self.target_units:
                     new_unit_index = insert_at + index + 1 + offset
-                    to_add += [(source_unit, new_unit_index)]
+                    to_add += [(proxy(source_unit), new_unit_index)]
             if delta > 0:
                 offset += delta
         return to_add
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/diff.py
```diff
@@ -251,10 +272,14 @@ def diff(self):
         """Return a dictionary of change actions or None if there are no
         changes to be made.
         """
-        diff = {"index": self.get_indexes_to_update(),
```
ta2-1: why can't we leave it as it was?

phlax: we can i was using `debug_sql` on it - thats why it changed

tests/models/translationproject.py
```diff
@@ -296,8 +300,16 @@ def test_tp_tool_update(po_directory, tp0, templates):
     tp0.stores.first().delete()
     tp0.stores.first().units.first().delete()
     unit = tp0.stores.first().units.first()
+    unit.source = multistring(["NEW TARGET", "NEW TARGETS"])
     unit.target = "NEW TARGET"
+    unit.context = "something-else"
     unit.save()
+    newunit = unit.__class__()
+    newunit.source = multistring(["OTHER NEW TARGET", "OTHER NEW TARGETS"])
+    newunit.target = "OTHER NEW TARGET"
```
dwaynebailey: Does this need some exercising around a multstring based target?

phlax: perhaps - although this is not something ever checked tested - the issue here was that the proxy class didnt have `hasplural` so would have failed when adding units to a store

phlax:
re queries - this PR increases db calls slightly - due to test changes. If we were testing more extensively with db updating it would have more impact on queries.

re timings - this PR seems to improve timings - altho these can be somewhat variable on a local system

before:

```
TESTS TOTAL test time: 195.282850981
TESTS TOTAL queries: 82177
```
after
```
TESTS TOTAL test time: 176.819938898
TESTS TOTAL queries: 82191
```

worth mentioning that this PR improves stability as much as performance - in that its not going back to the disk to check the units (i believe but i would need check that 100%)

pootle/apps/pootle_store/unit/proxy.py
```diff
@@ -39,3 +40,18 @@ def hasplural(self):
             self.source is not None
             and (len(self.source.strings) > 1
                  or getattr(self.source, "plural", None)))
+
+    def getnotes(self, origin=None):
+        return self.unit["%s_comment" % origin]
+
+    def getcontext(self):
```
ta2-1: @phlax have you considered to put these methods to DiffUnitProxy class?

phlax: i think they are useful for any unitproxy - no?

ta2-1: I used UnitProxy as base class but I didn't use any of its methods here
https://github.com/translate/pootle/pull/5202/commits/a56c885d290f12cda93cf25ff624bbbcb0775a54

phlax: fair point - im wondering if its useful outside of diff - here its used in the update too - and what we should call it

unho:
Commit "Add getcontext, hasplurals, isfuzzy, isobsolete and getid to UnitProxy" message doesn't match its changes.

pootle/apps/pootle_store/diff.py
```diff
@@ -32,6 +32,27 @@ def __eq__(self, other):
...
         return not self == other
 
+    def hasplural(self):
+        return (
+            self.source is not None
+            and (len(self.source.strings) > 1
+                 or getattr(self.source, "plural", None)))
+
+    def getnotes(self, origin=None):
+        return self.unit["%s_comment" % origin]
```
unho: How does this render in case no `origin` is specified? Perhaps it is better not to default to `None`?

phlax:
oops - need to move that method - and kill commit...

pootle/apps/pootle_store/diff.py
phlax: hm - i think it should return `""`

unho:
lgtm

pootle/apps/pootle_store/diff.py
```diff
@@ -271,12 +296,16 @@ def get_indexes_to_update(self):
...
+        proxy = (
+            isinstance(self.source_store, models.Model)
+            and DBUnit or FileUnit)
+
         for (insert_at, uids_add, next_index_, delta) in self.insert_points:
             for index, uid in enumerate(uids_add):
-                source_unit = self.source_store.findid(uid)
-                if source_unit and source_unit.getid() not in self.target_units:
+                source_unit = self.source_units.get(uid)
+                if source_unit and uid not in self.target_units:
```
ta2-1: uid ==  source_unit.getid() ?

phlax: yep

ta2-1:
lgtm

PR #5247, opened 2016-09-14
Implement an ISO 8601-aware `DateTimeField`

julen:
`ISODateTimeField` is a specialized `DateTimeField` which understands the ISO
8601 format. Unfortunately, Django's built-in field is not able to do so, even
if a custom input format is specified, mostly because Python 2 doesn't support
`%z` as a timezone marker.

This field is useful for forms dealing with strings coming from the client-side,
as they will conform to a simplification of the ISO 8601 format (defined in
ECMA-262).
This is what's produced by the `Date` object's `toJSON` method, as well as what
Django's (and hence Pootle's) JSON serializer generates. Note the `datetime`
object as parsed by the field  will add padding zeros to conform to 6-digit
milliseconds. Serializing will strip this down to 3 digits again.

Refs. http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15
Refs. https://code.djangoproject.com/ticket/11385

### DIFF
--- a/pootle/core/fields.py
+++ b/pootle/core/fields.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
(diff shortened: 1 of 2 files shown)


unho:
lgtm

PR #5251, opened 2016-09-15
Cleanup: try and stop travis updating stuff

phlax:
atm travis keeps reinstalling pytest, and py - trying to stop that...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,13 @@ cache:
   directories:
     - pootle/static/js/node_modules
     - pootle/assets
+    - $HOME/virtualenv/python2.7.9/bin
     - $HOME/virtualenv/python2.7.9/lib
     - $HOME/virtualenv/python2.7/bin
+before_install:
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -72,3 +72,4 @@ commands=
     py.test --cov-report= --cov=. -v -s -k test_debug_timing --debug-tests -
     python {toxinidir}/run_coveralls.py
     codecov -e TOXENV
+    pip install -r requirements/tests.txt
```
unho: But this is already being installed in Travis config, why installing here too?

phlax: i wasnt sure what the problem is (still not 100%)

i believe its basically that requirements parser in setup.py is installing the wrong pytest-django version (and maybe other stuff)

also, its installing to different directories because its being triggered by make/setup.py and not by pip

phlax:
ref https://github.com/travis-ci/travis-ci/issues/4873

PR #5253, opened 2016-09-15
[WIP] Bugfix: report changes after IME composition end in Chrome

julen:
Chrome 53 changed the order in which composition events are fired with respect to the input event, which affects the way we determine the end of IME composition.

Since at the moment there is no possibility to use the `isComposing` property across browsers, we need to work around this by calling the change handler manually after composition is over.

Refs.
https://chromium.googlesource.com/chromium/src/+/afce9d93e76f2ff81baaa088a4ea25f67d1a76b3%5E%21/
Refs. https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing

### DIFF
--- a/pootle/static/js/editor/components/RawFontTextarea.js
+++ b/pootle/static/js/editor/components/RawFontTextarea.js
@@ -262,6 +262,8 @@ const RawFontTextarea = React.createClass({
 
   handleComposition(e) {
     if (e.type === 'compositionend') {
+      // XXX: calling change handler to work around Chrome's changed behavior
+      this._handleChange(e.target.value);
       this.isComposing = false;
     } else {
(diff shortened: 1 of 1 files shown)


dwaynebailey:
@julen is there any way to test this?

julen:
Yes, to reproduce the issue Chrome 53+ needs to be used: type using an IME and commit the changes. The submit button will stay disabled, because the changes haven't been reported to the editor. With the PR, committing the changes will report them to the editor and the button will be enabled as expected.

Still, as mentioned in the description this is a workaround which I'm not very happy with, so I am stalling it in favor of investigating any option that doesn't require composition events at all.

julen:
Closing in favor of #5281.

PR #5256, opened 2016-09-15
Remove monkey patched for lazy_* gettext functions

dwaynebailey:
This removes the monkey patching of the lazy_* gettext functions.

Note that this means we won't get protective support for applications that we don't control i.e. things in our dependencies.


julen:
Looks good and I have no objections. Part of me though thinks it's questionable the usefulness of half-protection via `pootle.i18n.gettext._format_translation` and should rather be an all or nothing choice at least to ensure consistent behavior.

unho:
lgtm

PR #5262, opened 2016-09-16
Cleanup: remove en_tutorial_po fixture from store tests

phlax:
This speeds up tests and standardises the fixtures

### DIFF
--- a/tests/models/store.py
+++ b/tests/models/store.py
@@ -57,6 +57,7 @@ def _update_from_upload_file(store, update_file,
         store.update(store.file.store)
     test_store = getclass(upload)(upload.read())
     store_revision = parse_pootle_revision(test_store)
+    test_store._FOO = 1
     store.update(test_store, store_revision=store_revision,
                  user=user, submission_type=submission_type)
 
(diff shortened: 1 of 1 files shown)


tests/models/store.py
```diff
@@ -153,15 +152,14 @@ def test_update_ts_plurals(store_po, test_fs):
...
+def test_update_with_non_ascii(store0, test_fs):
+    store0.state = PARSED
+    orig_units = store0.units.count()
+    with test_fs.open(['data', 'po', 'tutorial', 'en',
                        'tutorial_non_ascii.po']) as f:
         store = getclass(f)(f.read())
-    en_tutorial_po.update(store)
-    assert en_tutorial_po.units[0].target == "Hèllö, wôrld"
+    store0.update(store)
+    assert store0.units[orig_units].target == "Hèllö, wôrld"
```
dwaynebailey: I realise it isn't this PR.  But the non-ascii should take things into the unicode range.  I fear that all these are in the first 127 chars so they're still technically single byte ascii, so not doing what is hoped here.

phlax: perhaps PR with some better unicode chars?

phlax: ☠

phlax:
closed in https://github.com/translate/pootle/commit/d5a3eecbc01dc5c44932d1227e245db56cc1d6d1

PR #5276, opened 2016-09-19
Tests: ensure we're running sqlite in memory

dwaynebailey:
sqlite in memory database is much faster to run than filebased.  This test is a canary and will fail if somehow the test environment changes to make it file based again.

### DIFF
--- a/tests/database.py
+++ b/tests/database.py
@@ -1,0 +1,17 @@
+import pytest
+# -*- coding: utf-8 -*-
+#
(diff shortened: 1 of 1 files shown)


phlax:
missing file header - otherwise gtm

tests/database.py
```diff
@@ -0,0 +1,17 @@
+import pytest
```
unho: I would say the import should be after the header.

Issue #5261, opened 2016-09-15
[LF] symbol-related issues in the raw editor


iafan:
During testing, we've found a few edge case issues with the new editor code that are worth documenting. I don't think they deserve being filed separately, so here they are in one list:

1. Place caret immediately before `LF` symbol and press Delete. The caret will be placed incorrectly after deleting the symbol.
2. Select a row ending with `LF` and press Delete. The first symbol of the selection will not be deleted.
3. Select some text sequence ending with `LF` and drag it with the mouse to some other place. The line break (`LF`) is lost.

Having said that, we already have a fix for this (a refactored code with more consistent approach to handling caret position and selection overall that doesn't exhibit these issues), but it may take some more time till we test it and integrate with the existing code.

dwaynebailey:
@translate/evernote do you have an ETA on a fix for this? We had planned to do a staging test server to expose to our users before live deployment.  We've decided to delay until this is fixed.  Result is a block on our deployment.

julen:
No ETA as I am aware of although I don't believe it would take much time provided the experiments Igor is doing are successful.

dwaynebailey:
@julen I'm still going to do some local testing to replicate these issues, I'll likely evaluate if I need to block on them based on how problematic I think they are and based on Igor's experiments.

iafan:
@dwaynebailey we basically have the new working code here: https://github.com/julen/raw-font-alternative, and the demo you can play with is here: http://julen.io/raw-font-alternative/  (note that this demo doesn't handle undo/redo by itself, see below).

This is a rewrite of event handlers for the textarea that is more compact, less intrusive wrt IME input and feels more natural when it comes to working with caret/selection. We need to integrate it back into the main component, where it will get support for the custom undo/redo stack that we already have.

We're committed to landing this ASAP. We'll schedule this with Julen.

julen:
Fixed by #5281.

PR #5281, opened 2016-09-22
Alternative method to handle raw font events

julen:
In this PR the [LF]-related issues reported in #5261 are taken care of by taking a different approach on how input events are handled. At the same time, this allows to avoid having to handle IMEs in application code (and workaround such as #5253).

The PR integrates the experiments @iafan did in an independent environment. Checklist:
* [x] Toggling the raw mode still has no effects
* [x] Values' state should totally be uncontrolled (to avoid conflicts, only rely on the DOM for this).
* [x] IE11 places the caret at the end after edits
* [x] Fix setting values from outside of the component tree
* [x] Make undo/redo remember cursor selections
* [x] Tests need to be adjusted
* [x] Dead code removal
* [x] Linting issue fixes
* [x] Cleanup the code for improved encapsulation of the functionality

**UPDATE**: This is now ready to check-in.


dwaynebailey:
@julen tested with the `escapes.po` file as well as issues raised in #5261, I didn't see any issues related to those.

dwaynebailey:
@julen lgtm based on my testing.  I'm not sure if anyone else wants to do a code review, I'll let them respond

unho:
Just doing a quick code review.

unho:
lgtm after quick code review and some testing on a staging server.

PR #5283, opened 2016-09-23
Use mysqlclient for MySQL database backend

dwaynebailey:
This is the recommended driver for MySQL since at least Django 1.7.  See
https://docs.djangoproject.com/en/1.9/ref/databases/#mysql-db-api-drivers

This brings with it Python 3 support which will be relevant sometime
soon.


unho:
lgtm. At the time of releasing this is something that must be prominently communicated in the release notes.

dwaynebailey:
@unho I'll add that to the relese notes as part of this PR

phlax:
and the install docs

Issue #5285, opened 2016-09-23
Leading and trailing whitespaces lost when submitting a translation


alexanderlaw:
When I try to submit a translation with the whitespaces at the start and the end of the string, these whitespaces not preserved in the store.
It makes impossible to translate postresql-messages, for example.
Look at https://babel.postgresql.org/po-9.6-branch/ecpg.pot
It contains messages like:
msgid "  -V, --version  output version information, then exit\n"

alexanderlaw:
Patch proposed:
https://github.com/translate/pootle/pull/5286/commits/d428ad8e524dd5cc0ebefd10826fa9e6b1f107ec

dwaynebailey:
Landed in d428ad8e524dd5cc0ebefd10826fa9e6b1f107ec

PR #5286, opened 2016-09-23
Preserve leading and trailing spaces in translation field  #5285

alexanderlaw:


dwaynebailey:
This seems to be a regression from https://github.com/django/django/commit/11cac1bd8ef7546ca32d9969d4348bf412dc6664 and our move to Django 1.9

@alexanderlaw thanks for picking this up and for the fix.  Would you be able to write tests for this?

alexanderlaw:
The test added.

dwaynebailey:
@alexanderlaw thanks once again for your fix and additionally for the tests.

PR #5287, opened 2016-09-23
[RFC] Release note rationalisation

dwaynebailey:
This tries to cleanup the release notes:

1. We only have one set of release notes for a release, just like Django
2. We have one for each bugfix release, just like Django
3. No idea what we do when we do a prerelease, but we should point to this
4. Changed wording so that we avoid obvious issues like this getting out of date.  Phrases like "this is the latest 2.7 release" are insane as they obviously go out of date.
5. All this because I wanted to add a oneliner about using mysqlclient instead of MySQLdb

Lastly, I really think we should be adding things to RN as part of things landing.  Instead of relying on the release manager to parse all the commits.  We're bound to forget or miss critical things that should be in the RN with the fix it later strategy. /sortofrant


julen:
[Totally support this](https://github.com/translate/pootle/pull/5167#issuecomment-243119160) 👍.

docs/releases/index.rst
```diff
@@ -14,9 +14,7 @@ If you are upgrading Pootle, you might want to see some tips to ensure your
 .. toctree::
    :maxdepth: 1
 
-   2.8.0b3 <2.8.0b3>
-   2.8.0b2 <2.8.0b2>
-   2.8.0b1 <2.8.0b1>
+   2.8.0
```
unho: This is not pointing to any doc at all.

unho:
I don't like at all dropping release notes for intermediate releases. What happens when someone tries to follow a link to some nonexistent release notes?

docs/releases/index.rst
dwaynebailey: Yes it is, they work just like normal links. Its just using the header form the 2.8.0.rst file

dwaynebailey:
@unho they won't find it.  We could try redirect to the actual release.  But I'm not going to keep something around that causes clutter and confusion just because it might break for someone.

dwaynebailey:
@unho we can use readthedocs redirects as needed for these pages.

unho:
lgtm

docs/releases/index.rst
```diff
@@ -11,34 +12,64 @@ Also be aware of the :ref:`important changes in the Translate Toolkit
...
 
+Final releases
+==============
+
+2.8 release
+-----------
+.. toctree::
+   :maxdepth: 1
+
+   2.8.0
```
julen: Note this is listed under _Final releases_, but it's not a final release. I see there's a warning in the _2.8.0_ doc's title, however that might not be enough to prevent confusion. I believe putting it under a _Upcoming releases_ or similar heading might help in that regards.

dwaynebailey: @julen good point.  I guess this is working on Django cause it's within their `dev` docs https://docs.djangoproject.com/en/dev/releases/

I'll figure something out, but ideally would like to push towards the Django way

PR #5292, opened 2016-09-27
Provide default for ACCOUNT_SESSION_REMEMBER

### LINKED ISSUE #5130: Keep users logged in for more than a session
I don't know if it depends on the user's permissions, but I have to login every time I restart the browser, which means daily on Nightly (or even more frequently).

It would be great to keep user logged in for a few days.

unho:
This forces sessions to be remembered, so users don't have to log in every time they restart the browser.

This gets back the behavior we had before moving to allauth.

Fixes #5130.

### DIFF
--- a/docs/releases/2.8.0.rst
+++ b/docs/releases/2.8.0.rst
@@ -149,6 +149,9 @@ Changes in settings
   This restricts how many units are shown in export views.
 - :setting:`POOTLE_SEARCH_BACKEND` was added, to allow configuring the search
   backend to be used.
+- Changed the default value for `ACCOUNT_SESSION_REMEMBER
+  <https://django-allauth.readthedocs.io/en/latest/configuration.html>`_ so now
+  sessions are always remembered.  
 
 
 Credits
--- a/pootle/settings/40-apps.conf
+++ b/pootle/settings/40-apps.conf
@@ -123,6 +123,9 @@ ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True
 # Automatically sign users in upon password reset completion
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 
+# True to always remember user session. False to not remember. None to ask user.
+ACCOUNT_SESSION_REMEMBER = True
+
 # Whether user sign up is enabled
 POOTLE_SIGNUP_ENABLED = True
 


pootle/settings/40-apps.conf
```diff
@@ -123,6 +123,9 @@ ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True
 # Automatically sign users in upon password reset completion
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 
+# True to always remember user session. False to not remember. None to ask user.
+ACOUNT_SESSION_REMEMBER = True
```
dwaynebailey: Is there a reason we aren't setting this to None? Would that require a UI change?  Just seems like the much more obvious way for Pootle the product and I'm not seeing why True was the one chosen.

pootle/settings/90-local.conf.template
```diff
@@ -237,3 +237,6 @@ POOTLE_MT_BACKENDS = [
 TEMPLATES[0]['DIRS'] = [
     # working_path(os.path.join('custom', 'templates')),
 ]
+
+# True to always remember user session. False to not remember. None to ask user.
+ACOUNT_SESSION_REMEMBER = True
```
dwaynebailey: We're already setting this is 40.  If we don't expect people to change it then lets leave it out.  If we do then lets comment it out so that they get the default we've configured.

unho: Removed.

unho:
Updated after the feedback received. @julen thanks for your input.

pootle/settings/40-apps.conf
dwaynebailey: sp ACCOUNT and elsewhere, explains why I couldn't find the config.

docs/releases/2.8.0.rst
```diff
@@ -149,6 +149,8 @@ Changes in settings
   This restricts how many units are shown in export views.
 - :setting:`POOTLE_SEARCH_BACKEND` was added, to allow configuring the search
   backend to be used.
+- Changed the default value for ``ACOUNT_SESSION_REMEMBER`` so now sessions are
```
dwaynebailey: Lets actually link to the upstream docs if they exist https://django-allauth.readthedocs.io/en/latest/configuration.html

dwaynebailey: sp

unho:
Updated and rebased.

docs/releases/2.8.0.rst
```diff
@@ -149,6 +149,9 @@ Changes in settings
   This restricts how many units are shown in export views.
 - :setting:`POOTLE_SEARCH_BACKEND` was added, to allow configuring the search
   backend to be used.
+- Changed the default value for `ACCOUNT_SESSION_REMEMBER
+  <https://django-allauth.readthedocs.io/en/latest/configuration.html>`_ so now
+  sessions always are remembered.
```
dwaynebailey: are always remembered

dwaynebailey:
lgtm, one minor wording issue

PR #5296, opened 2016-09-29
l10n: allow placeable reordering

dwaynebailey:
One string that's using %s for two placeables and is causing gettext warning, apart from not allowing translators to reorder in a very unlikely event.

### DIFF
--- a/pootle/static/js/auth/components/SocialVerification.js
+++ b/pootle/static/js/auth/components/SocialVerification.js
@@ -9,6 +9,7 @@
 import assign from 'object-assign';
 import React from 'react';
 
+import { t } from 'utils/i18n';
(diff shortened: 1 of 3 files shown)


pootle/static/js/auth/components/SocialVerification.js
```diff
@@ -13,6 +13,7 @@ import FormElement from 'components/FormElement';
 import FormMixin from 'mixins/FormMixin';
 
 import { gotoScreen, verifySocial } from '../actions';
+import { t } from 'utils/i18n';
```
julen: I would suggest to place this import in the block above, as it is an app-wide import and not a module-relative import.

PR #5302, opened 2016-09-29
Highlight spaces around [NBSP] and [TAB]

### LINKED ISSUE #5297: New editor doesn't clearly show spaces when you have [sp][nbsp]
In the new editor a string combining [sp] and [nbsp] such as `One[sp][nbsp]Two` creates a string where it is clear that you have a [nbsp] but it is not clear that there is a space before the [nbsp]. 

When we have two or more [sp] we use dotted spaces.  We should be using a dotted [sp] where there is a preceding or following WS otherwise you can't really see the [sp].

[nl] does show dotted space before and after the [nl] character

[nbsp] and [tab] are all candidates that can potentially hide a leading or trailing [sp].

julen:
This PR implements highlighting spaces around the [NBSP] and [TAB] symbols. An extra optimization for handling spaces is included as well.

Fixes #5297.

### DIFF
--- a/pootle/static/js/editor/utils/font.js
+++ b/pootle/static/js/editor/utils/font.js
@@ -193,7 +193,7 @@ export function raw2sym(value, { isRawMode = false } = {}) {
   // in raw mode, replace all spaces;
   // otherwise, replace two or more spaces in a row
   let newValue = isRawMode ?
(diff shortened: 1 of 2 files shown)


unho:
lgtm

PR #5304, opened 2016-09-30
Export view: fix highlighting of raw font

julen:
The font wasn't being loaded in the export view page hence default symbols
would be displayed instead of our custom ones.

--

To check this out, open up an export view page with units containing special characters that need to be highlighted. Previously a system default font would be used, with the PR the raw font symbols should be displayed.


unho:
lgtm

PR #5306, opened 2016-10-03
Requirements update: pytest, pydocstyle & pytz

dwaynebailey:
* pytest - mnor bug fixes
* pydocstyle - minor update with soe new disabled checks
* pytz - update for IANA tz update


unho:
lgtm

PR #5307, opened 2016-10-03
HTML highlight: allow `$` string between delimiters

julen:
In its previous shape, highlighting would be screwed due to the special meaning
of the dollar sign in regular expression replacement strings: they are treated
as replacement patterns.


unho:
Any quick way to test this?

julen:
If you wish, you can check out the new test against the old code.

unho:
lgtm

PR #5308, opened 2016-10-04
Reintroduce editor value's state

julen:
While we removed it motivated by the fact that the RawFontTextarea wasn't reading from it, it's a legitimate use-case other components willing to read from the state without manually resourcing to querying the DOM in an ad-hoc manner.


ta2-1:
thanks @julen 👍 
you had already made this PR when I realised that we get values from DOM.

unho:
lgtm

PR #5313, opened 2016-10-06
Leftovers from pootle_fs implementation

unho:
These have been detected when a newbie tried to setup pootle_fs.


phlax:
lgtm, nice one @unho

PR #5315, opened 2016-10-06
Remove django-debug-toolbar for dev

nootanghimire:
There were some problems concerning the particular django extension
while trying to run the dev build. I think it is better to remove
this since we could easily use other tools such as (i)pdb for debugging
. Removal of this will also make running the dev build easier.

### DIFF
--- a/pootle/settings/90-dev-local.conf.sample
+++ b/pootle/settings/90-dev-local.conf.sample
@@ -137,23 +137,6 @@ ASSETS_AUTO_BUILD = True
 INSTALLED_APPS += ['django_extensions']
 
 
-#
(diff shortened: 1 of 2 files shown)


unho:
Not sure if anyone is still using this. @julen ?

phlax:
lgtm from my side, ive always found the usefuleness of debug_toolbar to be outweighed by the performance hit it introduces

phlax:
...and ive experienced new devs having this issue

dwaynebailey:
Thanks @nootanghimire, landed in 0b5a5196a33a48e662ca32ef02ce64c6f5e52b50

Issue #5291, opened 2016-09-26
Review forms.CharField() occurances


dwaynebailey:
Django 1.9 introduced `strip=True` by default for CharField in forms.  Great idea except in localisation where we want to preserve exactly what the user submits.  We've fixed the main one in  d428ad8e524dd5cc0ebefd10826fa9e6b1f107ec via #5286 but this ticket is to review other occurances that would require us to specify `strip=False`, likely comments are the only ones.

julen:
I went through all existing `CharField`s in forms and the only conflictive place I can think of is the special characters field for languages. I'll PR the change required to make that work as expected.

dwaynebailey:
Looks like these are fixed based on @julen's review

PR #5317, opened 2016-10-06
Allow whitespace as part of special characters

julen:
With this PR, whitespace is now allowed as part of special characters. As a bonus, it also validates for uniqueness of inputted characters.

Refs. #5291.


unho:
lgtm

dwaynebailey:
lgtm, thanks @julen for picking up that special char instance of stripping.

PR #5318, opened 2016-10-06
Requirements: pin the stuff we build with, ensure that stuff we use is py3 and Dj1.9

dwaynebailey:
A few objectives here:

1. Pin anything we use in test, lint and dev - we don't want meta stuff breaking our dev cycle.  We can update linting and testing tools once we know the updates have no adverse consequences.  requires.io will pick up anything out of date pretty quickly.
2. Set a minimum version for tools we use that aren't versioned at all.  We have a few where for Django 1.9 support we actually do have a minimum.  In some places we've just set currently used version as the new minimum.
3. Ensure that all the packages we use are Python 3 compatible, thus eliminating that as a blockage.  Only cssmin has no Python 3 support at the moment.

### DIFF
--- a/requirements/_docs.txt
+++ b/requirements/_docs.txt
@@ -1,3 +1,3 @@
 # Documentation
 
-sphinx
+sphinx==1.4.8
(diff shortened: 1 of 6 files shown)


unho:
lgtm

PR #5325, opened 2016-10-07
Checks: check revision if redis server available

dwaynebailey:
Bail out of the revision check if there is no available redis server.
There is a check for the redis server that will flag the missing server.


nootanghimire:
lgtm... :+1:

dwaynebailey:
@nootanghimire thanks for the review

PR #5326, opened 2016-10-07
Requirements: break up the dbs and pin it all to the wall

dwaynebailey:
* DBs are now specified only in one place and range pinned
* All `deploy` is now range pinned
* `travis` is simply pinned


unho:
lgtm

PR #5328, opened 2016-10-07
Remove vftis

phlax:
as they are no longer required


pootle/apps/pootle_store/models.py
```diff
@@ -1526,9 +1525,6 @@ def get_parents(self):
         else:
             parents = [self.parent]
 
-        if 'virtualfolder' in settings.INSTALLED_APPS:
-            parents.extend(self.parent_vf_treeitems.all())
-
         return parents
```
unho: Nit: I would simplify this method:

```python
        if self.parent.is_translationproject():
            return [self.translation_project]

        return [self.parent]
```

pootle/apps/pootle_translationproject/models.py
```diff
@@ -392,17 +392,6 @@ def clear_all_cache(self, children=True, parents=True):
...
-            # VirtualFolderTreeItem can only have VirtualFolderTreeItem parents
-            # so it is necessary to flush their cache by calling them one by
-            # one.
-            from virtualfolder.models import VirtualFolderTreeItem
-            tp_vfolder_treeitems = VirtualFolderTreeItem.objects.filter(
-                pootle_path__startswith=self.pootle_path
-            )
-            for vfolder_treeitem in tp_vfolder_treeitems.iterator():
-                vfolder_treeitem.clear_all_cache(children=False, parents=False)
-
```
unho: Just remove the whole method. It will be inherited from parent class.

unho:
lgtm nits aside.

tests/views/project.py
```diff
@@ -65,9 +65,7 @@ def _test_translate_view(project, request, response, kwargs, settings):
             editor_extends="projects/base.html",
             check_categories=get_qualitycheck_schema(),
             previous_url=get_previous_url(request),
-            display_priority=(
-                VirtualFolderTreeItem.objects.filter(
-                    pootle_path__startswith=pootle_path).exists()),
+            display_priority=display_priority,
```
unho: Probably best to adjust this in order to test when vfolders are present.

PR #5338, opened 2016-10-09
Add pootle.core.views.admin

phlax:
Adds generic admin view and form view


pootle/core/views/admin.py
```diff
@@ -0,0 +1,13 @@
+
```
unho: No license notices?

PR #5343, opened 2016-10-10
Upgrade select2

phlax:
This is required in order to (easily) use xhr in selects with many options.

Also this seems to not have been upgraded for quite a while


pootle/templates/core/_breadcrumbs.html
```diff
@@ -14,7 +14,9 @@
     <option value=""></option>
     {% endif %}
     {% for lang_code, lang_name in ALL_LANGUAGES.items %}
-    <option value="{{ lang_code }}">{{ lang_name }}</option>
+    <option
+       value="{{ lang_code }}"
+       {% if lang_code == language.code %}selected="selected"{% endif %}>{{ lang_name }}</option>
```
unho: Nit: too indented, use two spaces indent. Same below

pootle/static/js/browser.js
```diff
@@ -193,11 +193,6 @@ function fixResourcePathBreadcrumbGeometry() {
 
 function fixDropdowns() {
   // We can't use `e.persisted` here. See bug 2949 for reference
-  const selectors = [sel.navigation, sel.language, sel.project, sel.resource];
-  for (let i = 0; i < selectors.length; i++) {
-    const $el = $(selectors[i]);
-    $el.select2('val', $el.data('initial-code'));
```
unho: Is the `initial-code` still needed? If not, drop too.

dwaynebailey:
Done various reviews of old vs new and we're almost 100% feature parity.

Difference that I've noticed:

1. x clear filter is larger on new, makes it easier to see and click
2. Search highlighting highlights but doesn't underline, not critical I don't think
3. Current selection highlighting is different on new.  New has current in grey.  Old has current selected using the selection highlighting of White on Blue. Old lost an idea of what is current when in operation.  New keeps this which is good.  Grey might not be the best choice in that it puts current selection to the back instead of the front.

I'm gtm on this. I think we'll see going forward if items 2 and 3 are issues at all.

dwaynebailey:
@phlax while we're using select2 from npm we seem to be still hosting select2.css plus our customisation, I would have thought we'd be able to drop the css for select2 and just override our needs?

PR #5344, opened 2016-10-10
[Django 1.9] fix admin permissions page

### LINKED ISSUE #5336: Permission form breaks (with latest django)
Attempting to use the permission form on current master creates the following error:

```
TypeError: object.__new__(generator) is not safe, use generator.__new__()
```

it appears to be caused by how the choices are set on PermissionSetForm

julen:
This PR fixes the permissions administration pages. Since the switch to Django 1.9 it wasn't possible to access these as generators were being used in `GroupedModelChoiceField`. Converting these to iterators solves the problem.

Fixes #5336.


unho:
Tested. GTM once linting issue picked by Travis is fixed. Thanks a lot for this one.

PR #5348, opened 2016-10-11
Minor nits: pootle_fs and others

unho:
- Docs changes
- Typo
- Add requirement for pootle_fs_git
- Moved `POOTLE_FS_WORKING_PATH`

### DIFF
--- a/pootle/.pootle_fs/tmp/README
+++ b/pootle/.pootle_fs/tmp/README
@@ -1,0 +1,5 @@
+This directory (pootle_fs) is the working directory for pootle_fs.
+
+WARNING: Do NEVER touch anything on this directory.
+
+WARNING: This directory can potentially get very large, so you need to place it somewhere with plenty of room.
--- a/requirements/_pootle_fs.txt
+++ b/requirements/_pootle_fs.txt
@@ -1,0 +1,4 @@
+# pootle_fs requirements
+
+# If you want to use pootle_fs git plugin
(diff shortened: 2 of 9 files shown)


requirements/_pootle_fs.txt
```diff
@@ -0,0 +1,4 @@
+# pootle_fs requirements
+
+# If you want to use pootle_fs git plugin
+-e git+https://github.com/translate/pootle_fs_git@pootle_fs_cleanups#egg=pootle-fs-git
```
unho: Maybe this should be disabled by default.

unho: Changed.

pootle/.pootle_fs/tmp/README
```diff
@@ -1 +1,3 @@
 This directory (pootle_fs) is the working directory for pootle_fs.
+
+NEVER touch anything on this directory.
```
phlax: i think we neen some more information here

this dir has the potential to get very large, so we need to communicate that they might want/need to place it somewhere specific and make the necessary adjustments to their settings

also rem that this folders ends up in an egg dir in the middle of nowhere in a production env, so im thinking there needs to be docs elsewhere explaining all of this

unho: Will adjust to highlight that both in docs and here.

phlax:
other than the docs changes, lgtm

PR #5352, opened 2016-10-12
Rm paid tasks

phlax:
As this functionality is unsupported, and the code is untested and has a history of security issues this PR removes it


pootle/static/js/admin/components/User/UserForm.js
```diff
@@ -159,8 +159,6 @@ const UserForm = React.createClass({
           <ul className="action-links">
             <li><a href={model.getProfileUrl()}>{gettext('Public Profile')}</a></li>
             <li><a href={model.getSettingsUrl()}>{gettext('Settings')}</a></li>
-            <li><a href={model.getStatsUrl()}>{gettext('Statistics')}</a></li>
```
phlax: this view is removed

unho: If removed then you might consider dropping `getStatsUrl` too.

phlax: yep, good point

dwaynebailey:
lgtm, withe caveat that we'll need to get something back for 2.8.0

unho:
I am not sure if removing all user's stats is something I really like, but won't block if that is considered to be back in a better way. lgtm once Travis passes.

phlax:
agreed - opening a ticket...

https://github.com/translate/pootle/issues/5356

PR #5353, opened 2016-10-12
Rm export views

phlax:
remove export views as these represent a major DOS vector on any pootle site


dwaynebailey:
gtm. Is `get_filter_name` really only used in Export View?

phlax:
yep, afaict

PR #5355, opened 2016-10-13
Use PEP0440 compatible release syntax ~=

dwaynebailey:
https://www.python.org/dev/peps/pep-0440/#compatible-release

requires.io doesn't seem to understand ~= so the rq.filter directives
remain.


dwaynebailey:
Landed in 55754f3aa6c665409c3b76e79102da0bd2bf918b

PR #5369, opened 2016-10-14
[requires.io] dependency update on stable/2.6.2 branch

dwaynebailey:


unho:
Provided requirements are good enough. No need to bump up.

PR #5372, opened 2016-10-14
[requires.io] dependency update on stable/2.5.1 branch

dwaynebailey:


unho:
We do not support this old release.

PR #5375, opened 2016-10-14
[requires.io] dependency update on stable/2.6.1 branch

dwaynebailey:


dwaynebailey:
Misconfigured requires.io, branch is not relevant

PR #5387, opened 2016-10-17
Canonical url

phlax:
Enforce setting of a CANONICAL_URL

This means that async operations -  eg notifications can authoritatively obtain the correct URL

This will look for a CANONICAL_URL, and if that is set use it - otherwise it will use the django.contrib.sites framework for obtaining the url

### DIFF
--- a/pootle/settings/30-site.conf
+++ b/pootle/settings/30-site.conf
@@ -15,6 +15,8 @@ MANAGERS = ADMINS
 
 DEBUG = False
 
+POOTLE_CANONICAL_URL = "http://localhost"
+
 # A list of strings representing the host/domain names that this Pootle server
 # can serve. This is a Django's security measure. More details at
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
--- a/pootle/apps/pootle_app/apps.py
+++ b/pootle/apps/pootle_app/apps.py
@@ -11,6 +11,8 @@ Pootle App Config
 See https://docs.djangoproject.com/en/1.9/ref/applications/
 """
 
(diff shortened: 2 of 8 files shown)


pootle/settings/30-site.conf
```diff
@@ -15,6 +15,8 @@ MANAGERS = ADMINS
 
 DEBUG = False
 
+CANONICAL_URL = ""
```
phlax: my feeling is that its not really a `POOTLE` thing - but yep it defo needs some docs

dwaynebailey: If we're expecting people to set this or use Sites.  Then this definitely needs to be in 90-local.conf.sample.  And I'm guessing now that we might want to to be localhost if we want Pootle to be able to run straight away after init?

We moved all our settings to the POOTLE_ namespace just to make sure its clear what we manage and what we don't.  Unless we're using something from Django I'd say its our thing and prevents any future conflict when Django decided CANONICAL_URL would b a great thing to add

phlax: re Sites - using it is undocumented -and has no ui - im figuring anyone using sites is already customizing heavily - probs noone

i am wondering tho if there is any possibility of a historical use that might mean someone has `sites` configured

dwaynebailey: @phlax yeah I get that, but now we can init/initdb without much effort, this would change that.  I think missing should fail, and yip if we just upgrade 'localhost' to a warning that would cover things I think.

unho:
lgtm. Docs can be landed afterwards.

PR #5390, opened 2016-10-18
Cleanup formtable (/more)

phlax:
more cleanups/updates/fixes for formtable

### DIFF
--- a/pootle/core/forms.py
+++ b/pootle/core/forms.py
@@ -16,6 +16,7 @@ from random import randint
 
 from django import forms
 from django.conf import settings
+from django.core.paginator import Paginator
 from django.utils.safestring import mark_safe
 
 from pootle.i18n.gettext import ugettext as _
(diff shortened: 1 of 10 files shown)


pootle/core/forms.py
```diff
@@ -154,23 +155,162 @@ def clean(self):
...
+    select_all_field = "select_all"
     actions = forms.ChoiceField(
         required=False,
         label=_("With selected"),
         widget=forms.Select(attrs={'class': 'js-select2'}),
-        choices=action_choices)
+        choices=(
+            ("", "----"),
+            ("reject", "Reject"),
+            ("accept", "Accept")))
```
unho: Choice labels should be translatable.

pootle/core/forms.py
```diff
@@ -154,23 +155,162 @@ def clean(self):
...
+    comment = forms.CharField(
+        label=_("add comment"),
```
unho: Nit: Start with capital letter: `Add comment`.

pootle/static/js/common.js
```diff
@@ -142,7 +142,40 @@ PTL.common = {
       }
     });
 
+    const updateFormtable = ($formtable) => {
+      $formtable.find('.js-formtable-messages').each(function () {
+        $(this).show();
+        $(this).find('ul.messages').each(function () {
+          $(this).html('<li>Updating data</li>');
```
unho: The message should be translatable.

pootle/core/forms.py
phlax: in this case it wants to be lower in the actual view - but ill update and make it lower in the template

unho:
lgtm after nits

PR #5405, opened 2016-10-19
Revision hashes

phlax:
Add a model for storing revision hashes or numbers against either the site, a model or an instance of a model

also can have different keys for above


pootle/apps/pootle_data/directory_data.py
```diff
@@ -21,7 +21,7 @@ class DirectoryDataTool(RelatedStoresDataTool):
 
     @property
     def context_name(self):
-        return self.context.pootle_path
+        return self.context.pootle_path.strip("/").replace("/", ".")
```
ta2-1: can you recall why it is necessary? 
`/af/prj0/dir.dir/` can't be distinct from `/af/prj/dir/dir/`

pootle/apps/pootle_data/project_data.py
```diff
@@ -20,22 +20,35 @@ def filter_data(self, qs):
...
 class ProjectSetDataTool(RelatedTPsDataTool):
     group_by = ("tp__project__code", )
+    cache_key_name = "projects"
 
     def get_root_child_path(self, child):
         return child[self.group_by[0]]
+
+    @property
+    def context_name(self):
+        return "ALL"
```
ta2-1: why not `projects`?

pootle/apps/pootle_data/utils.py
```diff
@@ -274,9 +274,11 @@ def aggregate_max_fields(self):
             for f
             in self.max_fields)
 
-    @property
+    @persistent_property
```
ta2-1: I see that revision updater in store_data save handler can update "checks" however I couldn't find where it is used.
IIUC persistent_property for checks should use "checks" key in rev_cache_key though I could misunderstand.

pootle/apps/pootle_data/directory_data.py
phlax: fair point ill revert

pootle/apps/pootle_data/utils.py
phlax: as they are updated together - removed the checks key for now

pootle/apps/pootle_data/project_data.py
phlax: the cache_key_name is projects

unho:
lgtm

PR #5412, opened 2016-10-20
Cleanup: remove data_tool.update calls from tests

phlax:
these calls to update should have been removed already as they now happen automatically

### DIFF
--- a/tests/pootle_data/data_updater_store.py
+++ b/tests/pootle_data/data_updater_store.py
@@ -62,8 +62,6 @@ def test_data_store_util_wordcount(store0):
     update_data = store0.data_tool.updater.get_store_data()
     for k in WORDCOUNT_KEYS:
         assert update_data[k] == updated_stats[k]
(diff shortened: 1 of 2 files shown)


unho:
I would like to have the explanation on the commit message for triggering `unit.save`. Other than that lgtm.

pootle/apps/pootle_store/utils.py
```diff
@@ -104,9 +104,7 @@ def add(self, unit, translation, user=None, touch=True,
                 mt_similarity=mt_similarity,
             )
             sub.save()
-
-            if touch:
-                unit.save()
+            unit.save()
```
dwaynebailey: I think you can safely remove touch paramater and docstring

phlax: actually it seems not i think - i think its there for when a suggestion is added during eg unit.update

it would be good to rethink this pattern a bit - but for now lets keep

PR #5419, opened 2016-10-21
Cleanup opti

phlax:
optimizations for browse views

### DIFF
--- a/pootle/apps/virtualfolder/utils.py
+++ b/pootle/apps/virtualfolder/utils.py
@@ -8,6 +8,8 @@
 
 from fnmatch import fnmatch
 
+from django.db.models import Max
+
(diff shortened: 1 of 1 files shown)


pootle/apps/virtualfolder/utils.py
```diff
@@ -195,6 +197,14 @@ class DirectoryVFDataTool(RelatedStoresDataTool):
     def context_name(self):
         return self.context.name
 
+    @property
+    def max_unit_revision(self):
+        # todo use through table for other stats?
```
unho: Please lets not add TODO notes referring to other code, instead better create GH issues. TODO notes re following code are acceptable, but this is not the case.

unho:
lgtm

PR #5421, opened 2016-10-21
Move to use setuptools extras_install

dwaynebailey:
Add a number of extra_installs and adapt documentation accordingly.


unho:
lgtm

PR #5425, opened 2016-10-22
Fix: send changed suggestion in suggestion feedback form

phlax:
this is a partial revert of c9632623 which appears to have broken the
feedback form.

the fix is probably not the best, but it works


unho:
Reviewing and tested, gtm.

PR #5426, opened 2016-10-22
Fix: add suggestion view uses form.cleaned_data

phlax:
as in title


unho:
Reviewed and tested. GTM

PR #5427, opened 2016-10-23
Remove permission caching

phlax:
This is insecure, and extremely inefficient and precludes the possibility
of running Pootle as any kind of distributed service

also adds various view optimizations


pootle/apps/pootle_translationproject/views.py
```diff
@@ -240,15 +239,19 @@ def get_context_data(self, *args, **kwargs):
...
+    def can_upload(self):
+        return (
             "import_export" in settings.INSTALLED_APPS
             and self.request.user.is_authenticated()
-            and check_permission('translate', self.request))
-        if has_upload:
-            if "po" in self.project.filetype_tool.valid_extensions:
-                ctx.update(handle_upload_form(self.request, self.tp))
+            and (self.request.user.is_superuser
+                 or "translate" in self.request.permissions
```
ta2-1: user with translate permissions only shouldn't be able to upload as different user. it looks like you don't differ translate and administrate here.

phlax: this is just the permission to show the widget

pootle/apps/pootle_translationproject/views.py
```diff
@@ -240,15 +239,19 @@ def get_context_data(self, *args, **kwargs):
...
 
-    def get_upload_widget(self):
-        ctx = {}
-        has_upload = (
+    @property
+    def can_upload(self):
+        return (
             "import_export" in settings.INSTALLED_APPS
             and self.request.user.is_authenticated()
-            and check_permission('translate', self.request))
```
ta2-1: I don't follow why you needed to replace this.

pootle/apps/pootle_translationproject/views.py
ta2-1: yes, I see . thanks. though there is another comment https://github.com/translate/pootle/pull/5427/commits/e0f0fe485d70497c271f2996d3cf14c5766b4ac5#r84593485

pootle/apps/pootle_translationproject/views.py
ta2-1: hmm... sorry, I don't follow where...
```
def check_permission(permission_codename, request):
    """Checks if the current user has `permission_codename`
    permissions.
    """
    if request.user.is_superuser:
        return True

    # `view` permissions are project-centric, and we must treat them
    # differently
    if permission_codename == 'view':
        path_obj = None
        if hasattr(request, 'translation_project'):
            path_obj = request.translation_project
        elif hasattr(request, 'project'):
            path_obj = request.project

        if path_obj is None:
            return True  # Always allow to view language pages

        return path_obj.is_accessible_by(request.user)

    return ("administrate" in request.permissions or
            permission_codename in request.permissions)
```

phlax: hmm fair point but this method is full of implict wierdness - ill change the comit message

pootle/apps/accounts/managers.py
```diff
@@ -97,8 +101,8 @@ def get_users_with_permission(self, permission_code, project, language):
             permissionset__positive_permissions__codename=permission_code
         )
 
-        language_path = language.directory.pootle_path
-        project_path = project.directory.pootle_path
+        language_path = "/%s/" % language.code
+        project_path = "/projects/%s/" % project.code
```
unho: `project.pootle_path` does the same: https://github.com/translate/pootle/blob/master/pootle/apps/pootle_project/models.py#L360-L362

unho:
lgtm after nits

PR #5431, opened 2016-10-23
User optimizations

phlax:
- language team caches permissions
- tp view adds select_related
- indeces on User
- upload widget uses lang team


unho:
It looks ok but Travis fails. Other than that lgtm.

PR #5436, opened 2016-10-24
Fix for stats on /projects/ page

phlax:
The stats are on the /projects/ page do not get invalidated - this fixes that.

also adds tests, and tests for additional cache keys, and similar tests for vfolder data_tools


unho:
lgtm

ta2-1:
lgtm

PR #5440, opened 2016-10-24
Rewrite of hacking docs to use installed Pootle

dwaynebailey:
Drop `./manage.py` and use `pootle` in dev docs i.e. consoldidate into one approach to running, hacking and deploying Pootle.


unho:
lgtm

PR #5445, opened 2016-10-25
Remove other manage.py references

dwaynebailey:
This remove almost all uses of manage.py

Except:

* tox.ini - if we use 'pootle' there then we don't seem to get the benefit of `91-travis.conf`
* Makefile - we should probably warn and exit if someone has not installed Pootle, but didn't want to block this for that.

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -17,7 +17,8 @@ will run:
 .. note::
 
   If you run Pootle from a repository checkout you can use the *manage.py* file
-  found in the root of the repository.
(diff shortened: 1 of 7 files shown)


docs/server/commands.rst
```diff
@@ -17,7 +17,8 @@ will run:
 .. note::
 
   If you run Pootle from a repository checkout you can use the *manage.py* file
-  found in the root of the repository.
+  found in the root of the repository.  Note that this model is deprecated and
```
unho: this model? this approach?

unho:
Nit aside, lgtm

Issue #5442, opened 2016-10-25
Use dict rather than RequestContext in render method


phlax:
something in pootle_store.views is triggering this warning

```
RemovedInDjango110Warning: render() must be called with a dict, not a RequestContext.
  RequestContext(self.request, context))
```

dwaynebailey:
Fixed in 1a1fb2ef27d34cf262bd617f73c57f1a7a0258f1

dwaynebailey:
There are some other cases fixed in the Django 1.10 cleanup, so lets keep this open for now

dwaynebailey:
It seems this is the one I was waiting for 87612a425a35b320a11470c76a58f7f88b519255 in `render_to_string`

PR #5446, opened 2016-10-25
Django 1.8: Template.render() expects  context=dict()

dwaynebailey:
This address #5442 

".. provide a render() method, however, the former takes a
django.template.Context as an argument while the latter expects a dict."

That is we need to move to use a context dict and not a Context
instance.

See
https://docs.djangoproject.com/en/1.10/releases/1.8/#rendering-templates-loaded-by-get-template-with-a-context


unho:
lgtm

PR #5449, opened 2016-10-26
Fix Pootle FS path processing to run on Windows

lawlor:


unho:
It looks like all the commits in this PR have been landed. Closing.

PR #5451, opened 2016-10-27
SubstringBase: reimplement old Django autoassignment process

### LINKED ISSUE #5268: Remove SubfieldBase from MultiStringField
In django 1.9 SubFieldBase is deprecated, and removed in 1.10

Related Stackoverflow - http://stackoverflow.com/questions/35166085/how-to-deal-with-subfieldbase-has-been-deprecated-use-field-from-db-value-inst

https://docs.djangoproject.com/en/1.9/ref/models/fields/#field-api-reference

afaict we can just safely remove - it seems it just needs to have to/from db methods

dwaynebailey:
Previously SubfieldBase would use to_python() on assignment to a custom
field.  With the removal of SubfieldBase this automagic no longer
happens.

We followed http://stackoverflow.com/a/39471064

Essentially we reimplement the dropped functionality and ensure that our
MultiStringField uses to_python() when values are assinged to it.

Fixes #5268

### DIFF
--- a/pootle/apps/pootle_store/fields.py
+++ b/pootle/apps/pootle_store/fields.py
@@ -48,10 +48,30 @@ def to_python(value):
         return multistring(value, encoding="UTF-8")
 
 
+class CastOnAssignDescriptor(object):
+    """
+    A property descriptor which ensures that `field.to_python()` is called on
+    _every_ assignment to the field.  This used to be provided by the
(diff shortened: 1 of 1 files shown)


unho:
I cannot say I understand the magic, but on test the annoying warning is finally gone. Nice work!

PR #5454, opened 2016-10-31
Rename translation_path to translation_mapping

phlax:
trying to reduce confusion about the role of the translation_mapping


dwaynebailey:
lgtm

PR #5455, opened 2016-10-31
Translation mapping presets

phlax:
Similar to the old project treestyle presets, this PR add presets to the fs ui translation mapping widget

### DIFF
--- a/pootle/apps/pootle_fs/forms.py
+++ b/pootle/apps/pootle_fs/forms.py
@@ -18,6 +18,17 @@ from .delegate import (
     fs_plugins, fs_translation_mapping_validator, fs_url_validator)
 
 
+FS_CHOICES = (
+    ("gnu", _("GNU-style"), "/po/<language_code>.<ext>"),
+    ("non-gnu",
+     _("non GNU-style"),
+     "/<language_code>/<dir_path>/<filename>.<ext>"),
+    ("django",
+     _("Django-style"),
+     "/locale/<lang_code>/LC_MESSAGES/<filename>.<ext>"),
+    ("custom", _("Custom"), ""))
+
+
 class ProjectFSAdminForm(forms.Form):
 
     fs_type = forms.ChoiceField(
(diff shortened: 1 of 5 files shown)


pootle/apps/pootle_fs/forms.py
```diff
@@ -18,6 +18,13 @@
     fs_plugins, fs_translation_mapping_validator, fs_url_validator)
 
 
+FS_CHOICES = (
+    ("gnu", "GNU-style", "/po/<language_code>.<ext>"),
+    ("non-gnu", "non GNU-style", "/<language_code>/<dir_path>/<filename>.<ext>"),
+    ("django", "Django-style", "/locale/<lang_code>/LC_MESSAGES/<filename>.<ext>"),
+    ("custom", "Custom"))
```
dwaynebailey: I'd say these should be localizable, not sure if it needs `lazy_gettext`

pootle/apps/pootle_fs/forms.py
```diff
@@ -30,9 +41,18 @@ class ProjectFSAdminForm(forms.Form):
...
+        required=False,
+        choices=(
+            [("", "-----"), ]
+            + [(x[0], x[1]) for x in FS_CHOICES]),
+        widget=forms.Select(
+            attrs={'class': 'js-select2 js-select-fs-mapping'}))
     translation_mapping = forms.CharField(
         help_text=_(
-            "The translation mapping for your filesystem"))
+            "The translation mapping for your filesystem"),
```
unho: Nit: I think this fits in the previous line.

unho:
Nits aside lgtm. Is there any live server to test?

pootle/apps/pootle_fs/forms.py
```diff
@@ -18,6 +18,17 @@
     fs_plugins, fs_translation_mapping_validator, fs_url_validator)
 
 
+FS_CHOICES = (
+    ("gnu", _("GNU-style"), "/po/<language_code>.<ext>"),
```
phlax: i think as its an example - lets use `po` - then users can use custom if they need to change

PR #5457, opened 2016-11-01
setup: ensure full PEP440 support

dwaynebailey:
pip and setuptools only fully support PEP440 following certain versions
and on some platforms they use older versions.  So we need to test that
we have the correct versions.

We are making use of the PEP440 ~= syntax because its clearer and less
cluttered, but that means that these deps should be upgraded if they are
too old.

We also need to catch these requirements before 'install_requires' and
we can't specify them in 'requirements/base.txt' because older tools
cannot parse those.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,10 +15,20 @@ from distutils.command.build import build as DistutilsBuild
 from distutils.core import Command
 from distutils.errors import DistutilsOptionError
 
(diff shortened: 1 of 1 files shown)


dwaynebailey:
@phlax this would have caught your install issue from yesterday

unho:
lgtm

PR #5464, opened 2016-11-02
Tp path

phlax:
** This PR is part of several to do the following **

    add tp_path to stores and dirs with just the tp_path bit

    add tp fk to all dirs

    make tp_id,tp_path unique_together

    optimize queries to use above

    gradually remove pootle_path as an attribute

    remove all dirs above tp level


unho:
lgtm

PR #5465, opened 2016-11-03
Appveyor CI for Windows

dwaynebailey:
Goal:

* Ensure Pootle and tests are runnable on Windows and stay that way, so that
* Developers on Windows can participate, work on and build Pootle

Work done:

* Add Appveyor CI for Windows testing
* Implement a number of minor changes to paths
* Implement quite a few abstractions in tests, but not complete (these could be done more cleanly I think)
* Disable any failing check related to Pootle FS only in Windows

Caveats:

* Pootle FS testing has mostly been disabled on Windows as there is a much deeper inspection needed to ensure that:

  - Local paths are indeed local and FS dependent.
  - translation_mapping are not FS dependent but are transformed when used on a local FS that needs a different structure.

### DIFF
--- a/tests/commands/pootle_runner.py
+++ b/tests/commands/pootle_runner.py
@@ -20,7 +20,11 @@ def test_pootle_noargs(capfd):
     """Pootle no args should give help"""
     call(['pootle'])
     out, err = capfd.readouterr()
-    assert "Type 'pootle help <subcommand>'" in out
+    # Expected:
+    #   Type 'pootle help <subcommand>'
+    # but 'pootle' is 'pootle-script.py' on Windows
+    assert "Type 'pootle" in out
+    assert " help <subcommand>'" in out
 
 
 @pytest.mark.cmd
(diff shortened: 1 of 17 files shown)


tests/commands/pootle_runner.py
```diff
@@ -20,7 +20,11 @@ def test_pootle_noargs(capfd):
     """Pootle no args should give help"""
     call(['pootle'])
     out, err = capfd.readouterr()
-    assert "Type 'pootle help <subcommand>'" in out
+    # Expcted:
```
unho: Typo

tests/commands/pootle_runner.py
```diff
@@ -20,7 +20,11 @@ def test_pootle_noargs(capfd):
...
+    #   Type 'pootle help <subcommand>'
+    # but 'pootle' is 'pootle-script.py' on Windows
```
unho: I guess we need to revisit this on a separate PR. TTK required to append the `.py` on previous versions, but on 2.0.0 it doesn't need it anymore.

unho:
Typos aside, lgtm

PR #5469, opened 2016-11-04
Add tp_path to Store

phlax:
see #5464 for longer explanation


pootle/apps/pootle_store/migrations/0021_set_tp_path.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def set_store_tp_path(apps, schema_editor):
+    stores = apps.get_model("pootle_store.Store").objects.all()
+    for store in stores:
+        parts = store.pootle_path.lstrip("/").split("/")
+        store.tp_path = '/'.join([""] + parts[2:])
```
unho: Nit, do not use `lstrip`:

```python
        parts = store.pootle_path.split("/")
        store.tp_path = '/'.join([""] + parts[3:])
```

unho:
Nit aside lgtm

pootle/apps/pootle_store/migrations/0021_set_tp_path.py
phlax: its necessary - else the `"/".join` adds another

PR #5470, opened 2016-11-04
Path opti

phlax:
adds indices for directory and store on tp_path

uses tp_paths in project.resources


pootle/apps/pootle_app/models/directory.py
```diff
@@ -80,6 +80,7 @@ class Meta(object):
         default_permissions = ()
         app_label = "pootle_app"
         index_together = ["obsolete", "pootle_path"]
+        unique_together = ["obsolete", "tp", "tp_path"]
```
unho: Can it be the same tp-tp_path combination be at the same time obsolete and not obsolete? If not I would drop `obsolete` from here.

pootle/apps/pootle_store/models.py
```diff
@@ -998,7 +998,9 @@ class Store(models.Model, CachedTreeItem, base.TranslationStore):
 
     class Meta(object):
         ordering = ['pootle_path']
-        unique_together = ('parent', 'name')
+        unique_together = [
+            ('parent', 'name'),
+            ["obsolete", "translation_project", "tp_path"]]
```
unho: I have the same doubt re `obsolete`.

Nit: Use a tuple, since as you can see it is what is being used in the migration.

pootle/apps/pootle_app/models/directory.py
```diff
@@ -97,22 +96,19 @@ def code(self):
...
+        try:
             return self.translationproject
-
-        aux_dir = self
-        while not aux_dir.is_translationproject() and aux_dir.parent is not None:
-            aux_dir = aux_dir.parent
-
-        return aux_dir.translationproject
+        except TranslationProject.DoesNotExist:
+            return
```
unho: `return None`? I would say make it explicit.

phlax: thats pointless - its a fundamental part of python

pootle/apps/pootle_app/models/directory.py
unho: ok

unho:
Nits aside, lgtm

PR #5471, opened 2016-11-04
A grab bag of static pages fixes

### LINKED ISSUE #5467: Static page editing save HTML outcome instead of raw text
To reproduce:

1. Server using Markdown for static pages and an existing static page written using Markdown
2. Minor edit the content of the static page (editing the title also seems to do this)
3. Saves fine
4. Edit the page again.  Static page text is now all in HTML. 
5. View static page on site.  Static page is now all garbled HTML

julen:
This PR includes a few fixes related to static pages:

* Adjusts a deprecated method call.
* Uses an `StaticPage` object directly to pass initial static page values instead of relying on form's internal attributes.
* Addresses the backwards-incompatible change by mplementing `to_python` in `MarkupField`.

Please check each commit message for further details.

Fixes #5467.

### DIFF
--- a/pootle/core/markup/fields.py
+++ b/pootle/core/markup/fields.py
@@ -120,7 +120,7 @@ class MarkupField(models.TextField):
         return value
 
     def value_to_string(self, obj):
-        value = self._get_val_from_obj(obj)
+        value = self.value_from_object(obj)
         return self.get_prep_value(value)
 
     def formfield(self, **kwargs):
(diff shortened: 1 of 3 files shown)


unho:
lgtm

dwaynebailey:
@julen tested this on a local install but it still seems to mess with data if it is written in Markdown.

dwaynebailey:
@julen https://github.com/translate/pootle/pull/5471#issuecomment-258487497 is caused by bleach having a very minimal set of allowed HTML tags.  Will PR that fix on its own.

Otherwise gtm

Issue #4180, opened 2015-11-12
Placeholder text is too wide for the input area


iafan:
When editing announcement pages, placeholder text for the `Virtual Path` parameter is not fully visible:

![screenclip 3](https://cloud.githubusercontent.com/assets/1728158/11130625/dfb77886-893a-11e5-973f-a23016e1aec4.png)

Let's increase the size of the input field to accommodate the placeholder text (mind that translations can be even longer).

unho:
The `virtual_path`  [model field ](https://github.com/translate/pootle/blob/master/pootle/apps/staticpages/models.py#L37) is 100 characters long. My guess is that setting 100 chars would the best way forward.

Also, as raised by some other people, the placeholder disappears as soon as you start typing, so it will be better to add a help text with the hint.

PR #5480, opened 2016-11-06
Resize virtual path field

prasastoadi:
Increase the size of virtual path input field as stated here https://github.com/translate/pootle/issues/4180


phlax:
hi @prasastoadi as there is already an open bug for this im gonna close as a duplicate

jayvdb:
@prasastoadi "codecov" will be much happier if you can add a unit test for `PageCreateView.get_form`

unho:
Note the placeholder text is translatable, so the length increase can be considerably larger.

unho:
This form field is for  [this model field ](https://github.com/translate/pootle/blob/master/pootle/apps/staticpages/models.py#L37) which is 100 characters long. My guess is that setting 100 chars would be enough.

unho:
Landed in https://github.com/translate/pootle/commit/262981d87d8c404a115ef6a0b0d2c2d7a324d9dc. Thanks for your contribution!

PR #5484, opened 2016-11-07
Add instructions for Windows dev environment setup

r-o-b-b-i-e:
I have added instructions for Windows environment setup to the hacking.rst document, alongside the normal dev environment setup instructions under a different header.

These are slightly 'fluid', in that some of the commands worked for some of us but not others, but where there are divergences and possible ambiguities I have tried to include explanations of what's going on and what alternatives to use.


docs/developers/hacking.rst
```diff
@@ -192,6 +192,165 @@ command) and running the :djadmin:`pootle runserver <runserver>` command.
...
+    (venv)> pip install C:\temp\lxml-3.6.4-cp27-cp27m-win32.whl
+    (venv)> pip install C:\temp\python_Levenshtein-0.12.0-cp27-none-win32.whl
+    (venv)> pip install C:\temp\scandir-1.2-cp27-none-win32.whl
+
+Now that all the requirements are lined up, we are ready to initialise Pootle.
+
+Clone your fork of the Pootle master using your favourite Windows implementation of Git so that you have a working 
+copy somewhere accessible on your computer (e.g. ``C:\git\pootle``).
+
+First, rename ``90-dev-local.conf.sample`` to ``90-dev-local.conf`` (in ``pootle\settings``) to enable a basic
```
unho: Copying it sounds like a better idea.

unho:
This addition largely expands the size of the hacking guide, so I would suggest offloading these notes to a separate file and then link it on the hacking docs. Moving the Linux instructions to a separate file might be a good idea as well.

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+
+    # lxml
+    # levenshtein
+    # scandir
+
+These three packages are difficult to build on Windows, so we will download pre-built versions to install manually 
+further on.
+
+Also change any ``~=`` compatible-version markers into ``==`` exact-version markers, 
+as Windows-based ``pip`` doesn't seem to like these.
```
dwaynebailey: Upgrading both pip and setuptools should solve these.

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+Install Nodejs
+^^^^^^^^^^^^^^
+
+Download the latest installer from:
+    https://nodejs.org/en/
+
+Pootle Set Up
+^^^^^^^^^^^^^
+
+.. note:: For convenient these instructions consistently specify paths ``C:\venv`` and ``C:\git\pootle``, but you can change these to suit your environment and needs.
```
dwaynebailey: convenience

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+further on.
+
+Also change any ``~=`` compatible-version markers into ``==`` exact-version markers, 
+as Windows-based ``pip`` doesn't seem to like these.
+
+To avoid permissions issues, required packages have to be downloaded and installed in separate steps. 
+By default, pip stores temporary files in your ``user\AppData`` folder which may not allow access in 
+all circumstances. There are also some packages that need to be downloaded and installed manually.
+Create a temporary folder somewhere that any user can access without permission issues (e.g. ``C:\temp``).
+If you are seeing "directory was not empty" or "file not found" errors during a ``pip install``, then these
```
dwaynebailey: I'd put these following commands in a `notes` block so that they're not part of the normal flow.

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+Clone your fork of the Pootle master using your favourite Windows implementation of Git so that you have a working 
+copy somewhere accessible on your computer (e.g. ``C:\git\pootle``).
+
+First, rename ``90-dev-local.conf.sample`` to ``90-dev-local.conf`` (in ``pootle\settings``) to enable a basic 
+configuration suitable for local hacking. Then initialise your local Pootle working copy into the virtualenv:
+
+.. code-block:: console
+
+    (venv)> cd C:\git\pootle
+    (venv)> pip install -e .
```
dwaynebailey: Note we've moved off the separate `-r` step on the Linux install docs.  Not sure if that would work on Windows because of some of the build issues you hit in previous steps.

But it would make sense to do:

`pip install -e .[dev]`

That will install all dev requirements.

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+
+At this point you should be able to initialise the Pootle demo database as normal:
+
+.. code-block:: console
+
+    (venv)> python manage.py migrate
+    (venv)> python manage.py initdb
```
dwaynebailey: Note that once you've done all `pip install .` above you actually have access to the `pootle` runner so you could do `pootle migrate` and `pootle initdb` but if you do that then the `cp 90-*` would need to be `pootle init` and your `90*` file would be in `~\.pootle\pootle.conf`

docs/developers/hacking.rst
```diff
@@ -188,6 +188,156 @@ virtual environment (using the :command:`workon` command) and running the
...
+
+Next, you will need to deactivate the virtual environment in order to set up the client-side bundles with NPM.
+
+.. code-block:: console
+
+    (venv)> deactivate
+    C:\git\pootle> cd pootle\static\js
+    C:\git\pootle\pootle\static\js> npm install
```
dwaynebailey: Why not just do that in another terminal?

r-o-b-b-i-e:
Okay, I have made a load of changes based on everyone's feedback:
- Split the setup instructions into two separate files to avoid the one page getting overly massive. Both are linked from the main Hacking guide and added to the Developers index page table of contents
- Moved the step about cloning your repo to be one of the earliest steps
- Changed the instruction to rename the 90-dev-local.conf.sample to suggest copying it instead
- Moved the download-and-build version of the pip requirements installation into a 'notes' block, so that anybody following the instructions will try the 'proper' command before breaking out of the normal flow to try the extended version
- Changed the instructions for pootle commands to use the 'pootle' runner, with a note block about using manage.py as an alternative if the runner doesn't work for some Windows reason
- Fixed a load of typos
- Squashed all the commits into one

unho:
lgtm

dwaynebailey:
@r-o-b-b-i-e thanks for this :)

PR #5485, opened 2016-11-08
Mixed bag of cleanups

phlax:
as in title

### DIFF
--- a/pootle/apps/pootle_data/store_data.py
+++ b/pootle/apps/pootle_data/store_data.py
@@ -27,7 +27,6 @@ class StoreDataTool(DataTool):
         stats = {
             v: getattr(self.context.data, k)
             for k, v in self.stats_mapping.items()}
-        stats["is_dirty"] = False
         stats["children"] = {}
         stats["lastaction"] = (
             self.context.data.last_submission
--- a/pootle/apps/pootle_misc/templatetags/common_tags.py
+++ b/pootle/apps/pootle_misc/templatetags/common_tags.py
@@ -15,10 +15,10 @@ from django.contrib.auth import get_user_model
 register = template.Library()
 
 
-@register.inclusion_tag('browser/_table.html', takes_context=True)
-def display_table(context, table):
+@register.inclusion_tag('browser/_table.html')
+def display_table(table, can_translate):
     return {
-        'can_translate': context["can_translate"],
+        'can_translate': can_translate,
         'table': table,
     }
 
--- a/pootle/apps/pootle_translationproject/models.py
+++ b/pootle/apps/pootle_translationproject/models.py
@@ -277,6 +277,9 @@ class TranslationProject(models.Model, CachedTreeItem):
         else:
             self.abs_real_path = None
         super(TranslationProject, self).save(*args, **kwargs)
+        if self.directory.tp_id != self.pk:
+            self.directory.tp = self
+            self.directory.save()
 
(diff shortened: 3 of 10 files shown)


pootle/apps/pootle_data/store_data.py
```diff
@@ -27,7 +27,6 @@ def object_stats(self):
         stats = {
             v: getattr(self.context.data, k)
             for k, v in self.stats_mapping.items()}
-        stats["is_dirty"] = False
```
unho: `is_dirty` appears in tests, not strictly related to this. Maybe other appearances of this can be dropped as well.

pootle/apps/pootle_translationproject/models.py
```diff
@@ -277,6 +277,9 @@ def save(self, *args, **kwargs):
         else:
             self.abs_real_path = None
         super(TranslationProject, self).save(*args, **kwargs)
+        if self.directory.tp_id != self.pk:
+            self.directory.tp = self
+            self.directory.save()
```
unho: Adding a migration for this would be helpful. I can contribute it.

pootle/apps/pootle_data/store_data.py
phlax: yep - they need to be - just done that

pootle/apps/pootle_translationproject/models.py
phlax: there is one already

pootle/apps/pootle_statistics/models.py
```diff
@@ -267,7 +267,6 @@ def get_submission_info(self):
             "displayname": displayuser.display_name,
             "username": displayuser.username,
             "display_datetime": dateformat.format(self.creation_time),
-            "iso_datetime": self.creation_time.isoformat(),
```
unho: Is this not being used in JavaScript code?

pootle/apps/pootle_translationproject/models.py
unho: There is one for when you added the field, but if a deployment has current master (without this PR) and ran that migration and then added a new TP, then that new TP won't have the tp in the directories newly added. Thus another migration is needed, even if it is repeating the same thing.

phlax: k - there is the remotest chance of a tp.directory which doesnt have the tp set

if you want to add a migration for this please do

phlax: i mean in almost every situation the directory would have been created - but testing brought out a situation where it wouldnt

pootle/apps/pootle_misc/templatetags/common_tags.py
```diff
@@ -15,10 +15,21 @@
 register = template.Library()
 
 
-@register.inclusion_tag('browser/_table.html', takes_context=True)
-def display_table(context, table):
+@register.inclusion_tag('browser/_progressbar.html')
+def progress_bar(total, fuzzy, translated):
```
unho: This is not being used anywhere.

unho: k

unho:
lgtm

pootle/apps/pootle_translationproject/models.py
unho: I will be adding a migration for this.

PR #5486, opened 2016-11-08
Increase bleach ALLOWED_TAGS to ensure Markdown renders correctly

dwaynebailey:
This builds on https://github.com/translate/pootle/pull/5471 and actually requires that to addres the MarkdownField issue

### DIFF
--- a/docs/server/settings.rst
+++ b/docs/server/settings.rst
@@ -229,6 +229,12 @@ Configuration settings for applications used by Pootle.
 
     POOTLE_MARKUP_FILTER = ('markdown', {})
 
+    POOTLE_MARKUP_FILTER = ('markdown', {
+                                'clean': {
+                                    'extra_tags': ['div'],
+                                },
+                            })
+
     POOTLE_MARKUP_FILTER = ('restructuredtext', {
                                 'settings_overrides': {
                                     'report_level': 'quiet',
--- a/pootle/core/markup/filters.py
+++ b/pootle/core/markup/filters.py
@@ -102,8 +102,8 @@ def apply_markup_filter(text):
           dictionary; some arguments may still be inferred as needed,
           however.
 
-    So, for example, to use Markdown with safe mode turned on (safe
-    mode removes raw HTML), put this in your settings file::
+    So, for example, to use Markdown with bleach cleaning turned on (cleaning
+    removes non-whitelisted HTML), put this in your settings file::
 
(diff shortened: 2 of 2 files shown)


docs/server/settings.rst
```diff
@@ -229,6 +229,12 @@ Configuration settings for applications used by Pootle.
 
     POOTLE_MARKUP_FILTER = ('markdown', {})
 
+    POOTLE_MARKUP_FILTER = ('markdown', {
+                                'clean': {
+                                    'allowed_tags': ['p', 'h1', 'h2'],
```
unho: IMHO it makes sense to add this as the default example in settings files.

dwaynebailey: Nope, since by default I don't actually want anyone to set anything

unho:
Other than nits, lgtm.

docs/server/settings.rst
unho: After checking the code below I agree it is ok.

pootle/core/markup/filters.py
```diff
@@ -133,7 +133,23 @@ def apply_markup_filter(text):
...
+                u'p', u'pre',
+                u'img',
+                u'hr',
+            ]
+
+            tags = MARKDOWN_DEFAULT_TAGS
+            if 'clean' in markup_kwargs:
+                if 'allowed_tags' in markup_kwargs['clean']:
+                    tags = markup_kwargs['clean']['allowed_tags']
+            html = bleach.clean(markdown.markdown(text, **markup_kwargs),
```
dwaynebailey: No, I specifically wanted someone to be on their own if they want to control the allowed tags.  Otherwise there is no way for them to fully override things.

pootle/core/markup/filters.py
```diff
@@ -133,7 +133,22 @@ def apply_markup_filter(text):
         elif markup_filter_name == 'markdown':
             import bleach
             import markdown
-            html = bleach.clean(markdown.markdown(text, **markup_kwargs))
+
+            # https://github.com/mozilla/bleach/blob/master/bleach/__init__.py#L27-L40
```
unho: This link will probably point to the wrong stuff quite soon. I suggest:

```python
            # Check for ALLOWED_TAGS in https://github.com/mozilla/bleach/blob/master/bleach/__init__.py
```

unho:
Other than nit just raised, lgtm.

PR #5487, opened 2016-11-08
Add dependency links for non-pypi, usually github dependencies

dwaynebailey:
This replaces https://github.com/translate/pootle/pull/5482

It works but feels like I could clean things up and the dependency_links function should also recurse.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,6 +60,15 @@ def parse_requirements(file_name, recurse=False):
     return requirements
 
 
+def parse_dependency_links(file_name):
+    dependency_links = []
+    for line in open(file_name, 'r').read().split('\n'):
+        if re.match(r'\s*-e\s+', line):
+            dependency_links.append(re.sub(r'\s*-e\s+', '', line))
+
+    return dependency_links
+
+
 class PyTest(TestCommand):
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -60,6 +60,15 @@ def parse_requirements(file_name, recurse=False):
     return requirements
 
 
+def parse_dependency_links(file_name):
+    dependency_links = []
+    for line in open(file_name, 'r').read().split('\n'):
+        if re.match(r'\s*-[ef]\s+', line):
```
dwaynebailey: Don't think we want `-f`

ta2-1: @dwaynebailey I trust you here

setup.py
```diff
@@ -51,7 +51,7 @@ def parse_requirements(file_name, recurse=False):
             continue
 
         if re.match(r'\s*-e\s+', line):
-            requirements.append(re.sub(r'\s*-e\s+.*#egg=(.*)$', r'\1', line))
+            requirements.append(re.sub(r'\s*-e\s+.*#egg=([^\d]*)-([\.\d]*)$', r'\1==\2', line))
```
ta2-1: line is too long

PR #5491, opened 2016-11-09
Fix: filter data in tp data tool correctly

phlax:
since switching to tp_path, the tp data tool was not filtering correctly


unho:
lgtm

PR #5493, opened 2016-11-09
Inline checks

phlax:
render check data inline rather than loading with xhr


pootle/static/js/editor/app.js
```diff
@@ -1795,11 +1794,11 @@ PTL.editor = {
 
   /* Gets the failing check options for the current query */
   getCheckOptions() {
-    StatsAPI.getChecks(this.settings.pootlePath)
```
unho: `getChecks` can be dropped as well.

phlax: yep

unho:
Reviewed and tested. lgtm

PR #5500, opened 2016-11-10
Some utilties

phlax:
- timesince util for getting nice formatted time delta - uses babel
- StatsDisplay view class

### DIFF
--- a/pootle/core/views/display.py
+++ b/pootle/core/views/display.py
@@ -7,7 +7,11 @@
 # AUTHORS file for copyright and authorship information.
 
 from django.utils.functional import cached_property
+from django.utils.html import escape, format_html
+from django.utils.safestring import mark_safe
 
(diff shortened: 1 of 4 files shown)


unho:
lgtm

pootle/core/views/display.py
```diff
@@ -33,3 +37,89 @@ def checks_by_category(self):
...
+            source=format_html(
+                u"<a href='{}'>{}</a>",
+                action["unit_url"],
+                escape(action["unit_source"])))
+        if action.get("check_name"):
+            params["check"] = format_html(
+                u"<a href='{}'>{}</a>",
+                action["checks_url"],
+                escape(action["check_display_name"]))
+        if (action["type"] == 2):
```
ta2-1: @phlax can you use constants instead of numbers here at least?

PR #5506, opened 2016-11-11
Fix: dont calculate percentages for objects with no units/data

phlax:
if you initalize a tp with no stores the view breaks, i assume it would do similar on other views if the totals are 0


unho:
lgtm

Issue #3360, opened 2014-11-19
Abort old async requests


julen:
In the admin app pending asynchronous requests are not taken into account, therefore under network conditions with some latency the app's behavior can be undefined. Pending requests need to be aborted before repeating a method to a resource.

PR #5511, opened 2016-11-11
Remove backbone js (part 1)

phlax:
backbone is hardly used and is a massive lib

need to check js templates for use of store - esp tm templates


unho:
Reviewed and tested. Gtm

julen:
FYI: safesync is used in the admin pages, see #3360 for reference.

PR #5518, opened 2016-11-12
Fix: not all stores have a count

phlax:
this prevents an error triggered if a store has no count


unho:
Reviewed and tested. Gtm

PR #5525, opened 2016-11-12
Optimization: cache browse tables

phlax:
cache browse tables use rev_cache_key


ta2-1:
lgtm after travis is green

unho:
gtm after travis is green too

PR #5526, opened 2016-11-12
pluralize_sugg only returns 3 arguments

dwaynebailey:
Picked from Django 1.10 migration.  Seems only in 1.10 does this fail, but it's been wrong for quite some time.  Not sure what or how this impact plural suggestions in the UI

### DIFF
--- a/pootle/templates/editor/units/edit.html
+++ b/pootle/templates/editor/units/edit.html
@@ -333,7 +333,7 @@
                 </a>
               {% endif %}
             </div>
-            {% for i, target, diff, title in sugg|pluralize_sugg %}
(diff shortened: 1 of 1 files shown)


ta2-1:
lgtm

phlax:
lgtm

unho:
It can return 4 arguments: https://github.com/translate/pootle/blob/5228e3d89df31492b7271836f1f494edc4b47d62/pootle/apps/pootle_store/templatetags/store_tags.py#L83

unho:
I am wrong, it is always 3. lgtm

dwaynebailey:
Landed in acb3c714024efe837fcb685887d2bbf2890c7a78

PR #5530, opened 2016-11-14
Django 1.10

dwaynebailey:
This brings Django 1.10 support to Pootle.

The concerns I have are:

* c523ad38608afd0a862d89c7b4f8d1b00a39b77b - not sure this is a correct fix, though it works, needs clicky testing I think

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@
 sudo: false
 language: python
 env:
-  - TOXENV=py27-django19-sqlite PYTHONPATH=$HOME/virtualenv/python2.7.9/lib/python2.7/site-packages  TOX_TESTENV_PASSENV="PYTHONPATH"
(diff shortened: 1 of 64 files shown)


pytest_pootle/factories.py
```diff
@@ -115,7 +115,6 @@ def fullname(self):
         # returns an incrementing index relative to the tp
         return 'Project %s' % Project.objects.count()
 
-    pootle_path = factory.LazyAttribute(lambda p: "/projects/%s" % p.code)
```
phlax: >  Not needed anymore 

how come?

dwaynebailey: No frigging idea :)

I think this was one we debugged together at PooCon but can't recall the reason

phlax: wfm 8/

dwaynebailey:
Blockers

* Requires and unreleased version of allauth.  To run in Django 1.10 allauth needs these changes https://github.com/pennersr/django-allauth/commit/b63e8db0e276bf3a4c6fe4742f1843d15632b125

PR #5536, opened 2016-11-14
Req: bump django-rq pin to v0.9.4

dwaynebailey:
Fixes and issue with rqworker for Django <1.10.  Note though that we
don't seem to use any of that functionality.

Seemed time to drop the range based versions which we've kept around
through the Django 1.9 upgrade.  Lets move to a pure pin.


unho:
lgtm

PR #5549, opened 2016-11-16
Cleanup: render_to_string & render

dwaynebailey:
This deals with deprecated Django 1.8 items that disappear in Django 1.10. So sort of a forward port from the Django 1.10 migration.


phlax:
lgtm

PR #5551, opened 2016-11-16
Django1.10 cleanups

dwaynebailey:
Backport some Django 1.10 cleanups

* Add on_delete everywhere as this becomes a required param in Django 1.10 and in 1.9 in most cases was added as an optional param.
* Some test cleanups
* Ensure we use STRICT_ALL_TABLES for MySQL in tests


phlax:
lgtm

PR #5553, opened 2016-11-16
Only show checks/latest info in browse view if available

phlax:
hides the "Latest..." and failing checks widgets if there is no data

also changest "Last action" -> "Last created" for the first of these to differentiate what the data is showing

### DIFF
--- a/pootle/templates/browser/index.html
+++ b/pootle/templates/browser/index.html
@@ -43,6 +43,7 @@
             </tbody>
           </table>
         </div>
+	{% if checks %}
         <div id="js-stats-checks" class="stats-checks">
           <h3>{% trans "Failing Checks" %}</h3>
           <div class="bd">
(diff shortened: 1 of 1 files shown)


pootle/templates/browser/index.html
```diff
@@ -69,24 +70,28 @@ <h3 class="top">{% trans "Translation Statistics" %}</h3>
...
 
       <div class="summary-3-col">
+	{% if stats.last_created_unit %}
         <div id="js-last-updated-wrapper">
           <h3 class="top">{% trans "Updates" %}</h3>
           <div class="bd">
             <div class="graph"></div>
             <div class="action-wrapper">
-              <label>{% trans "Last action:" %}</label>
+              <label>{% trans "Last created:" %}</label>
```
dwaynebailey: Is this only here when a unit is created? What happens if an existing unit changes (or is that just a creation anyway)?

PR #5556, opened 2016-11-16
Windows Pootle FS path processing fixes

lawlor:
This fixes some further problems relating to path mapping in Pootle FS on Windows.
- Replacing `"/"` with `os.sep` throughout finder.py
- Use `os.path.isabs` when validating if a path is absolute
- Allow both / and \ in paths on Windows to support both file paths and relative paths (e.g. when referring to a VCS)
- xfail one further test in fs_state.py which fails on Windows
- Use `os` methods throughout finder tests to make paths OS-agnostic so some further tests now pass on Windows


dwaynebailey:
@lawlor had a quick look at this and I few concerns for me.

We need to keep translation_mapping as being unix-like paths, it seems from the test adjustments that these are being os.path.sep'd.

My thoughts on a correct fix are this:

1. translation_mapping should be untouched regardless of platform
2. os.path.sep should only be need and tested when we're writing to disk.  I tried to create a mapper class that covered that specifically.

phlax:
@lawlor not sure if you saw - i had to xfail some more tests because the regex in finder was not parsing correctly on win, and its used now to derive cache_key

lawlor:
@phlax yes I saw. What this merge now does is
- xfail an additional test in fs_state.py
- allow both / and \ in paths on Windows to allow for local file paths as well as translation mappings
- use os.path.isabs when validating local filesystem path

Re: Codecov failure, it looks like this is because the windows logic in finder.py isn't reached. Should I write an additional test for the validator to ensure both bad_chars paths are tested?

phlax:
hi @lawlor 

> Should I write an additional test for the validator to ensure both bad_chars paths are tested?

yep, i reckon.

otherwise, lgtm, with the caveat that i dont have a win env to test on atm

lawlor:
@phlax it looks like the `TranslationMappingValidator` logic is tested in `tests\pootle_fs\validators.py`, but I've no way of mocking `os.path.sep` in there, because the validator is already set up before any tests are run. Any ideas on how best to test this (or is this test worthwhile)?

lawlor:
This is still falling short on code coverage as the Windows validation in the finder isn't tested, but the logic now validates paths, both relative and on the filesystem, on Windows successfully.

PR #5564, opened 2016-11-17
Fix select_related for non-relational fields

dwaynebailey:
Issue identified in Django 1.10 porting

This looks like the last one that can be backported from the Django 1.10 migration.


phlax:
lgtm

PR #5568, opened 2016-11-17
Fix: dont get stats for children that have none

phlax:
if a child - ie project/directory/language has no stores it will
genuinely have no stats.

This PR prevents failure in that case


unho:
lgtm

PR #5571, opened 2016-11-17
Use 'pip install -e .[xxx]' method when testing

dwaynebailey:
Test the way we're expecting users to install.  Using -r was
circumventing some of our -e magic.


unho:
lgtm

PR #5572, opened 2016-11-17
Fixes for panel caching

phlax:
uses correct locale code
adds incomplete stats for vfolders

### DIFF
--- a/pootle/apps/virtualfolder/panels.py
+++ b/pootle/apps/virtualfolder/panels.py
@@ -23,7 +23,14 @@ class VFolderPanel(ChildrenPanel):
 
     @property
     def vfdata(self):
-        return self.view.vfolders_data_view
+        vfdata = self.view.vfolders_data_view
+        if not vfdata or not vfdata.table_data.get("children"):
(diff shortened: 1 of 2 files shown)


unho:
Reviewed and tested, gtm once tests pass.

pootle/apps/virtualfolder/panels.py
```diff
@@ -23,7 +23,14 @@ def children(self):
 
     @property
     def vfdata(self):
-        return self.view.vfolders_data_view
+        vfdata = self.view.vfolders_data_view
+        if not vfdata.table_data.get("children"):
```
unho: Based on https://github.com/translate/pootle/blob/3104b810399389c214fa37d3b840739eafc2728a/pootle/apps/pootle_translationproject/views.py#L319-L322 I would say:

```python
        if not vfdata or not vfdata.table_data or not vfdata.table_data.get("children"):
```

phlax: nice @unho !

lets use `if self.view.has_vfolders:`

PR #5585, opened 2016-11-21
Increase Autosize version, fixes NS_ERROR_UNEXPECTED

### LINKED ISSUE #5439: NS_ERROR_UNEXPECTED in Firefox/Windows when navigating disabled edit rows
When browsing translation units as an anonymous user (when translation textarea is disabled), one might see NS_ERROR_UNEXPECTED errors in Firefox/Windows when navigating away from a unit. This is because of the https://bugzilla.mozilla.org/show_bug.cgi?id=889376 bug.

You may want to cherry-pick [this commit](https://github.com/evernote/zing/commit/a6f91f2887827f70fbdcd359e9368af0449d1c11) that mutes this specific error.

Another way to solve this would be to patch the autosizing textarea component not to use events, or not to disable textarea in the first place (one can imitate a disabled textarea by changing its background color and using `pointer-events: none` CSS style on it). I didn't test these alternative approaches, though.

phlax:
this fixes #5439 altho im wondering why autosize is being triggered at all

also fixes search bug

### DIFF
--- a/pootle/static/js/package.json
+++ b/pootle/static/js/package.json
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "autolinker": "^0.11.2",
-    "autosize": "^3.0.17",
+    "autosize": "^3.0.18",
     "classnames": "^1.2.0",
     "codemirror": "^5.7.0",
     "diff-match-patch": "^1.0.0",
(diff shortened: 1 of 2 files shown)


unho:
lgtm

Issue #5414, opened 2016-10-20
Select2 autoopens when clearing the dropdown selection


dwaynebailey:
This is a known issues https://github.com/select2/select2/issues/3320 with a few possible solution snippets at the end of the issue.

phlax:
fixed in #5590

PR #5590, opened 2016-11-22
Fix: dont open dropdown on clear click

phlax:
fixes annoying bug on select2 widgets #5414


unho:
Reviewed and tested, gtm.

PR #5596, opened 2016-11-22
Invalidate superusers language cache

unho:
Otherwise when adding or deleting a language or TP the languages
in the dropdown are not updated until cache expires naturally.


phlax:
as a temporary fix, lgtm - but we need to rethink this caching

PR #5606, opened 2016-11-23
Special chars various changes

### LINKED ISSUE #5400: Display hint on the special chars when hovering
It would be helpful to display a hint on which special char we are hovering. This is helpful since sometimes it is not easy to distinguish some of the special chars from others, and because sometimes different chars are represented with an empty box:

chars

Insert which symbol?

unho:
- Display character name and code on hovering
- Simplify template
- Prevent duplicate characters

Fixes #5400.

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -7,6 +7,7 @@
 # AUTHORS file for copyright and authorship information.
 
 import calendar
+import unicodedata
 
 from translate.lang import data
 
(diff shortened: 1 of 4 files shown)


phlax:
lgtm - defo needs some testing

dwaynebailey:
@unho looks good.  Not sure we need the hex, but can't see it hurting anyone.

I wonder if we should catch things like TAB and such, since those have a problem in that they don't actually have any `unicodedata.name`

Otherwise lgtm

PR #5610, opened 2016-11-24
Software cache keys

phlax:
add software cache_keys and namespacing

this allows certain modules to expire related caches when software changes without the need to flush the (entire) cache manually

### DIFF
--- a/pootle/core/views/browse.py
+++ b/pootle/core/views/browse.py
@@ -63,8 +63,9 @@ class PootleBrowseView(PootleDetailView):
     @property
     def cache_key(self):
         return (
-            "%s.%s.%s.%s.%s"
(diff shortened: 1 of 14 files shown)


pootle/core/views/browse.py
```diff
@@ -61,16 +61,6 @@ def states(self):
         return states
 
     @property
-    def cache_key(self):
-        return (
-            "%s.%s.%s.%s.%s"
-            % (self.page_name,
-               self.view_name,
-               self.object.data_tool.cache_key,
-               self.show_all,
```
unho: We are loosing `show_all`. Is that OK?

phlax: thanks @unho 8)

PR #5611, opened 2016-11-24
Add pytest-warnings

dwaynebailey:
This adds a pytest plugin to manage warnings.

There are a number of warning related fixes in this PR

Also a configuration which defaults to 'error' for warning.  Known deprecation warnings are suppressed for now.  There is also a guideline of how and when to suppress.


unho:
lgtm

PR #5612, opened 2016-11-24
Move to Django 1.10 middleware

dwaynebailey:
Use new style middleware, mostly don't change anything by using the MiddlewareMixin

Only last commit, the rest is #5611


unho:
lgtm

PR #5614, opened 2016-11-24
Fix: aggregate headline stats correctly for admins

phlax:
this resolves a bug where admins where seeing double for headline stats in tp directories


unho:
Based on testing I think it works, lgtm

PR #5619, opened 2016-11-24
Fix: add priority back into vfolder view table columns

### LINKED ISSUE #5615: Priority column is missing
Since the column reordering we've lost the priority column in the vfolders table

phlax:
fixes #5615


unho:
lgtm

PR #5650, opened 2016-11-28
Unit display

phlax:
adds a `unit_display` delegate provider.

This can be customised on a format specific basis

Allows customisation of unit display in timeline, user profile, browse views, admin uis etc


unho:
lgtm

PR #5652, opened 2016-11-28
Add cache_keys to pootle_fs.resources

phlax:
Adds a `pootle.fs.sync` revision hash to fs projects, which is expired when tracking info is changed.

This is used in combination with the filesystem hash for cache_keys in fs.resources


unho:
lgtm

PR #5658, opened 2016-11-29
[WIP] Humane numbers like N,NNN

dwaynebailey:
Uses django-babel to format numbers correctly.  Unfortunately a Django 1.10 compatible version of django-babel is not ready.

The other option is humanize but it only has limited support for languages.  We may need to go that route if django-babel doesn't gain Django 1.10 compatibility soon.


dwaynebailey:
Replaced by https://github.com/translate/pootle/pull/5664

PR #5673, opened 2016-12-05
Update staticpages mtime on save

### LINKED ISSUE #5672: Saving an announcement no longer changes its mtime
This is because it doesn't detect it having dirty fields https://github.com/translate/pootle/blob/master/pootle/apps/staticpages/models.py#L107

unho:
As a bonus we can drop a mixin.

Fixes #5672.


phlax:
lgtm once travis passes

Issue #5639, opened 2016-11-26
Dont store announcement info for anon user


phlax:
we seem to store "announcement" info for anon users

im not entirely clear what this announcement info is used for - but we dont want to store it for anon, and we dont want it to acummulate for other users

dwaynebailey:
The announcement info seems to be used to determine if the sidebar should be open or closed i.e. it remember a) if the user closed it and b) has the user seen this sidebar or not.

unho:
This info is stored in session now, even for anonymous. The point on keeping this data for any of the anonymous users as well is to allow them to keep the sidebar closed, and force it to display on next visit if any of the announcements changed since their last visit.

Therefore closing.

phlax:
@unho i think we want to look at this a bit further, and consider whether its actually useful for anon users to accumulate this data.

If not, there is a very strong argument for not doing so

dwaynebailey:
I'd agree with @phlax, a simple isOpen is enough for anonymous.  We really only want to convey the fact that the sidebar info has changed to users who can action that and likely that s logged in users.

Issue #5638, opened 2016-11-26
Dont trust cookie data


phlax:
atm there is parsing of cookie data that trusts its json - which is not a valid assumption

PR #5682, opened 2016-12-07
More cookies changes

### LINKED ISSUE #5637: Move session data into session
atm there is data like sidebar stuff that is stored in the actual cookie.

apart from slowing down normal users (they send all that data in *every* request), bots are acumulating cookie cruft and needlessly posting that at our sites.

if the cookie cruft gets too long it gets truncated and errors out

unho:
Theoretically this fixes #5637 and #5638. And also #5639.


phlax:
lgtm in terms of functionality - ie deleted cookie etc. Needs some test-fu tho, and some rebasing/commit renaming

PR #5702, opened 2016-12-18
Fix: fs.state returns correctly for pootle_staged when conflicting

phlax:
as in title


phlax:
closing - this was the wrong fix

PR #5707, opened 2016-12-19
Bump sw version for apps with browse views

phlax:
i think we need to bump all as the browse tables have changed recently


unho:
lgtm

PR #5717, opened 2016-12-20
pootle fs micro optimizations 8/

phlax:
- Optimization: remove default ordering in fs.resources.trackable_stores


unho:
lgtm once Travis passes

PR #5718, opened 2016-12-20
Fix droppy avatar

unho:
This works both for RTL and LTR. As a side effect we can drop
a CSS class.

This was tested locally using Chromium and Firefox.


phlax:
not tested but lgtm - nice work

PR #5719, opened 2016-12-20
Adjust README

unho:
This file is used as the PyPI package description, but most of this
trickery is not pleasant for PyPI.

The usage of |logo| still will present problems for PyPI, but that
will be handled on a followup commit.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,9 +1,27 @@
 |logo| Pootle
 =============
 
-|chat| |build| |health| |coverage| |requirements|
+.. image:: https://img.shields.io/gitter/room/translate/pootle.svg?style=flat-square
+    :alt: Join the chat at https://gitter.im/translate/pootle
+    :target: https://gitter.im/translate/pootle
 
-.. Resources: 
+.. image:: https://img.shields.io/travis/translate/pootle/master.svg?style=flat-square
+    :alt: Build Status
+    :target: https://travis-ci.org/translate/pootle
+
+.. image:: https://landscape.io/github/translate/pootle/master/landscape.svg?style=flat-square
+    :target: https://landscape.io/github/translate/pootle/master
+    :alt: Code Health
+
+.. image:: https://img.shields.io/coveralls/translate/pootle/master.svg?style=flat-square
+   :target: https://coveralls.io/github/translate/pootle?branch=master
+   :alt: Test Coverage
+
+.. image:: https://img.shields.io/requires/github/translate/pootle.svg?style=flat-square
+   :target: https://requires.io/github/translate/pootle/requirements/?branch=master
+   :alt: Requirements
+
+
 `Docs <http://docs.translatehouse.org/projects/pootle/en/latest/>`_ |
 `Changes <http://docs.translatehouse.org/projects/pootle/en/latest/releases/2.8.0.html>`_ |
 `Issues <https://github.com/translate/pootle/issues>`_ |
@@ -27,6 +45,7 @@ A number of translation projects for a number of languages can be hosted on
 Pootle.  Teams can manage their files, permissions, projects, and translate
 on-line.  Files can be downloaded for offline translation.
 
+
 Installation
 ------------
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -1,9 +1,27 @@
 |logo| Pootle
 =============
 
-|chat| |build| |health| |coverage| |requirements|
+.. image:: https://img.shields.io/gitter/room/translate/pootle.svg?style=flat-square
```
dwaynebailey: This change is rather sad to see. I made this alteration so that README.rst could be read by humans as well as machines.  So this revert essentially makes it unreadable again.  Is it not possible to handle the |replace| in the release script?  I do find it odd that pypi doesn't do |replace| since it's just docutils.

dwaynebailey: My goal in the fixes to the README where:

1. Its readable by humans
2. Its nice in github
3. It works in pypi (but that I was never able to test)

Just flagging that so that we actually try keep those in mind with any of these changes.

setup.py
```diff
@@ -267,6 +268,26 @@ def get_check_description(name, filterfunc):
         print("Checks templates written to %r" % (filename))
 
 
+def parse_long_description(filename):
+    readme_lines = open(
+        os.path.join(os.path.dirname(__file__), filename)
```
dwaynebailey: I'd leave this part outside the parser.

setup.py
```diff
@@ -267,6 +268,26 @@ def get_check_description(name, filterfunc):
         print("Checks templates written to %r" % (filename))
 
 
+def parse_long_description(filename):
+    readme_lines = open(
```
dwaynebailey: Can we use with so that we close the file properly, I know its not needed but its a good pattern to follow.

setup.py
```diff
@@ -267,6 +268,26 @@ def get_check_description(name, filterfunc):
...
+        os.path.join(os.path.dirname(__file__), filename)
+    ).read().split("\n")
+    pootle = "Pootle"
+    notes = "Release notes"
+    initial_title = [  # PyPI doesn't like |logo|
+        pootle,
+        "="*len(pootle)
+    ]
+    release_notes = [  # Missing in README.
```
dwaynebailey: They're not missing, there is a link that specifically says "Changes" which points to the exact same place this points to.  So I'm not 100% sure this is needed.

setup.py
unho: Sure. This whole function looks quick and dirty.

setup.py
unho: Will do.

setup.py
unho: I told you my mind does not associate "Changes" with "Release notes" :) I am undoing this.

README.rst
unho: Yes, it is sad. But we need this to work in PyPI too. I spent 2 hours trying to keep this and to PyPI to parse it without luck.

dwaynebailey: I was thinking more like `s/|chat|/...image` instead of this revert.

README.rst
```diff
@@ -1,9 +1,27 @@
...
+.. image:: https://img.shields.io/coveralls/translate/pootle/master.svg?style=flat-square
+   :target: https://coveralls.io/github/translate/pootle?branch=master
+   :alt: Test Coverage
+
+.. image:: https://img.shields.io/requires/github/translate/pootle.svg?style=flat-square
+   :target: https://requires.io/github/translate/pootle/requirements/?branch=master
+   :alt: Requirements
+
 
-.. Resources:
```
dwaynebailey: Does this comment prevent rendering? It was another one of those lets show this for users who won't se the rendering, without it there isn't much context.  But not a blocker

dwaynebailey:
lgtm.  I'd love to see how we can bring in some changes that get us back to human readable README.rst, but that can wait.

README.rst
unho: Yes, this broke PyPI too.

Issue #5728, opened 2016-12-21
Duplicate key name 'accounts_user_is_active_1b8b9f54_uniq


JohnnyKing94:
Before upgrading: 2.8.0b5
I was just upgrading pootle to 2.8.0rc1 and i can't do ```(env) $ pootle migrate```

```console
(pootle)root@server:~# pootle migrate
2016-12-21 22:53:32,753 INFO Loading custom settings from '/root/.pootle/pootle.conf'...
Operations to perform:
  Apply all migrations: account, accounts, auth, contenttypes, pootle_app, pootle_comment, pootle_config, pootle_data, pootle_format, pootle_fs, pootle_language, pootle_project, pootle_revision, pootle_statistics, pootle_store, pootle_translationproject, reports, sessions, sites, socialaccount, staticpages, virtualfolder
Running migrations:
  Applying accounts.0005_user_indeces...Traceback (most recent call last):
  File "/home/desktop/Pubblici/Envs/pootle/bin/pootle", line 11, in <module>
    load_entry_point('Pootle==2.8.0rc1', 'console_scripts', 'pootle')()
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/pootle/runner.py", line 315, in main
    django_settings_module='pootle.settings')
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/pootle/runner.py", line 295, in run_app
    management.execute_from_command_line(command)
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/django/core/management/base.py", line 294, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/home/desktop/Pubblici/Envs/pootle/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 204, in handle
    fake_initial=fake_initial,
[...]

PR #5737, opened 2016-12-23
Grabbag: README and auto update of pip install options

dwaynebailey:
* Adds README check to CI
* Address docs issues with extra parms, fake installs in upgrades
* Dynamically create extra options in Docs and in long_description

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -417,6 +417,7 @@ intersphinx_mapping = {
                'https://docs.djangoproject.com/en/%s/_objects/' % django_ver),
     'toolkit': ('http://docs.translatehouse.org/projects/translate-toolkit/en/latest/', None),
     'virtualenvwrapper': ('https://virtualenvwrapper.readthedocs.io/en/latest/', None),
+    'pip': ('https://pip.pypa.io/en/stable/', None),
 }
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,9 +12,11 @@
 # serve to show the default.
 
 import os
+import re
 import sys
 
 from django import __version__ as dj_version_actual
+from pootle.core.utils import version as pootle_version
 from translate.__version__ import sver as ttk_version_actual
 
 # If extensions (or modules to document with autodoc) are in another directory,
@@ -66,8 +68,7 @@ author = u'Pootle contributors'
 # built documents.
 #
 # The short X.Y version.
-from pootle.core.utils.version import get_docs_version
-version = get_docs_version()
+version = pootle_version.get_docs_version()
 # The full version, including alpha/beta/rc tags.
 from pootle import __version__
 release = __version__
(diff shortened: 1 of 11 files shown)


docs/server/installation.rst
```diff
@@ -129,6 +129,11 @@ Use :command:`pip` to install Pootle into the virtual environment:
 
 This will also fetch and install Pootle's dependencies.
 
+.. note:: pip requires :ref:`--pre <pip:install_--pre>` to install pre-release
+   versions of Pootle, i.e. alpha, beta and rc versions. You may require
+   :ref:`--process-dependency-links <pip:--process-dependency-links>` if Pootle
+   depends on unreleased versions of software.
```
unho: I would specify "third party software".

docs/server/upgrading.rst
```diff
@@ -185,12 +185,19 @@ Migrate your database schema
 ----------------------------
 
 Once you have updated your settings you can perform the database schema and
-data upgrade by running. This needs to be done in a few steps:
+data upgrade by running. This is done as follows:
+
+.. note:: The following fake migrations are only required when migrating from
```
unho: I would remove the "only".

docs/server/upgrading.rst
```diff
@@ -185,12 +185,19 @@ Migrate your database schema
...
+   Pootle 2.5.  Subsequent versions do not require these steps:
```
unho: I would say "2.6 or older" to cover all possibilities.

docs/conf.py
```diff
@@ -11,10 +11,12 @@
...
 # serve to show the default.
 
-import sys
 import os
+import re
+import sys
 
 from django import __version__ as dj_version_actual
+from pootle.core.utils import version as pootle_version
 from translate.__version__ import sver as ttk_version_actual
```
unho: Since imports are being rearranged it should be:

```python
import os
import re
import sys

from translate.__version__ import sver as ttk_version_actual

from django import __version__ as dj_version_actual

from pootle.core.utils import version as pootle_version
```

docs/server/upgrading.rst
dwaynebailey: I am trying to make doubly sure someone doesn't run this in a case where they are not needed, witness what happened the other night.  So I'd prefer to keep unless you have a strong enough wording.

unho:
Raised several nits. gtm.

docs/server/upgrading.rst
unho: k, keep it.

ta2-1:
@dwaynebailey IIUC this PR is related to https://github.com/translate/pootle/issues/5728 which can be closed.

docs/conf.py
dwaynebailey: This is taken from Sphinx and I'd like to keep its diff there minimal.  Its really not a Pootle thing.

unho: k

PR #5738, opened 2016-12-23
Auth: Allow unicode usernames

unho:
Auth: Allow unicode usernames

It is necessary to convert the cache key as well, since it doesn't
work with unicode usernames.

Newer allauth adapter uses the validators in User model, but until
it gets released most of this commit is needed.


pootle/apps/accounts/adapter.py
```diff
@@ -21,12 +22,14 @@
...
     """Reimplementation of DefaultAccountAdapter from allauth to change
-    ajax_response.
+    ajax_response and username validation.
 
     Differences:
       - the html key is removed for performance reasons
       - form_errors is renamed to errors
+      - Unicode usernames are allowed
     """
+    username_regex = re.compile(r'^[\w.@+-]+$', re.UNICODE)
```
dwaynebailey: Isn't it allcaps in allauth? I'd assume it should be 100% the same to actually work

pootle/apps/pootle_project/models.py
```diff
@@ -93,6 +93,7 @@ def cached_dict(self, user):
         else:
             cache_params = {'is_admin': user.is_superuser}
         cache_key = make_method_key('Project', 'cached_dict', cache_params)
+        cache_key = iri_to_uri(cache_key)
         projects = cache.get(cache_key)
```
dwaynebailey: Can we combine these lines? Feels like temp variable heaven

pootle/apps/accounts/adapter.py
unho: In other words. This is exactly as allauth 0.29.0 which is our pinned requirement.

pootle/apps/pootle_project/models.py
unho: Yes, we can.

tests/accounts/adapter.py
```diff
@@ -0,0 +1,18 @@
...
+import pytest
+
+from allauth.account.adapter import get_adapter
+
+
+@pytest.mark.django_db
+def test_account_adapter_unicode():
+    adapter = get_adapter()
+    adapter.clean_username('ascii')
+    adapter.clean_username(u'unicöde')
```
dwaynebailey: Can we add a user that's actually beyond latin1?

I'm worried that this won't test the user in other contexts, e.g. user stats, submissions, suggestions, etc.

PR #5743, opened 2016-12-28
Put unit and suggestion admin together in team page

unho:
Also add link to suggestion admin, and do not show links if the count
is zero (prevents hitting the database if not needed).


dwaynebailey:
For consistency I would:

1. Right align the numbers
2. Use locale formatting, since we've fixed that everywhere else lets not bring broken number formatting back
3. Use the styling we use on the browse page i.e. 0 is muted grey, certain number of colour lozenges.

phlax:
i agree with @dwaynebailey 's nits but otherwise lgtm

PR #5753, opened 2017-01-03
Update top_scorers template tag to use score data

phlax:
this is used in the welcome page

### DIFF
--- a/pootle/apps/pootle_app/views/index/index.py
+++ b/pootle/apps/pootle_app/views/index/index.py
@@ -7,15 +7,60 @@
 # AUTHORS file for copyright and authorship information.
 
 from django.contrib.auth import REDIRECT_FIELD_NAME
-from django.shortcuts import redirect, render
+from django.shortcuts import redirect
 from django.urls import reverse
+from django.utils.functional import cached_property
+from django.utils.translation import get_language
 from django.views.generic import TemplateView
 
+from pootle.core.decorators import persistent_property
+from pootle.core.delegate import revision, scores
 from pootle.i18n.override import get_lang_from_http_header
 from pootle_language.models import Language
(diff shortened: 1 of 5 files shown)


pootle/templates/welcome.html
```diff
@@ -30,7 +30,7 @@
 
       <div class="section">
         <h1>{% trans 'Top Contributors for the Last 30 Days' %}</h1>
-        {% top_scorers %}
+	{% top_scorers user score_data %}
```
unho: It was using the right indent.

pootle/apps/pootle_app/views/index/index.py
```diff
@@ -7,23 +7,66 @@
...
+            % (self.request.user.username,
+               self.revision,
+               self.request_lang))
+
+    @cached_property
+    def project_set(self):
+        user_projects = Project.accessible_by_user(self.request.user)
+        user_projects = (
+            Project.objects.for_user(self.request.user)
+                           .filter(code__in=user_projects))
```
unho: It is not the first time I see this pattern. I do wonder if it is better to move this to a new method in the manager.

phlax: not 100% if its right either - but it works 8/

unho:
lgtm

pootle/apps/pootle_app/views/index/index.py
unho: Yes, it works. Just raising that this looks like repeated code to me.

pootle/templates/welcome.html
phlax: it seems not - it should align with the h1

unho: Sure, that was what I meant. Now it is misaligned.

PR #5761, opened 2017-01-04
Fix: add translated, reviewed, suggested to score data

phlax:
these were missing, also localises the display


unho:
lgtm once tested adjusted.

PR #5762, opened 2017-01-04
Fix: add unformatted total_score for comparing whether its positive

phlax:
this is a temp fix, as it would be better for the +/- to be properly localised


unho:
lgtm

PR #5777, opened 2017-01-08
Term matcher migration2

phlax:
adds a migration to set terminology stems, and a receiver to keep them up to date


pootle/apps/pootle_terminology/receivers.py
```diff
@@ -0,0 +1,27 @@
...
+from pootle_store.constants import TRANSLATED
+from pootle_store.models import Unit
+
+
+@receiver(post_save, sender=Unit)
+def handle_unit_save(**kwargs):
+    unit = kwargs["instance"]
+    if unit.state != TRANSLATED:
+        return
+    is_terminology = (
```
unho: I would rename this to `is_in_terminology_tp`. Otherwise can be misleading since each regular project can have a `pootle-terminology` store which holds project-specific terminology, and we are leaving out that other terminology.

phlax: not sure whether we want to keep such terminology - but i figure it should be indexable here if we do

unho:
lgtm, nit aside.

pootle/apps/pootle_terminology/receivers.py
phlax: migration/utils adjusted to handle pootle-terminology.* files

PR #5786, opened 2017-01-09
Fix: prevent editor from scrolling off screen and limit jumpiness

phlax:
as in title


ta2-1:
@phlax I think this PR https://github.com/translate/pootle/pull/5764 can be reverted after you land the current one.

phlax:
closing in favour of #5791

PR #5792, opened 2017-01-11
Cleanups: picked up from py3 porting

dwaynebailey:
The following are simple cleanups taken from experimental Python 3 support.


unho:
lgtm

PR #5796, opened 2017-01-12
Fix: use user.pk in UserScores.cache_key, prevent unicode fail

phlax:
as in title


unho:
Reviewed and tested. GTM

ta2-1:
lgtm

PR #5798, opened 2017-01-12
Set rowcount from user.get_unit_rows only in unit.search.form

phlax:
hopefully fixes mismaches in batch size between server/client

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -395,7 +395,6 @@ def unit_comment_form_factory(language):
 
 class UnitSearchForm(forms.Form):
 
-    count = forms.IntegerField(required=False)
     offset = forms.IntegerField(required=False)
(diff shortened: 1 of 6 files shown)


pootle/apps/pootle_store/forms.py
```diff
@@ -462,14 +459,10 @@ def clean(self):
...
-            count = (
-                self.cleaned_data.get("count", self.default_count)
-                or self.default_count)
-            user_count = (
-                self.cleaned_data["user"].get_unit_rows()
-                or self.default_count)
-            self.cleaned_data['count'] = min(count, user_count)
+        user_count = (
+            self.cleaned_data["user"].get_unit_rows()
+            or self.default_count)
```
unho: What is `self.default_count` now?

phlax: not sure what we need to do about this - ill check

PR #5800, opened 2017-01-13
Add experimental on pootle fs project admin dropdown option

phlax:
as the code is currently subject to change and requires further testing


unho:
lgtm

PR #5802, opened 2017-01-13
Python 3 cleanups: print() and Octal

dwaynebailey:
Octal: use new syntax and prevent regression
print(): prevent regression, use `__future__` as needed and fix remaining print[sp] occurances


unho:
lgtm

Issue #4568, opened 2016-03-14
Templates TP displays need translation count despite it shouldn't be translated


unho:
![pending_templates](https://cloud.githubusercontent.com/assets/939880/13739830/b2a9b0ae-e9cd-11e5-993d-1e354f69ac74.png)

Picked from #4559.

Technically this should be skipped for any template language the project is using.

ta2-1:
There is more related information:
- `non .po` projects have non-zero stats;
- `.po` projects have zero stats;

![image](https://cloud.githubusercontent.com/assets/682845/15892750/24d8099c-2d84-11e6-8d3c-29a8a56bed59.png)

Neither `.po` or `non .po` projects don't work properly.

What do we want to get:
- Template project should have non-zero total stats;
- Template project have zero `need work`, `translated` stats;
- Template project stats should not be counted in parent (Project) level.

and 
- Probably we have to allow to list units from templates but it can be another issue.

unho:
Getting translations for templates TPs can be prevented by setting appropriate permissions when creating those template TPs.

PR #5813, opened 2017-01-15
Hide template stats in browse views

phlax:
fixes https://github.com/translate/pootle/issues/4568

### DIFF
--- a/pootle/apps/pootle_data/utils.py
+++ b/pootle/apps/pootle_data/utils.py
@@ -423,7 +423,11 @@ class RelatedStoresDataTool(DataTool):
         latest = dict(last_submission=None, last_created_unit=None)
         last_submission_pk = None
         last_created_unit_time = None
-        for child in stats["children"].values():
+        for path, child in stats["children"].items():
+            if path.startswith("templates-") or path == "templates":
+                if not self.context.pootle_path.startswith("/templates/"):
+                    del child["last_submission__pk"]
+                    continue
             for k in agg.keys():
                 agg[k] += child[k]
             if child.get("last_created_unit"):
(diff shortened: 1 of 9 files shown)


dwaynebailey:
lgtm just from visual/clicky testing.

Some observations, take or leave or flag as future issue:

* Do we want a progressbar on Project view? Indicate completeness and we don't have any concept of completeness for Templates
* We don't have expanded stats, OK.  But that means we can't see words vs units.  I realise we don't have that in normal.  But would be nice to have words and unit counts for templates.
* Templates TP listed when not logged in (shouldn't we hide it?)

phlax:
hi @dwaynebailey the first two points seem like new features and outside the scope of this PR

re
> Templates TP listed when not logged in (shouldn't we hide it?)

i wondered about this - but then i dont really see why - i think it could be useful and its not private data - not sure tho.

phlax:
maybe i misunderstood the first point - grokking

pootle/apps/pootle_data/utils.py
```diff
@@ -423,7 +423,11 @@ def aggregate_children(self, stats):
         latest = dict(last_submission=None, last_created_unit=None)
         last_submission_pk = None
         last_created_unit_time = None
-        for child in stats["children"].values():
+        for path, child in stats["children"].items():
+            if path.startswith("templates-") or path == "templates":
+                if not self.context.pootle_path.startswith("/templates/"):
```
ta2-1: @phlax can we use something like `self.context.is_template()` here?

pootle/apps/pootle_project/views.py
```diff
@@ -110,6 +110,12 @@ class ProjectBrowseView(ProjectMixin, PootleBrowseView):
     view_name = "project"
 
     @property
+    def is_templates_context(self):
+        return (
+            len(self.object_children) == 1
+            and self.object_children[0]["code"] == "templates")
```
ta2-1: @phlax I'm not sure I'm clear with logic here. Can you describe it?

pootle/apps/pootle_data/utils.py
phlax: not really because its not always a template - it can be a directory or a tp etc

unho:
lgtm

pootle/apps/pootle_project/views.py
phlax: ill add a comment

PR #5818, opened 2017-01-16
Templates in test env

phlax:
- Add templates into the test env
- update fixtures to use them
- add fixtures to block them
- update tests that need fixtures blocked


unho:
lgtm

PR #5832, opened 2017-01-17
Dont record pootle revision on fs.sync_rm if store doesnt exist

phlax:
this fixes a fail when syncing with removed stores


unho:
lgtm

PR #5844, opened 2017-01-19
Update store syncer

phlax:
moves the sync_stores specific logic back to Store.sync so Syncer class can more easily be reused


unho:
lgtm

ta2-1:
lgtm besides triggered error with empty store with no units which doesn't have last_sync_revision value.

phlax:
> empty store with no units which doesn't have last_sync_revision value.

this is fixed, and a test has been added

unho:
Tested a bit and seems to work. gtm

PR #5869, opened 2017-01-23
Raise CommandError in pootle commands if fail

### LINKED ISSUE #5731: Ensure update_stores command fails on error
Seems like update_stores may swallow (some) errors. This is not helpful for sysadmins 8/

phlax:
fixes #5731


unho:
lgtm

PR #5881, opened 2017-01-24
Limit store textfields to max 4096 chars

phlax:
there doesnt seem to be any limit on size of text/comment fields atm, which can cause problems


pootle/apps/pootle_store/models.py
```diff
@@ -198,7 +201,9 @@ class Unit(models.Model, base.TranslationUnit):
     store = models.ForeignKey("pootle_store.Store", db_index=True,
                               on_delete=models.CASCADE)
     index = models.IntegerField(db_index=True)
-    unitid = models.TextField(editable=False)
+    unitid = models.TextField(
+        editable=False,
+        validators=[MaxLengthValidator(4096)])
```
unho: Not sure if this is actually necessary. Is it possible to edit it from client side?

phlax: it doesnt really matter - it needs a limit - otherwise things fail

phlax: not sure what that limit should be tho

pootle/apps/pootle_store/models.py
```diff
@@ -214,10 +219,22 @@ class Unit(models.Model, base.TranslationUnit):
...
+        blank=True,
+        validators=[MaxLengthValidator(4096)])
+    translator_comment = models.TextField(
+        null=True,
+        blank=True,
+        validators=[MaxLengthValidator(4096)])
+    locations = models.TextField(
+        null=True,
+        editable=False,
+        validators=[MaxLengthValidator(4096)])
```
unho: This field is not editable from client side. I would say leave it out.

phlax: as above

dwaynebailey:
To give you some reassurance http://www.gamefaqs.com/boards/922345-section-z/64202650

At 4096 if you are translating in one unit you're definitely doing something wrong.

pootle/apps/pootle_store/models.py
```diff
@@ -214,10 +219,22 @@ class Unit(models.Model, base.TranslationUnit):
...
-    locations = models.TextField(null=True, editable=False)
-    context = models.TextField(null=True, editable=False)
+    developer_comment = models.TextField(
+        null=True,
+        blank=True,
+        validators=[MaxLengthValidator(4096)])
+    translator_comment = models.TextField(
+        null=True,
+        blank=True,
+        validators=[MaxLengthValidator(4096)])
```
unho: I would further restrict this. 4096 characters for a translator comment is like a whole novel.

PR #5900, opened 2017-01-26
Cleanup: make settings message DEBUG

phlax:
as in title


dwaynebailey:
gtm

unho:
I would prefer to keep it.

PR #5909, opened 2017-01-27
Docs: Pootle FS is in beta

dwaynebailey:
Try to prevent installation by users without understanding the
consequences.


unho:
lgtm

PR #5920, opened 2017-01-29
Unit and editor fixes

phlax:
- set unit index when resurrecting
- use min of total and max of slice in checking results

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -873,6 +873,7 @@ class Unit(models.Model, base.TranslationUnit):
             self.state = UNTRANSLATED
 
         self.update_qualitychecks(keep_false_positives=True)
+        self.index = self.store.max_index() + 1
         self._state_updated = True
         self._save_action = UNIT_RESURRECTED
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -873,6 +873,7 @@ def resurrect(self, is_fuzzy=False):
             self.state = UNTRANSLATED
 
         self.update_qualitychecks(keep_false_positives=True)
+        self.index = self.store.max_index() + 1
```
ta2-1: Put resurrected in the end always?
I feel that it will need to be changed :/

phlax: where would you suggest, it has no index while its obsolete?

ta2-1: Ideally index should come from a file though I'm not 100% sure.

phlax: im not sure if it works correctly - i need to test further - but i think the indeces get updated after resurrection. If not we might need to change the order in the update

PR #5924, opened 2017-01-30
Use kwargs to get path in tp browse view

### LINKED ISSUE #5923: Setting FORCE_SCRIPT_NAME breaks views
Seems like a number of views are not parsing path correctly when FORCE_SCRIPT_NAME is set.

phlax:
fixes #5923


dwaynebailey:
lgtm - clicky testing on staging and all seems to work fine.

PR #5932, opened 2017-01-30
Try and prevent django race condition on bulk delete/insert

### LINKED ISSUE #5730: IntegrityError: (1062, "Duplicate entry 'xxx-stats' for key > 'pootle_revision_content_type_id_xxx_uniq'")
This error has been spotted in the wild.

From code review its hard to see how its happening, and i ~~havent~~ managed to reproduce - *by clicking very fast on editor buttons*

~~It may be related to update_stores~~

phlax:
This is another attempt to fix #5730

The problem is almost certainly described here:

https://code.djangoproject.com/ticket/23576


unho:
lgtmed

PR #5936, opened 2017-01-31
[requires.io] dependency update on master branch

dwaynebailey:


dwaynebailey:
Enough of these, lets let requires.io open a new one on its own.

Issue #5679, opened 2016-12-07
Adapt README for release and branching


dwaynebailey:
The README is used in github and is used as the long description in releases.

It works well in and is written for `master`. It is not useful on release branches `stable/2.7.x` as its still pointing to badges that reflect `master` (we don't care about development branches).

For release long description it really should reflect the static nature of the release i.e. what is the state of the requirements as specified in the release. Currently its pointing to `master` and giving misleading information.

There are some issues:
* Github has no way to dynamically change references to the branch.  So likely we need to script those change to adjust the README in critical branches.
* Some of the upstream badges and references don't quite work with point releases.

My gut feeling is that we need a script that is able to read `master` README.rst and produce a version appropriate for a branch or tag.

PR #5937, opened 2017-01-31
[WIP] Update README for branches

dwaynebailey:
This is a bash script that updates the README.rst for branches.  Its horrible and nasty but at least it documents what we likely should do in a README update for a final release and to the README on a branch.

My idea was this:

* It should be converted to Python
* It should run as part of `setup.py` (a lot does already)
* That it should change to stable branch links when run by setup.py on a `stable/N.M.O` branch
* It should be run on stable branches to create a stable branch README, but also as part of Travis to check that it remains correct.

Putting this up so that whoever works on #5679 can do it without the linking, checking and testing that I had to do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -274,7 +274,7 @@ class BuildChecksTemplatesCommand(Command):
         print("Checks templates written to %r" % (filename))
 
 
-def parse_long_description(filename):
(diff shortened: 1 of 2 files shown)


dwaynebailey:
Replaced by #6569

PR #5943, opened 2017-02-01
Consolidate subs

phlax:
atm submissions are created in a number of places, this pr will rationalize that with respect to submissions created after units are changed


phlax:
this was landed in #6059

PR #5944, opened 2017-02-01
Frozen unit

phlax:
factors frozenunit into a frozen delegate


ta2-1:
lgtm after travis passes

PR #5946, opened 2017-02-02
Add UnitLifecycle

phlax:
adds a class for handling unit lifecycle events


phlax:
merged in https://github.com/translate/pootle/pull/5970

PR #5949, opened 2017-02-02
Set system user if unit users are deleted

phlax:
atm units are deleted if a user is deleted, this pr means units are kept and set to system


unho:
lgtm provided Travis passes

PR #5950, opened 2017-02-02
User delete 2

phlax:
prevents suggestions and submissions from being deleted if user is deleted


unho:
lgtm once travis passes

PR #5952, opened 2017-02-02
Migrate system user

phlax:
Currently the system user is created by initdb - but other code relies on system user existing whether this has been run or not.

This PR creates the system user in a migration where required.


pootle/core/initdb.py
```diff
@@ -110,15 +110,6 @@ def create_essential_users(self):
         }
         self._create_pootle_user(**criteria)
 
-        # The system user represents a system, and is used to
-        # associate updates done by bulk commands as update_stores.
-        criteria = {
-            'username': u"system",
-            'full_name': u"Pootle",
-            'is_active': True,
```
ta2-1: @phlax `is_active` is not set in the new code?

phlax: thanks, ill check...

pootle/apps/accounts/migrations/0008_add_system_user.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.5 on 2017-02-02 19:02
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def add_system_user(apps, schema_editor):
+    User = apps.get_model("accounts.User")
+    sysuser = users.filter(username="system").first()
```
unho: Use `exists()`

unho:
lgtm

PR #5956, opened 2017-02-06
Irish translation back to 100%

kscanne:


dwaynebailey:
Landed in 6ba06051c18f4580a3e8d07d42b87b39df41b0c9

@kscanne thanks once again

unho:
@kscanne We have landed some changes in Pootle strings, so the translation will need another small update. Sorry for the inconveniences.

kscanne:
No problem!  A quick question though... Since I've been submitting translations via PRs, I'm wondering about the status of the project here: http://pootle.locamotion.org/ga/pootle/     If a second contributor were to make changes to the translation through the Pootle server, would those be brought in to github automatically + periodically?  (I'm asking because we're trying to remove dependencies on me from our l10n processes!)

unho:
Right now Pootle translations are synchronized manually and on a non-regular basis, but we have plans to put automatic regular synchronization in the future. Since you are submitting complete translations through PRs we are most likely going to apply your changes that the ones coming from that server. If you want to avoid such situation you can directly translate in http://pootle.locamotion.org/ga/pootle/ so your changes are blended with other people contributions.

kscanne:
Thanks.  I'll continue with the PRs for now.

PR #5959, opened 2017-02-07
Add IOSChecker to the list of available checkers

unho:


dwaynebailey:
lgtm - is there anything translatable that needs to be part of this?

unho:
Surprisingly there is no translatable text. In the dropdown you just select the key for your checker in this dictionary.

PR #5965, opened 2017-02-10
i18n: Handle missing babel locale again

unho:
Fixes: 'UnknownLocaleError: unknown locale 'son'

### DIFF
--- a/pootle/i18n/dates.py
+++ b/pootle/i18n/dates.py
@@ -6,49 +6,12 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
-import locale as system_locale
-import os
 from datetime import datetime
 
-from babel.dates import format_timedelta
 
-from django.utils import translation
-
-
-class LocalDate(object):
-
-    def __init__(self):
-        if not self.locale_code and not os.name == "nt":
-            self.set_locale()
-
-    @property
-    def default_locale(self):
-        return translation.to_locale(translation.get_language())
-
-    def set_locale(self):
-        system_locale.setlocale(
-            system_locale.LC_ALL,
-            (self.default_locale, 'UTF-8'))
-
-    @property
-    def locale_code(self):
-        return system_locale.getlocale()[0]
-
-    def format_timesince(self, timestamp, locale=None):
-        return format_timedelta(
-            datetime.now()
-            - datetime.fromtimestamp(
-                timestamp),
-            locale=(
-                locale
-                or self.locale_code
-                or self.default_locale))
-
-
-localdate = LocalDate()
+from .formatter import get_locale_formats
 
 
 def timesince(timestamp, locale=None):
-    if locale:
-        locale = translation.to_locale(locale)
-    return localdate.format_timesince(timestamp, locale=locale)
+    timedelta = datetime.now() - datetime.fromtimestamp(timestamp)
+    return get_locale_formats(locale).timedelta(timedelta, format='long')
--- a/pootle/i18n/formatter.py
+++ b/pootle/i18n/formatter.py
@@ -15,8 +15,10 @@ from django.conf import settings
 from django.utils.translation import get_language, to_locale
 
 
-def get_locale_formats():
-    for language in [get_language(), settings.LANGUAGE_CODE, 'en-us']:
+def get_locale_formats(locale=None):
+    languages = [locale] if locale else []
+    languages += [get_language(), settings.LANGUAGE_CODE, 'en-us']
+    for language in languages:
         try:
             locale = babel_core.Locale.parse(to_locale(language))
             break
--- a/pootle/i18n/formatter.py
+++ b/pootle/i18n/formatter.py
@@ -15,7 +15,7 @@ from django.conf import settings
 from django.utils.translation import get_language, to_locale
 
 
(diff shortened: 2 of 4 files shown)


pootle/i18n/dates.py
```diff
@@ -35,14 +37,18 @@ def locale_code(self):
...
+                  or self.locale_code
+                  or self.default_locale)
+        try:
+            return format_timedelta(timedelta, locale=locale)
+        except UnknownLocaleError:
+            locale = settings.LANGUAGE_CODE.split("-")[0]
+            try:
+                return format_timedelta(timedelta, locale=locale)
+            except ValueError:
+                return format_timedelta(timedelta, locale='en')
```
dwaynebailey: This needs a test.

This seems like it's following the same pattern from the bits I coded.  Can you not refactor that out into a function that devolves to a workable locale.

pootle/i18n/dates.py
```diff
@@ -35,14 +37,18 @@ def locale_code(self):
...
-                or self.locale_code
-                or self.default_locale))
+        timedelta = datetime.now() - datetime.fromtimestamp(timestamp)
+        locale = (locale
+                  or self.locale_code
+                  or self.default_locale)
+        try:
+            return format_timedelta(timedelta, locale=locale)
+        except UnknownLocaleError:
+            locale = settings.LANGUAGE_CODE.split("-")[0]
```
dwaynebailey: Definitely a no no.  I'm pretty sure I used something that can give you a valid code.

tests/i18n/dates.py
```diff
@@ -6,21 +6,43 @@
...
+    assert timesince(timestamp) == format_timedelta(timedelta, locale=locale)
+
+
+@pytest.mark.parametrize('language,fallback', [
+    # Normal
+    ('af', 'af'),
+    ('en-za', 'en'),
+    ('en-us', 'en'),
+    # Missing in babel
+    ('son', 'en'),
```
dwaynebailey: Bad one to test, this test will be bust as soon as `son` comes to babel. Rather us a really broken one that won't ever appear in CLDR

tests/i18n/dates.py
```diff
@@ -6,21 +6,43 @@
...
-                localdate.locale_code
-                or localdate.default_locale)))
+    timedelta = datetime.now() - datetime.fromtimestamp(timestamp)
+    assert timesince(timestamp) == format_timedelta(timedelta, locale=locale)
+
+
+@pytest.mark.parametrize('language,fallback', [
+    # Normal
+    ('af', 'af'),
+    ('en-za', 'en'),
```
dwaynebailey: I'm guessing that this passes by accident rather than by design.  For numbers it will fail.  So really not sure what we're testing here.

pootle/i18n/formatter.py
```diff
@@ -15,12 +15,14 @@
 from django.utils.translation import get_language, to_locale
 
 
-def get_locale_formats():
-    for language in [get_language(), settings.LANGUAGE_CODE, 'en-us']:
+def get_locale_formats(locale=None):
+    languages = [locale] if locale else []
+    languages += [get_language(), settings.LANGUAGE_CODE, 'en-us', 'en']
```
dwaynebailey: Was there a specific reason to fallback to `en`?  Since our app is in `en-us` it doesn't make much sense to me.  And I suspect in CLDR `en-us == en` so kinda pointless step IIUC and just confuses the matter with no justification in the code or the commit

pootle/i18n/formatter.py
```diff
@@ -15,12 +15,14 @@
...
+    for language in languages:
         try:
             locale = babel_core.Locale.parse(to_locale(language))
             break
-        except UnknownLocaleError:
+        except (UnknownLocaleError, ValueError):
```
dwaynebailey: Do we want to catch ValueError? This sounds remarkably like `Except` catchall but without info I'm not sure.

tests/i18n/dates.py
```diff
@@ -6,21 +6,43 @@
...
+from pootle.i18n.dates import timesince
 
 
 def test_local_date_timesince(settings):
+    for language in [get_language(), settings.LANGUAGE_CODE, 'en-us']:
+        try:
+            locale = babel_core.Locale.parse(to_locale(language))
+            break
+        except UnknownLocaleError:
+            continue
```
dwaynebailey: Does this test anything? Can we actually have a test that flags the UnkownLocaleError so we know that we're succeeding in capturing that.

tests/i18n/dates.py
unho: We are testing here that it catches the error and falls back to default locale.

tests/i18n/dates.py
unho: Rearranged the test to drop this code.

pootle/i18n/formatter.py
unho: It seems `babel_support.Format(babel_core.Locale.parse(to_locale('en-us'))).timedelta(timedelta, format='long')` is not the same as `format_timedelta(timedelta, locale='en-us')`. The latter actually yields a `ValueError` and thus the default to `en`. Changed PR to remove all of that.

dwaynebailey:
lgtm

PR #5969, opened 2017-02-10
Store abstracts

phlax:
Move store models into abstract classes - this will make it easier to create eg `DeletedUnit`


ta2-1:
lgtm

PR #5971, opened 2017-02-10
Cleanup: remove submissions that have no usable data

phlax:
for example - no unit, or neither field nor type


unho:
This migration looks incomplete regarding the PR description.

phlax:
yep, i was testing to see what if any other migrations are required. Seems like this is sufficient for now

PR #5980, opened 2017-02-15
Unit meta

phlax:
add created_by field to separate unit table - `UnitSource`

also adds `created_with` field - atm we dont record this info, so it defaults to `SubmissionTypes.SYSTEM`

in future we can move more unit fields to this table - in particular fields that are only changed when unit is created


pootle/apps/pootle_store/abstracts.py
```diff
@@ -31,6 +32,35 @@
 fs = PootleFileSystemStorage()
 
 
+class AbstractUnitSource(models.Model):
+
+    class Meta(object):
+        abstract = True
+
+    unit = models.ForeignKey(
```
unho: Not OneToOne?

unho: Probably better leave as it is to avoid complications.

pootle/apps/pootle_store/abstracts.py
```diff
@@ -31,6 +32,35 @@
...
+        db_index=True,
+        unique=True,
+        null=False,
+        blank=False,
+        related_name="unit_source",
+        on_delete=models.CASCADE)
+
+    created_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=True,
```
unho: I would enforce having some value in all cases.

phlax: yep, ill update

pootle/apps/pootle_store/models.py
```diff
@@ -1,3 +1,4 @@
+
```
unho: Nit: remove this.

pootle/apps/pootle_store/models.py
```diff
@@ -321,6 +329,10 @@ def save(self, *args, **kwargs):
             self.submitted_on = None
 
         super(Unit, self).save(*args, **kwargs)
+        if created:
+            unit_source = self.unit_source.model(unit=self)
```
unho: I would say it is a convoluted way to instantiate `UnitSource`...

phlax: not sure i follow

pootle/apps/pootle_store/models.py
phlax: that would break pep8

unho: Weird, I didn't comment on this line, yet the comment show up here.

pootle/apps/pootle_store/abstracts.py
```diff
@@ -31,6 +32,36 @@
...
+    created_by = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        null=False,
+        blank=False,
+        db_index=True,
+        related_name='created_units',
+        default=get_system_user_id,
+        on_delete=models.SET(get_system_user))
+
+    created_with = models.IntegerField(
```
unho: What is this used for?

phlax: it will be used to record how the unit was created - ie `UPLOAD` `FS` `CLONE` etc

unho:
lgtm

PR #5982, opened 2017-02-16
Contact form fixes

unho:
Not only to fix bug in Django 1.10+ support, but to further tweak the forms.

### DIFF
--- a/pootle/apps/contact/forms.py
+++ b/pootle/apps/contact/forms.py
@@ -8,6 +8,7 @@
 
 from django import forms
 from django.conf import settings
+from django.template import loader
 
 from contact_form.forms import ContactForm as OriginalContactForm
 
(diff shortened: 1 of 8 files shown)


pootle/apps/contact/forms.py
```diff
@@ -18,10 +18,10 @@
 
 class ContactForm(MathCaptchaForm, OriginalContactForm):
 
-    field_order = ['name', 'email', 'subject', 'body', 'captcha_answer',
+    field_order = ['name', 'email', 'summary', 'body', 'captcha_answer',
```
dwaynebailey: `email_subject` might be a less confusing name for this as I don't know anyone who call the subject of an email a summary.

unho: The summary is actually something that gets appended with other stuff to conform the email subject.

dwaynebailey:
@unho what's actually holding up tests on this?

pootle/apps/contact/forms.py
dwaynebailey: Names are chosen to prevent confusion, I only reported this cause I actually was confused and checked.  I expect the next person to review to be similarly confused.  You changed it because of other code conflict, not because its not quite a subject.

unho: Any name can be used for the field, except one that clashes with methods or fields coming from the  contact_form API. `subject` is a method in the API so we can use it for a field if we actually need the method as well, which we do, thus the renaming to "summary" which is a naming that was already used for it in other parts of the code. IMHO this makes it easier to spot the difference between the summary and the subject.

unho: In other words, the patch to support Django 1.10+ requires setting the `subject()` method, which given the contact_form API effectively hides the `subject` field. In order to be able to use the `subject` field we need to rename to anything else (that doesn't clash with any reserved name in the contact_form API).

phlax:
other than minor nits raised in private channel - lgtm - nice work @unho

PR #5985, opened 2017-02-16
Check if using not supported Python version

unho:


pootle/checks.py
```diff
@@ -547,3 +549,15 @@ def check_mysql_timezones(app_configs=None, **kwargs):
...
     return errors
+
+
+@checks.register()
+def check_python3(app_configs=None, **kwargs):
+    errors = []
+    if sys.version_info >= (3,0):
+        errors.append(checks.Critical(
+            _("Pootle does not yet support Python 3."),
+            hint=_("Use a Python 2 virtualenv."),
```
dwaynebailey: Python 2.7

pootle/checks.py
```diff
@@ -547,3 +549,15 @@ def check_mysql_timezones(app_configs=None, **kwargs):
             if converted_now is None:
                 errors.append(missing_mysql_timezone_tables)
     return errors
+
+
+@checks.register()
+def check_python3(app_configs=None, **kwargs):
+    errors = []
+    if sys.version_info >= (3,0):
```
dwaynebailey: While we're at it you might as well check that its not < Python 2.7

pootle/checks.py
unho: Good point.

PR #5989, opened 2017-02-17
Cleanup: dont delete store/tp data if units/subs are removed

phlax:
altho data should be recalculated - it doesnt want to be auto destroyed if reffed units or submissions are removed


unho:
lgtm

PR #5993, opened 2017-02-19
[WIP] Separate unit fields into UnitChange table

phlax:
this will make the unit table easier to work with and hopefully provide some optimization


pootle/apps/pootle_statistics/models.py
```diff
@@ -400,9 +400,11 @@ def get_scorelogs(cls, submission):
             'submission': submission,
         }
 
-        translator_id = submission.unit.submitted_by_id
-        if submission.unit.reviewed_by_id:
-            reviewer_id = submission.unit.reviewed_by_id
+        translator_id = (
+            submission.unit.changed
+            and submission.unit.change.submitted_by_id)
+        if submission.unit.changed and submission.unit.change.reviewed_by_id:
```
unho: I was thinking on simplifying this, but better keep it as is.

pootle/apps/accounts/utils.py
```diff
@@ -222,17 +224,17 @@ def revert_units_commented(self):
                 # translator_comment, commented_by, and commented_on
                 last_comment = comments.latest('pk')
                 unit.translator_comment = last_comment.new_value
-                unit.commented_by_id = last_comment.submitter_id
-                unit.commented_on = last_comment.creation_time
+                unit.change.commented_by_id = last_comment.submitter_id
+                unit.change.commented_on = last_comment.creation_time
```
unho: Is it not possible to just use `change`?

phlax: not following

pootle/apps/accounts/utils.py
```diff
@@ -242,30 +244,34 @@ def revert_units_edited(self):
...
+            user = None
+            submitted_on = None
             if edits.exists():
-                last_edit = edits.latest("pk")
+                last_edit = edits.order_by("-creation_time", "-pk").first()
                 unit.target_f = last_edit.new_value
-                unit.submitted_by_id = last_edit.submitter_id
-                unit.submitted_on = last_edit.creation_time
+                user = last_edit.submitter
+                unit.change.submitted_by_id = last_edit.submitter
```
unho: Why not use `user`?

pootle/apps/accounts/utils.py
```diff
@@ -242,30 +244,34 @@ def revert_units_edited(self):
...
+                unit.change.submitted_on = last_edit.creation_time
+                submitted_on = last_edit.creation_time
```
unho: If you invert the order you can use `submitted_on` instead of `last_edit.creation_time`:
```python
                submitted_on = last_edit.creation_time
                unit.change.submitted_on = submitted_on
```

pootle/apps/pootle_store/utils.py
```diff
@@ -34,10 +34,14 @@ class FrozenUnit(object):
     """Freeze unit vars for comparison"""
 
     def __init__(self, unit):
+        submitter = (
+            unit.change.submitted_by
+            if unit.changed
```
unho: `unit.changed`?

phlax: yep unit has a `changed` property to check if there is a `change` object

pootle/apps/pootle_store/utils.py
```diff
@@ -161,6 +165,7 @@ def accept_suggestion(self, suggestion):
             create_subs[SubmissionFields.STATE] = [old_state, unit.state]
         create_subs[SubmissionFields.TARGET] = [old_target, unit.target]
         subs_created = []
+        unit.change, _ = UnitChange.objects.get_or_create(unit_id=unit.id)
```
unho:
```python
        unit.change = UnitChange.objects.get_or_create(unit_id=unit.id)[0]
```

pootle/apps/pootle_store/migrations/0037_set_unit_changed_with.py
```diff
@@ -0,0 +1,57 @@
...
+def set_unit_changed_with(apps, schema_editor):
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    UnitChange = apps.get_model("pootle_store.UnitChange")
+    unit_changes = {}
+    last_unit = 0
+    sub_data = subs.order_by("unit_id", "-creation_time", "-id").values_list(
+        "unit_id", "type")
+    for unit_id, sub_type in sub_data.iterator():
+        if unit_id != last_unit:
+            unit_changes[sub_type] = unit_changes.get(sub_type, [])
```
unho: This is getting a list that is larger with each iteration and assigning it again, why not:

```python
            if sub_type not in unit_changes:
                unit_changes[sub_type] = []
```

phlax: they both do the same lookup - i think the way its done is cleaner

pootle/apps/accounts/utils.py
unho: If you check a few lines above you are iterating changes, then you get the unit with `unit = change.unit` and now you call `unit.change` to set some fields on it. My question is if it is possible to just use `change` instead of `unit.change`.

phlax: ah, k

pootle/apps/pootle_store/migrations/0037_set_unit_changed_with.py
unho: k, not going to discuss this further.

pootle/apps/pootle_store/utils.py
unho: k

phlax:
this was landed in other prs

PR #5996, opened 2017-02-19
i18n: fix GNU string

leonardcj:


phlax:
hi @leonardcj - change looks good - but its stored in db - so will require a migration.

i can help with that if necessary

leonardcj:
It is not super critical, getting it into 2.8 release would be nice.

unho:
Landed in https://github.com/translate/pootle/commit/3adb262579a7307162648788c5a2f20880d9ac24 Thanks for your contribution.

PR #5998, opened 2017-02-20
[WIP] Add some tests for unit.change

phlax:
Ensure unit.change is set correctly when units change


phlax:
this is no longer required

PR #5999, opened 2017-02-20
Ensure unitchange is updated correctly when units change

phlax:
StoreUpdater adds user/submission time to save call


unho:
lgtm other than concern raised in chat.

PR #6011, opened 2017-02-21
Unit submit change

phlax:
add tests to ensure correct attribution etc on unit submit changes

### DIFF
--- a/pootle/apps/pootle_store/forms.py
+++ b/pootle/apps/pootle_store/forms.py
@@ -298,6 +298,7 @@ def unit_form_factory(language, snplurals=None, request=None):
             return super(UnitForm, self).clean()
 
         def save(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


pootle/apps/pootle_store/forms.py
```diff
@@ -308,11 +309,18 @@ def save(self, *args, **kwargs):
                     unit.reviewed_by = None
                     unit.reviewed_on = None
                     unit._log_user = self.user
+                suggestion = self.cleaned_data["suggestion"]
+                user = (
+                    suggestion.user
+                    if suggestion
+                    else self.user)
```
unho: So if we have suggestion, shall we set `suggestion.user` as `submitter` and `self.user` as the `reviewer` for the submission below?

phlax: let me check the tests

unho:
lgtm

PR #6012, opened 2017-02-22
Ensure correct change is recorded when qcs are um/muted

phlax:
as in title


pootle/apps/pootle_store/models.py
```diff
@@ -865,17 +865,21 @@ def toggle_qualitycheck(self, check_id, false_positive, user):
...
+            unit=self,
+            store=self.store,
+            type=sub_type,
+            quality_check=check)
         sub.save()
 
         # update timestamp
         # log user action
-        self.save()
+        self.save(submitted_by=user, submitted_on=update_time)
```
ta2-1: @phlax it'd be great to see when and why we set submitted_by described in one place.
previously you could get a real translator via `submitted_by`. I'm not sure if it's expected that you set this field when muting/unmuting checks.

phlax: @ta2-1 atm - there is a lot of assumptions about what does/not happen - the tests that have been added this week have been to challenge those assumptions and in code at least document how these things worked.

I would be really up for some docs around this - but for now my concern is ensuring it does what we expect, and our expectations have been thought through

PR #6014, opened 2017-02-22
Disable appveyor until next release of ttk

phlax:
appveyor was already a bit flaky - but it seems not to work with a
-e install of ttk from github


dwaynebailey:
lgtm

PR #6019, opened 2017-02-22
Remove unit._implicit_vars

phlax:
remove _state_updated, _target_updated, _source_updated from unit.save and other places


unho:
lgtm

PR #6020, opened 2017-02-23
Remove implicit vars from store models

phlax:
more cleaning up unit.save

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -285,7 +285,12 @@ class Unit(AbstractUnit):
         comment_updated = (
             kwargs.pop("comment_updated", None)
             or self._comment_updated)
+
         action = kwargs.pop("action", None) or getattr(self, "_save_action", None)
+        if self.state == OBSOLETE and self._frozen.state > OBSOLETE:
+            action = UNIT_OBSOLETE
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -872,10 +872,7 @@ def toggle_qualitycheck(self, check_id, false_positive, user):
                          unit=self, store=self.store, type=sub_type,
                          quality_check=check)
         sub.save()
-
-        # update timestamp
-        # log user action
-        self.save()
+        self.save(revision=Revision.incr())
```
ta2-1: @phlax why do we need to set new unit revision here?

phlax: not 100% - but that was what i figured the effect of having those other vars set was

phlax: @ta2-1 - looking at it tho - i think i just wanted to be sure that the revision was incremented - maybe we dont need to - maybe - its just stats that need a kick

unho:
lgtm

pootle/apps/pootle_store/models.py
unho: Incrementing it is the safest choice.

ta2-1:
lgtm

PR #6022, opened 2017-02-23
Remove _auto_translated from unit

phlax:
this separates out the update_wordcount method which seemed to be overloaed with auto translation logic

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -237,7 +237,6 @@ class Unit(AbstractUnit):
         self._rich_source = None
         self._rich_target = None
         self._comment_updated = False
-        self._auto_translated = False
         self._encoding = 'UTF-8'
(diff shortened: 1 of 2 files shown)


unho:
lgtm

pootle/apps/pootle_store/models.py
```diff
@@ -278,10 +277,7 @@ def save(self, *args, **kwargs):
         reviewed_on = kwargs.pop("reviewed_on", None)
         submitted_by = kwargs.pop("submitted_by", None)
         submitted_on = kwargs.pop("submitted_on", None)
-
-        auto_translated = (
-            kwargs.pop("auto_translated", None)
-            or self._auto_translated)
+        auto_translated = False
```
ta2-1: where is this var used?

phlax: further down - something to do with revision - im hoping we can get rid tho

PR #6023, opened 2017-02-23
Remove old wordcount

phlax:
seems like this was evernote specific - we can readd if we feel its useful


docs/server/settings.rst
```diff
@@ -501,7 +501,6 @@ Translation environment configuration settings.
   Current options:
 
   - Translate Toolkit (default) - translate.storage.statsdb.wordcount
-  - Pootle - pootle.core.utils.wordcount.wordcount
```
unho: I would remove `POOTLE_WORDCOUNT_FUNC` setting altogether if we are dropping this function.

ta2-1: I think it is a valid pattern to overwrite wordcount function. probs it can be implemented using getter.

phlax: yep agreed re getter

unho:
lgtm

PR #6029, opened 2017-02-24
Remove user arg from unit.save

phlax:
this has been replaced by submitted_by, created_by, reviewed_by

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -267,6 +267,7 @@ class Unit(AbstractUnit):
 
     def save(self, *args, **kwargs):
         created = self.id is None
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -1196,7 +1197,7 @@ def addunit(self, unit, index=None, user=None, update_revision=None,
         if self.id:
             newunit.save(
                 revision=update_revision,
-                user=user,
+                created_by=user,
```
unho: Is this the only place where `unit.save` was passed `user`?

phlax: afaict - i put a pdb in the save method and ran the full test suite to catch any

PR #6030, opened 2017-02-24
Unit change signals

phlax:
move unit mangling to unit pre_save handler

### DIFF
--- a/pootle/apps/pootle_store/models.py
+++ b/pootle/apps/pootle_store/models.py
@@ -278,64 +278,13 @@ class Unit(AbstractUnit):
         reviewed_on = kwargs.pop("reviewed_on", None)
         submitted_by = kwargs.pop("submitted_by", None)
         submitted_on = kwargs.pop("submitted_on", None)
-        was_fuzzy = self._frozen.state == FUZZY
         sysuser = get_user_model().objects.get_system_user()
-        auto_translated = False
-
-        if self.source_updated:
-            # update source related fields
-            self.source_hash = md5(self.source_f.encode("utf-8")).hexdigest()
-            self.source_length = len(self.source_f)
-            wc = self.update_wordcount()
-            if not wc and not bool(filter(None, self.target_f.strings)):
-                # auto-translate untranslated strings
-                self.target = self.source
-                self.state = FUZZY
-                auto_translated = True
-        if self.target_updated:
-            # update target related fields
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/receivers.py
```diff
@@ -28,6 +34,60 @@ def handle_suggestion_added(**kwargs):
...
+        unit.submitted_on = None
+
+    # Updating unit from the .po file set its revision property to
+    # a new value (the same for all units during its store updated)
+    # since that change doesn't require further sync but note that
+    # auto_translated units require further sync
+    update_revision = (
+        not unit.revision
+        or (not unit.revision_updated
+            and (unit.updated and not auto_translated)))
```
unho: I would invert the order and check `not auto_translated` first.

phlax: yep, k

pootle/apps/pootle_store/models.py
```diff
@@ -263,69 +276,15 @@ def save(self, *args, **kwargs):
         changed_with = kwargs.pop("changed_with", None) or SubmissionTypes.SYSTEM
         reviewed_by = kwargs.pop("reviewed_by", None)
         reviewed_on = kwargs.pop("reviewed_on", None)
-        revision = kwargs.pop('revision', None)
         submitted_by = kwargs.pop("submitted_by", None)
```
unho: I think this line needs changing too:

```python
        submitted_by = kwargs.pop("submitted_by", self.submitted_by)
```

pootle/apps/pootle_store/receivers.py
phlax: actually not - i think its more confusing having the neg at the front

pootle/apps/pootle_store/models.py
```diff
@@ -278,64 +278,13 @@ def save(self, *args, **kwargs):
         reviewed_on = kwargs.pop("reviewed_on", None)
         submitted_by = kwargs.pop("submitted_by", None)
```
unho: Do not forget to default to `self.submitted_by` here.

phlax: i would rather not as that field will be removed asap

phlax: ..and doesnt seem to break anything

PR #6036, opened 2017-02-25
Submission actions

phlax:
add submission actions and remove actions from the submission types field

also will require a migration to calc/set the actions and update the old type fields


pootle/apps/pootle_statistics/models.py
```diff
@@ -25,6 +27,17 @@
 SIMILARITY_THRESHOLD = 0.5
 
 
+class SubmissionActions(Enum):
+    EDIT = 1
+    REVIEW = 2
+    REVERT = 3
+    SUGG_ADD = 4
+    SUGG_ACCEPT = 5
+    SUGG_REJECT = 6
```
unho: I suggest name all the actions like `VERB_WHAT`, for example `REJECT_SUGGESTION` or `EDIT_TRANSLATION`.

pootle/apps/pootle_statistics/models.py
```diff
@@ -25,6 +27,17 @@
 SIMILARITY_THRESHOLD = 0.5
 
 
+class SubmissionActions(Enum):
+    EDIT = 1
+    REVIEW = 2
```
unho: Review what?

pootle/apps/pootle_statistics/models.py
```diff
@@ -25,6 +27,17 @@
...
+    REVERT = 3
```
unho: Revert what?

phlax:
@unho the "what" is expressed in the `field` field

unho:
Fair point re "what".

phlax:
this is no longer required

PR #6038, opened 2017-02-25
Add migration to drop source submissions

phlax:
these are not currently created/used, so there is no real way of
interpreting the data in these submissions


pootle/apps/pootle_statistics/migrations/0014_truncate_source_subs.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.5 on 2017-02-25 15:53
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+# this was the value of SubmissionTypes.REVERT
```
unho: Wrong comment.

pootle/apps/pootle_statistics/migrations/0014_truncate_source_subs.py
```diff
@@ -0,0 +1,24 @@
...
+SOURCE_FLAG = 1
+
+
+def truncate_source_field(apps, schema_editor):
```
unho: I would use a more descriptive name like `remove_source_field_submissions`

phlax: it doesnt remove the source field - it removes any rows that have a value in this column - which is close to being `truncated` i think

PR #6040, opened 2017-02-26
Cleanup submissions

phlax:
- Add migration to set revisions on subs with matching data
- Remove dupe subs


pootle/apps/pootle_statistics/migrations/0017_remove_duplicate_subs.py
```diff
@@ -0,0 +1,50 @@
...
+    for dupe in dupes.iterator():
+        count += 1
+        to_remove = list(
+            subs.filter(creation_time=dupe["creation_time"])
+                .filter(unit_id=dupe["unit_id"])
+                .filter(type=dupe["type"])
+                .filter(suggestion_id=dupe["suggestion_id"])
+                .filter(field=dupe["field"])
+                .values_list("id", flat=True))[1:]
+        subs.filter(id__in=to_remove).delete()
```
unho: Is it not safer to delete the submissions after iterating the queryset?

pootle/apps/pootle_statistics/migrations/0016_infer_sub_revisions.py
```diff
@@ -0,0 +1,42 @@
...
+
+def infer_submission_revisions(apps, schema_editor):
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    revisions = subs.filter(revision__gt=0).values_list(
+        "creation_time", "unit_id", "revision")
+    updated = 0
+    count = 0
+    total = revisions.count()
+    start = time.time()
+    for creation_time, unit_id, revision in revisions.iterator():
```
unho: If `revisions` is grouped by `unit` and sorted by `revision` then we can pick the lowest revision to update all the submissions without revision for that unit.

pootle/apps/pootle_statistics/migrations/0017_remove_duplicate_subs.py
phlax: its slower this way, but safer as otherwise the list of pks gets too long and "mysql goes away" and also we can swell memory

i can check the numbers tho - it might be safe to delete them together

phlax: i think its safer to do it this way

PR #6046, opened 2017-02-28
Remove SUGG_ADD submissions

phlax:
these duplicate data that is held on the suggestion model

this PR removes the data from subs table, and temporarily disables creation add from timeline and scores


ta2-1:
lgtm

PR #6048, opened 2017-02-28
Drop suggestion reject subs

phlax:
these dont contain any data that is not in the suggestion so this PR drops this sub type.

It also temporarily disables scoring/timeline from suggestion rejection actions


pootle/apps/pootle_statistics/migrations/0017_drop_reject_suggestion_subs.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.5 on 2017-02-28 13:11
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.db.models import F
```
unho: Unused import.

pootle/apps/pootle_statistics/models.py
```diff
@@ -232,10 +231,10 @@ def get_submission_info(self):
                                                           check_name),
                     'checks_url': reverse('pootle-checks-descriptions'),
                 })
-
-        if (self.suggestion and
-            self.type in (SubmissionTypes.SUGG_ACCEPT,
-                          SubmissionTypes.SUGG_REJECT)):
+        is_suggestion = (
+            self.suggestion
+            and (self.type == SubmissionTypes.SUGG_ACCEPT))
```
unho: Nit: don't use unnecessary parentheses.

pootle/apps/pootle_statistics/models.py
```diff
@@ -563,8 +556,6 @@ def get_sugg_accepted():
             TranslationActionCodes.REVIEW_PENALTY: lambda: (-1) * reviewCost,
             TranslationActionCodes.SUGG_ACCEPTED: get_sugg_accepted,
             TranslationActionCodes.SUGG_REVIEWED_ACCEPTED: lambda: reviewCost,
-            TranslationActionCodes.SUGG_REJECTED: get_sugg_rejected,
-            TranslationActionCodes.SUGG_REVIEWED_REJECTED: lambda: analyzeCost,
```
unho: Are these two fields of `TranslationActionCodes` still needed?

pootle/apps/pootle_statistics/models.py
phlax: in this case - i prefer a little extra bracket as i feet it makes the condition clearer

pootle/apps/pootle_statistics/models.py
phlax: im not sure we are going to use in the future - but the scorelog entries that contained them would be auto removed

PR #6051, opened 2017-03-01
Remove SUGG_ACCEPT submission

phlax:
this carries the same data as the suggestion and increases code complexity and db size

this PR will convert existing SUGG_ACCEPT to type ~~NORMAL~~ WEB (ie web editing) and ensure that submissions that are created from a suggestion are always associated.

there are also seems to be historical suggestions that dont have any submission data - so we might want to clean them up in some way.


pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
```diff
@@ -0,0 +1,68 @@
...
+WEB_FLAG = 1
+SUGG_ACCEPT_FLAG = 3
+
+
+def _assoc_state_accept_suggestion_subs(apps):
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    suggestions = apps.get_model("pootle_store.Suggestion").objects
+    accepted = suggestions.filter(state__name="accepted").values(
+        "review_time", "pk", "unit_id", "reviewer")
+    state_acc = subs.filter(type=SUGG_ACCEPT_FLAG).filter(field=3)
```
ta2-1: is `3` target or state?

ta2-1: state

phlax: ill change to a constant to make it more obv

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
```diff
@@ -0,0 +1,68 @@
...
+
+    for suggestion in accepted.iterator():
+        suggestion["submitter"] = suggestion["reviewer"]
+        suggestion["creation_time"] = suggestion["review_time"]
```
ta2-1: microsecond accuracy?

phlax: if everything else matches - then i think we can assume its the correct - suggestion

phlax: in fact its not even an assumption - if the same user changed the state of the same unit at exactly the same time, by accepting a suggestion - then *it is* a dupe - and the suggestion is the same

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
```diff
@@ -0,0 +1,68 @@
...
+    # review time doesnt match
+    sugg_acc.exclude(suggestion__isnull=True).exclude(
+        creation_time=F("suggestion__review_time")).delete()
+
+    # reviewer doesnt match
+    sugg_acc.exclude(suggestion__isnull=True).exclude(
+        submitter=F("suggestion__reviewer")).delete()
+
+    # no suggestion
+    sugg_acc.filter(suggestion__isnull=True).count()
```
ta2-1: delete()

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
```diff
@@ -0,0 +1,68 @@
...
+        "review_time", "pk", "unit_id", "reviewer")
+    state_acc = subs.filter(type=SUGG_ACCEPT_FLAG).filter(field=3)
+
+    for suggestion in accepted.iterator():
+        suggestion["submitter"] = suggestion["reviewer"]
+        suggestion["creation_time"] = suggestion["review_time"]
+        suggestion_id = suggestion["pk"]
+        del suggestion["review_time"]
+        del suggestion["reviewer"]
+        del suggestion["pk"]
```
unho: Maybe I am wrong, but I feel it safer to create a temporary dict to set this data instead of reusing `suggestion`.

unho: In other words, explicit better than implicit.

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
unho: Why not do this one first? It will save you from having to exclude on the two previous querysets.

phlax: fair point

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
phlax: im not entirely sure what is more efficient - by i figured reusing existing dict is

pootle/apps/pootle_statistics/models.py
```diff
@@ -554,33 +532,10 @@ def get_paid_wordcounts(self):
...
-
-            return None, None
-
         def get_edited():
             return None, reviewed_words
 
         return {
             TranslationActionCodes.NEW: lambda: (translated_words, None),
             TranslationActionCodes.EDITED: get_edited,
-            TranslationActionCodes.REVIEWED: lambda: (None, reviewed_words),
```
unho: Why is this line removed?

pootle/apps/pootle_statistics/migrations/0019_convert_accept_sugg_subs.py
unho: I raised this because I was worried about what else can be in `suggestion`, thus my ask to make it explicit what we pass below. Anyway I notice now that it is used for filtering so it looks less concerning to me. I am happy to let it go.

phlax: the variables are explicit above in the `values_list`

pootle/apps/pootle_statistics/models.py
phlax: as i think it wouldnt work any longer as it did - and not thinking we will need it again

ta2-1: @phlax it looks like it is used still...

phlax: k @ta2-1 @unho - ill put this back for now

unho:
lgtm after addressing the concerning points.

PR #6053, opened 2017-03-01
Req: bump allauth pin to 0.31.0

dwaynebailey:
Need some second opinions.  Here is the ChangeLog https://github.com/pennersr/django-allauth/blob/master/ChangeLog.rst#0310-2017-02-28

I can't see anything problematic but I fear our test coverage is slim.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,7 +4,7 @@
 Django~=1.10.5  # rq.filter: <1.11
 
 # Django apps
-django-allauth==0.30.0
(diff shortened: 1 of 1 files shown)


unho:
Instead of just relying on the changelog I would also check https://github.com/pennersr/django-allauth/compare/76df7ee3...3d8f14e1

unho:
@dwaynebailey Reviewed and I saw no changes that can affect. lgtm

PR #6060, opened 2017-03-02
Silence field.W342 warning

dwaynebailey:
We don't use a OneToOneField for pootle_store.UnitSource.unit fore
technical reasons.  So silnce this to prevent us hiding other warnings.
Unfortunately we can't do this on a per field basis, so the whole check
is blocked.


unho:
lgtm

PR #6062, opened 2017-03-03
Submission unit cycle

phlax:
clean up of UnitLifecycle to ensure that subs are created in the correct order and with correct data

### DIFF
--- a/pootle/apps/pootle_store/utils.py
+++ b/pootle/apps/pootle_store/utils.py
@@ -391,54 +391,64 @@ class UnitLifecycle(object):
 
     def sub_comment_update(self, **kwargs):
         _kwargs = dict(
-            creation_time=self.unit.mtime,
+            creation_time=self.unit.change.commented_on,
             unit=self.unit,
-            submitter=self.unit.commented_by,
+            submitter=self.unit.change.commented_by,
             field=SubmissionFields.COMMENT,
-            type=SubmissionTypes.SYSTEM,
-            old_value=self.original.translator_comment,
-            new_value=self.unit.translator_comment)
+            type=self.unit.change.changed_with,
+            old_value=self.original.translator_comment or "",
+            new_value=self.unit.translator_comment or "")
         _kwargs.update(kwargs)
         return self.create_submission(**_kwargs)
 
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_store/utils.py
```diff
@@ -391,54 +391,64 @@ def save_subs(self, subs):
...
+            old_value=self.original.target or "",
             new_value=self.unit.target_f)
         _kwargs.update(kwargs)
         return self.create_submission(**_kwargs)
 
     def sub_state_update(self, **kwargs):
+        is_review = (
+            self.unit.change.reviewed_on
+            and (self.unit.change.reviewed_on
+                 >= self.unit.change.submitted_on))
```
unho: This looks like it can be a property of `unit.change`.

tests/pootle_store/utils.py
```diff
@@ -171,14 +200,13 @@ def test_unit_lifecycle_update_target(store0, member):
...
-    unit.submitted_by = member
-    unit.save()
+    unit.save(submitted_by=member)
     sub_target_update = lifecycle.get(Unit)(unit).sub_target_update()
     assert isinstance(sub_target_update, Submission)
     assert sub_target_update.unit == unit
     assert sub_target_update.translation_project == store0.translation_project
     assert sub_target_update.revision == unit.revision
-    assert sub_target_update.submitter == unit.submitted_by
+    assert sub_target_update.submitter == unit.change.submitted_by
```
unho: No `assert` to ensure it is `member`?

unho:
Nits aside, lgtm.

pootle/apps/pootle_store/utils.py
phlax: no - unit change is a model - and i think that would be confusing

tests/pootle_store/utils.py
phlax: i can add

pootle/apps/pootle_store/utils.py
unho: Models have convenience properties, but I can live with this as is.

PR #6064, opened 2017-03-03
Display meaningful error message when registering existing username

### LINKED ISSUE #5925: Form validation on login form exposes variables
if you try to signup with an existing username you get this

```
%(model_name)s with this %(field_label)s already exists.
```

unho:
Fixes #5925.


dwaynebailey:
lgtm, but is there anything upstream that needs to be fixed?

unho:
No, nothing needs fixing upstream.

This comes from our `User` model inheriting from `AbstractBaseUser` instead of `AbstractUser` (both Django abstracts, where `AbstractUser` inherits from `AbstractBaseUser`). `AbstractUser` includes this error message, but we don't inherit from that model to avoid having some unwanted fields in our `User` model.

unho:
Reported failing interpolation in https://code.djangoproject.com/ticket/27898#ticket

pootle/apps/accounts/models.py
```diff
@@ -60,6 +60,10 @@ class User(AbstractBaseUser):
                            _('Enter a valid username.'),
                            'invalid')
         ],
+        # Remove when fixed in Django: https://code.djangoproject.com/ticket/27898#ticket
```
dwaynebailey: That's not true, we always want a human readable error here I'd assume.  The Django issue is about the interpolation which we will no longer use.

dwaynebailey: Please try use something like `FIXME` as tools can flag that in future.  Apart from the fact that we don't need this comment at all.

dwaynebailey:
Minor nits otherwise lgtm

pootle/apps/accounts/models.py
unho: k

PR #6066, opened 2017-03-03
Update unit_lifecycle to use dict params not kwargs for order

phlax:
this allows the order in which subs are created to be determinate


ta2-1:
lgtm after tests pass
(nit: put comment or adjust tests to show how ordering can be used here)

unho:
lgtm

phlax:
@ta2-1 ordering cant be used as such - other than that the `calculate_change` method will create an ordereddict - and that is what determines the order in which the subs are created

PR #6068, opened 2017-03-03
Remove review rates

phlax:
These are no longer used in Pootle


unho:
lgtm

ta2-1:
Let me post it here just to highlight some points how scorelog `rate` and `review_rate` fields were used.
Those fields allowed to change score coefficients without touching real user score values.
Any configuration change can be applied for future scores. It might be important if we want reliable and consistent user scoring history.


There are hardcod`EDIT_COEF`, `REVIEW_COEF`

phlax:
> Let me post it here just to highlight some points how scorelog rate and review_rate fields were used.

There were a few problems with the existing implementation - one was that it measured change by checking the current state of units to calculate change - and not the state after the change.

Another problem was that "rates" were set per-user - this meant that one persons score was not necessarily comparable with another.

ta2-1:
@phlax all I wanted to say that if we need such feature as frozen past user scores after we change configuration we might want to add similar fields in future. Probs it can be more than two fields as it was previously but all configuration params.

PR #6070, opened 2017-03-04
Improve response handling for invalid suggestion form

phlax:
non-permissions get 404
otherwise 403 and helpful message

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -14,7 +14,7 @@ from translate.lang import data
 
 from django import forms
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


pootle/apps/pootle_store/views.py
```diff
@@ -688,7 +688,9 @@ def form_valid(self, form):
             self.get_context_data(form=form))
 
     def form_invalid(self, form):
-        raise PermissionDenied(form.errors)
+        if form.non_field_errors():
+            raise Http404
+        raise Http400(form.errors)
```
unho: Does this raise a 403 error?

phlax: no - title of PR updated

PR #6071, opened 2017-03-04
Add update_target flag to sugg_review.accept

phlax:
this allows us to suppress setting the target when accepting a
suggestion - so an edited version can be submitted

### DIFF
--- a/pootle/apps/pootle_store/utils.py
+++ b/pootle/apps/pootle_store/utils.py
@@ -199,9 +199,17 @@ class SuggestionsReview(object):
                 creation_time=make_aware(timezone.now()))
         return (suggestion, True)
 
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/utils.py
```diff
@@ -199,9 +199,17 @@ def add(self, unit, translation, user=None):
                 creation_time=make_aware(timezone.now()))
         return (suggestion, True)
 
-    def accept_suggestion(self, suggestion):
+    def accept_suggestion(self, suggestion, update_unit):
+        current_time = timezone.now()
```
unho: I am wondering if this should be enclosed in `make_aware`.

PR #6077, opened 2017-03-04
Submission times

phlax:
move suggestion saving to after unit to ensure consistent times

### DIFF
--- a/pootle/apps/pootle_store/utils.py
+++ b/pootle/apps/pootle_store/utils.py
@@ -191,17 +191,8 @@ class SuggestionsReview(object):
                 creation_time=make_aware(timezone.now()))
         return (suggestion, True)
 
-    def accept_suggestion(self, suggestion, update_unit):
-        current_time = timezone.now()
-        accepted = SuggestionState.objects.get(name="accepted")
+    def update_unit_on_accept(self, suggestion):
         unit = suggestion.unit
-        translation_project = unit.store.translation_project
-        suggestion.state_id = accepted.id
-        suggestion.reviewer = self.reviewer
-        suggestion.review_time = current_time
-        suggestion.save()
-        if not update_unit:
-            return
 
         # Save for later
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/utils.py
```diff
@@ -212,17 +203,26 @@ def accept_suggestion(self, suggestion, update_unit):
         unit.target = suggestion.target
         if unit.state == FUZZY:
             unit.state = TRANSLATED
-
+        unit.submitted_by = suggestion.user
+        unit.submitted_on = current_time
+        unit.reviewed_by = self.reviewer
+        unit.reviewed_on = unit.submitted_on
+        unit.save(
+            submitted_by=suggestion.user,
```
unho: This was already set a few lines above, is it necessary to pass it to `save()`?

pootle/apps/pootle_store/utils.py
```diff
@@ -212,17 +203,26 @@ def accept_suggestion(self, suggestion, update_unit):
...
             create_subs[SubmissionFields.STATE] = [old_state, unit.state]
         subs_created = []
         for field in create_subs:
             kwargs = {
-                'creation_time': current_time,
-                'translation_project': translation_project,
-                'submitter': self.reviewer,
+                'creation_time': unit.change.reviewed_on,
+                'translation_project': unit.store.translation_project,
+                'submitter': suggestion.user,
```
unho: Submissions for accepting suggestions were previously credited to the reviewer, but now to the suggester. Is that ok?

pootle/apps/pootle_store/utils.py
phlax: yes - because its used to set unit change/unit_source objects - the above lines will be removed pretty soon

unho: As long as this code is simplified as much as possible, I am happy.

PR #6087, opened 2017-03-08
Add cache expiry for tp deletion

### LINKED ISSUE #6085: Delete a TP from an old style project and the Project page stays cached
1. Create a new TP
2. TP is available
3. Delete TP
4. Project page still shows project listed - though it should be gone
5. Going to supposedly deleted TP and we get 404

We're not expiring cache when a TP is deleted.

phlax:
and some tests to ensure that cache is cleared

fixes #6085


unho:
lgtm

PR #6094, opened 2017-03-13
Submission attr

phlax:
ensure correct attribution and submission times in store update

### DIFF
--- a/pootle/apps/pootle_store/updater.py
+++ b/pootle/apps/pootle_store/updater.py
@@ -153,6 +153,13 @@ class UnitUpdater(object):
                 and self.unit.source != self.newunit.source)
 
     @cached_property
+    def resurrected(self):
+        return (
+            self.newunit
+            and not self.newunit.isobsolete()
+            and self.unit.isobsolete())
+
+    @cached_property
     def state_updated(self):
         # `fuzzy` state change is checked here.
         # `translated` and `obsolete` states are handled separately.
@@ -161,8 +168,6 @@ class UnitUpdater(object):
 
     @cached_property
     def target_updated(self):
-        if not self.newunit:
-            return False
         if not self.update.store_revision:
             return True
         if self.unit.target == self.newunit.target:
@@ -220,7 +225,7 @@ class UnitUpdater(object):
     def set_unit(self):
         if self.translator_comment_updated:
             self.set_commented()
-        if self.target_updated:
+        if self.target_updated or self.resurrected:
             self.set_submitted()
         self.save_unit()
         self.record_submission()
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/updater.py
```diff
@@ -153,6 +153,12 @@ def should_update_source(self):
                 and self.unit.source != self.newunit.source)
 
     @cached_property
+    def resurrected(self):
+        return (
+            self.newunit
+            and self.unit.isobsolete() != self.newunit.isobsolete())
```
ta2-1: this condition will be `True` for opposite change as well. Probs we need to be explicit here or just rename the method.

pootle/apps/pootle_store/updater.py
```diff
@@ -222,20 +228,23 @@ def set_unit(self):
...
             and self.update.store_revision is not None
             and (self.update.store_revision > self.unit.revision
                  and not self.update.resolve_conflict == POOTLE_WINS)
             and not self.newunit.isobsolete())
 
     def update_unit(self):
+        reordered = False
         suggested = False
         updated = False
         need_update = (self.should_unobsolete
```
unho: Unrelated, but `unobsolete` probably should be renamed to `resurrect`.

pootle/apps/pootle_store/updater.py
```diff
@@ -222,20 +228,23 @@ def set_unit(self):
             self.set_commented()
         if self.target_updated:
             self.set_submitted()
+        if self.resurrected:
```
unho: Nit: fold in previous `if`.

pootle/apps/pootle_store/updater.py
unho: This comparison might return True when unit is obsoleted.

unho: The condition needs to be changed. It returns `True` when unit is obsoleted.

pootle/apps/pootle_store/updater.py
```diff
@@ -153,6 +153,13 @@ def should_update_source(self):
                 and self.unit.source != self.newunit.source)
 
     @cached_property
+    def resurrected(self):
+        return (
+            self.newunit
+            and self.newunit.isobsolete()
+            and self.unit.isobsolete() != self.newunit.isobsolete())
```
unho: Nit: `and not self.unit.isobsolete()`

unho:
lgtm

pootle/apps/pootle_store/updater.py
phlax: i got this the wrong way round - but i got what you meant - and updated...

PR #6099, opened 2017-03-13
Adjust test to expect correct suggestion submitter

phlax:
seems like we are setting different things in different places


ta2-1:
lgtm

PR #6107, opened 2017-03-15
Add pootle_log.Log

phlax:
base log class for retrieving LogEvents

### DIFF
--- a/pootle/apps/pootle_log/utils.py
+++ b/pootle/apps/pootle_log/utils.py
@@ -143,3 +143,77 @@ class Log(object):
             end=kwargs.get("end"),
             field="unit__creation_time")
         return created_units
+
+    def get_created_units(self, **kwargs):
+        for created_unit in self.filtered_created_units(**kwargs):
+            yield self.event(
+                created_unit.unit,
+                created_unit.created_by,
+                created_unit.unit.creation_time,
+                "unit_created",
+                created_unit)
+
+    def get_submissions(self, **kwargs):
+        for submission in self.filtered_submissions(**kwargs):
+            event_name = "state_changed"
+            if submission.field == SubmissionFields.CHECK:
+                event_name = (
+                    "check_muted"
+                    if submission.new_value == "0"
+                    else "check_unmuted")
+            elif submission.field == SubmissionFields.TARGET:
+                event_name = "target_updated"
+            elif submission.field == SubmissionFields.SOURCE:
+                event_name = "source_updated"
+            elif submission.field == SubmissionFields.COMMENT:
+                event_name = "comment_updated"
+            yield self.event(
+                submission.unit,
+                submission.submitter,
+                submission.creation_time,
+                event_name,
+                submission)
+
+    def get_suggestions(self, **kwargs):
+        for suggestion in self.filtered_suggestions(**kwargs):
+            add_event = (
+                (not kwargs.get("start")
+                 or (suggestion.creation_time >= kwargs.get("start"))
+                 and (not kwargs.get("end")
+                      or (suggestion.creation_time < kwargs.get("end")))
+                 and (not kwargs.get("user")
+                      or (suggestion.user == kwargs.get("user")))))
+            review_event = (
+                not suggestion.state.name == "pending"
+                and (not kwargs.get("start")
+                     or (suggestion.review_time >= kwargs.get("start"))
+                     and (not kwargs.get("end")
+                          or (suggestion.review_time < kwargs.get("end")))
+                     and (not kwargs.get("user")
+                          or (suggestion.reviewer == kwargs.get("user")))))
+            if add_event:
+                yield self.event(
+                    suggestion.unit,
+                    suggestion.user,
+                    suggestion.creation_time,
+                    "suggestion_created",
+                    suggestion)
+            if review_event:
+                event_name = (
+                    "suggestion_accepted"
+                    if suggestion.state.name == "accepted"
+                    else "suggestion_rejected")
+                yield self.event(
+                    suggestion.unit,
+                    suggestion.reviewer,
+                    suggestion.review_time,
+                    event_name,
+                    suggestion)
+
+    def get_events(self, **kwargs):
+        for event in self.get_created_units(**kwargs):
+            yield event
+        for event in self.get_suggestions(**kwargs):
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,141 @@ def __init__(self, unit, user, timestamp, action, value,
...
+    def created_unit_model(self):
+        return UnitSource.objects
+
+    @property
+    def suggestion_model(self):
+        return Suggestion.objects
+
+    @property
+    def submission_model(self):
+        return Submission.objects
```
unho: This and the ones above are not models, but managers.

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,141 @@ def __init__(self, unit, user, timestamp, action, value,
...
+
+    def filter_path(self, qs, path, field="unit__store__pootle_path"):
+        return (
+            qs.filter(**{"%s__startswith" % field: path})
+            if path is not None
+            else qs)
+
+    def filter_store(self, qs, store, field="unit__store_id"):
+        return (
+            qs.filter(**{"%s" % field: store})
```
unho: Why not just simpler:

```python
            qs.filter(**{field: store})
```

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,141 @@ def __init__(self, unit, user, timestamp, action, value,
...
+            if store is not None
+            else qs)
+
+    def filter_times(self, qs, start, end, field="creation_time"):
```
unho: Shouldn't the default values for `start` and `end` be explicit here?

```python
    def filter_times(self, qs, start=None, end=None, field="creation_time"):
```

unho: Same for other methods.

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,141 @@ def __init__(self, unit, user, timestamp, action, value,
...
+
+    def filtered_submissions(self, **kwargs):
+        submissions = self.submission_model
+        submissions = self.filter_store(
+            self.submission_model,
+            kwargs.get("store"))
+        submissions = (
+            self.filter_user(
+                submissions,
+                kwargs.get("user")))
```
unho: Nit: This fits in a single line. I guess it was kept like this to ease reading.

pootle/apps/pootle_log/utils.py
phlax: yep

pootle/apps/pootle_log/utils.py
phlax: yep

pootle/apps/pootle_log/utils.py
phlax: quite a lot of refactoring - so probs it was longer previously

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,141 @@ def __init__(self, unit, user, timestamp, action, value,
...
+    @property
+    def created_unit_model(self):
+        return UnitSource.objects.select_related("unit", "created_by")
+
+    @property
+    def suggestion_model(self):
+        return Suggestion.objects.select_related("unit", "user", "reviewed_by")
+
+    @property
+    def submission_model(self):
```
unho: These are still managers.

pootle/apps/pootle_log/utils.py
unho: I think the same for `user` and other params used for filtering.

pootle/apps/pootle_log/utils.py
unho: I am ok with keeping this as is.

pootle/apps/pootle_log/utils.py
phlax: i think they might actually be qs as they have select_related

unho: You are right. But definitely not models ;)

unho:
Just wondering if there should be any filtering by unit as well.

phlax:
> Just wondering if there should be any filtering by unit as well.

i also wondered, but decided against on the basis that you should use UnitLog in that case

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,145 @@ def __init__(self, unit, user, timestamp, action, value,
...
+        return Submission.objects.select_related("unit", "submitter")
+
+    @cached_property
+    def event(self):
```
unho: Nit: rename to `event_model`?

pootle/apps/pootle_log/utils.py
```diff
@@ -18,3 +24,145 @@ def __init__(self, unit, user, timestamp, action, value,
...
+            for n
+            in ["SOURCE", "TARGET", "STATE", "COMMENT", "CHECK"]}
+
+    @cached_property
+    def subtypes(self):
+        return {
+            getattr(SubmissionTypes, n): n.lower()
+            for n in ["WEB", "UPLOAD", "SYSTEM"]}
+
+    def filter_path(self, qs, path, field="unit__store__pootle_path"):
```
unho: `path=None`

pootle/apps/pootle_log/utils.py
phlax: thats really not a model!

im gonna rename the others

pootle/apps/pootle_log/utils.py
```diff
@@ -143,3 +143,56 @@ def filtered_created_units(self, **kwargs):
...
+                 and (not kwargs.get("end")
+                      or (suggestion.creation_time < kwargs.get("end")))
+                 and(not kwargs.get("user")
+                     or (suggestion.user == kwargs.get("user")))))
+            review_event = (
+                (not kwargs.get("start")
+                 or (suggestion.review_time >= kwargs.get("start"))
+                 and (not kwargs.get("end")
+                      or (suggestion.review_time < kwargs.get("end")))
+                 and(not kwargs.get("user")
```
unho: no space before parentheses

unho:
lgtm

PR #6112, opened 2017-03-15
Score events

phlax:
updates tool and adds adapter for scoring

### DIFF
--- a/pootle/apps/pootle_score/apps.py
+++ b/pootle/apps/pootle_score/apps.py
@@ -19,4 +19,5 @@ class PootleScoreConfig(AppConfig):
 
     def ready(self):
         importlib.import_module("pootle_score.getters")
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_score/updater.py
```diff
@@ -6,8 +6,26 @@
...
         self.user = kwargs['user']
         self.store = kwargs['store']
+
+    @property
+    def event_class(self):
+        return LogEvent
+
+    @cached_property
+    def logs(self):
+        return log.get(self.store.__class__)(self.store)
```
ta2-1: do we need to pass `user` here?

phlax: actually - not here

phlax:
this was landed in #6135

PR #6117, opened 2017-03-16
Add update logic for score updater

phlax:
as in title


pootle/apps/pootle_score/updater.py
```diff
@@ -6,9 +6,57 @@
...
 
 class StoreScoreUpdater(object):
 
     def __init__(self, store, *args, **kwargs):
         self.store = store
         self.user = kwargs['user']
+
+    @property
+    def event_class(self):
+        return LogEvent
```
unho: Just wondering if it is actually needed to use a property to what it looks like a constant value. Is it not possible to just use a class attribute?

phlax: yep

PR #6140, opened 2017-03-18
Removed setting of the script prefix

claudep:
This was fixed in Django 1.10.


phlax:
lgtm, thanks @claudep

PR #6142, opened 2017-03-18
Replaced force_unicode by force_text

claudep:
Those are aliases, but force_text is compatible with Python 3.


claudep:
May I push this one?

dwaynebailey:
@claudep lgtm

claudep:
Thanks, merged in 251b100

PR #6143, opened 2017-03-18
Removed unneeded binary prefixes in migration files

claudep:


unho:
Since these were added automatically by Django I wonder if it would be problematic to drop them.

claudep:
No it's not. This was bugs in Django (see https://code.djangoproject.com/ticket/22564 and https://code.djangoproject.com/ticket/26385).

unho:
@claudep Thanks for the background. lgtm.

PR #6155, opened 2017-03-19
Improve unit save attribution

phlax:
allow commented_by to be passed in, and ensure correct reviewed_by (maybe)

### DIFF
--- a/tests/pootle_score/receivers.py
+++ b/tests/pootle_score/receivers.py
@@ -28,7 +28,7 @@ def test_user_tp_score_update_suggestions(store0, member, member2):
         date=timezone.now().date())
     old_suggested = current_score.suggested
 
-    old_score = current_score.score
+    old_score = round(current_score.score, 2)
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_store/models.py
```diff
@@ -318,18 +332,21 @@ def save(self, *args, **kwargs):
...
-                self.change.submitted_by = submitted_by or sysuser
+                self.change.submitted_by = submitted_by
                 self.change.submitted_on = submitted_on
             if reviewed_by is not None:
-                self.change.reviewed_by = reviewed_by or submitted_by
+                self.change.reviewed_by = reviewed_by
             if reviewed_on is not None or reviewed_by is not None:
-                self.change.reviewed_on = reviewed_on or submitted_on
+                self.change.reviewed_on = reviewed_on
+            if self.comment_updated:
```
unho: This `if` is a few lines above, why repeat it here?

PR #6160, opened 2017-03-20
Add testing for comment updates and submits

phlax:
which seems to be broken 8/


phlax:
closed in favour of #6161

PR #6170, opened 2017-03-21
README: use codecov badge

dwaynebailey:
Not sure if there is agreement on this, but PRing to see if there is any pushback.


phlax:
no objections - but codecov has gone down 8/

PR #6179, opened 2017-03-23
Migration cleanup

phlax:
existing migration was not correctly migrating data


unho:
lgtm once Travis passes.

PR #6184, opened 2017-03-23
Contact tests

unho:


tests/forms/contact.py
```diff
@@ -35,27 +38,35 @@ def test_contact_form(admin, rf, mailoutbox):
     assert message.from_email == settings.DEFAULT_FROM_EMAIL
     reply_to = u'%s <%s>' % (data['name'], data['email'])
     assert reply_to == message.extra_headers['Reply-To']
-    subject = u'[%s] %s' % (settings.POOTLE_TITLE, data['email_subject'])
+    assert [recipient_email] == message.recipients()
+    assert message.subject.startswith(u'[%s] ' % settings.POOTLE_TITLE)
```
phlax: why not just check `message.subject == subject` ?

unho: Not strictly necessary, for this line, but it is an error that haunted us for some time so I wanted it to be explicit. What you suggest is next line, btw.

tests/views/contact.py
```diff
@@ -0,0 +1,74 @@
...
+    url = '%s?report=%s' % (reverse('pootle-contact-report-error'), unit.pk)
+    client.login(username=admin.username, password=admin.password)
+    response = client.get(url)
+
+    assert 'contact_form_title' in response.context
+    url = reverse('pootle-contact-report-error')
+    assert response.context['contact_form_url'] == url
+
+    with pytest.raises(AttributeError):
+        assert response.context['form'].unit == unit
```
phlax: i think it would be clearer to
```
assert not hasattr(response.context["form"], "unit")
```

tests/views/contact.py
```diff
@@ -0,0 +1,74 @@
...
+        assert response.context['view'].unit == unit
```
phlax: if the first of these assertions raised, the second would not be run

tests/views/contact.py
```diff
@@ -0,0 +1,74 @@
...
+
+from pootle_store.models import Unit
+
+
+@pytest.mark.django_db
+def test_contact_report_form_view(client, admin, rf):
+    unit = Unit.objects.last()
+    url = '%s?report=%s' % (reverse('pootle-contact-report-error'), unit.pk)
+    client.login(username=admin.username, password=admin.password)
+    response = client.get(url)
```
phlax: whenever i see `client.get` or `client.post` im curious as to what the `response.status_code` is

tests/views/contact.py
```diff
@@ -0,0 +1,74 @@
...
+
+    with pytest.raises(AttributeError):
+        assert response.context['view'].unit is None
+    assert 'context' not in response.context['view'].get_initial()
+
+
+@pytest.mark.django_db
+def test_contact_report_form_view_blank_unit(client, admin, rf):
+    url = '%s?report=' % reverse('pootle-contact-report-error')
+    client.login(username=admin.username, password=admin.password)
```
phlax: im wondering if this form works differently for different users - its always testing with admin here

unho: The form doesn't care about users. But view that calls it, yes. Will look into expanding view tests.

tests/views/contact.py
unho: Will add some asserts for this.

tests/views/contact.py
unho: Good catch.

tests/views/contact.py
unho: Will change.

tests/views/contact.py
phlax: with view tests its generally a good idea to use the `request_users` fixture to test for each

tests/forms/contact.py
```diff
@@ -44,6 +44,40 @@ def test_contact_form(admin, rf, mailoutbox):
...
+    form = ContactForm(request=request, data=data)
+    assert not form.is_valid()
+
+    data['email_subject'] = "a" * 100
+    form = ContactForm(request=request, data=data)
+    assert form.is_valid()
+
+
+@pytest.mark.django_db
+def test_contact_form_mandatory_fields(admin, rf, mailoutbox):
```
phlax: i would use `_required_` as that is what django users

phlax:
lgtm

PR #6194, opened 2017-03-27
Ensure POOTLE_TITLE is Unicode

### LINKED ISSUE #6193: Handle unicode in `POOTLE_TITLE`
as reported by @hirnschmalz, it seems as tho using unicode in `POOTLE_TITLE` fails

dwaynebailey:
This ensures that setting POOTLE_TITLE to a Unicode string will work in
emails. allauth expects this to be unicode.

Fixes #6193 

This is the initial fix.  We might want to review the other settings we have to see if any are impacted.

### DIFF
--- a/pootle/settings/10-base.conf
+++ b/pootle/settings/10-base.conf
@@ -5,7 +5,7 @@
 # Dummy translate function so we can extract text
 _ = lambda x: x
 
-POOTLE_TITLE = _("Pootle Translation Server")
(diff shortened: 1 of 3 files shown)


unho:
The change looks gtm, but I am just wondering if it would be worth adding a Django check to ensure the setting is a unicode string.

phlax:
lets land this and raise a ticket/change existing for adding checks

lgtm

PR #6198, opened 2017-03-27
Tests: Use member for testing

unho:
Avoid testing with admin unless we actually need to test superuser.


phlax:
lgtm if travis agrees

PR #6205, opened 2017-03-30
Optis and cleanups

phlax:
as in title

### DIFF
--- a/pootle/apps/pootle_store/fields.py
+++ b/pootle/apps/pootle_store/fields.py
@@ -14,6 +14,7 @@ from translate.misc.multistring import multistring
 
 from django.db import models
 from django.db.models.fields.files import FieldFile, FileField
+from django.utils.functional import cached_property
 
 from pootle.core.utils.multistring import (parse_multistring,
                                            unparse_multistring)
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_store/fields.py
```diff
@@ -152,65 +135,23 @@ def realpath(self):
...
+
+        fileclass = self.instance.syncer.file_class
+        classes = {
+            str(self.instance.filetype.extension): fileclass,
+            str(self.instance.filetype.template_extension): fileclass}
+        return factory.getobject(
+            self.path,
+            ignore=self.field.ignore,
+            classes=classes)
         return self._store_tuple.store
```
ta2-1: return after return

pootle/apps/pootle_store/fields.py
```diff
@@ -152,65 +135,23 @@ def realpath(self):
         else:
             return self._store_tuple.realpath
```
unho: `_store_tuple` can be dropped too, now that it is not populated. `realpath` would need merging with `_get_realpath`.

pootle/apps/pootle_store/fields.py
unho: This line is never reached.

unho:
lgtm

PR #6208, opened 2017-03-31
Opti: dont get suggestion.state if not required

phlax:


ta2-1:
lgtm

pootle/apps/pootle_store/utils.py
```diff
@@ -176,14 +176,14 @@ def add(self, unit, translation, user=None):
...
         user = user or User.objects.get_system_user().id
-        pending = SuggestionState.objects.get(name="pending")
         try:
             suggestion = Suggestion.objects.pending().get(
                 unit=unit,
                 user_id=user,
                 target_f=translation)
             return (suggestion, False)
         except Suggestion.DoesNotExist:
+            pending = SuggestionState.objects.get(name="pending")
```
unho: You can just get the `id` using `values_list('id', flat=True)` but not sure if that will optimize it further.

phlax: yep - maybe a fraction

PR #6210, opened 2017-03-31
Remove __repr_ as it is not needed

dwaynebailey:
If `__repr__` is present X.objects.all() will use it.  But in that case
you are resppnsible to create a printable string, i.e. UTF8 NOT Unicode.

Without `__repr__` X.objects.all() will use `__unicode__` and will handle
all the UTF8 decode/encode required.

Moral of the story, just don't create a `__repr__` for any objects, rather
create `__unicode__` and all will be happy and well.


phlax:
lgtm

PR #6213, opened 2017-04-01
Cleanup: remove old resource caching

phlax:
this also saves a couple of hits on db when stores are saved


unho:
lgtm

phlax:
tested live - and not seeing any real performance degradation - we can add appropriate caching if we see a need

PR #6216, opened 2017-04-01
Opti: dont iterate projects and tps in base cmd

phlax:
minor speed up for command runner

### DIFF
--- a/pootle/apps/pootle_app/management/commands/__init__.py
+++ b/pootle/apps/pootle_app/management/commands/__init__.py
@@ -14,6 +14,7 @@ from django.core.management.base import BaseCommand, CommandError
 from pootle.runner import set_sync_mode
 from pootle_language.models import Language
 from pootle_project.models import Project
+from pootle_translationproject.models import TranslationProject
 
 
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_app/management/commands/__init__.py
```diff
@@ -141,20 +142,21 @@ def handle_all(self, **options):
...
+            tps = tps.filter(project__code__in=self.projects)
 
-            if self.languages:
-                tp_query = tp_query.filter(language__code__in=self.languages)
+        if self.languages:
+            tps = tps.filter(language__code__in=self.languages)
 
-            for tp in tp_query.iterator():
-                self.do_translation_project(tp, **options)
+        for tp in tps.iterator():
```
unho: It is a personal nit, but I would sort the TPs in some way, so when we run the commands we get the TPs processed in a predictable order, for example process the TPs for a given project first, and then for the next one...

unho: Damn, I see it is already done.

unho:
lgtm

PR #6217, opened 2017-04-01
Remove ALL_LANGUAGES from context_processors

phlax:
this was not very optimal


unho:
Broken localized names was reported in https://github.com/translate/pootle/issues/6218 Tested and filtering languages depending on regular user/superuser works ok.

gtm, despite it would be better to have tests.

phlax:
we can add some tests when the l10n is fixed ref #6218

PR #6236, opened 2017-04-08
Generalized force_bytes call to md5 arguments

claudep:


ta2-1:
@claudep PR looks good.
The only doubt was about `unitid_hash` data which is stored in DB and I checked that adding `force_bytes` wouldn't change a real `md5` value.

PR #6239, opened 2017-04-10
Use styling instead of l10n

unho:


dwaynebailey:
@unho lgtm, my one concern though is that with >>> and <<< (not a fan tbh) I'm not sure if they flip the right way for RTL and that they behave as expected as in they actually swap places.

unho:
I am not a fan of >>> and <<<  either, but @phlax wanted those back. The styling added in this PR is specifically aimed to fix rendering of those in RTL.

PR #6257, opened 2017-04-20
Cleanup: remove condition for target change in updater

phlax:
this condition was added - but we are going to need to do this differently so removing


ta2-1:
lgtm

PR #6265, opened 2017-04-21
Fix: make run_given_checks work correctly

phlax:
with TeeChecker

and correctly separate translated units in mass checker

### DIFF
--- a/pootle/apps/pootle_misc/checks.py
+++ b/pootle/apps/pootle_misc/checks.py
@@ -971,7 +971,19 @@ def run_given_filters(checker, unit, check_names=None):
     failures = {}
 
     for functionname in check_names:
-        filterfunction = getattr(checker, functionname, None)
+        if isinstance(checker, checks.TeeChecker):
+            for _checker in checker.checkers:
+                filterfunction = getattr(_checker, functionname, None)
+                if filterfunction:
+                    checker = _checker
+                    checker.str1 = data.normalized_unicode(unit.source) or u""
+                    checker.str2 = data.normalized_unicode(unit.target) or u""
+                    checker.language_code = unit.language_code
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_misc/checks.py
```diff
@@ -971,7 +971,19 @@ def run_given_filters(checker, unit, check_names=None):
...
+            for _checker in checker.checkers:
+                filterfunction = getattr(_checker, functionname, None)
+                if filterfunction:
+                    checker = _checker
+                    checker.str1 = data.normalized_unicode(unit.source) or u""
+                    checker.str2 = data.normalized_unicode(unit.target) or u""
+                    checker.language_code = unit.language_code
+                    checker.hasplural = unit.hasplural()
+                    checker.locations = unit.getlocations()
+                    break
```
unho: This means that only the first checker in teechecker is run, but the purpose of teechecker is to give the ability to run checks from several checkers. This `break` defuses that.

phlax: it means the first one with a matching check

unho: I see now that the code is correct, despite it only taking the first filter matching the provided name.

unho: Is `filterfunction` unset if none of the checkers has any matching filter?

phlax: yep - and as a consequence any exsting checks would be deleted

phlax: ie

```
pootle calculate_checks --check-name=my_old_check_thats_no_longer_used
```
would delete the corresponding check

unho:
lgtm

PR #6286, opened 2017-04-24
Opti: use data.revision in store diff

phlax:
it should be safe to use `store.data.max_unit_revision` in store diff (if its not it means that a store wasnt updated correctly)


unho:
lgtm

PR #6321, opened 2017-05-02
Debug utils

phlax:
with this PR running any code from tests,  you can do:

```
pdb.set_trace()
```
without the need to `import pdb` first

similarly you can do

```
with debug_sql():
    debug_some_code()
```

without importing `debug_sql` first

As a further convenience, if you want to pdb etc at a certain code point, but only when it follows a certain code path you can do

```
_trace.debug = True
```
higher/previously in the stack you want to trace

and then
```
if _trace.debug: pdb.set_trace()
```
at the codepoint you want to break on.

these additions should be (more) production safe as pyflakes will see `_trace` `pdb` and `debug_sql` as missing locals if found anywhere in the code

### DIFF
--- a/pootle/apps/pootle_data/utils.py
+++ b/pootle/apps/pootle_data/utils.py
@@ -10,10 +10,9 @@ from django.db import models
 from django.db.models import Sum
 from django.utils.functional import cached_property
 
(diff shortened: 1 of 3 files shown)


ta2-1:
lgtm

PR #6326, opened 2017-05-03
Fix: embarassing mem leak in bulk_operations

phlax:
as in title 8/

### DIFF
--- a/pootle/apps/pootle_translationproject/contextmanagers.py
+++ b/pootle/apps/pootle_translationproject/contextmanagers.py
@@ -25,10 +25,9 @@ from pootle_store.models import Store
 
 class Updated(object):
     data = None
-    checks = set()
+    checks = None
     tp_data = False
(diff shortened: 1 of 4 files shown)


pootle/apps/pootle_translationproject/contextmanagers.py
```diff
@@ -25,10 +25,9 @@
 
 class Updated(object):
     data = None
-    checks = set()
+    checks = None
     tp_data = False
-    dirs = set()
```
unho: No `dirs`?

phlax:
```
(pootle)1 phlax@3ba624ac16c9:~/.virtualenvs/pootle/pootle$ git grep updated.dirs
(pootle)1 phlax@3ba624ac16c9:~/.virtualenvs/pootle/pootle$
```
not used i think

PR #6336, opened 2017-05-04
Bulk checks cleanup

phlax:
minor cleanups for checks

also passes through units to be updated correctly to store checker


unho:
lgtm

PR #6343, opened 2017-05-05
Add contextmanager for debugging mem usage/retention

phlax:
in blocks of code

can be used as in:

```python
from pootle.core.debug import memusage

with memusage() as usage:
     leak_some_mem()

print(usage)
```


unho:
lgtm

PR #6359, opened 2017-05-09
Migration hacks

phlax:
squash pootle_statistics migrations 11-23


unho:
lgtm

pootle/apps/pootle_store/migrations/0035_set_created_by_again.py
```diff
@@ -3,45 +3,70 @@
...
 
 
-def set_unit_created_by(apps, schema_editor):
-    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+UNIT_SOURCE_SQL = (
+    "INSERT INTO `pootle_store_unit_source` (`unit_id`, `created_with`, `created_by_id`) "
+    "(SELECT `id` as `unit_id`, "
+    "  %s as `created_with`, "
+    "  %s as `created_by_id` "
+    "  from `pootle_store_unit`)")
```
unho: Nit: Use `FROM` and `AS` instead of lowercase.

pootle/apps/pootle_store/migrations/0035_set_created_by_again.py
```diff
@@ -52,4 +77,4 @@ class Migration(migrations.Migration):
 
     operations = [
         migrations.RunPython(set_unit_created_by),
-]
+    ]
```
unho: Nit: undo this change.

pootle/apps/pootle_store/migrations/0034_unitsource_squashed_0037_unitsource_fields.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class Migration(migrations.Migration):
+
+    replaces = [
+        (b'pootle_store', '0034_unitsource'),
+        (b'pootle_store', '0035_set_created_by_again'),
+        (b'pootle_store', '0036_unitchange'),
+        (b'pootle_store', '0037_unitsource_fields'),
+        (b'pootle_store', '0046_unit_source_one_to_one.py')]
```
unho: Shouldn't all these be removed?

dwaynebailey: `replaces =` not `depends =`

pootle/apps/pootle_store/migrations/0034_unitsource_squashed_0037_unitsource_fields.py
```diff
@@ -0,0 +1,123 @@
...
+    cursor.execute(UNIT_SOURCE_SQL % (SubmissionTypes.SYSTEM, sysuser))
+
+
+def create_with_orm(apps, creators, sysuser):
+    UnitSource = apps.get_model("pootle_store.UnitSource")
+    units = apps.get_model("pootle_store.Unit").objects.all()
+
+    def _unit_source_create(pk):
+        return dict(
+            unit_id=pk,
```
phlax: this needs to copy fields

pootle/apps/pootle_store/migrations/0034_unitsource_squashed_0037_unitsource_fields.py
```diff
@@ -0,0 +1,123 @@
...
+
+def set_unit_created_by(apps, schema_editor):
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    sysuser = get_system_user_id()
+    # type 10 is the now deleted UNIT_CREATE
+    creators = dict(
+        subs.filter(type=10)
+            .exclude(submitter_id=sysuser)
+            .values_list("unit_id", "submitter"))
+    if schema_editor.connection.vendor in ["mysql", "postgres"]:
```
phlax: mysql only

pootle/apps/pootle_store/migrations/0034_unitsource_squashed_0037_unitsource_fields.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def update_with_orm(apps, creators):
+    UnitSource = apps.get_model("pootle_store.UnitSource")
+
+    def _set_created_by(unit_source):
+        unit_source.created_by_id = creators.get(unit_source.unit_id)
+        return unit_source
+    Batch(UnitSource.objects, batch_size=500).update(
+        list(UnitSource.objects.filter(unit_id__in=creators)),
```
phlax: creators.keys() would be a bit more explicit

pootle/apps/pootle_store/migrations/0035_set_created_by_again.py
```diff
@@ -3,45 +3,70 @@
...
-                created_by_id=creators.get(pk, sysuser))
-             for pk
-             in units[offset: offset + step].values_list("id", flat=True)])
-        logger.debug(
-            "added %s/%s in %s seconds"
-            % (offset + step, total, (time.time() - start)))
-        if offset > total:
-            break
-        offset = offset + step
+    if schema_editor.connection.vendor in ["mysql", "postgres"]:
```
phlax: mysql only

pootle/apps/pootle_store/migrations/0035_set_created_by_again.py
```diff
@@ -3,45 +3,70 @@
...
-    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+UNIT_SOURCE_SQL = (
+    "INSERT INTO `pootle_store_unit_source` (`unit_id`, `created_with`, `created_by_id`) "
+    "(SELECT `id` as `unit_id`, "
+    "  %s as `created_with`, "
+    "  %s as `created_by_id` "
+    "  from `pootle_store_unit`)")
+
+
+def create_with_sql(sysuser, schema_editor):
```
phlax: use the methods etc from 0034

phlax: not sure that is right/possible

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+    cursor.execute(CLEANUP_STORE_DATA_SQL)
+    cursor.execute(CLEANUP_TP_DATA_SQL)
+    cursor.execute(CLEANUP_SQL)
+
+
+def cleanup_subs_with_orm(apps):
+    # type 10 is the previous SubmissionTypes.UNIT
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    (subs.filter(field=1)
+     | subs.filter(type__in=[2, 8, 9, 10])).delete()
```
unho: This doesn't include `unit__isnull=True`.

phlax: good point - it needs to batch also

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+        subs.filter(**suggestion).update(suggestion_id=suggestion_id)
+
+
+def _delete_redundant_accept_sugg_subs(apps):
+    subs = apps.get_model("pootle_statistics.Submission").objects
+    subs = subs.filter(type=SUGG_ACCEPT_FLAG)
+    # these appear to be caused by dupes
+    to_delete = (
+        subs.filter(suggestion__isnull=True)
+        | subs.exclude(creation_time=F("suggestion__review_time"))
```
phlax: im wondering if this can be combined with the deletion above

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class Migration(migrations.Migration):
+
+    replaces = [
+        (b'pootle_statistics', '0011_cleanup_submissions'),
+        (b'pootle_statistics', '0012_remove_create_subs'),
+        (b'pootle_statistics', '0013_remove_revert_flag'),
+        (b'pootle_statistics', '0014_truncate_source_subs'),
+        (b'pootle_statistics', '0015_remove_system_scorelogs'),
```
unho: I don't see any code dropping `system` scorelogs in this squashed migration.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: I guess it is ok since unit creation submissions are no longer deleted either.

unho: Wrong. Unit creation subs are deleted.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+
+def cleanup_subs_with_sql(schema_editor):
+    cursor = schema_editor.connection.cursor()
+    cursor.execute(CLEANUP_STORE_DATA_SQL)
+    cursor.execute(CLEANUP_TP_DATA_SQL)
+    cursor.execute(CLEANUP_SQL)
+
+
+def cleanup_subs_with_orm(apps):
+    # type 10 is the previous SubmissionTypes.UNIT
```
unho: IMHO it would be better to use constants instead of comments. It helps to understand the code better.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: Good point.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+        migrations.RemoveField(
+            model_name='scorelog',
+            name='rate',
+        ),
+        migrations.RemoveField(
+            model_name='scorelog',
+            name='review_rate',
+        ),
+        migrations.DeleteModel(
+            name='ScoreLog',
```
unho: I have never tried it, but is it not possible to just drop the `ScoreLog` model instead of deleting fields from it and then dropping the `ScoreLog` model as we are doing here?

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+    subs = apps.get_model("pootle_statistics.Submission").objects.all()
+    (subs.filter(field=1)
+     | subs.filter(type__in=[2, 8, 9, 10])).delete()
+
+
+def cleanup_subs(apps, schema_editor):
+    if schema_editor.connection.vendor == "mysql":
+        cleanup_subs_with_sql(schema_editor)
+    else:
+        cleanup_subs_with_orm(apps)
```
unho: `cleanup_subs_with_orm` and `cleanup_subs_with_sql` don't look like they do the same stuff.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: `subs.filter(suggestion__isnull=True)` can't be moved above because it depends on first other code populating this field.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def _delete_redundant_accept_sugg_subs(apps):
+    subs = apps.get_model("pootle_statistics.Submission").objects
+    subs = subs.filter(type=SUGG_ACCEPT_FLAG)
+    # these appear to be caused by dupes
+    to_delete = (
+        subs.filter(suggestion__isnull=True)
+        | subs.exclude(creation_time=F("suggestion__review_time"))
+        | subs.exclude(submitter=F("suggestion__reviewer")))
```
unho: For these three OR conditions I would keep the comments from original migration as they clearly state what means each condition. This one for example is "reviewer doesnt match".

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,175 @@
...
+        (b'pootle_statistics', '0016_drop_add_suggestion_subs'),
+        (b'pootle_statistics', '0017_drop_reject_suggestion_subs'),
+        (b'pootle_statistics', '0018_remove_submission_store'),
+        (b'pootle_statistics', '0019_convert_accept_sugg_subs'),
+        (b'pootle_statistics', '0020_convert_checks_subs'),
+        (b'pootle_statistics', '0021_remove_paid_rates'),
+        (b'pootle_statistics', '0022_set_suggestion_submitters'),
+        (b'pootle_statistics', '0023_remove_scorelog')]
+
+    dependencies = [
```
unho: I would also add explicit dependency on `('pootle_store', '0035_set_created_by_again'),`

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
```diff
@@ -0,0 +1,437 @@
...
+        (b'pootle_store', '0042_add_default_suggestion_states'),
+        (b'pootle_store', '0043_suggestion_state'),
+        (b'pootle_store', '0044_set_new_suggestion_states'),
+        (b'pootle_store', '0045_remove_suggestion_tmp_state'),
+        (b'pootle_store', '0046_unit_source_one_to_one'),
+        (b'pootle_store', '0047_remove_old_unit_fields'),
+        (b'pootle_store', '0048_set_change_commented'),
+        (b'pootle_store', '0049_remove_unit_commented'),
+        (b'pootle_store', '0050_set_change_reviewed'),
+        (b'pootle_store', '0051_remove_unit_reviewed'),
```
unho: Missing `('pootle_store', '0051_remove_unit_commented'),`

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: the scorelogs model is deleted, so its not necessary

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: i think this is how you are supposed to do it, but i can check

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: im pretty sure that would create a circular dependency

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
phlax: i think this is correct

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: other than in the way that you commented above ?

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: actually - realised it wouldnt - but it would be pointless

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
```diff
@@ -0,0 +1,437 @@
...
+        migrations.AlterField(
+            model_name='unitchange',
+            name='reviewed_by',
+            field=models.ForeignKey(null=True, on_delete=models.SET(pootle.core.user.get_system_user), related_name='reviewed', to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AlterField(
+            model_name='unitchange',
+            name='submitted_by',
+            field=models.ForeignKey(null=True, on_delete=models.SET(pootle.core.user.get_system_user), related_name='submitted', to=settings.AUTH_USER_MODEL),
+        ),
```
unho: These three `AlterField` can probably be folded into `CreateModel` that are even further above.

unho: Note the code to populate the fields might need to be altered.

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
```diff
@@ -0,0 +1,437 @@
...
+        migrations.AddField(
+            model_name='suggestion',
+            name='state',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='suggestions', to='pootle_store.SuggestionState'),
+        ),
+        migrations.RunPython(set_suggestion_states),
+        migrations.RemoveField(
+            model_name='suggestion',
+            name='tmp_state',
+        ),
```
unho: Re this sequence of commands to make the `Suggestion.state` to be a FK, I wonder if it is simpler/better to instead:

- Create `tmp_state` FK field,
- Populate `tmp_state` FK field from `state` plain field,
- Remove old `state` plain field,
- Rename `tmp_state` FK field to `state`.

It looks simpler than current:

- Create new `tmp_state` plain field
- Copy data from old `state` plain field
- Remove `state` plain field
- Add new `state` FK field
- Populate new `state` FK field
- Remove `tmp_state` plain field

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: Yes, I think so. As far as I can see `cleanup_subs_with_orm` doesn't provide an alternative to `CLEANUP_STORE_DATA_SQL` or `CLEANUP_TP_DATA_SQL`.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: k

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
unho: ok, wfm leaving as it is.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: it doesnt need to - django takes care of that

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
phlax: they cant - they set the related fields that were previously used by unit attributes

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
phlax: im kinda happy that this works and is not too inefficient

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
unho: k

pootle/apps/pootle_store/migrations/0027_unit_created_by_squashed_0055_fill_unit_source_data.py
unho: I only suggested it in order to reduce the number of operations needed in order to ease reviewing the squashing. My proposal looks like it should be more performant too.

phlax: the only difference performance wise is storing the fk - but to be a performance boost i think it would need to be stored as an int not fk

phlax: actually - using rename is probs faster too

phlax: ill update

phlax: updated to use rename, seems to work and a bit faster, thanks

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: seems like not required, removed

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: k

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,191 @@
...
+
+# Old types
+REVERT_TYPE = 2
+SUGG_ACCEPT_TYPE = 3
+SUGG_ADD_TYPE = 8
+SUGG_REJECT_TYPE = 9
+UNIT_CREATE_TYPE = 10
+
+# Fields
+SOURCE_FIELD = 1
```
unho: Unused?

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,191 @@
...
+    _cleanup_accept_suggestion_subs(apps)
+
+
+class Migration(migrations.Migration):
+
+    replaces = [
+        (b'pootle_statistics', '0011_cleanup_submissions'),
+        (b'pootle_statistics', '0012_remove_create_subs'),
+        (b'pootle_statistics', '0013_remove_revert_flag'),
+        (b'pootle_statistics', '0014_truncate_source_subs'),
```
unho: I don't see the changes from this migration anywhere.

phlax: this one has been excluded - i raised a pr to also remove the code from that migration because i dont think this was the correct thing to do

phlax: but i didnt land the pr - because it also removed other code which we want to keep #6179

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,191 @@
...
+    "DELETE FROM `pootle_app_submission` "
+    " WHERE (`unit_id` IS NULL"
+    "        OR `pootle_app_submission`.`type` in (%s))")
+
+
+def cleanup_subs_with_sql(schema_editor):
+    cursor = schema_editor.connection.cursor()
+    _old_types = ", ".join(str(t) for t in OLD_TYPES)
+    cursor.execute(CLEANUP_STORE_DATA_SQL % _old_types)
+    cursor.execute(CLEANUP_TP_DATA_SQL % _old_types)
```
unho: Why these two SQL queries above don't have a matching ORM call in `cleanup_subs_with_orm`?

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: https://github.com/translate/pootle/pull/6179 is about `pootle_store` migrations. I don't see the relevance.

ta2-1: I'm not clear how #6179 is related to the 0014 migration but I agree to skip it.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: as i explained previously, the orm takes care of nulling these columns

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: my bad - that should have been #6368 - apologies for the confusion

phlax: specifically this commit https://github.com/translate/pootle/pull/6368/commits/b67825e456deee76994bafe431d53ee5edf21e38 - which i can add to this PR

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
phlax: yep, i can remove

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: I would add some comment on code clearly stating this.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
unho: +1 on adding the commit in this PR.

pootle/apps/pootle_statistics/migrations/0011_cleanup_submissions_squashed_0023_remove_scorelog.py
```diff
@@ -0,0 +1,191 @@
...
+
+def _cleanup_accept_suggestion_subs(apps):
+    # state/target changes that happened from suggestion had type=3,
+    # and state change subs did not have their suggestion associated
+    # They also credited the reviewer not the submitter
+    # we need to:
+    # - associate any state change subs with relevant suggestions
+    # - set the suggester as submitter for all type=3 submissions
+    # - update type to WEB for all type=3 subs that have suggestions
+    # any remaining submissions of this type are likely dupes and will be deleted
```
unho: Where are these deleted?

phlax: here 

https://github.com/translate/pootle/pull/6359/files#diff-6662dc20f40422a0dbe59d8ee6bd9663R55

and here

https://github.com/translate/pootle/pull/6359/files#diff-6662dc20f40422a0dbe59d8ee6bd9663R74

unho: ok

unho:
lgtm

ta2-1:
lgtm

PR #6375, opened 2017-05-14
Add user events to user profile util

phlax:
for use in user profile views

### DIFF
--- a/pootle/apps/pootle_profile/utils.py
+++ b/pootle/apps/pootle_profile/utils.py
@@ -6,9 +6,14 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+from datetime import timedelta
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_profile/utils.py
```diff
@@ -18,6 +23,18 @@ def __init__(self, user):
         self.user = user
 
     @cached_property
+    def avatar(self):
+        return render_as_template(
+            "{% load common_tags %}{% avatar username email_hash 20 %}",
```
unho: I must say I am not particularly happy about having template code in views.

phlax: i dont really see a problem with it - altho it does feel something of an antipattern - its more efficient than using another inclusion tag i think

unho:
lgtm

PR #6378, opened 2017-05-15
l10n: Change team roles labels

unho:
After talking with some translators they found previous naming misleading.
New labels align with the Pootle concepts of translating and suggesting
offered to users in the translation editor.


phlax:
i would really rather you didnt do this for all of the reasons discussed previously

PR #6381, opened 2017-05-16
User profile view

phlax:
update user profile view, makes use of new profile inclusion tags, adds team info to user profile

### DIFF
--- a/pootle/templates/user/profile.html
+++ b/pootle/templates/user/profile.html
@@ -2,109 +2,24 @@
 {% load core i18n profile_tags assets common_tags %}
 
 {% block body_id %}profile-page{% endblock %}
-
 {% block header_block %}{% endblock %}
 
 {% block content %}
-<div id="user-profile" lang="{{ LANGUAGE_CODE }}">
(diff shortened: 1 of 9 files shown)


pootle/templates/user/profile.html
```diff
@@ -115,12 +30,12 @@ <h3 class="user-position">
...
-    userData: {{ user.field_values|to_js }},
-    appRoot: '{{ user.get_absolute_url }}',
-  });
-});
+    $(function() {
+        PTL.user.init({
+            userData: {{ user.field_values|to_js }},
+            appRoot: '{{ user.get_absolute_url }}',
+        });
+    });
```
unho: Why the whitespace here?

phlax: because it should be indented from the script tag

unho: Not necessarily, but even having that in account in templates we indent by 2 spaces at a time and here you are using 4 spaces.

unho:
lgtm

PR #6384, opened 2017-05-16
Only use last 30 days to calculate user score

phlax:
atm score is calculated from all tpscores, which are stored by date

if we score on all then calculating scores becomes increasingly expensive

also, this restores previous behaviour


dwaynebailey:
I think 30 days was the right way to go.  It means scores are shifting and we reward current activity not historic activity.  Since it restoring what we did previously I don't think there is much need to explain this to users which helps also.

unho:
lgtm

PR #6385, opened 2017-05-16
Only use revision to match submissions to accepted suggestion

phlax:
when a suggestion is accepted, the system needs to associate the
submissions that were created.

previously it did this in a post_save handler by matching revision
and review_time, but this seems to be flaky on mysql where it lacks
microsecond support

this moves the association to the suggestion review util, and only
uses the units revision to match the submissions


dwaynebailey:
Tested locally and this solves the Suggest, Accept scenario where the Accept event didn't show any meta info and description.  lgtm based on that test.

PR #6393, opened 2017-05-17
i18n: ensure strings are Unicode in mute/unmute timeline messages

dwaynebailey:
Certainly self.check_url, does not interpolate into the Mute, Unmute
strings if it is Unicode, it works for ASCII.  Using u"" solves the
issues and strings interpolate correctly.


ta2-1:
lgtm

dwaynebailey:
Landed in 666aafeb8c3a503bbf6c9542afd31e31681bd695 and 981949045fd2bafc608603c33e07d5d0317c5657

PR #6396, opened 2017-05-17
Dont score suggestions with no creation_time

phlax:
this resolves an issue where suggestions are being scored for current day after refresh_scores and affects historical suggestions


unho:
lgtm though not tested.

PR #6403, opened 2017-05-18
l10n: Simplify translatable strings

unho:
This avoids translators breaking the UI with translations, and also
presents them simpler strings to translate.

The removed tags and CSS is not needed to render the tooltips.


dwaynebailey:
lgtm, not 100% sure why we need force_escape this is coming via babel via CLDR and will be `10,0%` like strings.  It kinda feels potentially like overkill, but not a blocker.

unho:
@dwaynebailey `force_escape` is being used because this is not only stuff coming from babel but also to sanitize the translations.

Issue #4184, opened 2015-11-13
Make Markdown the default content format


dwaynebailey:
At the moment we fall back to HTML for `POOTLE_MARKUP_FILTER`.  This seems like a safe default but means that most people then are required to write HTML and might not even know that they can use simpler Markdown if they configure things.

I'd like us to change this so that:
1. Markdown is the default and is installed by default.
2.  Choosing HTML is a concious decision.

This would mean that the default is sane and simple for anyone writing pages.  You do lose complete control, and that's for the HTML people, but you gain a very simple way of adding pages.

dwaynebailey:
@translate/evernote can we have some thoughts on this?

julen:
I totally love markdown. Having said that, I believe HTML is a saner default because:
* It's more widespread than other markup formats, and more people are familiar with it.
* It requires no extra dependencies.

> people then are required to write HTML and might not even know that they can use simpler Markdown if they configure things

Arguably the same can be said about any default markup format. Is discoverability perhaps the real issue?

cc @iafan for extra thoughts.

phlax:
@julen i think the problem is that atm users are by default encouraged to edit html in a non-html editor - which even for experienced coders can be a pita. Although  i can definitely understand the reasoning to offer html input.

Having markdown as the default seems to offer the right balance of user-editability

julen:
The editor highlights the markup mode you are on and the live preview is immediate (this is even an extra advantage for HTML as it requires no server round-trips). This has been a recent improvement; did you miss this feature perhaps?

phlax:
no i see that - but im still not that keen on editing HTML outside my editor - and generally against asking end users to edit it without an WYSIWYG editor

iafan:
Can we probably use both at the same time? Markdown itself allows for inline HTML. We could use Markdown by default, seamlessly adding HTML tags and CSS styles as they need them.

With that being said (and if everyone agrees that this is something we'd like to explore), I'd suggest the following:

1. Make Markdown->HTML conversion in JavaScript, to make sure there's no round-trip to the server during the preview.
2. Make this conversion always on (so that both `None` and `markdown` values for `POOTLE_MARKUP_FILTER`) work the same; marking one of those as obsolete. This will give a good config upgrade path.

iafan:
With that, we might also want to discontinue the support for `restructuredtext` option.

unho:
And also drop `textile`.

julen:
[...]

PR #6431, opened 2017-05-24
Direct users to Markdown for staticpages

dwaynebailey:
Goal is to allow us to migrate users off anything else onto Markdown for staticpages.  Reason, raw HTML is a vector, other formats require lots of support with limited value.

* Introduce explicit 'html' format type
* Warn people about None format type.  So they will make it 'html' and we can deprecate that, or they will move to markdown.
* Warn that html, textile and restructuredtext are now deprecated.
* Make markdown default for new installs, default config is still None (so that we can communicate the deprecation changes)

Lots of checks introduced to make this a reality.

For a future release we will finally remove actual support for these

### DIFF
--- a/pootle/checks.py
+++ b/pootle/checks.py
@@ -263,46 +263,6 @@ def check_settings(app_configs=None, **kwargs):
             id="pootle.W010",
         ))
 
-    try:
-        markup_filter = settings.POOTLE_MARKUP_FILTER[0]
-    except AttributeError:
-        errors.append(checks.Warning(
-            _("POOTLE_MARKUP_FILTER is missing."),
-            hint=_("Set POOTLE_MARKUP_FILTER."),
-            id="pootle.W012",
-        ))
-    except (IndexError, TypeError, ValueError):
-        errors.append(checks.Warning(
-            _("Invalid value in POOTLE_MARKUP_FILTER."),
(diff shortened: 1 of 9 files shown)


pootle/checks.py
```diff
@@ -387,6 +387,12 @@ def check_settings_markup(app_configs=None, **kwargs):
                            "POOTLE_MARKUP_FILTER."),
                     id="pootle.W015",
                 ))
+        if markup_filter is None:
+                errors.append(checks.Warning(
+                    _("POOTLE_MARKUP_FILTER set to 'None' is deprecated."),
+                    hint=_("Set your markup to 'html' explicitly."),
+                    id="pootle.W025",
+                ))
```
unho: Nit: I would put this a few lines above where `if markup_filter is not None:` is,  and make it a `if..else`

unho:
lgtm after nit

unho:
Also don't forget to reference to #4184 issue in some commit.

pootle/checks.py
dwaynebailey: The sections are too long and too indented, thus I specifically chose to put this here.  Its very clear and explicit.

unho: I don't think it will change indentation at all. But I can live without this nit.

PR #6445, opened 2017-05-30
Bump django contact form

unho:
django-contact-form 1.4.1 is already out with some code cleanups that would allow to remove some patching we have.


phlax:
lgtm

PR #6448, opened 2017-05-31
Get units opti

phlax:
optimization where get_units initially just retrieves ordered pks, and then retrieves data for those units.

this prevents huge temporary tables being created with all of the associated data


ta2-1:
lgtm

PR #6449, opened 2017-05-31
Get edit unit opti

phlax:
- adds select relateds
- makes expensive calls in the view rather than templates
- prevents if exists: get queryset, pattern
- improve get_alt_src_langs


ta2-1:
lgtm

PR #6454, opened 2017-06-02
Edit optis

phlax:
some optimizations for edit/submit views


pootle/apps/pootle_score/updater.py
```diff
@@ -67,9 +68,15 @@ def select_for_update(self, qs):
         return qs.select_related("tp")
 
     def update_scores(self, objects):
+        users = (
+            set(user
+                for user
```
unho: Nit: Call it `user_id` since that is what it is.

unho:
lgtm

PR #6471, opened 2017-06-07
Fix: language list needs to be a dict

### LINKED ISSUE #6469: AttributeError: 'QuerySet' object has no attribute 'keys'
The following appears when you first go to `/` and Pootle tries to redirect you to your language based on your Accept-Lang headers.  Note this will set a cookie so you need to have cookies reset to replicate.

```python
AttributeError: 'QuerySet' object has no attribute 'keys'
  File "pootle_app/views/index/index.py", line 94, in dispatch
    else self.languages).values_list('code', flat=True))
  File "pootle/i18n/override.py", line 81, in get_lang_from_http_header
    for lang in supported.keys():

AttributeError: 'QuerySet' object has no attribute 'keys'
```

dwaynebailey:
This fixes #6469 

This path is called at `/`.  Our code tries to redirect you to a `/$lang` if your header says you want that language and we have it hosted on the server.  If not you get directed to `/projects`.  This makes use of a cookie so to replicate you need to set the cookie to None.

### DIFF
--- a/pootle/apps/pootle_app/views/index/index.py
+++ b/pootle/apps/pootle_app/views/index/index.py
@@ -89,9 +89,9 @@ class IndexView(View):
         if lang is None:
             lang = get_lang_from_http_header(
                 request,
-                (self.all_languages
-                 if request.user.is_superuser
-                 else self.languages).values_list('code', flat=True))
+                dict((self.all_languages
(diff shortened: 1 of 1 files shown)


phlax:
lgtm

Issue #6243, opened 2017-04-12
Running the Pootle tests from Makefile


ymmd:
Hi,

I'm trying to get the Pootle tests running (since I hope it will help us figuring out why we [can't comment on untranslated units](https://github.com/translate/pootle/issues/6231) on our server).

We didn't install Pootle via pip, since we may need to tweak the authentication. Instead we cloned the git repository, making our directory look like this:

    /var/www/pootle2/ (the repository we cloned)
      - env/ (our "normal" virtual environment)
         - bin/
            - pootle (a pootle runner file that I created so that we could use the pootle command)
            - [...]
         - include/ [...]
         - lib/ [...]
         - local/
            - lib/
               - python2.7/
                   - site-packages/ [...]
                   - [...]
            - [...]
      - tests/ [...]
      - pootle/ [...] (the django app)
      - [...]

We have set up the virtualenv "env/" as mentioned in the [installation instructions](http://docs.translatehouse.org/projects/pootle/en/latest/server/installation.html#setting-up-the-virtual-environment)

When running the tests via the Makefile, first the assets are being built. This requires to be in an active virtualenv, so I we used env for this. I ran these two commands:

    source env/bin/activate
    export PYTHONPATH=/var/www/pootle2
    make test

First the assets are successfully being built. But as soon as the tests start, it fails:

    ...
    chmod 664 /var/www/pootle2/pootle/assets/.webassets-cache/*
    virtualenv pootle_test_env && \
    source pootle_test_env/bin/activate && \
    export PYTHONPATH=/var/www/pootle2
    New python executable in /var/www/pootle2/pootle_test_env/bin/python
    Installing setuptools, pip, wheel...done.
    /bin/sh: 2: source: not found
    Makefile:77: recipe for target 'test' failed
    make: *** [test] Error 127

I then deactivated the virtualenv, but now I can't make the tests anymore:

    $ make test
    npm cache clear
    rm -rf pootle_test_env
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "pootle/settings.py", line 27, in <module>
        execfile(os.path.abspath(f))
      File "/var/www/pootle2/pootle/settings/95-outro.conf", line 39, in <module>
        from pootle.i18n import override
      File "pootle/i18n/override.py", line 15, in <module>
        from django.utils import translation
    ImportError: No module named django.utils

---

Running the tests directly via `pytest` worked with these steps:

    export PYTHONPATH=/var/www/pootle2
    virtualenv test_env
[...]

PR #6479, opened 2017-06-08
Test: remove 'make test' target

dwaynebailey:
Prefer direct pytest execution

Part of #6243


phlax:
lgtm, lets limit the num of ways to do things

PR #6481, opened 2017-06-10
Add issue template

dwaynebailey:
Just slightly tired of having to extract useful info from bug reporters, hopefully this sets them on the right direction where the reporting guidelines just haven't helped.


unho:
lgtm

julen:
You'll need to rename the file to _.github/issue_tem**p**late.md_ or it won't work 😉.

PR #6492, opened 2017-06-12
Make LINGUAS generation check part of Travis runs

dwaynebailey:
Various cleanups to LINGUAS and LINGUAS generation script.  Adds the test to tox to allow us to fail if a language should/shouldn't be in LINGUAS and is missing/included.


unho:
lgtm

PR #6495, opened 2017-06-13
Output git hash with --version

dwaynebailey:
If available lets get the hash, as `--version` is not that helpful when
someone might be running a git checkout.

Before:
```
$ pootle --version
Pootle 2.8.0rc5 (Django 1.10.7, Translate Toolkit 2.1.0)
```

After:
```
$ pootle --version
Pootle 2.8.0rc5 [d1a9feb] (Django 1.10.7, Translate Toolkit 2.1.0)
```


phlax:
lgtm, should we add similar for ttk?

dwaynebailey:
@phlax opinions on hash in about page?

TTK is a good idea but would be harder as we don't have the Pootle versioning code in there.  Lots of ifs ito if it would work as an installed package with `-e git....`

Issue #6341, opened 2017-05-05
Broken PO files are partially loaded without error


dwaynebailey:
I mistakenly made a PO file with duplicate units.  The files are mostly like:

N/M.po where N=1 to 3 and M=1 to 20

A bug in my script meant that my files had more than 100 units and they where duplicated.

1. The files loaded but when checking only the first 100 units are actually loaded
2. No error was reported and update_stores did not abort or report any issue

Expected:
1. PO fil with duplicate units should throw and error
2. update_stores should report that error and abort update or that specific store.

phlax:
after some discussion, it was agreed to warn rather than fail

PR #6498, opened 2017-06-13
Add warning to update on duplicates in source file

phlax:
and adds a test for handling dupes on update


unho:
#6341 is about reporting error and aborting. This PR is only about reporting, so my question is if aborting is already handled, otherwise this PR doesn't fully fix the error.

phlax:
well, i think the concensus seemed to be that it shouldnt fail, so altho it doesnt fix the original intent of the ticket, it does fix the issue

PR #6501, opened 2017-06-13
Add refresh_scores to upgrade docs

### LINKED ISSUE #5906: Add info about updating scores to upgrade docs
As we removed a migration to update scores, we probs need to notify users upgrading from any versions before that migration was introduced, to update their scores

phlax:
fixes #5906


dwaynebailey:
lgtm if you can slip in below that will allow user to easily get to the actual command docs.  But not a blocker.
```
:djadmin:`refresh_scores`
```

PR #6503, opened 2017-06-13
Skip memusage tests by default

phlax:
these tests are quite useful, but are taking a lot of time, and some fail occasionally - so making them skip by default

### DIFF
--- a/pytest_pootle/plugin.py
+++ b/pytest_pootle/plugin.py
@@ -8,6 +8,8 @@
 
 from pkgutil import iter_modules
 
+import pytest
+
 from . import fixtures
 from .fixtures import (
     core as fixtures_core, formats as fixtures_formats,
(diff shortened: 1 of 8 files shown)


dwaynebailey:
lgtm

pytest_pootle/plugin.py
```diff
@@ -55,3 +63,9 @@ def pytest_configure(config):
             fixtures_fs))
     for plugin in pytest_plugins:
         config.pluginmanager.import_plugin(plugin)
+
+
+def pytest_runtest_setup(item):
+    marker = item.get_marker("pootle_memusage")
+    if marker is not None and not item.config.getoption("--memusage"):
```
dwaynebailey:
```
os.environ["TRAVIS_EVENT_TYPE"] != "cron"
```
Would allow execution of these tests if we're running under Travis cron.  Just not sure that's the best way to configure this.

phlax: to  run the test with memusage - add the flag ```--memusage```

dwaynebailey: Not sure how to achieve that in tox env.

phlax: add to the py.test command

PR #6504, opened 2017-06-14
Use pycountry for langnames

### LINKED ISSUE #6218: Fix display localized language names in dropdown
Noticed as part of https://github.com/translate/pootle/pull/6217 testing that it wasn't working.

unho:
It is more standard and ensures there are less problems retrieving a translation for it.

Along with https://github.com/translate/translate/pull/3636, it fixes #6218.

### DIFF
--- a/pootle/apps/pootle_language/models.py
+++ b/pootle/apps/pootle_language/models.py
@@ -8,11 +8,15 @@
 
 from collections import OrderedDict
 
+from translate.lang.data import get_language_iso_fullname
+
+from django.conf import settings
 from django.db import models
 from django.urls import reverse
+from django.utils import translation
 from django.utils.functional import cached_property
 
-from pootle.core.delegate import data_tool
+from pootle.core.delegate import data_tool, language_code, site_languages
 from pootle.core.mixins import TreeItem
 from pootle.core.url_helpers import get_editor_filter
 from pootle.i18n.gettext import language_dir, tr_lang, ugettext_lazy as _
@@ -77,8 +81,24 @@ class Language(models.Model, TreeItem):
 
     @property
     def name(self):
-        """Localized fullname for the language."""
-        return tr_lang(self.fullname)
+        """localised ISO name for the language or fullname
+        if request lang == server lang, and fullname is set. Uses code
+        as the ultimate fallback
+        """
+        site_langs = site_languages.get()
+        server_code = language_code.get()(settings.LANGUAGE_CODE)
+        request_code = language_code.get()(translation.get_language())
+        use_db_name = (
+            not translation.get_language()
+            or (self.fullname
+                and server_code.matches(request_code)))
+        if use_db_name:
+            return self.fullname
+        iso_name = get_language_iso_fullname(self.code) or self.fullname
+        return (
+            site_langs.capitalize(tr_lang(iso_name))
+            if iso_name
+            else self.code)
 
     @property
     def direction(self):
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -31,6 +31,7 @@ pytz==2017.2
 rq==0.8.0
 scandir==1.5
 stemming==1.0.1
+pycountry==17.5.14
 
 # Markup: Markdown filter for POOTLE_MARKUP_FILTER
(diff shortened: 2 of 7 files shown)


pootle/apps/pootle_language/models.py
```diff
@@ -76,9 +78,20 @@ def pootle_path(self):
         return '/%s/' % self.code
 
     @property
+    def iso_name(self):
+        """ISO full name for the language."""
+        lang_code = simplify_to_common(self.code)
```
dwaynebailey: What about `en_ZA` `ca@valentian` and similar?

unho: Just added more code in https://github.com/translate/translate/pull/3636 to handle `en_ZA` and `es-ES` by retrieving separatedly the language name and the country name. But for `ca@valencia` I am afraid it is not possible to do anything, so it falls back to language fullname as defined in Pootle database.

pootle/apps/pootle_language/models.py
```diff
@@ -76,9 +78,14 @@ def pootle_path(self):
...
     @property
+    def iso_fullname(self):
+        """ISO fullname for the language."""
+        return get_language_iso_fullname(self.code) or self.fullname
+
+    @property
     def name(self):
         """Localized fullname for the language."""
-        return tr_lang(self.fullname)
+        return tr_lang(self.iso_fullname)
```
dwaynebailey: Correct me if I'm wrong.  But this looks like if your code is found in TTK you are going to get the `iso_fullname` regardless if you have a simpler `fullname`.  While this is great in another language, it is a serious problem in English.

unho: This has been sorted out.

tests/models/language.py
```diff
@@ -28,3 +34,64 @@ def test_language_specialchars_uniqueness():
...
+    afrikaans.save()
+    assert afrikaans.fullname == new_fullname
+
+    settings.LANGUAGE_CODE = "cy"
+    ui_lang = to_locale(settings.LANGUAGE_CODE)
+    with translation.override(ui_lang):
+        actual_ui_lang = to_locale(get_language())
+        assert actual_ui_lang == ui_lang
+        assert actual_ui_lang == "cy"
+        assert simplify_to_common(actual_ui_lang) != "en"
```
phlax: not sure what this is testing - i think it needs to be a +ve test

tests/models/language.py
```diff
@@ -28,3 +34,64 @@ def test_language_specialchars_uniqueness():
...
+
+@pytest.mark.django_db
+def test_language_iso_fullname(afrikaans):
+    new_fullname = "FULLNAME"
+    fullname = afrikaans.fullname
+    afrikaans.fullname = new_fullname
+    afrikaans.save()
+    assert afrikaans.fullname == new_fullname
+
+    settings.LANGUAGE_CODE = "cy"
```
phlax: not sure if you need settings here, but if you do use the fixture rather than import

tests/models/language.py
```diff
@@ -28,3 +34,64 @@ def test_language_specialchars_uniqueness():
...
+    ui_lang = to_locale(settings.LANGUAGE_CODE)
+    with translation.override(ui_lang):
```
phlax: i think this would be a lot easier to read if it was

```
with translation.override("cy"):
   ...
```
ui_lang is not used, and you could get rid of the actual_ which i think would simplify

requirements/base.txt
```diff
@@ -40,4 +40,4 @@ Markdown==2.6.8
 # Note: also adjust pootle/checks::TTK_MINIMUM_REQUIRED_VERSION
 translate-toolkit==2.1.0
 # If you want to use Translate Toolkit 'master'
-#-e git://github.com/translate/translate.git#egg=translate-toolkit-2.0.0
+#-e git+https://github.com/translate/translate.git#egg=translate-toolkit-2.2.0
```
dwaynebailey: Please add the why to the commit message, it isn't required, it's only required for older pip versions.  Cause git: is faster we probably want to bring that back at some time.

PR #6509, opened 2017-06-16
Release notes 2.8.0rc6

unho:
Release notes 2.8.0rc6 version


dwaynebailey:
lgtm, there are a few not quite Englishisms in the release notes, not sure how best to report and fix.

Do we really need to reset app versions?

unho:
Not really needed to reset app versions, but feels the safer approach.

PR #6510, opened 2017-06-16
Lang code util

phlax:
adds a lang code util for parsing/matching language codes

there are some similar functions in ttk, but they needs some work and are not easy to change or extend

once the ttk functions are cleaned up we might be able to shift most of the logic from here

### DIFF
--- a/tests/core/language.py
+++ b/tests/core/language.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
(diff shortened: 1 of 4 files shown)


tests/core/language.py
```diff
@@ -0,0 +1,65 @@
...
+    server_code = LanguageCode(settings.LANGUAGE_CODE)
+    request_code = LanguageCode(translation.get_language())
+    return server_code.matches(request_code, ignore_dialect=ignore_dialect)
+
+
+def test_language_server_request_match(settings):
+    # default LANGUAGE_CODE = "en-us"
+    with translation.override(settings.LANGUAGE_CODE):
+        assert _server_lang_matches_request()
+        assert _server_lang_matches_request(ignore_dialect=False)
```
unho: Why these are not asserted against `is True` ?

unho:
lgtm

PR #6524, opened 2017-06-23
Dont null submitter for untranslated units

phlax:
we currently set submitter to none if the unit is untranslated

this means we lose the info about who made whatever change, and breaks
submissions. this change prevents that from happening


unho:
lgtm

unho:
lgtm once Travis passes.

PR #6538, opened 2017-06-29
Add explicit message for auto-translate matched TM

### LINKED ISSUE #5824: filter=untranslated also shows fuzzy strings
There is no way to only have untranslated messages in the list of units to go through. There is a filter fuzzy that shows only stings marked as fuzzy, and there is a filter "incomplete" that (from my understanding) is meant to be fuzzy and untranslated combined. However, "untranslated" behaves the similar to "incomplete", at least it also shows fuzzy strings, which is not expected.

(incomplete and untranslated have different string counts though, so there is another difference)
In any case doesn't reflect what http://docs.translatehouse.org/projects/pootle/en/latest/developers/glossary.html claims to be...

Sample URLs:
https://translations.documentfoundation.org/te/libo_ui/translate/#filter=untranslated at the moment of writing this 8866 strings, the first being presented is a fuzzy one.
https://translations.documentfoundation.org/te/libo_ui/translate/#filter=incomplete at the moment of writing this 10699 strings

interestingly enough, the amount of fuzzy strings is 1833, the difference between the two. So the stats/count reflect what would be expected, but the units offered for translations isn't filtered properly?

phlax:
currently untranslated units with 100% matches in TM are shown as fuzzy and the matched translation is inserted into the editor.

this is confusing, and doesnt reflect the units state in the db

this PR ~~prevents the translation from being copied to the editor, but instead changes the colour of the editor widget to reflect that there is a matched translation~~ adds a message on auto-translated units

fixes #5824

### DIFF
--- a/pootle/static/css/editor.css
+++ b/pootle/static/css/editor.css
@@ -348,6 +348,12 @@ div.translate-container
     clear: both;
 }
 
+
+div.translate-container.error
+{
+    background-color: blue;
+}
+
 div.translate-container.error
 {
     background-color: #fdd;
@@ -1060,7 +1066,7 @@ html[dir="rtl"] .suggestion-original
     border-bottom: 1px solid #d9d9d9;
     border-bottom-left-radius: 3px;
     border-bottom-right-radius: 3px;
-    margin: 0 auto 0.5em;
+    margin: 0 auto 0.1em;
 }
 
 .editor-area-wrapper:first-child
@@ -1727,4 +1733,17 @@ td.translate-delete-terminology input
 .right
 {
     float: right;
+}
+
+
+.js-auto-matched-translation textarea
+{
+    color: #aaa;
+}
+
+.auto-match-msg
+{
+    font-weight: bold;
+    margin-bottom: 0.5em;
+    font-style: italic;
 }
--- a/pootle/templates/editor/units/edit.html
+++ b/pootle/templates/editor/units/edit.html
@@ -185,6 +185,9 @@
 
             {% block target_field %}
             <div class="js-mount-editor"></div>
+	    <div class="js-auto-match-msg auto-match-msg hide">
+	      {% trans "Showing match from similar translations" %}
+	    </div>
 
             <script type="text/javascript">
               PTL.reactEditor.init({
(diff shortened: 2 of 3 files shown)


julen:
I understand the current behavior might result confusing for someone coming across to it for the first time, although have you considered to indicate in a different way the fact the translation has been pre-populated?
By adding explicit messaging, using different colors, or anything that makes the distinction explicit can greatly help. Ripping the feature off hurts quite a bit OTOH — personally this has saved me hundreds of clicks. My 2c.

phlax:
hi @julen - i would probs go with explicit messaging - but was mostly going on advice on this.

it turns out that just removing - also removes the highlighting of the matched TM - so i think this PR will not land as is - and some reconsideration will happen

phlax:
@julen pr updated with a different approach

pootle/templates/editor/units/edit.html
```diff
@@ -185,6 +185,10 @@
 
             {% block target_field %}
             <div class="js-mount-editor"></div>
+	    <div class="js-auto-match-msg hide"
+		 style="font-weight: bold; margin-bottom: .5em; font-style: italic;">
```
unho: Nit: put styling in CSS stylesheet.

unho:
lgtm despite the nit.

pootle/templates/editor/units/edit.html
phlax: oh yep

pootle/static/css/editor.css
```diff
@@ -1728,3 +1734,16 @@ td.translate-delete-terminology input
...
     float: right;
 }
+
+
+.js-auto-matched-translation textarea
+{
+    color: #aaa;
+}
+
+.js-auto-match-msg
```
unho: Nit: Don't use `js-` selector for CSS.

phlax: its used by js also - and other places do this already

it saves adding more classes

phlax: actually - i suppose i can add another class without having to add/remove in js

phlax: well for the 2nd at least

unho: I realize that it has been done like this a few times. But the agreement years ago was to have separate classes for JS and CSS code, even if that meant having two classes in the same element, like

```html
<p class="something js-something">Hi</p>
```

PR #6542, opened 2017-06-30
Generate language-specific check descriptions

### LINKED ISSUE #5809: Missing language-specific check descriptions
When running `./setup.py build_checks_templates` the Romanian specific checks descriptions are not added.

unho:
Fixes #5809.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -216,6 +216,7 @@ class BuildChecksTemplatesCommand(Command):
             CHECK_NAMES, EXCLUDED_FILTERS)
         from translate.filters.checks import (TeeChecker, StandardChecker,
                                               StandardUnitChecker)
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -209,6 +209,33 @@ def initialize_options(self):
     def finalize_options(self):
         pass
 
+    def get_language_checkers(self):
+        from translate.lang.factory import getlanguage
+
+        def get_languages_with_module():
```
dwaynebailey: I think move this whole function into translate.lang.factory as there is already a getlanguage.  Just create a getlanguages function that returns classes for all of these.

unho:
This was landed on master. The message above reflects my mistake when merging.

Issue #6199, opened 2017-03-27
Catch Unicode related issues in settings


dwaynebailey:
This PR https://github.com/translate/pootle/pull/6194 addressed a problem in Python 2 where a string in the configuration was not marked as Unicode but contained Unicode related settings.  This breaks things elsewhere.

Ideally we want to prevent these in future.  Either we:

1. Use `unicode_literal` to ensure that all strings are Unicode regardless of `u` marking
2. Add a Django check

PR #6543, opened 2017-06-30
WIP: Use uncode_literals in settings

dwaynebailey:
Try to prevent unicode strings in settings from breaking things

Attempt at #6199


dwaynebailey:
There are some Tracebacks in initdb.  My gut feeling is lets leave this until we actually move to Python 3.

PR #6545, opened 2017-07-03
Phrase and case sensitive search

### LINKED ISSUE #6475: Exact search does not return exact search results
For example searching for "tab" returns results containing "table" or "database" which are partial matches.

phlax:
separate out phrase and case sensitive search in translation ui

fixes #6475


unho:
lgtm

PR #6549, opened 2017-07-03
Docs: Add note for refresh_scores command

unho:
Add note for `refresh_scores` command warning about the possibility of collision.


phlax:
lgtm

Issue #6507, opened 2017-06-15
Wrong count of incompletes for empty source strings


julen:
In the context of https://mozilla.locamotion.org/, the statistics for incomplete counts in _Firefox 55 (Beta)_ don't look correct, as they report incomplete counts for empty source strings.

Examples:
* https://mozilla.locamotion.org/cy/firefox_beta/
* https://mozilla.locamotion.org/eu/firefox_beta/

This is not happening in the _Firefox (Nightly)_ project, which in fact was the source for the beta project.

dwaynebailey:
@julen Nightly was cloned to Beta, so should carry all translations, check states, etc across.  Does nightly use ZWNJ for those or are they truly blank?

julen:
No ZWNJ that I'm aware of. Note how the same strings are reported as translated in nightly.

milupo:
How should empty strings marked that they are shown as translated? Until lately, I translated them adding a space character and I removed the space character then again. I sent a message to m.dev.l10n, too.

news://news.mozilla.org:119/Lt2dnYwk3Og9it3EnZ2dnUU7-IfNnZ2d@mozilla.org

"How to mark empty strings as translated?", from June 13th.

julen:
Thanks for the fix 👍 Is this going to be deployed anytime soon? FWIW there are still incomplete counts being reported.

unho:
@julen Thanks for raising it. A deployment with this fix was just carried over. Correct stats should be displayed shortly.

ta2-1:
@julen I'd reopen this issue because of https://github.com/translate/pootle/pull/6553#issuecomment-314745123

PR #6553, opened 2017-07-04
Fix: dont set units with empty source to wc=1

phlax:
not entirely why this was setting unit wordcounts to 1 if source is empty, but it seems to break stuff and didnt seem to happen historically, so fix seems ok

fixes: #6507


unho:
lgtm

unho:
lgtm

ta2-1:
I have 2 concerns here:
- state of a unit with empty source is ignored (though empty units should be marked as translated explicitly). if you need to set specific translation for units with empty source you won't be able to find them as untranslated.
https://mozilla.locamotion.org/eu/firefox_beta/translate/#filter=untranslated
- wordcount function can ignore a specific set of words such as `Pootle` which shouldn't be translated.

phlax:
@ta2-1 not really following the issue you raise here

afaict the wordcount should just be what the wordcount function says - irrespective of state - similar with ignored words really

what do you suggest would be a better way to store wordcounts ?

ta2-1:
@phlax 
There is inconsistence:
If you see zero incomplete wordcount you don't expect to see untranslated or fuzzy units.
And I'd say it's critical. 

I agree that setting wordcount to `1` is not perfect solution but it was made intentionally.

As perfect solution I can think about splitting incomplete unit count and incomplete word count and having two different numbers in stats.

Speaking about fixing mlo issue after cloning `ff_beta` I think empty unit states should be set to `translated` manually (though I'm not sure how it will affect `.po` files).

phlax:
> As perfect solution I can think about splitting incomplete unit count and incomplete word count and having two different numbers in stats.

yep, i was just chatting with @dwaynebailey re this, and that was the conclusion i came to. Not sure how quickly we are gonna add that though

PR #6556, opened 2017-07-05
i18n: Use string instead of unit

unho:
Translators are used to string, and is also the most used translation
for this thing throughout Pootle.


fitojb:
Looking forward to see this merged.

phlax:
lgtm

PR #6565, opened 2017-07-07
Rid us of manage.py

dwaynebailey:
* Remove from docs as much as possible
* Warning in ./manage.py - sadly suppressed by default in Python
* Various other docs fixes picked up in doing this cleanup

### DIFF
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -579,6 +579,17 @@ Gets, sets, lists, appends and clears pootle configuration settings.
   Treat data as JSON when getting, setting, or appending values.
 
 
+.. django-admin:: schema
+
+schema
+^^^^^^
+
+.. versionadded:: 2.8
+
+Dumps a JSON representation for the Pootle database schema, currently only
+MySQL, for debugging and comparison to a reference database schema.
+
+
 .. _commands#translation-memory:
 
 Translation Memory
--- a/docs/server/commands.rst
+++ b/docs/server/commands.rst
@@ -14,12 +14,6 @@ will run:
 
     $ pootle help
 
-.. note::
-
-  If you run Pootle from a repository checkout you can use the *manage.py* file
-  found in the root of the repository.  Note that this approach is deprecated
-  and may not be supported in future.
-
 
 .. _commands#managing_pootle_projects:
 
@@ -1443,7 +1437,7 @@ For the full details on how to configure cron, read your platform documentation
 (for example ``man crontab``). Here is an example that runs the
 :djadmin:`calculate_checks` command daily at 02:00 AM::
 
-    00 02 * * * www-data /var/www/sites/pootle/manage.py calculate_checks
+    00 02 * * * www-data source /var/www/sites/pootle/env/bin/activate; pootle calculate_checks
 
 Test your command with the parameters you want from the command line. Insert it
 in the cron table, and ensure that it is executed as the correct user (the same
--- a/manage.py
+++ b/manage.py
@@ -9,14 +9,16 @@
 
 import os
 import sys
+import warnings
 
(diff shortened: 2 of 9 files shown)


docs/server/commands.rst
```diff
@@ -1443,7 +1437,7 @@ For the full details on how to configure cron, read your platform documentation
 (for example ``man crontab``). Here is an example that runs the
 :djadmin:`calculate_checks` command daily at 02:00 AM::
 
-    00 02 * * * www-data /var/www/sites/pootle/manage.py calculate_checks
+    00 02 * * * www-data . /home/dwayne/.virtualenvs/mozilla/bin/activate; pootle calculate_checks
```
unho: mozilla? I would use a different name for the venv.

unho: Also, is the space after the dot ok?

manage.py
```diff
@@ -9,14 +9,16 @@
...
 import os
 import sys
-
-from django.core.management import execute_from_command_line
+import warnings
 
 from pootle import syspath_override  # noqa
+from pootle.core.log import cmd_log
+
+from django.core.management import execute_from_command_line
```
unho: If this import doesn't need to be imported after Pootle ones then it is incorrectly placed. If it needs to be here I will put a comment before stating it.

manage.py
```diff
@@ -9,14 +9,16 @@
...
 
 
 if __name__ == "__main__":
-    from pootle.core.log import cmd_log
+    warnings.warn("Deprecated use 'pootle' command instead", DeprecationWarning)
```
unho: `"Deprecated. Use 'pootle' command instead."`

pootle/runner.py
```diff
@@ -14,11 +14,13 @@
 import sys
 from argparse import SUPPRESS, ArgumentParser
 
+from pootle.core.cache import PERSISTENT_STORES
+from pootle.core.log import cmd_log
```
unho: If these imports need to be before the Django ones I will put a comment stating it clearly.

pootle/tools/docs_check_commands.sh
```diff
@@ -1,25 +1,25 @@
...
-    $BASE_DIR/manage.py --help | \
+    pootle | \
     sed -E -n "/^\[$app\]$/,/^\[/ p" | \
     sed "/^\[/d" | \
     sed "s/^[ ]*//g"
 done)
 
 django_commands=$(
-    $BASE_DIR/manage.py --help | \
+    pootle | \
```
unho: Is `--help` not necessary here and above?

docs/server/commands.rst
```diff
@@ -579,6 +579,17 @@ Gets, sets, lists, appends and clears pootle configuration settings.
...
 
+.. django-admin:: schema
+
+schema
+^^^^^^
+
+.. versionadded:: 2.8
+
+Dumps a JSON representation for the Pootle database schema, currently only
+MySQL, for debugging and comparision to a reference database schema.
```
unho: Typo: comparison.

unho:
The major concern is the order of the imports. Other than that there only the raised nits. Once those fixed lgtm.

pootle/tools/docs_check_commands.sh
dwaynebailey: Specifically not in fact, --help give help, nothing gives list of possible commands.  Odd but true.

docs/server/commands.rst
dwaynebailey: `. blah` == `source blah` and yes it is correct, it is != `./`

manage.py
dwaynebailey: Sorted

PR #6567, opened 2017-07-13
Make initial migrations conditional for accounts/pootle_tp

phlax:
this allows us to simplify the docs for migrations from 2.5, which was causing some confusion


unho:
lgtm after travis passes.

PR #6568, opened 2017-07-14
Update data cli

phlax:
adds a cli command `update_data` and removes the update_data migrations, which were no longer working


unho:
lgtm

PR #6571, opened 2017-07-18
Don't create tp if tp directory already exists

### LINKED ISSUE #6081: Give better feedback when trying to initialize a tp which already has directories
If you try to initialize a TP from templates where there are directories already present for that project/lang then it fails - which is correct - but we need to improve feedback

phlax:
Currently the tp is created but not populated.

This prevents the tp being created, also improves error handling
in the project form.

Fixes #6081

### DIFF
--- a/pootle/apps/pootle_project/forms.py
+++ b/pootle/apps/pootle_project/forms.py
@@ -6,6 +6,8 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import os
+
 from django import forms
 from django.db import connection
 from django.forms.models import BaseModelFormSet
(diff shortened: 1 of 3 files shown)


pootle/apps/pootle_project/forms.py
```diff
@@ -84,6 +86,22 @@ def __init__(self, *args, **kwargs):
         self.fields["project"].queryset = self.fields[
             "project"].queryset.filter(pk=project_id)
 
+    def clean(self):
+        if not self.cleaned_data.get("id"):
+            exists_on_file = (
+                self.cleaned_data["language"].code
+                in os.listdir(self.cleaned_data["project"].get_real_path()))
+            if exists_on_file:
```
unho: `exists_on_disk`?

pootle/apps/pootle_project/forms.py
```diff
@@ -84,6 +86,22 @@ def __init__(self, *args, **kwargs):
...
+                errordict = dict(
+                    lang=self.cleaned_data["language"].code,
+                    path=os.path.join(
+                        self.cleaned_data["project"].get_real_path(),
+                        self.cleaned_data["language"].code))
+                raise forms.ValidationError(
+                    _("Cannot create translation project for language "
+                      "'%(lang)s', path '%(path)s' already exists",
```
unho: Nit: `directory '%(path)s' already exists`

phlax: its not necessarily a directory

unho:
lgtm after nits.

PR #6582, opened 2017-07-21
Add critical_checks_active in process submission xhr

### LINKED ISSUE #6550: Muted checks don't act muted in Pootle 2.8.0rc6
In Pootle 2.8.0rc5, after translating a string and triggering some warnings (namely, XML tag mismatch), I could mute that check and/or press Ctrl+Enter to navigate to the next string.

With rc6, even if I mute that check for that translation, Ctrl+Enter still flashes that check to me and does NOT take me to the next string, regardless of how many times I press that key combo.

phlax:
fixes #6550

### DIFF
--- a/pootle/apps/pootle_store/views.py
+++ b/pootle/apps/pootle_store/views.py
@@ -690,7 +690,9 @@ class UnitSubmitJSON(UnitSuggestionJSON):
             result = dict(
                 checks=_get_critical_checks_snippet(self.request, form.unit),
                 user_score=self.request.user.public_score,
-                newtargets=[target for target in form.unit.target.strings])
+                newtargets=[target for target in form.unit.target.strings],
+                critical_checks_active=(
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_store/views.py
```diff
@@ -690,7 +690,8 @@ def get_context_data(self, *args, **kwargs):
             result = dict(
                 checks=_get_critical_checks_snippet(self.request, form.unit),
                 user_score=self.request.user.public_score,
-                newtargets=[target for target in form.unit.target.strings])
+                newtargets=[target for target in form.unit.target.strings],
+                form.unit.get_active_critical_qualitychecks().exists())
```
unho: Shouldn't this have a key?

phlax: oh yep, i hadnt tested the pr yet - but i guess it will fail

unho:
Reviewed and tested. gtm

PR #6585, opened 2017-07-21
Use 'pootle init' always instead of 90-*local.conf

### LINKED ISSUE #5210: Clarify the location of configuration files
altho the current docs do provide most of the information, it can be confusing for new users to know where and when which files are used, and where is best to customize

dwaynebailey:
There are some more of these.  But it might mean adding a `pootle init --dev` option so we can get rid of `cp 90-local-dev.conf.sample pootle.conf`

@phlax might be best for me to idle this till your where is `pootle.conf` change lands.

Fixes #5210

### DIFF
--- a/docs/developers/setup_windows.rst
+++ b/docs/developers/setup_windows.rst
@@ -51,10 +51,6 @@ Clone your fork of the Pootle master using your favourite Windows
 implementation of Git so that you have a working copy somewhere accessible on
 your computer (e.g. ``C:\git\pootle``).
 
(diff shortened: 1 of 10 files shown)


docs/developers/setup_windows.rst
```diff
@@ -115,6 +111,7 @@ Linux system.
 
 .. code-block:: console
 
+    (venv)> pootle init --dev
```
phlax: im wondering if we should use `(pootle_env)` here too to be consistent

phlax:
minor nit, but otherwise lgtm

PR #6587, opened 2017-07-21
Add extra attr overrides to Markdown cleaning

dwaynebailey:
* Allows users to add extra default bleach attr filtering.
* Adds tests for Markdown parsing and cleaning

### DIFF
--- a/docs/server/settings.rst
+++ b/docs/server/settings.rst
@@ -269,6 +269,14 @@ Configuration settings for applications used by Pootle.
     POOTLE_MARKUP_FILTER = ('markdown', {
                                 'clean': {
                                     'extra_tags': ['div'],
+                                    'extra_attrs': {
+                                        '*': ['style']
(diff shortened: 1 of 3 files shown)


docs/server/settings.rst
```diff
@@ -269,6 +269,15 @@ Configuration settings for applications used by Pootle.
     POOTLE_MARKUP_FILTER = ('markdown', {
                                 'clean': {
                                     'extra_tags': ['div'],
+                                    'extra_attrs': {
+                                        '*': ['class'],
+                                        '*': ['style']
```
phlax: seems like class and style are still in docs

dwaynebailey: That's find I think, these are custom settings, not defaults.

unho:
I don't understand the raised issue, but if that is not a blocker, lgtm from my side.

Issue #6187, opened 2017-03-24
$ 400 bounty to add user permission to allow projects creation in Pootle


pyg77:
Just to let you know that we have created a bounty to allow projects creations per user in Pootle.

https://www.bountysource.com/issues/42027191-add-user-permission-to-allow-projects-creation-in-pootle

The value of the bounty is $400, if anyone wants to work on it

dwaynebailey:
Hi @pyg77 feel free to ping me on https://gitter.im/dwaynebailey

ta2-1:
If we are happy with current admin project UI I think it’s worth to reuse it and:

- add new user projects URL (`/user/projects/`) available for non admins
- add new template based `template/admin/projects.html` but extended from standard layout
- add `can_create_projects` permission
- handle `can_create_projects` permission in `ProjectAPIView`
- add filter for projects by new permissions for a user in `ProjectAPIView`

giliam:
Thanks @ta2-1 for this information. Do you agree that the URL and template should be in pootle_project app?

ta2-1:
@giliam `pootle_project` is ok for `urls`.  And templates are stored in `pootle/templates/` directory.

giliam:
Ok! I will work on it (in https://github.com/translate/pootle/compare/master...giliam:feature/project_creation_auth) and come back to you as soon as something seems to be working.

pyg77:
(just a quick note to thanks @ta2-1 for his advices, and @giliam that the bounty is still open :) I saw https://github.com/translate/pootle/compare/master...giliam:feature/project_creation_auth and it seem that the work is going well! :) )

PR #6593, opened 2017-07-24
Project creation auth for user without Admin rights

giliam:
Hello!
A new feature so that a user with a new special perms who isn’t site administrator can create a project and edit it (issue #6187)
I changed the API to use the administration page for projects. Now the project list either display all projects for Administrator users or only administrated projects for normal users.
I am new to pootle so I guess this code review will take some time.
Thank you in advance for your help and comments!
Gilïam

P.S.: Sorry, I didn't have a lot of time these last two weeks.


ldidry:
It seems to be a very good job 👍 

Small bug: the "My projects" link, when clicking on my avatar (on a non-admin user) isn't displayed when being on a project's page (https://exemple.org/projects/project_name/). I didn't verified that it's displayed on all other places you can go.

giliam:
Thank you Ididry!

I couldn't reproduce this bug. You are logged as non-admin user (who has the right to create project?) and you go to some project page (is it a project created by this user or by another user?). You click on the right top menu and the "My projects" link is not displayed. Is it displayed anywhere else?

Thanks again :)

ldidry:
How to reproduce on my test installation:

- log in with a non-admin user, with the right to create projects
- created a project (or already have one created)
- go to my project (from "My projects" link or from the drop-downs on top of the page): no "My projects" link displayed
- go to a project that doesn't belong to me: "My projects" link is displayed in the top right drop-down or anywhere else

giliam:
I think I understand the problem. It already exists a permission called _add_project_ in the database which is specific to the folder pootle_project. The permission I added has to be in the root directory as it is a permission global to the whole site, just as _administrate_ or _review_ (other permissions visible in the _permissions_ admin panel). 
When in a specific folder, you are in the directory _pootle_project_ and thus, the permissions don't seem to have _add_project_ global permission I added.

I see three ways of getting out of this:
1. Either I change my permission's name (but I don't know if it will work) to create_project for example, and the conflict will maybe not happen
2. Either I create some specific function for the template to check if the user has the permission requested using the root directory
3. Either I delete the specific permission which is not used.

I would probably try the first possibility but maybe it is not the best one. Any pootle expert advice?

phlax:
hi @giliam i havent had chance to check out the PR properly yet.

We are just about to get a stable release (finally!) - hopefully this should be out in the next week. Once thats done I would be up for assisting with landing this and addressing any issues that come up re permissions etc.

From an architectural point of view, the problem i see is that for each project it currently creates a directory in the language directories, so we need to work out the best way to structure personal projects, and how we can integrate that to pootle's namespace.

giliam:
Hi @phlax!
Thank you for your comment and I hope the release is going well :)
I just fixed the issue given by @ldidry by renaming add_project to create_project (to have a clean database) and by using a specific tag to check for the permission. I also fixed the issues given by travis tests (but I still don't know how to fix the code coverage patch test).
I will still be available for further questions/work on this pr, specially on this question of architecture.

ldidry:
@phlax Any news?

PR #6596, opened 2017-07-27
Move AMAGAMA_SOURCE_LANGUAGES to settings

phlax:
and add `en-US`

fixes a problem reported by @chazotte in gitter channel

### DIFF
--- a/pootle/settings/60-translation.conf
+++ b/pootle/settings/60-translation.conf
@@ -51,6 +51,9 @@ POOTLE_MT_BACKENDS = [
 # the trailing slash.
 AMAGAMA_URL = 'https://amagama-live.translatehouse.org/api/v1/'
 
+# List of available source languages in AMAGAMA
+AMAGAMA_SOURCE_LANGUAGES = ('en', 'en_US', 'en-US')
+
 # See 90-local.conf.template for example configuration for local TM server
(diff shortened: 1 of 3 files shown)


pootle/settings/60-translation.conf
```diff
@@ -51,6 +51,9 @@ POOTLE_MT_BACKENDS = [
 # the trailing slash.
 AMAGAMA_URL = 'https://amagama-live.translatehouse.org/api/v1/'
 
+# List of available source languages in AMAGAMA
+AMAGAMA_SOURCE_LANGUAGES = ('en', 'en_US', 'en-US')
```
unho: amaGama only supports `en` as source language. See https://amagama-live.translatehouse.org/api/v1/languages/

unho: Is not a problem to specify unsupported sources languages though, just raising that default setting is probably better to reflect what is supported by the main amaGama server.

phlax: it supports en-us and en_us afaict - maybe others

unho:
Please add new setting to release notes.

Other than this and other raised nit, lgtm.

phlax:
ive not added to docs/release notes because its not configurable as is - altho im not clear about other `en` dialects, so i think its better as a setting

PR #6604, opened 2017-08-01
Including select2 localization

JohnnyKing94:
I don't know if it's possible to serve the local lang.js file on your server creating some sort of virtual path (like `/select2_i18n`) which points the local path (`/static/js/node_modules/select2/src/js/select2/i18n`). For the moment i use cloudflare CDN


unho:
@JohnnyKing94 Thanks for the contrib, but as you raised it is better for Pootle to serve this translations, thus I am closing this PR in favor of https://github.com/translate/pootle/pull/6608

PR #6608, opened 2017-08-04
Expose select2 l10n

unho:


pootle/apps/pootle_app/assets.py
```diff
@@ -50,6 +54,18 @@
 register('js_editor', js_editor)
 
 
+rel_path = os.path.join('js', 'node_modules', 'select2', 'dist', 'js', 'i18n')
+select2_l10n_dir = os.path.join(settings.WORKING_DIR, 'static', rel_path)
+l10n_files = [os.path.join(rel_path, f)
+              for f in os.listdir(select2_l10n_dir)
+              if (os.path.isfile(os.path.join(select2_l10n_dir, f)) and
```
phlax: can we put and/or operators at beginning of line please

dwaynebailey:
@unho the first commit IIUC is completely uneeded.  This will get overrtten/created when you run `make assets` and there is no need to keep this in our repo.

Makefile
```diff
@@ -44,6 +45,7 @@ travis-assets:
 		cd ${JS_DIR} && \
 		npm install && \
 		cd ${CWD}; \
+	    cp ${JS_DIR}/node_modules/select2/dist/js/i18n/*.js ${JS_DIR}/select2_l10n/; \
```
dwaynebailey: Please use whatever whitespace the file is using

pootle/apps/pootle_app/assets.py
```diff
@@ -50,6 +54,19 @@
 register('js_editor', js_editor)
 
 
+rel_path = os.path.join('js', 'select2_l10n')
+select2_l10n_dir = os.path.join(settings.WORKING_DIR, 'static', rel_path)
```
dwaynebailey: Just curious, if we are already using this in static/ dir.  Why can't we just pick these out of node_modules dir in static dir?

unho: When we build the packages (for PyPI), the `node_modules` is not present. That is the reason why we rely on this new `select2_l10n` subdirectory within `static/js/` directory.

PR #6610, opened 2017-08-09
Update to work with apache 2.4 out of the box

benbankes:
This update allows static assets to be displayed out of the box when using Apache 2.4.  Apache 2.2 users will receive an error (among the other errors they may already have to fix), which will lead them directly to the comment.

Of course, this might not be the best approach, as the PR process may bring to light.

This was found while installing Pootle on Ubuntu 16.04 with the default version of Apache (2.4.18)


phlax:
hi @benbankes thanks for this

we mostly recommend using uwsgi + nginx these days, but the docs have not really been updated to reflect that.

lgtm, although im wondering whether we need to just drop the apache docs and add some nginx ones

benbankes:
@phlax Thanks for taking a look and responding.  Sounds good to me.  I hope the nginx documentation can be used by users who are more familiar with apache.

dwaynebailey:
Thanks @benbankes

PR #6613, opened 2017-08-16
Allow git command to be missing

dwaynebailey:
Since we introduced the git hash into the version we're expecting to have the git command.  On a docker install this might specifically be excluded, so make calling these version helpers more robust.

### DIFF
--- a/pootle/core/utils/version.py
+++ b/pootle/core/utils/version.py
@@ -236,14 +236,17 @@ def get_git_branch():
 
 @lru_cache()
 def get_git_hash():
-    """Returns the current git commit hash.
+    """Returns the current git commit hash or None.
 
     >>> get_git_hash()
     'ad768e8'
     """
-    return _shell_command(
+    git_hash = _shell_command(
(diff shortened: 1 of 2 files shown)


pootle/core/utils/version.py
```diff
@@ -233,14 +236,17 @@ def get_git_branch():
...
     'ad768e8'
     """
-    return _shell_command(
+    git_hash = _shell_command(
         ['/usr/bin/git', 'rev-parse', '--verify', '--short', 'HEAD']
-    ).strip()
+    )
+    if git_hash:
+        return git_hash.strip()
+    return None
```
unho: Why not just return a `''` here?

unho:
Other than nit, lgtm.

phlax:
lgtm

pootle/core/utils/version.py
dwaynebailey: Cause I wanted None to indicate that you don't have a hash/undefined, which is different from empty string, I reasoned.

PR #6615, opened 2017-08-16
Version caches

phlax:
allows caches to be properly expired for new versions


unho:
lgtm after Travis passes.

PR #6623, opened 2017-08-19
Fs initdb

phlax:
Use pootle fs with localfs adapter for initial projects

### DIFF
--- a/pootle/core/initdb.py
+++ b/pootle/core/initdb.py
@@ -7,19 +7,22 @@
 # AUTHORS file for copyright and authorship information.
 
 import logging
+import os
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
+from django.db import transaction
 from django.utils.translation import ugettext_noop as _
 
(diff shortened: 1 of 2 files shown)


pootle/core/initdb.py
```diff
@@ -45,13 +48,13 @@ def _init_db(self, create_projects=True):
...
         self.create_root_directories()
         self.create_template_languages()
+        self.create_default_languages()
         if create_projects:
             self.create_terminology_project()
         self.create_pootle_permissions()
         self.create_pootle_permission_sets()
         if create_projects:
             self.create_default_projects()
-        self.create_default_languages()
```
unho: Moving this up looks like fixing a bug. Maybe put in a separate commit is better.

unho:
lgtm

pootle/core/initdb.py
phlax: not a bug - its necessary for the langs to be present for fs tp to sync

unho: Sorry, I was under the assumption that creating the projects would immediately trigger importing translations from disk and that therefore the TPs would only be created for the existing languages at that time.

PR #6626, opened 2017-08-21
Remove changed langs command

phlax:
this command ~duplicates functionality in the list_languages command


unho:
Please also move it from available commands to "Deprecated commands" section in doc about commands. 

Other than that lgtm once Travis passes.

PR #6627, opened 2017-08-21
Cleanups for tests in prep for switch to pootle fs

phlax:
- Setup fs config for all projects in pytest env

### DIFF
--- a/tests/models/revision.py
+++ b/tests/models/revision.py
@@ -11,16 +11,10 @@ import pytest
 from pootle.core.models import Revision
 from pootle_store.models import Unit
 
-from .unit import _update_translation
-
 
 @pytest.mark.django_db
 def test_max_revision(revision, project0_nongnu, store0):
     """Tests `max_revision()` gets the latest revision."""
-    store0.sync()
-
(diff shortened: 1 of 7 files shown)


tests/models/revision.py
```diff
@@ -45,18 +41,17 @@ def test_max_revision(revision, project0_nongnu, store0):
 def test_revision_incr(store0):
     """Tests revision is incremented when units change."""
     previous_revision = Revision.get()
-    db_unit = _update_translation(store0, 0, {'target': [u'Fleisch']},
-                                  sync=False)
-
+    db_unit = store0.units.exclude(target_f="").first()
```
unho: Why exclude units with blank target?

unho:
lgtm once Travis passes.

tests/models/revision.py
phlax: not sure it matters, but it was to get a translated unit

unho: I don't think it matters, just curious why the qs was being filtered at all. I am happy leaving this as is provided Travis passes.

PR #6633, opened 2017-08-22
Added create_project pootle command

jjmcarrascosa:
Added a create_project command (create_project.py in pootle/apps/pootle_app/management/commands/ ) to start analyzing the requirements to make a command for creating projects, without login as an administrator in the website.

### DIFF
--- a/pootle/apps/pootle_app/management/commands/create_project.py
+++ b/pootle/apps/pootle_app/management/commands/create_project.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+import traceback
+
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from django.db import transaction
+from django.core.management.base import BaseCommand
+from django.core.exceptions import ValidationError
+from django.core.management import CommandError
+from django.core.validators import validate_email
+
+from pootle.core.delegate import formats
+from pootle_fs.delegate import fs_plugins
+from pootle_fs.presets import FS_PRESETS
+from pootle_language.models import Language
+from pootle_project.models import Project, PROJECT_CHECKERS
+
+
+class Command(BaseCommand):
+    help = "Creates a new project from command line."
+
+    def add_arguments(self, parser):
+        format_names = formats.get().keys()
+        fs_plugins_names = fs_plugins.gather().keys()
+
+        parser.add_argument(
+            'code',
+            action='store',
+            help='Project code',
+        )
+        parser.add_argument(
+            "--name",
+            action="store",
+            dest="name",
+            help="Project name",
+        )
+        parser.add_argument(
+            "--filetype",
+            action="append",
+            dest="filetypes",
+            choices=format_names,
+            default=[],
+            help="File types: {0}. Default: po".format(
+                 " | ".join(format_names)),
+        )
+        mapping_group = parser.add_mutually_exclusive_group(required=True)
+        mapping_group.add_argument(
+            "--preset-mapping",
+            action="store",
+            dest="preset_mapping",
+            choices=FS_PRESETS.keys(),
+            help="Filesystem layout preset: {0}".format(
+                 " | ".join(FS_PRESETS.keys())),
+        )
+        mapping_group.add_argument(
+            "--mapping",
+            action="store",
+            dest="mapping",
+            help="Custom filesystem layout",
+        )
+        parser.add_argument(
+            "--fs-type",
+            action="store",
+            dest="fs_type",
+            default="localfs",
+            choices=fs_plugins_names,
+            help="Filesystem type: {0}".format(" | ".join(fs_plugins_names)),
+        )
+        parser.add_argument(
+            "--fs-url",
+            action="store",
+            dest="fs_url",
+            default="",
+            help="Filesystem path or URL.",
+        )
+        parser.add_argument(
+            "--source-language",
+            action="store",
+            dest="sourcelang",
+            default="en",
+            help=("Source language. Examples: [en | es | fr | ...]."
+                  "Default: %(default)s"),
+        )
+        parser.add_argument(
+            "--report-email",
+            action="store",
+            default="",
+            dest="contact",
+            help="Contact email for reports. Example: admin@mail.com.",
+        )
+        parser.add_argument(
+            "--disabled",
+            action="store_true",
+            dest="disabled",
+            help="Does the project start disabled?",
+        )
+        parser.add_argument(
+            "--checkstyle",
+            action="store",
+            dest="checkstyle",
+            choices=PROJECT_CHECKERS.keys(),
+            default="standard",
+            help="Quality check styles. Example: {0}. Default: %(default)s"
+                 .format(" | ".join(PROJECT_CHECKERS.keys())),
+        )
+
+    @transaction.atomic
+    def handle(self, **options):
+        """Imitates the behaviour of the admin interface
+        for creating a project from command line.
+        """
+        self.check_project_options(**options)
+        self.set_project_config(
+            self.create_project(**options), **options)
+
+    def check_project_options(self, **options):
+        if options["contact"]:
+            try:
+                validate_email(options["contact"])
+            except ValidationError as e:
+                if options["traceback"]:
+                    traceback.print_exc()
+                raise CommandError(e)
+        if options["fs_type"] != "localfs" and options["fs_url"] == "":
+            raise CommandError("Parameter --fs-url is mandatory "
+                               "when --fs-type is not `localfs`")
+
+    def create_project(self, **options):
+        try:
+            return Project.objects.create(
+                code=options["code"],
+                fullname=options["name"] or options["code"].capitalize(),
+                checkstyle=options["checkstyle"],
+                source_language=self.get_source_language(**options),
+                filetypes=options["filetypes"] or ["po"],
+                report_email=options["contact"],
+                disabled=options["disabled"])
+        except ValidationError as e:
+            if options["traceback"]:
+                traceback.print_exc()
+            raise CommandError(e)
+
+    def get_source_language(self, **options):
+        try:
+            return Language.objects.get(code=options["sourcelang"])
+        except Language.DoesNotExist:
+            raise CommandError("Source language %s does not exist"
+                               % options["sourcelang"])
+
+    def set_project_config(self, project, **options):
+        mapping = (
+            dict(default=FS_PRESETS[options["preset_mapping"]][0])
+            if options["preset_mapping"]
+            else dict(default=options["mapping"]))
+        try:
+            project.config["pootle_fs.translation_mapping"] = mapping
+            project.config["pootle_fs.fs_type"] = options["fs_type"]
+            if options["fs_url"]:
+                project.config["pootle_fs.fs_url"] = options["fs_url"]
+        except ValidationError as e:
+            if options["traceback"]:
+                traceback.print_exc()
+            raise CommandError(e)
--- a/tests/commands/create_project.py
+++ b/tests/commands/create_project.py
@@ -1,0 +1,314 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import pytest
+
+from django.core.management import call_command, CommandError
+
+from pootle.core.plugin import provider
+from pootle_fs.delegate import fs_plugins
+from pootle_fs.plugin import Plugin
+from pootle_fs.presets import FS_PRESETS
+from pootle_project.models import Project
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_defaults(settings):
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.fullname == "Foo0"
+    assert foo0.checkstyle == "standard"
+    assert foo0.filetypes.first().name == "po"
+    assert foo0.source_language.code == "en"
+    assert foo0.disabled is False
+    assert foo0.report_email == ""
+
+    assert (
+        foo0.config["pootle_fs.translation_mapping"]
+        == dict(default=FS_PRESETS["gnu"][0]))
+    assert foo0.config["pootle_fs.fs_type"] == "localfs"
+    assert (
+        foo0.config["pootle_fs.fs_url"]
+        == ("{POOTLE_TRANSLATION_DIRECTORY}%s"
+            % foo0.code))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_existing_project(settings):
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu")
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo0",
+            "--preset-mapping=gnu")
+    assert (
+        "'code': [u'Project with this Code already exists.'"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_title(capfd):
+    call_command(
+        "create_project",
+        "foo1",
+        "--name=Bar1",
+        "--preset-mapping=gnu")
+    foo1 = Project.objects.get(code="foo1")
+    assert foo1.fullname == "Bar1"
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_checkstyle(capfd):
+    call_command(
+        "create_project",
+        "foo0",
+        "--checkstyle=minimal",
+        "--preset-mapping=gnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.checkstyle == "minimal"
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo1",
+            "--preset-mapping=gnu",
+            "--checkstyle=DOESNOTEXIST")
+    assert (
+        "Error: argument --checkstyle: invalid choice: u'DOESNOTEXIST'"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_filetypes(capfd):
+    call_command(
+        "create_project",
+        "foo0",
+        "--filetype=ts",
+        "--preset-mapping=gnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.filetypes.first().name == "ts"
+    call_command(
+        "create_project",
+        "foo1",
+        "--filetype=ts",
+        "--filetype=po",
+        "--preset-mapping=gnu")
+    foo1 = Project.objects.get(code="foo1")
+    assert (
+        list(foo1.filetypes.values_list("name", flat=True))
+        == ["ts", "po"])
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo2",
+            "--preset-mapping=gnu",
+            "--filetype=DOESNOTEXIST")
+    assert (
+        "Error: argument --filetype: invalid choice: u'DOESNOTEXIST'"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_report_email(capfd):
+    call_command(
+        "create_project",
+        "foo0",
+        "--report-email=foo@bar.com",
+        "--preset-mapping=gnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.report_email == "foo@bar.com"
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo0",
+            "--report-email=foo@bar",
+            "--preset-mapping=gnu")
+    assert (
+        "CommandError: [u'Enter a valid email address.']"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_source_language(language0):
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--source-language=%s" % language0.code)
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.source_language == language0
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo0",
+            "--preset-mapping=gnu",
+            "--source-language=DOESNOTEXIST")
+    assert (
+        "CommandError: Source language DOESNOTEXIST does not exist"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_disabled():
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--disabled")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.disabled is True
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_preset_mapping():
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=nongnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.translation_mapping"]
+        == dict(default=FS_PRESETS["nongnu"][0]))
+    with pytest.raises(CommandError) as e:
+        # cant specify both mapping and preset-mapping
+        call_command(
+            "create_project",
+            "foo1",
+            "--preset-mapping=gnu",
+            "--mapping='/<language_code>/<dir_path>/<filename>.<ext>'")
+    assert (
+        "Error: argument --mapping: not allowed with argument --preset-mapping"
+        in str(e))
+    with pytest.raises(CommandError) as e:
+        # must specify one or other mapping or preset
+        call_command(
+            "create_project",
+            "foo1")
+    assert (
+        "Error: one of the arguments --preset-mapping --mapping is required"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_mapping():
+    call_command(
+        "create_project",
+        "foo0",
+        "--mapping=/<language_code>/<dir_path>/<filename>.<ext>")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.translation_mapping"]
+        == dict(default=u'/<language_code>/<dir_path>/<filename>.<ext>'))
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo1",
+            "--mapping=NOTAMAPPING")
+    assert (
+        "Translation mapping 'NOTAMAPPING'"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_fs_type():
+
+    class DummyPlugin(Plugin):
+        pass
+
+    @provider(fs_plugins)
+    def dummy_fs_plugin(**kwargs):
+        return dict(dummyfs=DummyPlugin)
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo0",
+            "--preset-mapping=gnu",
+            "--fs-type=dummyfs")
+    assert (
+        "Parameter --fs-url is mandatory when --fs-type is not `localfs`"
+        in str(e))
+
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--fs-type=dummyfs",
+        "--fs-url=DUMMYURL")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.fs_type"]
+        == "dummyfs")
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo1",
+            "--preset-mapping=gnu",
+            "--fs-type=DOESNOTEXIST")
+    assert (
+        "Error: argument --fs-type: invalid choice"
+        in str(e))
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_fs_url():
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--fs-url={POOTLE_TRANSLATION_DIRECTORY}special_foo0")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.fs_url"]
+        == "{POOTLE_TRANSLATION_DIRECTORY}special_foo0")
+
+    call_command(
+        "create_project",
+        "foo1",
+        "--preset-mapping=gnu",
+        "--fs-url=/abs/path/to/foo1")
+    foo1 = Project.objects.get(code="foo1")
+    assert (
+        foo1.config["pootle_fs.fs_url"]
+        == "/abs/path/to/foo1")
+
+    with pytest.raises(CommandError) as e:
+        call_command(
+            "create_project",
+            "foo2",
+            "--preset-mapping=gnu",
+            "--fs-url=NOTANABSOLUTEPATH")
+    assert (
+        "Enter an absolute path "
+        in str(e))
+    with pytest.raises(Project.DoesNotExist):
+        Project.objects.get(code="foo2")
--- a/tests/commands/minus_minus_help.py
+++ b/tests/commands/minus_minus_help.py
@@ -16,6 +16,7 @@ CORE_APPS_WITH_COMMANDS = (
 
 
 @pytest.mark.cmd
+@pytest.mark.django_db
 @pytest.mark.parametrize("command,app", [
     (command, app)
     for command, app in get_commands().iteritems()


pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,87 @@
...
+            help="Does the project start disabled? [ True | False ]",
+        )
+
+    def handle(self, **options):
+        self.create_project(**options)
+
+    def create_project(self, **options):
+        """Immitates the behaviour of the admin interface for creating a project from command line."""
+
+	from pootle_language.models import Language
```
phlax: could you move these imports to top of file, they need to be below django imports and in alpha order (or tests will fail)

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,87 @@
...
+	from pootle_project.models import Project
+
+	source_language_obj = Language.objects.get(code=options["sourcelang"]) 
+	
+        if "name" not in options:
+          options["name"] = options["code"].capitalize()
+		
+	new_project = Project.objects.create(code=options["code"], fullname=options["name"], source_language=source_language_obj)
```
phlax: you will need to make this statement multiline or it will fail pep8

also re fullname, rather than setting above as you have done you could do...

...`, fullname=options["name"] or options["code"].capitalize(), `...

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -10,7 +10,8 @@
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
 
 from django.core.management.base import BaseCommand
-
+from pootle_language.models import Language
+from pootle_project.models import Project
```
unho: Need two blank lines after imports.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -10,7 +10,8 @@
 os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
 
 from django.core.management.base import BaseCommand
-
```
unho: Need 1 blank line between import groups.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def handle(self, **options):
+        self.create_project(**options)
+
+    def create_project(self, **options):
+        """Immitates the behaviour of the admin interface for creating a project from command line."""
+
+	source_language_obj = Language.objects.get(code=options["sourcelang"]) 
+	
+	new_project = Project.objects.create(code=options["code"], fullname=options["name"] or options["code"].capitalize(), source_language=source_language_obj)
```
phlax: this line needs to be split like

```
new_project = Project.objects.create(
    code=options["code"],
    fullname=options["name"] or options["code"].capitalize(),
    source_language=source_language_obj)

File: pootle/apps/pootle_app/management/commands/create_project.py [hunk 6]
```diff
@@ -76,12 +77,6 @@ def handle(self, **options):
...
-	from pootle_language.models import Language
-	from pootle_project.models import Project
-
 	source_language_obj = Language.objects.get(code=options["sourcelang"]) 
 	
-        if "name" not in options:
-          options["name"] = options["code"].capitalize()
-		
-	new_project = Project.objects.create(code=options["code"], fullname=options["name"], source_language=source_language_obj)
+	new_project = Project.objects.create(code=options["code"], fullname=options["name"] or options["code"].capitalize(), source_language=source_language_obj)
```
unho: This still is too long, so will fail pep8 test.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,82 @@
...
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+
+import os
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from django.core.management.base import BaseCommand
```
phlax: leave a blank line between django imports and pootle ones

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,88 @@
...
+from pootle_language.models import Language
+from pootle_project.models import Project
+
+
+class Command(BaseCommand):
+    help = "Creates a new project from command line."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--code',
```
unho: I think this is a mandatory argument so specify `required=True`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,88 @@
...
+            action="store",
+            dest="quality",
+            default="standard",
+            help="Quality management",
+        )
+        parser.add_argument(
+            "--filetype",
+            action="store",
+            dest="filetype",
+            help="File type: [ po/pot | xliff | xlf | ts | lang ]",
```
unho: `po/pot` might be difficult to type. I suggest to use just `po`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        parser.add_argument(
+            "--structure",
+            action="store",
+            dest="structure",
+            default="gnu",
+            help="Catalogue structure: [ automatic | gnu | no-gnu | filesystem ]",
```
unho: We have used `non-gnu` for years, so I would stick to it.

Do we want `automatic`?

What is `filesystem`?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+    def handle(self, **options):
+        self.create_project(**options)
+
+    def create_project(self, **options):
+        """Immitates the behaviour of the admin interface
+            for creating a project from command line."""
+
+        source_language_obj = Language.objects.get(code=options["sourcelang"])
```
unho: What if the passed code doesn't exist in Pootle? I guess in that case you should trigger some error:

```python
    try:
        source_language_obj = Language.objects.get(code=options["sourcelang"])
    except Language.DoesNotExist:
        raise CommandError("Source language %s does not exist" % options["sourcelang"])
```

pootle/apps/pootle_app/management/commands/create_project.py
phlax: these need to match `filetype` names

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) Pootle contributors.
+#
+# This file is a part of the Pootle project. It is distributed under the GPL3
+# or later license. See the LICENSE file for a copy of the license and the
+# AUTHORS file for copyright and authorship information.
+import os
```
unho: Nit: Add a blank line between imports and comments.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: these need to match the preset mappings, or be an actual mapping

unho: If I am correct these are the supported preset mappings https://github.com/translate/pootle/blob/538300f9e13cc18f54f79644819995a9e5511334/pootle/apps/pootle_fs/forms.py#L23-L31 So `gnu`, `non-gnu`, `django`.

pootle/apps/pootle_app/management/commands/create_project.py
unho: If I am correct these are https://github.com/translate/pootle/blob/5228e3d89df31492b7271836f1f494edc4b47d62/pootle/apps/pootle_format/default.py#L10-L30

pootle/apps/pootle_app/management/commands/create_project.py
phlax: yep

phlax: but best not to hardcode - ie use `FS_CHOICES.keys()` or similar

unho: I agree, provided the `custom` key is dropped.

phlax: hmm, we need to figure out how to handle translation_mappings, which would be the `custom` option in  terms of presets

pootle/apps/pootle_app/management/commands/create_project.py
phlax: there is a special api for getting these, they can be registered by 3rd party software (might be same for presets actually - i need to check)

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            "--name",
+            action="store",
+            dest="name",
+            help="Project name",
+        )
+        parser.add_argument(
+            "--filetype",
+            action="append",
+            dest="filetypes",
+            choices=formats.get().keys(),
```
phlax: probably better to do `formats.get()` once at the top of this method

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            action="store",
+            dest="mapping",
+            help="Custom mapping layout",
+        )
+        parser.add_argument(
+            "--fs-type",
+            action="store",
+            dest="fs_type",
+            default="localfs",
+            choices=fs_plugins.gather().keys(),
```
phlax: probably better to do fs_plugins.gather() once at the top of this method

pootle/apps/pootle_app/management/commands/create_project.py
unho: Addressed.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+os.environ['DJANGO_SETTINGS_MODULE'] = 'pootle.settings'
+
+from django.core.management.base import BaseCommand
+from django.core.exceptions import ValidationError
+from django.core.management import CommandError
+from django.core.validators import validate_email
+
+from pootle.core.delegate import formats
+from pootle_fs.presets import FS_PRESETS
+from pootle_fs.delegate import fs_plugins
```
unho: This import should be before the previous one.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+from pootle_language.models import Language
+from pootle_project.models import Project
+from pootle_project.models import PROJECT_CHECKERS
```
unho: Combine this import with previous one.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            choices=fs_plugins.gather().keys(),
+            help="Filesystem type: {0}".format(
+                " | ".join(fs_plugins.gather().keys())),
+        )
+        parser.add_argument(
+            "--fs-url",
+            action="store",
+            dest="fs_url",
+            default="",
+            help="Filesystem url.",
```
unho: Nit: It should be `URL`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+        parser.add_argument(
+            "--source-language",
+            action="store",
+            dest="sourcelang",
+            default="en",
+            help="Source language. Examples: [en | es | fr | ...]",
```
unho: I would print the default using argparse's `%default` placeholder https://docs.python.org/2/library/argparse.html#help

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            help="Does the project start disabled?",
+        )
+        parser.add_argument(
+            "--checkstyle",
+            action="store",
+            dest="checkstyle",
+            choices=PROJECT_CHECKERS.keys(),
+            default="standard",
+            help="Quality check styles. Example: {0}"
+                 .format(" | ".join(PROJECT_CHECKERS.keys())),
```
unho: Nit: I would include default in help string.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            default="localfs",
+            choices=fs_plugins.gather().keys(),
+            help="Filesystem type: {0}".format(
+                " | ".join(fs_plugins.gather().keys())),
+        )
+        parser.add_argument(
+            "--fs-url",
+            action="store",
+            dest="fs_url",
+            default="",
```
unho: Do we need to specify a default here?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            action="store",
+            dest="name",
+            help="Project name",
+        )
+        parser.add_argument(
+            "--filetype",
+            action="append",
+            dest="filetypes",
+            choices=formats.get().keys(),
+            default=[],
```
unho: Does this need a default?

pootle/apps/pootle_app/management/commands/create_project.py
unho: Also, what about Filesystem path, for example for `localfs`?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+        """Immitates the behaviour of the admin interface
+            for creating a project from command line."""
+
+        try:
+            source_language_obj = Language.objects.get(code=options["sourcelang"])
+        except Language.DoesNotExist:
+            raise CommandError("Source language %s does not exist"
+                               % options["sourcelang"])
+
+        if(options["contact"]):
```
unho: There is no need for parentheses. Also need a whitespace after `if`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            except ValidationError as e:
+                raise CommandError(e)
+
+        new_project = Project.objects.create(
+            code=options["code"],
+            fullname=options["name"] or options["code"].capitalize(),
+            checkstyle=options["checkstyle"],
+            source_language=source_language_obj,
+            filetypes=options["filetypes"] or ["po"],
+            report_email=options["contact"] or "",
```
unho: I would put the default in the argument definition above.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: ... or path

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            disabled=options["disabled"])
+
+        if(options["preset_mapping"]):
```
unho: Again no need for parentheses and also whitespace needed after `if`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            new_project.config["pootle_fs.translation_mapping"] = dict(
+                default=FS_PRESETS[options["preset_mapping"]][0])
+        else:
+            new_project.config["pootle_fs.translation_mapping"] = dict(
+                default=options["mapping"])
```
unho: Shouldn't this be sanitized in any way?

pootle/apps/pootle_app/management/commands/create_project.py
phlax: not here (the default uses the project code) - but perhaps in the help

pootle/apps/pootle_app/management/commands/create_project.py
phlax: it cant - without the default always being included

pootle/apps/pootle_app/management/commands/create_project.py
unho: Same below.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+
+        if(options["fs_type"]):
```
unho: We have a default for this argument, so this check is unnecessary, just assign it.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: its supposed to be in the config setter - i need to follow up - see disabled test

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            new_project.config["pootle_fs.fs_type"] = options["fs_type"]
+
+        if(options["fs_url"]):
+            new_project.config["pootle_fs.fs_url"] = options["fs_url"]
```
unho: Is this not checked before assignment?

phlax: it should raise an error if the assignment is wrong - i need to follow up

unho: Maybe I am a bit mislead, but is it not necessary to save the config at the end?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            choices=PROJECT_CHECKERS.keys(),
+            default="standard",
+            help="Quality check styles. Example: {0}"
+                 .format(" | ".join(PROJECT_CHECKERS.keys())),
+        )
+
+    def handle(self, **options):
+        self.create_project(**options)
+
+    def create_project(self, **options):
```
unho: Nit: I would combine this method into `handle`.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            dest="preset_mapping",
+            choices=FS_PRESETS.keys(),
+            help="Preset filesystem layout: {0}".format(
+                 " | ".join(FS_PRESETS.keys())),
+        )
+        mapping_group.add_argument(
+            "--mapping",
+            action="store",
+            dest="mapping",
+            help="Custom mapping layout",
```
unho: `Custom filesystem layout`. Also I would include some help on how that layout must look like.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+
+from pootle.core.plugin import provider
+from pootle_fs.delegate import fs_plugins
+from pootle_fs.plugin import Plugin
+from pootle_project.models import Project
+
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_defaults(settings):
```
unho: Is it possible to create a project without specifying a "FS path or URL"?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+from pootle_project.models import Project
+from pootle_project.models import PROJECT_CHECKERS
+
+
+class Command(BaseCommand):
+    help = "Creates a new project from command line."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'code',
```
unho: This argument must be required.

pootle/apps/pootle_app/management/commands/create_project.py
unho: I don't understand, what is always included?

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+                 " | ".join(FS_PRESETS.keys())),
+        )
+        mapping_group.add_argument(
+            "--mapping",
+            action="store",
+            dest="mapping",
+            help="Custom mapping layout",
+        )
+        parser.add_argument(
+            "--fs-type",
```
unho: This argument should be required as well.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            action="store",
+            dest="fs_type",
+            default="localfs",
+            choices=fs_plugins.gather().keys(),
+            help="Filesystem type: {0}".format(
+                " | ".join(fs_plugins.gather().keys())),
+        )
+        parser.add_argument(
+            "--fs-url",
```
unho: This argument should be required.

pootle/apps/pootle_app/management/commands/create_project.py
unho: I don't follow, how the "FS path or URL" gets a default from the project code? Are you maybe assuming the the "FS backend" is `localfs`?

pootle/apps/pootle_app/management/commands/create_project.py
unho: Ok, then change the help to "Filesystem path or URL."

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            action="store",
+            dest="fs_url",
+            default="",
+            help="Filesystem url.",
+        )
+        parser.add_argument(
+            "--source-language",
```
unho: This should be required.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            help="Contact email for reports. Example: admin@mail.com",
+        )
+        parser.add_argument(
+            "--disabled",
+            action="store_true",
+            dest="disabled",
+            help="Does the project start disabled?",
+        )
+        parser.add_argument(
+            "--checkstyle",
```
unho: This should be required.

pootle/apps/pootle_app/management/commands/create_project.py
unho: When it comes to a command I would expect all provided input is sanitized before doing anything. With this pattern you might end creating a project with partial or wrong config.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_existing_project(settings):
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu")
+
+    with pytest.raises(ValidationError):
```
unho: Shouldn't this raise a `CommandError` saying that the project already exists?

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_title(capfd):
+    call_command(
+        "create_project",
+        "foo1",
+        "--name=Bar1",
+        "--preset-mapping=gnu")
+    foo1 = Project.objects.get(code="foo1")
+    assert foo1.fullname == "Bar1"
```
unho: Test also name not being specified.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.filetypes.first().name == "ts"
+
+    call_command(
+        "create_project",
+        "foo1",
+        "--filetype=ts",
+        "--preset-mapping=gnu")
+    foo1 = Project.objects.get(code="foo1")
+    assert foo1.filetypes.first().name == "ts"
```
unho: Why two valid calls that are essentially the same?

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+    call_command(
+        "create_project",
+        "foo0",
+        "--report-email=foo@bar.com",
+        "--preset-mapping=gnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.report_email == "foo@bar.com"
+
+    with pytest.raises(CommandError):
+        # invalid email address
```
unho: Instead of trusting that `CommanError` here is because of invalid email address I would rather check that the error message in the exception contains such error. See second example in https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+        call_command(
+            "create_project",
+            "foo0",
+            "--report-email=foo@bar",
```
unho: Nit: I would use something that is much more obvious it is not a valid email, like `NOT_AN_EMAIL`.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+@pytest.mark.cmd
+@pytest.mark.django_db
+def test_cmd_create_project_source_language(language0):
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--source-language=%s" % language0.code)
+    foo0 = Project.objects.get(code="foo0")
+    assert foo0.source_language == language0
```
unho: I would also test the source language not being specified.

unho: Forget about this, already done in first test.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+@pytest.mark.django_db
+def test_cmd_create_project_preset_mapping():
+    call_command(
+        "create_project",
+        "foo0",
+        "--preset-mapping=nongnu")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.translation_mapping"]
+        == dict(default=u'/<language_code>/<dir_path>/<filename>.<ext>'))
```
unho: Can't this value be retrieved from the `FS_PRESETS`?

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+    with pytest.raises(CommandError):
+        call_command(
+            "create_project",
+            "foo1",
+            "--preset-mapping=gnu",
+            "--mapping=u'/<language_code>/<dir_path>/<filename>.<ext>'")
```
unho: I would say the `u` prefix shouldn't be there.

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+        call_command(
+            "create_project",
+            "foo1")
```
unho: Two command calls in the same `pytest.raises`?

pootle/apps/pootle_app/management/commands/create_project.py
phlax: no - the config automatically saves

tests/commands/create_project.py
phlax: yes - as in the project form it defaults to `{POOTLE_TRANSLATION_DIRECTORY}$projectcode`

pootle/apps/pootle_app/management/commands/create_project.py
phlax: no - its not - it defaults to `localfs` - no need to specify it

pootle/apps/pootle_app/management/commands/create_project.py
phlax: no it defaults to `en`

tests/commands/create_project.py
```diff
@@ -0,0 +1,269 @@
...
+        "create_project",
+        "foo0",
+        "--preset-mapping=gnu",
+        "--fs-type=dummyfs")
+    foo0 = Project.objects.get(code="foo0")
+    assert (
+        foo0.config["pootle_fs.fs_type"]
+        == "dummyfs")
+
+    with pytest.raises(CommandError):
```
unho: As with the other `pytest.raises(CommandError)` I would check for the exception message to ensure it fails for the reason we are testing.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: see above comment on the test

unho:
Overall it looks ok, despite there are still a few things to address. Great job people!

tests/commands/create_project.py
phlax: that is tested in the defaults test

tests/commands/create_project.py
phlax: ill remove one of these

tests/commands/create_project.py
phlax: hmm, i think it better in this case to use a more borderline example

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: @phlax discovered a bug here. https://bugs.python.org/issue16399 append action with default list adds to list instead of overridin

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,149 @@
...
+            help="Project name",
+        )
+        parser.add_argument(
+            "--filetype",
+            action="append",
+            dest="filetypes",
+            choices=formats.get().keys(),
+            default=[],
+            help="File types: {0}".format(
+                 " | ".join(formats.get().keys())),
```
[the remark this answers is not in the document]
jjmcarrascosa: I have included it, I hope this is not misleading for other people who may read that we don't set a default but in the help we say it's "po".

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: I have added "Default: "

pootle/apps/pootle_app/management/commands/create_project.py
phlax: hmm, lets make this command atomic, and then it wont - if it triggered an error i would have caught it

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: positional arguments are always required. Actually, there is no "Required=true" for this kind of arguments, argparse raises an error.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: no - it has sane defaults

jjmcarrascosa: But we have a `default="standard"`here...

pootle/apps/pootle_app/management/commands/create_project.py
phlax: yep, fair point we need to make this required only if the fs-type is not `localfs`

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,148 @@
...
+            dest="fs_url",
+            default="",
+            help="Filesystem path or URL.",
+        )
+        parser.add_argument(
+            "--source-language",
+            action="store",
+            dest="sourcelang",
+            default="en",
+            help="Source language. Examples: [en | es | fr | ...].\
```
unho: Avoid using `\` for continuing strings. Use parentheses instead:

```python
        help=("Source language. Examples: [en | es | fr | ...]."
              "Default: %(default)s"),
```

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,148 @@
...
+                     Default: %(default)s",
+        )
+        parser.add_argument(
+            "--report-email",
+            action="store",
+            dest="contact",
+            help="Contact email for reports. Example: admin@mail.com. Default: ",
```
unho: There is no default here.

jjmcarrascosa: You told to put a default in help, but we don't have any default. I am not sure we can put a default email. Imho it's better to leave it blank? Or it's mandatory to have one?

pootle/apps/pootle_app/management/commands/create_project.py
unho: I was asking you to add the `default` above:

```python
        parser.add_argument(
            "--report-email",
            action="store",
            default="",
            dest="contact",
            help="Contact email for reports. Example: admin@mail.com.",
        )
```

and then change this line to:

```python
            report_email=options["contact"],
```

jjmcarrascosa: Now I got you. Done.

pootle/apps/pootle_app/management/commands/create_project.py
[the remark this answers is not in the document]
unho: Maybe it is a good idea to put a comment in the code to clarify why this happens.

pootle/apps/pootle_app/management/commands/create_project.py
unho: Shouldn't this be changed to use `fs_plugins_names`?

pootle/apps/pootle_app/management/commands/create_project.py
unho: In this case I wouldn't include the default in the help.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            new_project.config["pootle_fs.translation_mapping"] = dict(
+                default=options["mapping"])
+
+        new_project.config["pootle_fs.fs_type"] = options["fs_type"]
+
+        if options["fs_url"]:
+            try:
+                new_project.config["pootle_fs.fs_url"] = options["fs_url"]
+            except ValidationError as e:
```
unho: I still think that all input should be validated before doing any change. That way we can avoid the situation where a project gets created but its config is not properly set. Have in mind this command only allows to create new projects, but not altering already existing projects.

Also if we can verify beforehand if the input is wrong or not, why not just tell the user to fix the mistakes before committing any change?

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: Well actually it's already there, I've just forgotten to put there `format_names`. Done.

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: Sure, my fault. Done.

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: I've done this with a if block, not sure if it can be done otherwise

pootle/apps/pootle_app/management/commands/create_project.py
jjmcarrascosa: Yes, that's how it was initially. Ok.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,160 @@
...
+            help="Custom filesystem layout",
+        )
+        parser.add_argument(
+            "--fs-type",
+            action="store",
+            dest="fs_type",
+            default="localfs",
+            choices=fs_plugins_names,
+            help="Filesystem type: {0}".format(
+                " | ".join(fs_plugins_names)),
```
unho: Nit: Move this to previous line.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,171 @@
...
+            help="File types: {0}. Default: po".format(
+                 " | ".join(format_names)),
+        )
+        mapping_group = parser.add_mutually_exclusive_group(required=True)
+        mapping_group.add_argument(
+            "--preset-mapping",
+            action="store",
+            dest="preset_mapping",
+            choices=FS_PRESETS.keys(),
+            help="Preset filesystem layout: {0}".format(
```
unho: Nit: "Filesystem layout preset:"

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,171 @@
...
+            dest="checkstyle",
+            choices=PROJECT_CHECKERS.keys(),
+            default="standard",
+            help="Quality check styles. Example: {0}. Default: %(default)s"
+                 .format(" | ".join(PROJECT_CHECKERS.keys())),
+        )
+
+    @transaction.atomic
+    def handle(self, **options):
+        """Immitates the behaviour of the admin interface
```
unho: Typo: Imitates

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,171 @@
...
+            return Language.objects.get(code=options["sourcelang"])
+        except Language.DoesNotExist:
+            raise CommandError("Source language %s does not exist"
+                               % options["sourcelang"])
+
+    def set_project_config(self, project, **options):
+        mapping = (
+            dict(default=FS_PRESETS[options["preset_mapping"]][0])
+            if options["preset_mapping"]
+            else dict(default=options["mapping"]))
```
unho: Is the custom mapping validated in any way?

phlax: yes, when its set

unho: nvm, just saw the test.

pootle/apps/pootle_app/management/commands/create_project.py
```diff
@@ -0,0 +1,171 @@
...
+        self.set_project_config(
+            self.create_project(**options), **options)
+
+    def check_project_options(self, **options):
+        if options["contact"]:
+            try:
+                validate_email(options["contact"])
+            except ValidationError as e:
+                if options["traceback"]:
+                    traceback.print_exc()
```
unho: I am wondering if the traceback on validation might be needed in any scenario. I mean, it either fails the validation or passes it, seeing the traceback does not convey any useful information, afaict, besides the one that the commanderror with the validation error message will already provide.

unho:
lgtm, nits aside.

pootle/apps/pootle_app/management/commands/create_project.py
phlax: i think the problem is that when you catch and reraise with CommandError - it swallows the original error and makes the `--traceback` option useless - i think its better to have the option to print the traceback should you require

unho: Ok.

phlax:
gtm, nice work!

unho:
lgtm

PR #6644, opened 2017-08-29
Docs adjustments for Pootle FS and misc changes

### LINKED ISSUE #6637: Document switch to pootle fs for file syncing
As we have now removed the old (project_tree) update/sync code paths and switched to pootle fs we need to go through docs and ensure that they are compatible/correct

unho:
This PR fixes #6637.


docs/server/project_setup.rst
```diff
@@ -116,7 +116,25 @@ unless you need to further customize your project.
...
+Please not that **just creating the project is not enough**, now it is
+necessary to set the the filesystem configuration. So now click on the
+``Filesystems`` link below the project edit form and set the following:
+
+.. image:: ../_static/pootle_fs_link.png
+
+
+* **Filesystem backend** to ``localfs``
+* **Backend URL or path** to the value of
+  :setting:`POOTLE_TRANSLATION_DIRECTORY` + project code, e.g.
```
dwaynebailey: This is not true, you can set it to anything accessible, it is not related to `POOTLE_TRANSLATION_DIRECTORY` at all.

docs/server/project_setup.rst
```diff
@@ -116,7 +116,25 @@ unless you need to further customize your project.
...
+  :file:`/path/to/pootle/translations/my-project` in our example
+* **Translation mapping presets** to ``GNU-style``
+
+And then click on the **Update filesystem configuration** button to save the
+Pootle FS configuration for the project.
+
+Creating the project doesn't actually import all the translations to
 Pootle. To do that you need to run :djadmin:`update_stores` on the command line
```
dwaynebailey: Are we going to revisit the `update_stores` part later?

docs/server/project_setup.rst
```diff
@@ -78,7 +78,7 @@ Below you can see an example with two projects using the GNU layout:
 Among the regular translation files there are two files named 
 :file:`templates.pot`. These are the template (master or reference) files that
 contain the original strings. Usually these files contain only English strings,
-it is much less confusing to use the term ``templates`` than e.g. ``en`` or
+but it is much less confusing to use the term ``templates`` than e.g. ``en`` or
```
dwaynebailey: I'd use the `however`

docs/features/migrate_to_pootle_fs.rst
```diff
@@ -34,10 +34,15 @@ following:
...
-* *Backend URL or path* to the value of
+* **Filesystem backend** to ``localfs``
+* **Backend URL or path** to the value of
   :setting:`POOTLE_TRANSLATION_DIRECTORY` + MYPROJECT, e.g.
   :file:`/path/to/pootle/translations/MYPROJECT`
+* **Translation path mapping** to the one your project uses.
+
+  .. note:: You can instead pick one of the **Translation mapping presets** to
+     easily set your project's **Translation path mapping** in case your
+     project actually matches one of the available presets.
```
dwaynebailey: I hate seeing repeated paragraphs, they're just open to get out of sync and out of date.  Do we need to explain something that is on the actual form? And I'm not sure why we need a note, that is to make something off topic or highlight, none of which is happening here. Just add "You can use one of the presets or specify your own mapping", but even that seems redundant as we're stepping them through a task list and that info is on the screen.

docs/server/commands.rst
```diff
@@ -233,7 +233,12 @@ When the :option:`--reset` option is used , all score log data is removed and
 sync_stores
 ^^^^^^^^^^^
 
-.. versionchanged:: 2.7
+.. versionchanged:: 2.9
+
+
+.. note:: Since version 2.9 all projects are managed by Pootle FS and therefore
```
dwaynebailey: Don't we want to mark these as deprecated?

docs/server/project_setup.rst
phlax: it kinda is in that this is the default directory

docs/server/commands.rst
phlax: not sure tbh - they could be useful as a collection of api calls

docs/server/project_setup.rst
unho: Changed.

docs/server/project_setup.rst
unho: Since this is an example I tried to keep it as simple and familiar as possible. Note the text doesn't tell in any way that it can't have any other value here.

docs/server/project_setup.rst
unho: Given that these commands are working with Pootle FS I was wondering if we actually want to drop them now that everything *works* with Pootle FS or if they would stay around a bit more.

So in short this stays as it is until we agree when to deprecate the commands formally.

docs/features/migrate_to_pootle_fs.rst
unho: I removed those in favor of simpler text.

docs/server/commands.rst
unho: Given that these commands now use Pootle FS I was wondering if we actually want to drop them or if they would stay around a bit more to provide easier way to sync projects.

So in short I would leave this as it is until we agree when to deprecate or drop the commands formally.

PR #6651, opened 2017-08-30
Add back increment_unsynced_revision algo

phlax:
increment unsynced revisions after fs.sync where update="pootle"

that way the unsynced units should be showing as unsynced and will update the file on next sync

this makes update_stores/sync_stores work independently

### DIFF
--- a/pootle/apps/pootle_fs/plugin.py
+++ b/pootle/apps/pootle_fs/plugin.py
@@ -399,7 +399,8 @@ class Plugin(object):
         return response
 
     @responds_to_state
-    def sync_merge(self, state, response, fs_path=None, pootle_path=None):
+    def sync_merge(self, state, response, fs_path=None,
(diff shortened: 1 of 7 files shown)


pootle/apps/pootle_fs/plugin.py
```diff
@@ -524,8 +526,12 @@ def sync(self, state, response, fs_path=None, pootle_path=None, update="all"):
         """
         self.sync_rm(
             state, response, fs_path=fs_path, pootle_path=pootle_path)
-        self.sync_merge(
-            state, response, fs_path=fs_path, pootle_path=pootle_path)
+        if update in ["all", "pootle"]:
```
unho: Nit: Unnecessary to add this `if` since it already exists immediately below.

unho:
lgtm even without nit provided Travis passes.

PR #6653, opened 2017-08-31
Releasing 2.8.1

unho:


README.rst
```diff
@@ -7,7 +7,7 @@
 .. Resources
 
 `Docs <http://docs.translatehouse.org/projects/pootle/en/stable-2.8.0/>`_ |
-`Changes <http://docs.translatehouse.org/projects/pootle/en/stable-2.8.0/releases/2.8.0.html>`_ |
+`Changes <http://docs.translatehouse.org/projects/pootle/en/stable-2.8.0/releases/2.8.1.html>`_ |
```
dwaynebailey: Pretty sure this should be `stable-2.8` or similar

unho: I will have a look into the script to see if we can force using branches like `stable-2.8.x` which is what we are aiming for now.

PR #6656, opened 2017-09-01
Xliff roundtrip

phlax:
add test to ensure that xliff files can roundtrip


unho:
It looks like the test fails.

PR #6665, opened 2017-09-05
Fix for unit obsoletion

phlax:
dont obsolete units that have been updated in diff class, but then do exclude them if not pootle wins in updater


dwaynebailey:
gtm, based on live testing

PR #6672, opened 2017-09-06
Make unit required field for Submission

phlax:
as in title


unho:
@phlax This missed the data migration to remove all existing submissions with no unit. Otherwise people upgrading from a legacy deployment will have a hard time.

phlax:
@unho there is already a data migration for that - i went with the assumption that it would not have occurred since that migration

unho:
@phlax You are right. My apologies for the noise.

PR #6680, opened 2017-09-06
Lang discovery adjustments

unho:


phlax:
tested locally and this fixes a few issues. Needs some tests tho. Gtm once test/s added

tests/i18n/override.py
```diff
@@ -0,0 +1,168 @@
...
+    # Test header with unsupported language.
+    request = rf.get("", HTTP_ACCEPT_LANGUAGE='FAIL')
+    assert get_lang_from_http_header(request, SUPPORTED_LANGUAGES) is None
+
+
+@pytest.mark.django_db
+def test_get_lang_from_session(rf, client):
+    # Test no session.
+    request = rf.get("")
+    assert not hasattr(request, 'session')
```
phlax: i think this assert is not what is being tested - better add a comment to say that there is no session

tests/i18n/override.py
```diff
@@ -0,0 +1,168 @@
...
+    response = client.get("")
+    request = response.wsgi_request
+    request.session[LANGUAGE_SESSION_KEY] = 'en-US'
+    assert get_lang_from_session(request, SUPPORTED_LANGUAGES) == 'en-US'
+
+    # Test cookie with longer underscore language code for a supported
+    # language.
+    response = client.get("")
+    request = response.wsgi_request
+    request.session[LANGUAGE_SESSION_KEY] = 'gl_ES'
```
phlax: missing assert ?

tests/i18n/override.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Test cookie with longer hyphen language code for a supported language.
+    assert get_lang_from_session(request, SUPPORTED_LANGUAGES) is None
+    response = client.get("")
+    request = response.wsgi_request
+    request.session[LANGUAGE_SESSION_KEY] = 'fr-FR'
+    assert get_lang_from_session(request, SUPPORTED_LANGUAGES) is None
```
phlax: should this not fallback to `fr` ?

phlax:
im not seeing in any of the current tests anything testing for the fallback codes that we are adding. Apart from seeing the fallback working, it would also be good to see something where it doesnt fallback - eg `spam-EGGS`

tests/i18n/override.py
unho: No, here we are telling the session to use this particular language. Maybe there is a better way to achieve this, but this works.

tests/i18n/override.py
unho: Probably yes, but will need to change the function to support that. Will look into it.

tests/i18n/override.py
unho: Will do

tests/i18n/override.py
unho: It results that the missing assert is a few lines below.

tests/i18n/override.py
unho: Updated PR to handle this.

tests/i18n/override.py
unho: Fixed now.

unho:
Updated with all raised concerns.

phlax:
lgtm

PR #6684, opened 2017-09-07
i18n: Align field labels

unho:


pootle/apps/pootle_app/forms.py
```diff
@@ -55,7 +55,7 @@ class ProjectForm(forms.ModelForm):
     source_language = forms.ModelChoiceField(label=_('Source Language'),
                                              queryset=Language.objects.none())
     fs_plugin = forms.ChoiceField(
-        label=_('FS Plugin'),
+        label=_('Filesystem backend'),
```
phlax: this isnt actually displayed anywhere - not sure if we can just remove it

unho: I think you are right. I will adjust the PR.

pootle/apps/pootle_app/forms.py
```diff
@@ -52,11 +52,8 @@ def clean_specialchars(self):
 
 class ProjectForm(forms.ModelForm):
 
-    source_language = forms.ModelChoiceField(label=_('Source Language'),
```
phlax: this one should also probably be removed

phlax:
lgtm

PR #6685, opened 2017-09-07
Add some fail handling to fs migration

phlax:
currently if the fs and db are not in sync, this migration can fail. This PR fixes that, and makes an optimisation

### DIFF
--- a/pootle/apps/pootle_fs/migrations/0002_convert_localfs.py
+++ b/pootle/apps/pootle_fs/migrations/0002_convert_localfs.py
@@ -15,6 +15,9 @@ from django.db import migrations
 from translate.lang.data import langcode_re
 
 
+logger = logging.getLogger(__name__)
+
+
 def _file_belongs_to_project(project, filename):
     ext = os.path.splitext(filename)[1][1:]
     filetype_extensions = list(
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_fs/migrations/0002_convert_localfs.py
```diff
@@ -88,12 +91,15 @@ def _set_project_config(Config, project_ct, project):
...
         defaults=dict(
             value="localfs"))
-    Config.objects.update_or_create(
-        content_type=project_ct,
-        object_pk=project.pk,
-        key="pootle_fs.translation_mappings",
-        defaults=dict(
-            value=dict(default=_get_translation_mapping(project))))
+    old_translation_path = settings.POOTLE_TRANSLATION_DIRECTORY
+    proj_trans_path = os.path.join(old_translation_path, project.code)
```
unho: There is no need to set again these two variables.

pootle/apps/pootle_fs/migrations/0002_convert_localfs.py
```diff
@@ -106,30 +112,52 @@ def convert_to_localfs(apps, schema_editor):
...
+            translation_project__project=project).exclude(file="").exclude(obsolete=True)
         _set_project_config(Config, project_ct, project)
         project.treestyle = "pootle_fs"
         project.save()
+
+        if project.disabled:
+            continue
+        if not os.path.exists(proj_trans_path):
+            logger.warn(
+                "Missing project ('%s') translation directory '%s', skipping",
```
unho: I would be more clear on what is being skipped. Someone might think the project is skipped while it was changed a few lines above.

unho:
I would put the optimization in a separate commit.

PR #6687, opened 2017-09-07
Add warning in sync_stores if there are unresolved conflicts

### LINKED ISSUE #6676: sync_stores/update_stores
For 2.9.0 we're making Pootle FS the main path into Pootle, but `sync_stores` and `update_stores` are kept for backward compatibility.

But we have situations where Pootle FS can't do what sync/update does in that it may do more or may touch the db/file when sync/update would only do it on one side.

So we need to warn:
1. That this mode is deprecated
2. Where there are conflicts e.g. we should warn and bail or warn and allow it to continue

Essentially, we want people to know when we can't do what they expect.

phlax:
fixes #6676

### DIFF
--- a/pootle/apps/pootle_app/management/commands/sync_stores.py
+++ b/pootle/apps/pootle_app/management/commands/sync_stores.py
@@ -6,11 +6,15 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import logging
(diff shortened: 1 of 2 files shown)


unho:
@phlax Note that this doesn't fully fix 6676 since the first part of that issue is "Warn that `sync_stores` and ` update_stores` commands are deprecated" unless I am missing something.

PR #6702, opened 2017-09-09
Dont redirect on submitting last unit if task is not complete

### LINKED ISSUE #4795: Translating last unit says "congratulations" regardless of whether task is actually finished
it should only say congrats, if the task is actually completed i would have thought

phlax:
fixes #4795


unho:
Reviewed and tested. It works ok.

But I am not sure if the new behavior is enough since after submitting the translation I would expect some change. Maybe we can also show a notification saying "There are unfinished strings above." ?

unho:
@dwaynebailey Maybe you can share your thoughts on also displaying some notification "There are unfinished strings above." since just staying in the unit might be confusing for the translators.

PR #6703, opened 2017-09-09
Use langcode.po when initing from templates.pot

phlax:
currently in init_from_templates if the template is called templates.pot (ie gnu-style layout) it will create a store called templates.po - this means all files will called templates.po and wont be able to sync to disk.

this PR uses langcode.po when initing from a template with this name, which is the expected behaviour for gnu-style projects

### DIFF
--- a/pootle/apps/pootle_translationproject/models.py
+++ b/pootle/apps/pootle_translationproject/models.py
@@ -250,11 +250,20 @@ class TranslationProject(models.Model, CachedTreeItem):
         pootle_path = ".".join(
             [posixpath.splitext(pootle_path)[0],
              template_store.filetype.extension.name])
-        store, __ = self.stores.get_or_create(
-            parent=self.create_parent_dirs(pootle_path),
-            pootle_path=pootle_path,
-            name=posixpath.basename(pootle_path))
-        return store
+        name = posixpath.basename(pootle_path)
+        if name in ["template.po", "templates.po"]:
+            # gnu-style layout
+            # use language code instead of template name
+            name = ".".join(
+                [self.language.code,
+                 template_store.filetype.extension.name])
+            dirname = posixpath.dirname(pootle_path)
+            pootle_path = posixpath.join(dirname, name)
+        if not self.stores.filter(pootle_path=pootle_path).exists():
+            return self.stores.create(
+                parent=self.create_parent_dirs(pootle_path),
+                pootle_path=pootle_path,
+                name=name)
 
     def init_from_templates(self):
         """Initializes the current translation project files using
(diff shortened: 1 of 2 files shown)


pootle/apps/pootle_translationproject/models.py
```diff
@@ -250,23 +250,33 @@ def init_store_from_template(self, template_store):
...
+            name = ".".join(
+                [self.language.code,
+                 template_store.filetype.extension.name])
+            dirname = posixpath.dirname(pootle_path)
+            pootle_path = posixpath.join(dirname, name)
+        if not self.stores.filter(pootle_path=pootle_path).exists():
+            return self.stores.create(
+                parent=self.create_parent_dirs(pootle_path),
+                pootle_path=pootle_path,
+                name=name)
```
unho: This `return` has changed previous behavior.

pootle/apps/pootle_translationproject/models.py
```diff
@@ -250,23 +250,33 @@ def init_store_from_template(self, template_store):
         pootle_path = ".".join(
             [posixpath.splitext(pootle_path)[0],
              template_store.filetype.extension.name])
-        store, __ = self.stores.get_or_create(
-            parent=self.create_parent_dirs(pootle_path),
-            pootle_path=pootle_path,
-            name=posixpath.basename(pootle_path))
-        return store
+        name = posixpath.basename(pootle_path)
+        if name in ["template.po", "templates.po"]:
```
unho: There should be a better way to check if this project uses a gnu-style layout. Note that you might have a non-gnu project which has files like `template.pot`.

phlax: there is no longer gnu/nongnu etc - so this is the way to check that

re po/pot - the template(s) does have pot suffix - this is the name of the store that will be created

unho: Well, in the form we have a translation mapping preset for `gnu` style, can't we check if project's translation path mapping matches against it?

phlax: no - not really the preset mapping could be (almost) anything as long as it has <language_code> somewhere and ends with `<ext>`

pootle/apps/pootle_translationproject/models.py
phlax: and is handled in the calling method appropriately

pootle/apps/pootle_translationproject/models.py
phlax: sorry, i misunderstood the point about non-gnu templates.pot - in that case (which is a pretty corner-case) i think this is still the right way to deal with it

unho: The preset for `gnu` style is `/po/<language_code>.<ext>`. The preset is hardcoded, so I am pretty sure that whatever the translation path mapping the project uses, if it matches this preset we can tell the project layout is `gnu`. I think adding this check to the `if` condition will avoid that corner case I talked about.

phlax: way too arbitrary - this is the correct solution - whenever it sees a template pot with name `template` (which is equiv of a lang code) it uses the lang code instead

PR #6705, opened 2017-09-09
Deprecate update/sync stores commands

### LINKED ISSUE #6638: Add tests and check flags for update_stores/sync_stores
Update/sync stores has been switched to use pootle fs, although not all command flags are currently used.

We need to make sure that the flags work or are removed if that is appropriate. We also need some better testing around these commands

Probably we want to also add a deprecation notice on these commands as the pootle fs api provides a richer interface to these activities

### LINKED ISSUE #6690: Deprecate update/sync stores
Wondering what is best with these commands.

on the one hand they are quite useful for grouping common operations

on the other, it would be better for users to learn the more powerful fs api, and grouping can be done in other ways

phlax:
also:

- warn that force no longer has any effect
- fix the overwrite option

fixes #6638 
fixes #6690

### DIFF
--- a/pootle/apps/pootle_app/management/commands/sync_stores.py
+++ b/pootle/apps/pootle_app/management/commands/sync_stores.py
@@ -28,9 +28,7 @@ class Command(PootleCommand):
             action='store_true',
             dest='overwrite',
             default=False,
-            help="Don't just save translations, but "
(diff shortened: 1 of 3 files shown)


unho:
Tests need to be adjusted, but other than that lgtm.

Issue #6639, opened 2017-08-28
Improve projects add/edit ui for fs


phlax:
The ui for setting up fs for a project is a bit flaky and could do with some love

PR #6706, opened 2017-09-10
Set default project path for localfs in project form

phlax:
as in title

ref #6639

### DIFF
--- a/pootle/apps/pootle_app/forms.py
+++ b/pootle/apps/pootle_app/forms.py
@@ -6,11 +6,13 @@
 # or later license. See the LICENSE file for a copy of the license and the
 # AUTHORS file for copyright and authorship information.
 
+import os
 import re
 import urlparse
 from collections import OrderedDict
 
 from django import forms
+from django.conf import settings
 from django.contrib.auth import get_user_model
 
 from pootle.i18n.gettext import ugettext_lazy as _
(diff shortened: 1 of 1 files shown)


pootle/apps/pootle_app/forms.py
```diff
@@ -93,8 +95,13 @@ def clean_code(self):
         return self.cleaned_data['code'].strip()
 
     def save(self, commit=True):
+        created = bool(not self.instance.pk)
         project = super(ProjectForm, self).save(commit=commit)
         project.config["pootle_fs.fs_type"] = self.cleaned_data["fs_plugin"]
+        if created and self.cleaned_data["fs_plugin"] == "localfs":
```
unho: What if an already existing project changed its plugin to `localfs`. I would just drop the `created` bit.

phlax: if we do that it will overwrite any custom path

unho: Sure, but that is what we want, isn't it?

phlax: i mean it would overwrite an already configured project

unho: mmm, you are right. But the first part of my initial comment in this thread still stands.

phlax: if you change the fs_type then it would be kinda up to you to update the url too. This is just setting a default

phlax:
obsoleted by #6710

PR #6707, opened 2017-09-10
Fs placeholder

phlax:
if a `localfs` path starts with `{POOTLE_TRANSLATION_DIRECTORY}` it will interpolate the appropriate setting value.

this means that for a site that has several localfs projects, the translation directory setting can be changed without having to go back through all of the projects and update the paths

### DIFF
--- a/pootle/apps/pootle_fs/plugin.py
+++ b/pootle/apps/pootle_fs/plugin.py
@@ -13,6 +13,7 @@ import uuid
 
 from bulk_update.helper import bulk_update
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db import transaction
 from django.utils.functional import cached_property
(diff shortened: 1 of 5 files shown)


pootle/apps/pootle_fs/plugin.py
```diff
@@ -82,7 +83,16 @@ def finder_class(self):
...
+        fs_type = self.project.config["pootle_fs.fs_type"]
+        fs_url = self.project.config["pootle_fs.fs_url"]
+        parse_placeholder = (
+            fs_type == "localfs"
+            and fs_url.startswith("{POOTLE_TRANSLATION_DIRECTORY}"))
+        if parse_placeholder:
+            return os.path.join(
+                settings.POOTLE_TRANSLATION_DIRECTORY,
+                fs_url[30:])
+        return fs_url
```
unho: Shorter version:

```python
        fs_url = self.project.config["pootle_fs.fs_url"]
        if self.project.config["pootle_fs.fs_type"] == "localfs":
            return fs_url
        if fs_url.startswith("{POOTLE_TRANSLATION_DIRECTORY}"):
            return os.path.join(
                settings.POOTLE_TRANSLATION_DIRECTORY,
                fs_url[30:])
        return fs_url
```

unho:
Tests fail. Also the explanation on the forms will need to be expanded to include the new behavior. And docs are also needed.

pootle/apps/pootle_fs/plugin.py
phlax: your version wouldnt work

phlax:
re docs - there is a ticket for updating the docs - i would rather not deal with that in this PR

pootle/apps/pootle_fs/plugin.py
unho: k

Issue #5916, opened 2017-01-29
Check pootle fs docs


phlax:
Docs seem to be wrong in places - eg here the pootle-wins flag is the wrong way roung

http://docs.translatehouse.org/projects/pootle/en/latest/features/using_pootle_fs.html#overwrite-pootle-with-filesystem-version

we also need to check that overwrite still works

PR #6721, opened 2017-09-12
Pootle FS docs adjustments

unho:
This includes screenshots updates, adjustments because Pootle FS is now the only way and is no longer beta, deprecation of `sync_stores`, `update_stores`...

References #5916.


docs/features/pootle_fs_add_project.rst
```diff
@@ -39,7 +39,9 @@ is outlined here:
   - Set the **Filesystem backend** to ``localfs`` as we are working against
     files stored on the local filesystem.
   - Set the **Path or URL** to point to the translation files on
-    Pootle's local filesystem, e.g. ``/path/to/translations/MYPROJECT/``
+    Pootle's local filesystem, e.g. ``{POOTLE_TRANSLATION_DIRECTORY}MYPROJECT``
+    (``{POOTLE_TRANSLATION_DIRECTORY}`` will be transparently replaced by the
+    value of :setting:`POOTLE_TRANSLATION_DIRECTORY` setting)
```
dwaynebailey: I would add the `{POOTLE_TRANSLATION_DIRECTORY}` as a second example.  The way its written now someone will assume that they must use `{POOTLE_TRANSLATION_DIRECTORY}` but it is optional

docs/server/project_setup.rst
```diff
@@ -130,9 +130,9 @@ necessary to set the the filesystem configuration. So now click on the
 
 
 - **Filesystem backend** to ``localfs``
-- **Path or URL** to the value of
-  :setting:`POOTLE_TRANSLATION_DIRECTORY` + project code, e.g.
-  :file:`/path/to/pootle/translations/my-project` in our example
+- **Path or URL** to ``{POOTLE_TRANSLATION_DIRECTORY}my-project``
+  (``{POOTLE_TRANSLATION_DIRECTORY}`` will be transparently replaced with the
+  value of :setting:`POOTLE_TRANSLATION_DIRECTORY` setting)
 - **Path mapping preset** to ``GNU style``
```
dwaynebailey: Mmm we seem to have duplicated stuff here.  Can we rationalise it into one place?

docs/server/project_setup.rst
```diff
@@ -176,8 +176,10 @@ dropdown:
    .. image:: ../_static/project_form_bottom_links.png
 
 
-The existing languages enabled for the project are listed. In our example we
-are adding **Arabic** to the project:
+The existing languages enabled for the project are listed, including an
+optional mapping to the language code used in the filesystem (in the screenshot
+below you can see that ``fr_FR`` on filesystem maps to ``fr`` on Pootle). In
+our example we are adding **Arabic** to the project:
```
dwaynebailey: Please break out a new idea into its own section.  Mapping is important, lets write a section on it if needed instead of hiding it inside an example that has nothing to do with mapping.

docs/server/project_setup.rst
```diff
@@ -106,12 +106,20 @@ will see a **New Project** button:
 
 
 Click on that button and the **Add Project** form will be displayed. Enter the
-new project's details. **Code** must match the name of the directory within
+new project's details.
+
+
+**Code** must match the name of the directory within
 :setting:`POOTLE_TRANSLATION_DIRECTORY` that contains the project translation
 files, in our example :file:`my-project`. Also you must specify the
```
dwaynebailey: This section is not tur anymore, it's just a code

docs/features/pootle_fs_add_project.rst
```diff
@@ -23,35 +23,31 @@ Synchronizing against any version control system requires you add some
...
-    ``/<language_code>/<dir_path>/<filename>.<ext>`` or pick one of the
-    existing **Path mapping presets**.
-
-  .. note:: If you are setting up Pootle FS for a VCS then configure as
-     follows:
-
-     - Set the **Filesystem backend** to the required VCS backend.
-     - Set the **Path or URL** to point to the repository, e.g.
-       ``git@github.com:user/repo.git``
+Just follow the :ref:`Creating a project <project_setup#creating-the-project>`
```
dwaynebailey: Surely we don't need this anymore, we only need the one add a project section.

dwaynebailey:
There are some bigger issues of doc duplication, so while the corrections might be good, the correct solution is to merge.  Biggie is we previously needed to explain Pootle FS project setup apart from normal, now there is only one way to setup projects.  So we don't need two sets of the same docs.

PR #6724, opened 2017-09-12
Invert k,v in language mapping dictionary

phlax:
currently the lang mappings are the wrong way round, this inverts them

also fixes a situation where you cannot unset a mapping


unho:
lgtm

PR #6747, opened 2017-09-13
Add character count to edit page

### LINKED ISSUE #6725: Add current character count when entering translations
We're trying to use Pootle to translate metadata strings for an app in the iOS AppStore. The metadata includes individual messages for the app name, subtitle, description, privacy URL and so on and there are different limits on the number of characters allowed in each of them. For instance, an app's name can be no more than 30 characters. 

When entering translations, it would be really helpful to see the current number of characters that the translation uses as you type to ensure that you're not exceeding the limit. This could maybe fit on the lower right corner of the input view. You currently have timeline / comment / raw on the left. Current character count could just be a small label that floats to the right on the same line.

# Environment (i.e. 'pootle --version', DB, OS, Browser):

Pootle 2.8.0

Johennes:
This adds an auto-updating label for the current character count to the bottom of the editing text area.

Closes #6725

This obviously needs more polish. I just mean this to be a basis for discussion for what I requested in #6725 and how to achieve it. Here is an example of how it looks.

Example

### DIFF
--- a/pootle/core/templatetags/core.py
+++ b/pootle/core/templatetags/core.py
@@ -25,6 +25,13 @@ def to_js(value):
     return mark_safe('JSON.parse("%s")' % escapejs(jsonify(value)))
 
 
+@register.filter
+def map_to_lengths(value):
+    """Maps a list value by replacing each element with its length.
+    """
+    return [len(e) for e in value]
+
+
 @register.inclusion_tag('includes/formtable.html')
 def formtable(formtable):
     return dict(formtable=formtable)
--- a/pootle/static/js/editor/app.js
+++ b/pootle/static/js/editor/app.js
@@ -915,6 +915,23 @@ PTL.editor = {
       this.checkSimilarTranslations();
       this.similarityTimer = null;  // So we know the code was run
     }, 200);
+
+    this.updateCharacterCount();
+  },
+
+  getCharacterCountString() {
+    if (ReactEditor.stateValues.length === 1) {
+      const counter = ReactEditor.stateValues[0].length;
+      const format = ngettext('%s character', '%s characters', counter);
+      return interpolate(format, [counter]);
+    }
+    const counters = ReactEditor.stateValues.map(e => e.length).join(' \ ');
+    const format = gettext('%s characters');
+    return interpolate(format, [counters]);
+  },
+
+  updateCharacterCount() {
+    $('#js-char-count').html(this.getCharacterCountString());
   },
 
   isTextareaValueDirty() {
--- a/pootle/templates/editor/units/edit.html
+++ b/pootle/templates/editor/units/edit.html
@@ -212,9 +212,23 @@
               {% endblock %}
               <a class="js-toggle-raw"><i class="icon-raw-mode" title="{% trans 'Raw View' %}"></i></a>
             </div>
-            {% if unit.istranslated or unit.isfuzzy %}
-	    {% if unit.change.submitted_by and unit.change.submitted_on %}
             <div class="unit-meta">
+              <span id="js-char-count">
+                {% if unit_values|length == 1 %}
+                  {% blocktrans count counter=unit_values|first|length %}
+                    {{ counter }} character
+                  {% plural %}
+                    {{ counter }} characters
+                  {% endblocktrans %}
+                {% else %}
+                  {% blocktrans with counter=unit_values|map_to_lengths|join:" / " %}
+                    {{ counter }} characters
+                  {% endblocktrans %}
+                {% endif %}
+              </span>
+              {% if unit.istranslated or unit.isfuzzy %}
+              {% if unit.change.submitted_by and unit.change.submitted_on %}
+              <br>
               <time
                 title="{{ unit.change.submitted_on|dateformat }}"
                 datetime="{{ unit.change.submitted_on.isoformat }}">{{ unit.change.submitted_on|relative_datetime_format }}</time>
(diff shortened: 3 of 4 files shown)


pootle/templates/editor/units/edit.html
```diff
@@ -201,6 +201,7 @@
                 alternativeSources: {{ altsrcs|to_js }},
                 fileType: '{{ filetype }}',
                 suggestions: {{ suggestions_dict|to_js }},
+                onChange: function(values) { $('#char-count-{{ unit.id }}').html(values.map(function(e) { return e.length; }).join(' \ ')); },
```
phlax: can we move javascript to one of the js files - probs `/static/js/editor/app.js` you can use jquery to listen for change

phlax: also, this prevents the submit button from working

Johennes:
Ok, next iteration. This time it should actually work. :)

I couldn't find a good way to handle units with multiple plural forms (= multiple textareas). As a cheap workaround I just joined all the counters with `" / "`. Maybe it's worth rethinking the placement of the counter label in general. I'm currently failing to see a better placement though.

Unrelated question: Would it be preferable to use `"xx characters"` instead of just `"xx"` for the label?

pootle/templates/editor/units/edit.html
```diff
@@ -215,6 +215,13 @@
             {% if unit.istranslated or unit.isfuzzy %}
 	    {% if unit.change.submitted_by and unit.change.submitted_on %}
             <div class="unit-meta">
+              <span id="char-count">
```
unho: Change the id to `js-char-count`.

pootle/static/js/editor/app.js
```diff
@@ -915,6 +915,13 @@ PTL.editor = {
       this.checkSimilarTranslations();
       this.similarityTimer = null;  // So we know the code was run
     }, 200);
+
+    this.updateCharacterCount()
+  },
+
+  updateCharacterCount() {
+    let count = ReactEditor.stateValues.map(e => e.length).join(' \ ')
+    $('#char-count').html(count)
```
unho: Travis detects several errors with this Javascript code:

- Must use semicolons at end of relevant lines
- Use `const` instead of `let` because it never gets reassigned

unho:
Tested this and it only works for strings that have been previously translated, shouldn't it be better to always show the character count?

I suggest to use `"xx characters"` instead of just `"xx"` for the label because otherwise people won't understand what the count is. Note it needs to be translatable.

Johennes:
@unho Thanks for the review. I changed the span's ID and addressed those JS issues. Will look into adding translatable "characters" and not hiding it on untranslated strings now.

Johennes:
Ok, I made the label read `"xx Characters"` and prevented it from being hidden for untranslated strings. I also put it on its own line analogous to the editing time and reviewer.

pootle/templates/editor/units/edit.html
```diff
@@ -212,9 +212,15 @@
               {% endblock %}
               <a class="js-toggle-raw"><i class="icon-raw-mode" title="{% trans 'Raw View' %}"></i></a>
             </div>
-            {% if unit.istranslated or unit.isfuzzy %}
-	    {% if unit.change.submitted_by and unit.change.submitted_on %}
             <div class="unit-meta">
+              <span id="js-char-count">
+                {% blocktrans with count=unit_values|map_to_lengths|join:" / " %}
+                  {{ count }} Characters
```
unho: Nit: Use `characters` (in lowercase)

unho: Also this needs to be able to handle plurals

Johennes: Yes, so the thing with the plurals is that for messages with more than one text area, the string will read `"12 / 13 / 17 / ... characters"` so I wasn't sure how to properly handle plurals on that one.

pootle/static/js/editor/app.js
```diff
@@ -915,6 +915,15 @@ PTL.editor = {
       this.checkSimilarTranslations();
       this.similarityTimer = null;  // So we know the code was run
     }, 200);
+
+    this.updateCharacterCount();
+  },
+
+  updateCharacterCount() {
+    const counters = ReactEditor.stateValues.map(e => e.length).join(' \ ');
+    const format = gettext('%s characters');
```
unho: This needs plurals too:

```javascript
    const counters = ReactEditor.stateValues.map(e => e.length).join(' \ ');
    const string = nt('%s character', '%s characters', counters, [counters]);
    $('#js-char-count').html(string);
```

unho: This might require an import at the top:

```javascript
import { nt } from 'utils/i18n';
```

pootle/templates/editor/units/edit.html
unho: IMHO it is ok to always using "characters" if there is more than a textarea, but if there is only a single textarea I would use the plurals.

unho: Or maybe it is better to show the character count for each textarea instead of having the ` / `, but not sure about the impact in the UI. @phlax Any input on this?

Johennes:
Changed it to correctly pluralize when there is only one text area and always use "characters" when there are multiple.

phlax:
lgtm

Johennes:
I think this build error has nothing to do with my changes now, or does it? The log mentions Python files I didn't touch.

unho:
Yes, the failing tests are unrelated to the changes in this PR.

phlax:
@Johennes the last test run has linting errors - that one needs to be fixed - but the other fail is unrelated

Johennes:
@phlax I'm sorry but I'm struggling with reading the logs. Could you please point me to where I could see the linting issues?

phlax:
@Johennes its fixed - sorry i should have checked

Johennes:
Ah, no worries. :)

pootle/core/templatetags/core.py
```diff
@@ -25,6 +25,13 @@ def to_js(value):
     return mark_safe('JSON.parse("%s")' % escapejs(jsonify(value)))
 
 
+@register.filter
+def map_to_lengths(value):
+    """Maps a list value by replacing each element with its length.
```
phlax: this could do with a quick test

Johennes: Added a few basic tests. I hope I put them into the right directory.

PR #6752, opened 2017-09-15
Fs presets

phlax:
makes fs_presets a dict, which makes it easier to work with and gives unique lowercase keys for project command


unho:
lgtm provided Travis passes.

PR #6757, opened 2017-09-19
Ensure filetypes are added to new project in correct order

phlax:
this is important when choosing a filetype for a file if there are more than one filtypes registered with the same extension


unho:
lgtm

PR #6758, opened 2017-09-19
Fs type validation

phlax:
adds validation for setting fs_type config and adjusts/removes tests as appropriate


unho:
lgtm provided Travis passes

PR #6760, opened 2017-09-20
More docs adjustments for Pootle FS

### LINKED ISSUE #5916: Check pootle fs docs

unho:
Addresses the concerns raised in #6721 that still were pending.

Fixes #5916.

### DIFF
--- a/docs/server/project_setup.rst
+++ b/docs/server/project_setup.rst
@@ -303,7 +303,9 @@ language. The first step is to save all the Pootle translations to disk:
 
 .. code-block:: console
 
-    (env) $ pootle sync_stores --project=my-project
+    (env) $ pootle fs fetch my-project
+    (env) $ pootle fs resolve my-project --overwrite --pootle-wins
+    (env) $ pootle fs sync my-project --update=fs
 
 
 Then update all those translations on disk against the newer templates. We
--- a/docs/server/project_setup.rst
+++ b/docs/server/project_setup.rst
@@ -291,7 +291,9 @@ update the template translations in the Pootle database.
 
 .. code-block:: console
 
-    (env) $ pootle update_stores --project=my-project --language=templates
+    (env) $ pootle fs fetch my-project --fs-path=my-project/templates/*
+    (env) $ pootle fs resolve my-project --overwrite --fs-path=my-project/templates/*
+    (env) $ pootle fs sync my-project --update=pootle --fs-path=my-project/templates/*
 
 
 This command will ensure that new strings are added to the project and any
@@ -328,7 +330,9 @@ to Pootle:
 
 .. code-block:: console
 
-    (env) $ pootle update_stores --project=my-project
+    (env) $ pootle fs fetch my-project
+    (env) $ pootle fs resolve my-project --overwrite
+    (env) $ pootle fs sync my-project --update=pootle
 
 
 .. note:: If your project languages contain many translations you might want to


docs/features/index.rst
```diff
@@ -13,7 +13,6 @@ Pootle FS
    :maxdepth: 2
 
    pootle_fs
-   pootle_fs_install_plugins
```
dwaynebailey: Not sure it makes sense to move VCS plugins instructions to migration.  It might make sense when someone migrates from 2.8 -> 2.9.  But after that its a normal install situation, and someone won't be migrating.  It's a Pootle FS setup issue.

docs/features/migrate_to_pootle_fs.rst
```diff
@@ -79,13 +79,36 @@ plugin.  For examples for Git:
 This is done once for the whole Pootle server.
 
 
+.. _migrate_to_pootle_fs#connect-with-vcs:
```
dwaynebailey: Same problem with this move.  Migration is once off and now setup instructions are hidden here.  The correct approach is the other way around.  Migration, once off, should point to general setup instructions as needed.

docs/features/migrate_to_pootle_fs.rst
```diff
@@ -38,6 +38,31 @@ files on disk will likely mess with direct Pootle FS change detection.
 You may want to look at the format adaptors for future massaging or formats.
 
 
+.. _migrate_to_pootle_fs#replacing-update_stores-and-sync_stores:
+
+Replacing update_stores and sync_stores
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:command:`update_stores` can be replaced with the following set of Pootle FS
```
dwaynebailey: Please always use `:djadmo:` for Pootle and Django commands, this automatically links to them in the docs.  `:command:` is just styling and for commands like `git`, etc

docs/features/migrate_to_pootle_fs.rst
```diff
@@ -38,6 +38,31 @@ files on disk will likely mess with direct Pootle FS change detection.
...
+commands:
+
+.. code-block:: console
+
+  (env) $ pootle fs fetch $MYPROJECT
+  (env) $ pootle fs resolve $MYPROJECT --overwrite
+  (env) $ pootle fs sync $MYPROJECT --update=pootle
+
+
+:command:`sync_stores` can be replaced with the following set of Pootle FS
```
dwaynebailey: `:djadmin:`

docs/features/pootle_fs.rst
```diff
@@ -89,7 +89,9 @@ and :djadmin:`sync_stores`, to allow translations to be pushed into Pootle or
 pulled from Pootle.
 
 These two commands still exist but are now deprecated. We will be phasing these
-out in the long term to make everything use Pootle FS.
+out in the long term to make everything use Pootle FS, but you can
+:ref:`replace these commands with Pootle FS
+<migrate_to_pootle_fs#replacing-update_stores-and-sync_stores>`.
```
dwaynebailey: Pootle FS equivalents.

docs/features/migrate_to_pootle_fs.rst
```diff
@@ -135,16 +135,26 @@ Configure the project to use VCS
...
+- Ensure you have committed all the translation files to your version control
+  system.
+- Set the project's **Filesystem backend** to the appropriate VCS backend.
+- Set the URL to your version control repository in the project's **Path or
+  URL**, e.g. ``git@github.com:user/repo.git``.
+- Synchronize as follows:
+
+  .. code-block:: console
+
+    (env) $ pootle fs fetch --force $MYPROJECT
```
dwaynebailey: There is no `--force` in `fetch` anymore.

docs/server/project_setup.rst
```diff
@@ -124,15 +124,51 @@ files, in our example :file:`my-project`. Also you must specify the
...
+and they might require the installation of additional Pootle FS plugins.
+
+**Path or URL** either is a path pointing to the translation files on the disk
+if you are using the ``localfs`` **Filesystem backend**, or is a URL pointing
+to the VCS repository. The path can be an absolute path like
+``/path/to/translations/my-project/`` or it can use the
+``{POOTLE_TRANSLATION_DIRECTORY}`` placeholder if you are using the ``localfs``
+**Filesystem backend**, like ``{POOTLE_TRANSLATION_DIRECTORY}my-project``
+(the placeholder will be transparently replaced by the value of the
+:setting:`POOTLE_TRANSLATION_DIRECTORY` setting).
```
dwaynebailey: I'd add a similar verbose instance of a proper non-localfs backend e.g. git.  Though you do that later which might be enough, just worried this leaves people with a localfs impression, rather than a fuller picture.

docs/server/project_setup.rst
```diff
@@ -139,6 +139,17 @@ project.
 Once you are done click on the **Save** button below the form to create the
 project and save its Pootle FS configuration.
 
+If not all of your project's language codes match those available in Pootle,
```
dwaynebailey: If all of your .... language codes **do not** match...

docs/server/project_setup.rst
```diff
@@ -139,6 +139,17 @@ project.
...
+then you must add language mapping configurations for those languages. You can
+do that by clicking on the **Languages** link that is displayed below your
+project form:
+
+.. image:: ../_static/project_form_bottom_links.png
+
+
+and then provide a different language code in the **Filesystem language code**
+field for those languages whose codes need to be mapped.
+
```
dwaynebailey: I think the above really want an example of what we mean or why this is important.  Something like:

This allows you to maintain consistent language codes on Pootle that can map to any codes that you have on various projects.  Examples might be `af` (Afrikaans) on Pootle but `af-ZA` in version control.  W can now map `af` to `af-ZA`.  Other common mappings might be `fr-FR` to `fr` (French) and `es_ES` to `es` (Spanish).  Or `en_GB` to `en-GB` where we are using different separators.

docs/server/project_setup.rst
```diff
@@ -94,6 +94,13 @@ translation files for your new project. Make sure you have a
 Creating the project
 ++++++++++++++++++++
 
+.. note:: If you want to integrate your project with a version control
+   repository you need to :ref:`install the Pootle FS plugins for the VCS
+   <migrate_to_pootle_fs#install-vcs-plugins>` and :ref:`connect Pootle FS with
+   the VCS repository <migrate_to_pootle_fs#connect-with-vcs>` before creating
+   the project in Pootle.
```
dwaynebailey: I like that VCS setup doesn't crowd this out, but again like I said earlier, I don't think this crucial info should be buried in migration.  It does need its own doc as it's a once off thing.

docs/server/project_setup.rst
```diff
@@ -195,6 +195,19 @@ of the Pootle server:
     (env) $ pootle update_stores --project=my-project
 
 
+.. note:: Alternatively you can do this with the following set of Pootle FS
+   commands:
```
dwaynebailey: This shouldn't be alternatively, this should be the default way.  So I'd remove all `update_stores` found earlir and replace it with what you have here.

docs/server/project_setup.rst
```diff
@@ -195,6 +195,19 @@ of the Pootle server:
...
+
+   .. code-block:: console
+
+     (env) $ pootle fs state my-project
```
dwaynebailey: You'd need `fetch` first.

docs/server/project_setup.rst
```diff
@@ -186,6 +186,12 @@ project form:
...
 field for those languages whose codes need to be mapped.
 
+
+.. _project_setup#import-translations:
+
+Import the translations
++++++++++++++++++++++++
+
 Creating the project doesn't actually import all the translations to
 Pootle. To do that you need to run :djadmin:`update_stores` on the command line
```
dwaynebailey: If Pootle FS is default, we shouldn't have any `update_stores` anymore for basic instructions.

docs/features/index.rst
unho: Will have separate docs for "migrate to Pootle FS" and "integrate with VCS".

docs/features/migrate_to_pootle_fs.rst
unho: I used `:command:` on purpose for those commands that are deprecated, but will change to `:djadmin:`.

docs/server/project_setup.rst
unho: This explanation is added by a later commit.

unho:
Merged. Will create another PR for the changes not merged.

PR #6791, opened 2017-11-27
Move subcommand to dj.subcommand

phlax:
I have separated and released the subcommand functionality here http://github.com/phlax/dj.subcommand

This PR uses that module instead of hosting the code in pootle. This will reduce code, make tests quicker etc. This makes sense as the code rarely changes, and is not specific to pootle.

There is a fix in the upstream code already.


pootle/core/management/subcommands.py
```diff
@@ -6,241 +6,21 @@
...
 from django.utils.functional import cached_property
 
-from pootle.core.delegate import subcommands
+from dj import subcommand
 
+from pootle.core.delegate import subcommands
 
-class SubcommandsError(CommandError):
 
+class SubcommandsError(subcommand.subcommands.SubcommandsError):
```
unho: Do we need to define here a exception that has been defined in third party package?

unho:
lgtm after nit, provided Travis passes.

pootle/core/management/subcommands.py
phlax: i did this to avoid code changes elsewhere, but maybe it wasnt necessary - ill check

PR #6793, opened 2017-12-11
Import django_bulk_update instead of bulk_update

erosennin:
The module name of django-bulk-update has changed from `bulk_update` to `django_bulk_update` in version 2.0.0 ([see changelog](https://github.com/aykut/django-bulk-update/blob/master/CHANGELOG)). Pootle still uses the old package name. This still works most of the time, because `django_bulk_update-2.2.0-py2.py3-none-any.whl` on PyPI includes the old `bulk_update` module for backward compatibility. However, the source tree of django-bulk-update does not contain the `bulk_update` module anymore, and Pootle currently won't work if django-bulk-update is installed from source. This PR fixes that.

### DIFF
--- a/pootle/apps/pootle_fs/plugin.py
+++ b/pootle/apps/pootle_fs/plugin.py
@@ -11,7 +11,7 @@ import os
 import shutil
 import uuid
 
-from bulk_update.helper import bulk_update
+from django_bulk_update.helper import bulk_update
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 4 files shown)


phlax:
@erosennin looks good - but im wondering if we need to bump requirements to ensure that pootle uses the newer version of bulk update

erosennin:
@phlax looks like we already have `django-bulk-update==2.2.0` in `requirements/base.txt`, it is currently the latest version of django-bulk-update.

PR #6795, opened 2017-12-24
OC locale: some updates

Quenty31:
I couldnt bear to see the mistake "i a" instead of "Fa XX minutas"


Quenty31:
oh oh, why didn't it work?

PR #6821, opened 2018-04-05
Requirement updates django

phlax:
Update base requirements and adjust related tests

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -5,42 +5,42 @@
 Django~=1.10.8  # rq.filter: <1.11
 
 # Django apps
-django-allauth==0.33.0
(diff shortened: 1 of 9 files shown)


requirements/base.txt
```diff
@@ -2,7 +2,7 @@
 
 # Note: also adjust pootle/constants::DJANGO_MINIMUM_REQUIRED_VERSION
 # and docs/releases/M.N.O.rst
-Django~=1.10.8  # rq.filter: <1.11
+Django~=1.11.12  # rq.filter: <1.13
```
unho: @phlax This would probably install latest Django 1.12.x where the intention seems to be to make it 1.11.x

Issue #6887, opened 2019-01-27
Django-Redis DataError Invalid type 'CacheKey'


arky:
# Steps to reproduce:

Pootle Linux development environment setup fails at 'pootle migrate' step with  "redis.exceptions.DataError: Invalid input of type: 'CacheKey'. Convert to a byte, string or number first" error. 

# Results (Expected/Actual):
```$ pootle migrate
Traceback (most recent call last):
  File "/home/arky/.venvs/pootle_test/bin/pootle", line 11, in <module>
    load_entry_point('Pootle', 'console_scripts', 'pootle')()
  File "/home/arky/Code/Django/pootle/pootle/runner.py", line 365, in main
    run_app(project='pootle', django_settings_module='pootle.settings')
  File "/home/arky/Code/Django/pootle/pootle/runner.py", line 320, in run_app
    caches = settings.CACHES.keys()
  File "/home/arky/.venvs/pootle_test/local/lib/python2.7/site-packages/django/conf/__init__.py", line 56, in __getattr__
    self._setup(name)
  File "/home/arky/.venvs/pootle_test/local/lib/python2.7/site-packages/django/conf/__init__.py", line 41, in _setup
    self._wrapped = Settings(settings_module)
  File "/home/arky/.venvs/pootle_test/local/lib/python2.7/site-packages/django/conf/__init__.py", line 129, in __init__
    raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
(pootle_test) arky@Think:~/Code/Django/pootle$ pootle init 
File already exists at '/home/arky/.venvs/pootle_test/pootle.conf', overwrite? [Ny] n
2019-01-27 16:07:56,478 ERROR File already exists, not overwriting.
(pootle_test) arky@Think:~/Code/Django/pootle$ nano ~/.venvs/pootle_test/pootle.conf 
(pootle_test) arky@Think:~/Code/Django/pootle$ pootle migrate
Traceback (most recent call last):
  File "/home/arky/.venvs/pootle_test/bin/pootle", line 11, in <module>
    load_entry_point('Pootle', 'console_scripts', 'pootle')()
  File "/home/arky/Code/Django/pootle/pootle/runner.py", line 365, in main
    run_app(project='pootle', django_settings_module='pootle.settings')
  File "/home/arky/Code/Django/pootle/pootle/runner.py", line 346, in run_app
[...]

PR #6895, opened 2019-04-12
Update django-redis version

phlax:
to <3 as it seems to break django-redis


arky:
@phlax Cute pun on version number. 

I had tried fixing this in local environment sometime back (https://github.com/translate/pootle/issues/6887). Both pinning redis to (pip install redis==2.10.6) or upgrading django-redis (pip install django-redis==4.10.0) works for me.

phlax:
@arky - thanks ive updated the `django-redis` version instead

PR #6926, opened 2024-08-26
Updates README to note that Pootle is not actively maintained

### LINKED ISSUE #6920: Status of that project?
The last commit was in June 2022 (last year).

Is the project still maintained somehow?

linghengqian:
- Fixes #6920 .
- Updates README to note that Pootle is not actively maintained. See https://github.com/translate/pootle/issues/6920#issuecomment-1525567199 and https://github.com/translate/pootle/issues/6920#issuecomment-1525596832 .

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,6 +1,14 @@
 |logo| Pootle
 =============
 
+Maintainers needed
+------------------
+
+Pootle is in the process of looking for potential maintainers to take over.
+If you are interested in maintaining Pootle please open an issue at `Pootle Issues <https://github.com/translate/pootle/issues>`_.
+
+Until then, users can use `Weblate <https://github.com/WeblateOrg/weblate>`_ as an alternative.
+
 |chat| |build| |health| |coverage| |requirements|
 
 


linghengqian (commented):
- CC @phlax .

README.rst
```diff
@@ -1,6 +1,9 @@
 |logo| Pootle
 =============
 
+Pootle is in the process of looking for potential maintainers to take over. Until then, users can use `Weblate <https://github.com/WeblateOrg/weblate>`_ as an alternative.
+------------------
```
phlax, line 5: in rst the underline needs to be at least as long as the sentence - usually best to make them the same length

phlax (commented):
thanks for this @linghengqian - not 100% about status of docs publishing but would be good to land this i think

probably we should add something like

> If you are interested in maintaining Pootle please open an issue at ...

README.rst
linghengqian: - I modified the document. To be honest I had never used reStructuredText before.

phlax (commented):
how about not making these headers?

perhaps add a header like

```rst

Maintainers needed
---------------------------

...
```

and then add the points you have as the body of the paragraph

linghengqian:
> and then add the points you have as the body of the paragraph

- How about this? I updated the commit.
- image

phlax (approved):
lgtm, thanks @linghengqian

